PR #19, opened 2024-03-24
Add MLKEM C reference code

cothan:

This MLKEM reference code is taken directly from Kyber standard branch https://github.com/pq-crystals/kyber/tree/11d00ff1f20cfca1f72d819e5a45165c1e0a2816
with Kyber 90s variants and KEX removed.

### DIFF
--- a/mlkem/api.h
+++ b/mlkem/api.h
@@ -1,0 +1,66 @@
+#ifndef API_H
+#define API_H
+
+#include <stdint.h>
+
+#define pqcrystals_kyber512_SECRETKEYBYTES 1632
+#define pqcrystals_kyber512_PUBLICKEYBYTES 800
+#define pqcrystals_kyber512_CIPHERTEXTBYTES 768
+#define pqcrystals_kyber512_KEYPAIRCOINBYTES 64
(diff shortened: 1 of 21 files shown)


mlkem/api.h
```diff
@@ -0,0 +1,66 @@
+#ifndef API_H
```
mkannwischer, line 1: Please add the SPDX identifiers already, please? 
```
// SPDX-License-Identifier: Apache-2.0
```

cothan: I plan to separate the SPDX identifiers in a separate commit. 

Btw, I believe for MLKEM it is: 
```
// SPDX-License-Identifier: Apache-2.0 or CC0-1.0
```

right?

mkannwischer: No. We need to state in the text the original license which indeed is Apache-2.0 or CC0-1.0.
But we do need to state here that we are using it under Apache-2.0 and all modifications are going to be covered by that. I learned that only this week, so I'll still have to make the changes accordingly.

cothan: Nice. I'll add the Apache 2.0 License ASAP. CC0-1.0 is problematic to some organizations.

PR #24, opened 2024-03-27
Add astyle test to Github Workflow

cothan:

This PR will enforce future code to be formatted by Astyle, as provision by `.astylerc` at the root of the project. 


cothan:
This PR should fail until the other PR format the whole repo with astyle.

mkannwischer:
A problem we faced for https://github.com/pq-code-package/mlkem-c-embedded is that different versions of astyle format the code differently.
I see two options to solve this:
- We fix to the version that the Github runners have (and document it in the CONTRIBUTING documentation)
- We use something like nix to make sure we have full control over which version is used. 

Both are fine for me for now.

cothan:
The ubuntu-latest provide Astyle 3.1, while in other project my astyle version is 3.4.10. 
So I add Nix package, same as what `mlkem-c-embedded` provide.

flake.nix
```diff
@@ -0,0 +1,61 @@
...
+      perSystem = { pkgs, system, inputs', ... }:
+        let
+          core = with pkgs; [
+            # formatter & linters
+            astyle # 3.4.10
+
+            # build dependencies
+            gcc-arm-embedded-13 # arm-gnu-toolchain-13.2.rel1
+            openocd # 0.12.0
+            python311Packages.pyserial # 3.5
```
mkannwischer, line 28: Those are not needed for the AArch64 repo. Please remove.

cothan: Done in https://github.com/pq-code-package/mlkem-c-aarch64/pull/24/commits/e55890ea93839417830ae6db2b1c7d8e1dc08716

potsrevennil:
Hi, I came across this PR, I saw that you're using the nix CI setup step from `mlkem-c-embedded` `main` branch. I've did some refactor on the nix setup in CI since the original one is a bit mess, although not yet merged into the main branch, I would recommend to use this refactored version instead.

https://github.com/pq-code-package/mlkem-c-embedded/blob/bde05c9a0b4bbc902a889d82eb6f9913a1181117/.github/actions/setup-nix/action.yaml

cothan:
hi @potsrevennil , nice, thanks for the action.yaml, I will add it immediately.

mkannwischer (approved):
Thanks. LGTM

PR #29, opened 2024-03-29
Add checksum KAT to Github Workflow

cothan:

This PR should be merged after #24 . 

This PR add:

- `gen_KAT.c`: A way to deterministic generate KAT file. 
- `gen_NISTKAT.c`: Generate NISTKAT file to compare with official/upstream ML-KEM KATs.
- `checksum.sh`: A bash script run executable and compare output of executable with an hash. 
- `Github workflow` to test checksum of MLKEM512, 768, 1024 from 10,000 iterations.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,6 +28,9 @@ jobs:
       shell: nix develop .#ci -c bash -e {0}
       run: |
           make mlkem
+          ./checksum.sh ./test/test_kyber512 880e1fe907682fb5a9029fc7da114131a0740f4f3328d6ddfae46728b5ac8db0
+          ./checksum.sh ./test/test_kyber768 2b32e7f5aa91fe47b0ca9a1d5aa70d7dc7c4ff1ffcb7e8967e51bf15cc5d2095
+          ./checksum.sh ./test/test_kyber1024 87da552c0878350fa7643a6b49996893d4dba4d8be2badb32c4298c5c2b512c6
     - name: Compare gen_KAT with known hash
       shell: nix develop .#ci -c bash -e {0}
       run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
     - name: Build targets
       shell: nix develop .#ci -c bash -e {0}
       run: |
-          make
+          make mlkem
     - name: Compare gen_KAT with known hash
       shell: nix develop .#ci -c bash -e {0}
       run: |
@@ -35,3 +35,7 @@ jobs:
         ./checksum.sh ./test/gen_KAT512  ec4ac397e595ac7457cb7d8830921faf3290898a10d7dd3864aab89ea61fe9a3
         ./checksum.sh ./test/gen_KAT768  9a0826ad3c5232dfd3b21bc4801408655c565a491b760f509b2ee2cd7180babe
         ./checksum.sh ./test/gen_KAT1024 6dafb867599b750a6a831b03e494cf41dea748c78a0e275e7b268bbb893cf37d
+    - name: Clean up
+      shell: nix develop .#ci -c bash -e {0}
+      run: |
+        make clean
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ on:
     branches: [ '*' ]
   pull_request:
     branches: [ "main" ]
+
 jobs:
   build_test:
     runs-on: ubuntu-latest
@@ -20,10 +21,17 @@ jobs:
           if [[ ${#err} != 0 ]]; then
             echo "$err" | while IFS= read -r file; do
               echo "::error file={"$file"},title={checking}::Formatted $file"
-            done
-            exit 1
+              done
+              exit 1
           fi
     - name: Build targets
       shell: nix develop .#ci -c bash -e {0}
       run: |
           make
+    - name: Compare gen_KAT with known hash
+      shell: nix develop .#ci -c bash -e {0}
+      run: |
+        make kat; 
+        ./checksum.sh ./test/gen_KAT512  ec4ac397e595ac7457cb7d8830921faf3290898a10d7dd3864aab89ea61fe9a3
+        ./checksum.sh ./test/gen_KAT768  9a0826ad3c5232dfd3b21bc4801408655c565a491b760f509b2ee2cd7180babe
+        ./checksum.sh ./test/gen_KAT1024 6dafb867599b750a6a831b03e494cf41dea748c78a0e275e7b268bbb893cf37d
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,48 +3,48 @@
 name: Build
 on:
   push:
-    branches: [ '*' ]
+    branches: ["*"]
   pull_request:
-    branches: [ "main" ]
+    branches: ["main"]
 jobs:
   build_test:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
-    - name: Setup nix
-      uses: ./.github/actions/setup-nix
-    - name: Astyle
-      shell: nix develop .#ci -c bash -e {0}
-      run: |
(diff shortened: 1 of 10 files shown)


.github/actions/setup-nix~ef376a0 (refactor build.yml)
mkannwischer, line 1: This file should be removed

mkannwischer (changes requested):
Please rebase on top of main and resolve any conflicts. 

Do you think it's possible to switch to actual NISTKATs? 
Unfortunately, the hashes in the kyber repo are not up to date, but you can generate them yourself from the code in https://github.com/pq-crystals/kyber/tree/standard. 
You would have to re-implement it along the lines of https://github.com/PQClean/PQClean/blob/master/test/crypto_sign/nistkat.c - which should not be too hard. This will be much easier to match against the official testvectors that will be published by NIST togethr


MLKEM512: 
`4b88ac7643ff60209af1175e025f354272e88df827a0ce1c056e403629b88e04` 
MLKEM768:
`21b4a1e1ea34a13c26a9da5eeb9325afb5ca11596ca6f3704c3f2637e3ea7524`
MLKEM1024: 
`6471398b0a728ee1ef39e93bb89b526fbf59587a3662edadbcfc6c88a512cd71`

cothan:
Good idea. I will add NISTKAT test. I see it's no harm to add independent test as long as our checksum match NISTKAT. 
The reason I didn't add it as first place was the AES part to generate random data, which can be done with FIPS202 (already included in XOF of MLKEM). 
I want to separate the AES part out of the code, perhaps out of the repo so the actual code does not have any dependency with AES, do you have any idea? 

We can always fall back to add a `common/nistrng.c` and a `common/README.md` say that `nistrng.c` is only used for NIST KAT testing.

test/gen_KAT.c
```diff
@@ -0,0 +1,64 @@
+#include <stddef.h>
```
mkannwischer, line 1: Please add the SPDX license header.

checksum.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
mkannwischer, line 1: Please add the SPDX license header.

Makefile
mkannwischer, line 1: Please also add the SPDX license header to the Makefile.

.github/workflows/build.yml
```diff
@@ -20,10 +21,24 @@ jobs:
...
           fi
     - name: Build targets
       shell: nix develop .#ci -c bash -e {0}
       run: |
-          make
\ No newline at end of file
+          make mlkem
+          ./checksum.sh ./test/test_kyber512 880e1fe907682fb5a9029fc7da114131a0740f4f3328d6ddfae46728b5ac8db0
+          ./checksum.sh ./test/test_kyber768 2b32e7f5aa91fe47b0ca9a1d5aa70d7dc7c4ff1ffcb7e8967e51bf15cc5d2095
+          ./checksum.sh ./test/test_kyber1024 87da552c0878350fa7643a6b49996893d4dba4d8be2badb32c4298c5c2b512c6
```
mkannwischer, line 33: Do I understand correctly that this is to make sure that the tests don't print `ERROR`?
I think it is more robust to test for the return code. 

```suggestion
          ./test/test_kyber512
          ./test/test_kyber768
          ./test/test_kyber1024
```

mkannwischer (changes requested):

.github/workflows/build.yml
cothan: Actually, this is to make sure that when the internal checks are successful, then it will print out information. 

When error happen, return code is non 0, also the hash won't match. 

Return 0 -> print infor -> check hash -> success
Return non 0 -> check hash -> fail.

Makefile
cothan: added in https://github.com/pq-code-package/mlkem-c-aarch64/pull/29/commits/83a92e6dfbe34baee8a6fd25f7e202897ad6d27c

checksum.sh
cothan: added in https://github.com/pq-code-package/mlkem-c-aarch64/pull/29/commits/83a92e6dfbe34baee8a6fd25f7e202897ad6d27c

test/gen_KAT.c
cothan: added in https://github.com/pq-code-package/mlkem-c-aarch64/pull/29/commits/83a92e6dfbe34baee8a6fd25f7e202897ad6d27c

.github/workflows/build.yml
mkannwischer: I think that is confusing. Better to just check for the return code.

cothan: done in  https://github.com/pq-code-package/mlkem-c-aarch64/pull/29/commits/b76247fcc61c6b0d1322cacfe120603d104524ff

.github/workflows/build.yml
```diff
@@ -20,10 +21,31 @@ jobs:
           if [[ ${#err} != 0 ]]; then
             echo "$err" | while IFS= read -r file; do
               echo "::error file={"$file"},title={checking}::Formatted $file"
-            done
-            exit 1
+              done
+              exit 1
```
mkannwischer, line 25: what happened here?

.github/workflows/build.yml
```diff
@@ -6,6 +6,7 @@ on:
     branches: [ '*' ]
   pull_request:
     branches: [ "main" ]
+
```
mkannwischer, line 9:
```suggestion
```

test/nistrng/aes.c
mkannwischer, line 1: What do you think about using openssl for AES here? Then we do not have to include AES in the sources - but I am not sure how easy that is to get to work on MacOS.

test/nistrng/rng.c
```diff
@@ -0,0 +1,102 @@
+//
+//  rng.c
+//
+//  Created by Bassham, Lawrence E (Fed) on 8/29/17.
+//  Copyright © 2017 Bassham, Lawrence E (Fed). All rights reserved.
+//  Modified for liboqs by Douglas Stebila
+//
```
mkannwischer, line 7: I do not like this one at all. 
We'll have to replace that. We are working on that for https://github.com/pq-code-package/mlkem-c-embedded and I will get back to you once it's done so you can plug it in here.

mkannwischer (changes requested):

test/nistrng/rng.c
mkannwischer: We have a reimplemented version now here: https://github.com/pq-code-package/mlkem-c-embedded/pull/30 
Once that is merged you can use it here instead of the one you currently have.

.github/workflows/build.yml
cothan: eh, it's my linter reformat the file.

test/nistrng/aes.c
cothan: We can add OpenSSL here, I think it's a bit more complicated to support OpenSSL on Aarch64, now it's MacOS and Linux, in near future it's Android, iPhone and Windows ARM. 
One approach is to pull the AES file from somewhere to do the CI test, exclude the KAT test in the repo. This approach is preferable since it's can be shared between all ML-KEM code packages. 

What do you think?

test/nistrng/rng.c
cothan: done in https://github.com/pq-code-package/mlkem-c-aarch64/pull/29/commits/e7d6d8d4f7982aab1db7cda447ca69addb287d52

.github/workflows/build.yml
cothan: fixed in https://github.com/pq-code-package/mlkem-c-aarch64/pull/29/commits/8e876910eb78296ab887f26ae78af8b6e9cc26e6

test/nistrng/aes.c
mkannwischer: Okay, let's leave it in for now and try to eliminate it later.

mkannwischer (approved):
Thanks for making the changes. This looks good to me now.

Issue #34, opened 2024-05-10
Automate benchmarking and add to CI


hanno-becker:
Depends on: #28 

Acceptance criteria:
* Automated benchmarking of MLKEM-C-AArch64 implementations on the listed platforms (see [README](https://github.com/pq-code-package/mlkem-c-aarch64)) is available and can be run by maintainers and by CI. 
* Automated benchmarking is added to CI.

Steps:
- [x] Plan access to various benchmarking platforms. For some, the maintainers own boards with the right CPU, but we need to figure how we want to access them remotely from CI. For others, such as Graviton instances, we likely need to setup EC2 accounts.
- [x] Provide scripts for running benchmarks and preparing results
- [x] Add benchmarking to CI

planetf1:
A few related items
 - Github does have ARM based runners. Currently in private beta, expected to go public in July. (As an enterprise (PQCA) we can have access). There are currently costs involved, and a [request is open ](https://github.com/PQCA/TAC/issues/20)with PQCA TAC.
  - still not appropriate for benchmarking, and I'm not sure of what the actual processor support is
  - if funding is needed for other CI support, we can go through the TAC, but would need to get some estimates.
  - not sure how applicable above is, given this is a very hw specific area

hanno-becker:
@planetf1 Leaving aside the benchmarking platforms for which the maintainers have development boards, we would need budget for AWS EC2 instances (Graviton2, Graviton3, Apple M1) for benchmarking and CI. Is there precedent for how to get funding for this in PQCA/LF?

planetf1:
The process in general would be to raise the issue at the TSC (this becomes more relevant when there are more projects so we can consolidate & ensure there's awareness, but given the active projects are working together less of an issue).

Then we need to raise it with the PQCA TAC for general review, and they then either approve (if within existing budget) or raise a request with the governing board (if not).

Sounds like administrivia, but apart from the last step I think we can do quickly. Also it's my broad understanding, we're just getting started. I do join all the PQCA meetings 

Do you have an estimate on resource usage for the AWS EC2 instances ?
Any thoughts on how this may grow over time (so we plan for say a year) ?
Are there alternatives to EC2 (we may be asked)?
When do you need it (yesterday?)
What's the impact if you don't have it?

There's a PQCA TAC on Wed, so if we can get info together for that I can add an agenda item to the discussion

hanno-becker:
[...]

PR #38, opened 2024-05-16
Vectorize Rejection Sampling

cothan:

# Improve the performance of Rejection Sampling

I have not benchmark this code, this code is refactored and cleaned up from my previous implementation of Kyber.

It's seem like we need to switch to CircleCI to have ARM-based runner. 


cothan:
I tried to rewrite the sampler to meet the layout of Keccak. 
It's seem like we can't use the layout friendly to Keccak, this make the loading from sampling very difficult. 

The sample will need 3 bytes, to produce 2 coefficients. 

If we layout like this:

`0 - 1 - 2 -3 - 4 - 5 -6 - 7 | 8 - 9 - a - b - c - d - e - f`.

With `0 -> to 7` belongs to `keccak_way_1` and `8->f` belongs `keccak_way_2`, then it's not possible to align the byte we need:

Bytes: `0, 1, 2`, `3, 4, 5` create 4 coeffcients. Missing one byte added to `6, 7` to complete the sampling.

Similarly, `keccak_way_2` also needs one additional byte to complete the sampling. 

Thus, the memory layout of Keccak must be `All data of keccak_way_1 || All data of keccak_way_2 || ...` The interleave will make the sampler more difficult.

cothan:
We need ARM-runner so build_test can run.  #34

mkannwischer:
We can just cross compile and emulate using qemu. 
See for example here: https://github.com/slothy-optimizer/pqax/blob/main/.github/workflows/tests.yaml. 
Do you mind setting that up? We won't have Arm runners any time soon.

cothan:
I have a tight schedule this week. Perhaps next week I'll set it up.

hanno-becker:
@cothan @mkannwischer This PR reminds me that we should discuss intrinsics vs. ASM.

As hinted at in the README, my preference is to write 'clean' ASM and automatically micro-optimize it, e.g. using SLOTHY. The main reasons are: (1) _Assurance:_ There is more tooling for the verification of ASM than Intrinsics, at least to my knowledge. Moreover, verification can be split into separate verification of the clean code and the micro-optimizations. (2) _Performance:_ Past experience shows that the best implementations are obtained by writing ASM by hand. Moreover, by splitting development into manual writing of clean code and automatic micro-optimization, development time should not be much higher than when writing intrinsics (This may be false in general if there are a lot of stack spills, but this is not the case for MLKEM).

I would therefore propose that we do not use intrinsics but clean ASM throughout, including this PR. @cothan Would you be willing to port your intrinsics code to ASM? Might be a useful learning experience as well.

Thoughts?

cothan:
Hi @hanno-becker , 

Good idea. Originally this function was implemented in Intrinsic since it's quicker for me to debug in my VSCode. I Don't mind writing ASM at all. I will port this implementation to ASM. Do we have a Python template to generate ASM that I can take look at?

hanno-becker:
@cothan I tend to write ASM by hand, but making extensive use of `.req` and `.macro` directives to keep things modular and readable. We have plenty of examples in the [SLOTHY](https://github.com/slothy-optimizer/slothy) repo.

cothan:
Update: I have my ASM working and will update this PR very soon!

hanno-becker:
Awesome @cothan !

cothan:
Oh, with macos-latest I thought it will run on arm64. But it's not.

PR #44, opened 2024-05-29
Wrap up Nix configuration, setup arm64 github runner

potsrevennil:

I've specifically specified to install arm64 gcc in nix

I've set macos-latest as our GitHub runner for now, according to the [official document](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories), `macos-latest` is currently using M1 CPU with 7 GB RAM, which should help #38 for now. 

But as https://github.com/pq-code-package/mlkem-c-aarch64/issues/27 mentioned, we still would like a build system that support cross compilation and testing in qemu in the future

The `checksum.sh` script is moved under `scripts/` folder, and renamed as `checksum` for convenience. In the nix shell, `checksum` should be available instead of typing the whole path. 

In the ci action for setting up nix, I switched to use `DeterminateSystems/nix-installer-action` instead, as `nixbuild/nix-quick-install-action` currently doesn't support installing `nix` on macOS.

I've also added a summary for workflows, since I'm a bit lazy to jump into ci steps to see what's going on everytime :joy:

- **refactor nix configuration**
- **use macos-latest(arm64) as ci runner**
- **Add summary for github action**

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -12,32 +12,36 @@
     };
   };
 
-  outputs = inputs@{ flake-parts, nixpkgs, ... }:
+  outputs = inputs@{ flake-parts, ... }:
     flake-parts.lib.mkFlake { inherit inputs; } {
       imports = [ ];
       systems = [ "x86_64-linux" "aarch64-linux" "aarch64-darwin" "x86_64-darwin" ];
-      perSystem = { pkgs, system, inputs', ... }:
+      perSystem = { pkgs, ... }:
         let
-          core = with pkgs; [
-            # formatter & linters
-            astyle # 3.4.10
-            nixpkgs-fmt
-            shfmt
-          ];
+          core = builtins.attrValues {
+            inherit (pkgs)
+              gcc13
+
+              # formatter & linters
+              astyle# 3.4.10
+              nixpkgs-fmt
+              shfmt;
+          };
         in
         {
-          devShells.default = with pkgs; mkShellNoCC {
-            packages = core ++ [
-              direnv
-              nix-direnv
-            ];
+          devShells.default = pkgs.mkShellNoCC {
+            packages = core ++ builtins.attrValues {
+              inherit (pkgs)
+                direnv
+                nix-direnv;
+            };
 
             shellHook = ''
               export PATH=$PWD/scripts:$PWD/scripts/ci:$PATH
             '';
           };
 
-          devShells.ci = with pkgs; mkShellNoCC {
+          devShells.ci = pkgs.mkShellNoCC {
             packages = core;
             shellHook = ''
               export PATH=$PWD/scripts:$PWD/scripts/ci:$PATH
@@ -53,4 +57,3 @@
       };
     };
 }
-
--- a/flake.nix
+++ b/flake.nix
@@ -16,11 +16,8 @@
     flake-parts.lib.mkFlake { inherit inputs; } {
       imports = [ ];
       systems = [ "x86_64-linux" "aarch64-linux" "aarch64-darwin" "x86_64-darwin" ];
-      perSystem = { pkgs, system, ... }:
+      perSystem = { pkgs, ... }:
         let
-          # NOTE: For installing pkgs specific for aarch64 (not sure if there is any better way of doing this)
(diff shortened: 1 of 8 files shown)


hanno-becker (commented):
I am not familiar with nix, but superficially and from looking at the GitHub actions run on the PR itself, it looks good.

flake.nix
```diff
@@ -12,32 +12,40 @@
...
+      perSystem = { pkgs, system, ... }:
         let
-          core = with pkgs; [
-            # formatter & linters
-            astyle # 3.4.10
-            nixpkgs-fmt
-            shfmt
-          ];
+          # NOTE: For installing pkgs specific for aarch64 (not sure if there is any better way of doing this)
+          aarch64-system = "aarch64-" + pkgs.lib.lists.last (pkgs.lib.strings.splitString "-" system);
```
hanno-becker, line 22: Can you say a bit more about what's happening here? Why don't we use the native compiler?

potsrevennil: Hi, @hanno-becker thanks for the comment.
So the nix file `flake.nix` is just for specifying a deterministic development environment for developers using different machines here, including local development and CI. (it can be used as a build system as well, but that's not the case here). 

Developers or CI runners of this project might be using `x86-64` machines, which is why in here we are specifying that we support for developing and running the project on these machines (`x86_64-linux`, `aarch64-linux`, `aarch64-darwin`, `x86_64-darwin`). https://github.com/pq-code-package/mlkem-c-aarch64/blob/9f1bb4fa732a6e9b80fb63161266f8d26d246926/flake.nix#L18

So it will cause an issue if we didn't specify which compiler to use, as the nix shell of the `x86-64` machines will use the native compiler which is not the `arm64` compiler that we expected

potsrevennil: Actually `nix` also provide ways to achieve cross compilation, it would be cleaner I think, but I'm not yet familiar with this not sure if there is any caveat, and also it took so so long on my macOS..., so I gave up on this solution lol

potsrevennil: oh no... I think I'm wrong about something...
so the fix will not work on x86_64 machine... cross compile might be the solution
but there is this issue https://github.com/NixOS/nixpkgs/issues/244871..

I'll have to think how to work around this

potsrevennil: @hanno-becker @mkannwischer 

So... seems like, in nix, cross compile `gcc` is the idiomatic nix solution, but in nix the `gcc` toolchain is built from scratch instead of downloading some pre-built binary. 

While native `gcc` in nix is cached at `cache.nixos.org` this will not be an issue, but cross compile `gcc` is not cached.. which means we will need to build the whole gcc toolchain if doing this with nix, and this would be unacceptable in CI.

Instead I suggest we just fetch the `aarch64` `gcc` from https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads, then in CI it should be acceptable (but I'd like to do some experiment to make sure). Then we're left to choose between
1. use the gcc downloaded from the website for `x86-64` and `aarch64` machines
2. use the gcc downloaded from the website for `x86-64` and use nix native gcc for `aarch64` machines

I'm fine with both options, and would like to here from you.

Also, I'd suggest that, I'd would remove this confusing part https://github.com/pq-code-package/mlkem-c-aarch64/blob/9f1bb4fa732a6e9b80fb63161266f8d26d246926/flake.nix#L22-L23
just to make CI to work for `aarch64` machine in this PR, and solve the issue for `x86-64` machine in another PR (for example solve it together the emulation), so that we can merge this PR first.

hanno-becker: Using native gcc for Mac and aarch64-linux, and pre-built Arm cross compiler from https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads for x86 machines sounds good to me.

potsrevennil: Now I made nix to use native gcc provided by nix for `aarch64-linux`.
The rest of the systems will use the native gcc provided by the system, so `aarch64-darwin` will be using native apple clang.

https://github.com/pq-code-package/mlkem-c-aarch64/blob/2b8204f66805231fd99cdbe205383ebadf5f9e70/flake.nix#L30-L33

hanno-becker (approved):

PR #45, opened 2024-05-31
Use pre-built cross-compiled aarch64 gcc for x86-64 linux machine

potsrevennil:

Made x86-64 linux machine to use the pre-built corss-compile aarch64 gcc
Added x86-64 linux GitHub runner, it will only cross compile the project, emulation on qemu will be done in other PR

- **add arm gnu gcc nix spec for x86_64**
- **use the pre-built cross-compiled gcc for x86_64 machines in ci**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,9 +11,11 @@ jobs:
     strategy:
       matrix:
         system: [macos-latest, pqcp-arm64]
+        skipped: [false]
         include:
         - system: ubuntu-latest
           cross_prefix: aarch64-none-linux-gnu-
+          skipped: true
     runs-on: ${{ matrix.system }}
     steps:
       - uses: actions/checkout@v4
@@ -35,7 +37,7 @@ jobs:
         run: |
           make CROSS_PREFIX=${{ matrix.cross_prefix }} mlkem
 
-          if [ -z ${{ matrix.cross_prefix }} ]; then
+          if [ ${{ matrix.skipped }} != true ]; then
             ./test/test_kyber512
             ./test/test_kyber768
             ./test/test_kyber1024
@@ -49,7 +51,7 @@ jobs:
 
           make CROSS_PREFIX=${{ matrix.cross_prefix }} kat
 
-          if [ -z ${{ matrix.cross_prefix }} ]; then
+          if [ ${{ matrix.skipped }} != true ]; then
             cat >> $GITHUB_STEP_SUMMARY <<-EOF
               $(checksum ./test/gen_KAT512  ec4ac397e595ac7457cb7d8830921faf3290898a10d7dd3864aab89ea61fe9a3)
               $(checksum ./test/gen_KAT768  9a0826ad3c5232dfd3b21bc4801408655c565a491b760f509b2ee2cd7180babe)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   build_kat:
     strategy:
+      fail-fast: false
       matrix:
         include:
         - system: macos-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,10 +12,12 @@ jobs:
       matrix:
         system: [macos-latest, pqcp-arm64]
         skipped: [false]
+        expect_arch: [ arm64 ]
         include:
         - system: ubuntu-latest
           cross_prefix: aarch64-none-linux-gnu-
           skipped: true
+          expect_arch: x86_64
     name: build_kat (${{ matrix.system }})
     runs-on: ${{ matrix.system }}
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -8,14 +8,21 @@ on:
     branches: ["main"]
 jobs:
   build_test:
-    runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - os: ubuntu-latest
```
hanno-becker, line 15: We're implicitly relying on github runners for `ubuntu-latest` to be x86_64 here, right?

.github/workflows/build.yml
```diff
@@ -8,37 +8,72 @@ on:
...
-          ./checksum.sh ./test/gen_KAT512  ec4ac397e595ac7457cb7d8830921faf3290898a10d7dd3864aab89ea61fe9a3
-          ./checksum.sh ./test/gen_KAT768  9a0826ad3c5232dfd3b21bc4801408655c565a491b760f509b2ee2cd7180babe
-          ./checksum.sh ./test/gen_KAT1024 6dafb867599b750a6a831b03e494cf41dea748c78a0e275e7b268bbb893cf37d
+          cat >> $GITHUB_STEP_SUMMARY <<-EOF
+            ## KAT tests
+          EOF
+
+          make CROSS_PREFIX=${{ matrix.cross_prefix }} kat
+
+          if [ -z ${{ matrix.cross_prefix }} ]; then
```
hanno-becker, line 57: Perhaps add a `skipped` for now otherwise.

.github/workflows/build.yml
```diff
@@ -8,37 +8,72 @@ on:
...
-        run: |
-          make clean
+          cat >> $GITHUB_STEP_SUMMARY <<-EOF
+            ## NISTKAT tests
+          EOF
+
+          make CROSS_PREFIX=${{ matrix.cross_prefix }} nistkat
+
+          if [ -z ${{ matrix.cross_prefix }} ]; then
+            cat >> $GITHUB_STEP_SUMMARY <<-EOF
```
hanno-becker, line 74: Same here: Add `skipped` or similar until we run the tests in emulation.

cothan:
I would say it's easier to install zig and perform cross-compile. ` -target x86_64-native` or ` -target aarch64-linux` and zig can be installed easily.

.github/workflows/build.yml
potsrevennil: maybe we can check with `uname -m`, but what to do if it's not x86_64, just fail immediately ?

potsrevennil:
zig seems like a nice solution for this!

.github/workflows/build.yml
hanno-becker: Yes

hanno-becker:
@cothan 

> I would say it's easier to install zig and perform cross-compile.  -target x86_64-native or  -target aarch64-linux and zig can be installed easily.

What specifically would you suggest to be changed compared to the present state of the PR?

.github/workflows/ci.yml
```diff
@@ -11,14 +11,20 @@ jobs:
     strategy:
       matrix:
         system: [macos-latest, pqcp-arm64]
+        skipped: [false]
+        include:
+        - system: ubuntu-latest
```
hanno-becker, line 16: Just to be sure I understand the meaning of this: It means we add `ubuntu-latest` to the `system` array, and set `skipped` to `false` for `macos-latest` and `pqcp-arm64`, but to `true` to `ubuntu-latest`?

potsrevennil:
If we're using `zig` for cross compiling, other than updating the dependencies, adding something like this in the `Makefile` is all we need I think. (Then we can get rid of the `arm-gcc-gnu.nix` file that I added in this PR)

```
CC := zig cc
TARGET ?=

ifdef TARGET
        ZFLAGS += -target $(TARGET)
endif

...:
  $(CC) $(ZFLAGS) ...
```

Then in CI we will need to specify the target when cross compilation is needed (i.e. `make TARGET=aarch64-linux-gnu`)

hanno-becker (approved):

hanno-becker:
@potsrevennil Is this ready to go in?

potsrevennil:
@hanno-becker yes, if we ever want `zig` I can file another pr

.github/workflows/ci.yml
potsrevennil: yes, exactly

potsrevennil:
@hanno-becker yes, I think I can file another pr for `zig` if we ever want that

hanno-becker:
@potsrevennil Ready to merge once the DCO is fixed.

potsrevennil:
@hanno-becker yes, this is ready for merge, I've just updated to fail the job if the architecture is not as expected.

For `zig`, I think I can file another pr if we ever want to use `zig` for cross compiling.

hanno-becker:
Thank you @potsrevennil, merged.

Are you doing the follow-up adding qemu-based tests?

potsrevennil:
@hanno-becker Yes, I will do the follow-up qemu-based tests, but got stucked by other issue unfortunately, so maybe will not finish it very soon

PR #47, opened 2024-06-04
Add NEON assembly for rejection sampling

cothan:

Improve the performance of Rejection Sampling. 

I should re-arrange the instructions, but I think when we have SLOTHY figures our the best instruction sequences, so I left it as future work. 

Performance improvement: average 8%.


hanno-becker:
> I should re-arrange the instructions, but I think when we have SLOTHY figures our the best instruction sequences, so I left it as future work.

Yes, please write the code as clean as possible for now -- micro-optimization will be left to SLOTHY. (And if you have time, please feel free to try that out yourself)

mlkem/rej_uniform_asm.s
```diff
@@ -0,0 +1,268 @@
...
+*              - const uint8_t *buf:  pointer to input buffer
+*                                     (assumed to be uniform random bytes)
+*              - unsigned int buflen: length of input buffer in bytes
+*              - unsigned int *consumed: the length of consumed buffer
+* Returns number of sampled 16-bit integers (at most len)
+**************************************************/
+.align 4
+.global     rej_uniform_asm
+.global     _rej_uniform_asm
+rej_uniform_asm:
```
hanno-becker, line 32: Don't you need some preamble/postamble for stack stash/restore?

mlkem/rej_uniform.c
```diff
@@ -0,0 +1,369 @@
...
+*              - unsigned int *consumed: the length of consumed buffer
+* Returns number of sampled 16-bit integers (at most len)
+**************************************************/
+static unsigned int rej_uniform_large(int16_t *r, unsigned int len,
+                                     const uint8_t *buf, unsigned int buflen,
+                                     unsigned int *consumed) {
+
+    // Ensure alignment, the minimum stack size is (len + 32) * sizeof(int16_t)
+    // Since we don't want to use malloc/free, we assume the maximum len = KYBER_N and use stack.
+    int16_t stack[KYBER_N + 32] = {0};
```
hanno-becker, line 289: There's no need to do this here -- you can allocate stack as needed in the ASM?

cothan: You're right, I'm thinking of using `memcpy` to copy data from stack to `r`,  so I forgot to move the stack and memcpy to asm.

cothan: done in https://github.com/pq-code-package/mlkem-c-aarch64/pull/47/commits/3cf987be5dc19702a271ee94a7585fd202fde681

cothan:
Okay, we will need to integrate arm runner to run this code in CI.

hanno-becker:
@potsrevennil Didn't #44 add an Arm runner?

hanno-becker:
@cothan I see CI for https://github.com/pq-code-package/mlkem-c-aarch64/pull/47/commits/9ee3e95a2486218faffb882940094f7b4e082033 failing because of linting, not because of an architecture problem.

hanno-becker:
@cothan Can you mark the PR as DRAFT until you think it is ready to be merged, and clean up the history beforehand?

cothan:
Sure!

hanno-becker:
@cothan If you rebase on `main`, we should now be able to see separate CI jobs for macos-latest, pqcp-arm64, cbmc, and linting.

mlkem/rej_uniform_asm.s
hanno-becker, line 1: @mkannwischer @cothan 

I'd suggest we have a separate directory `mlkem/asm` for all assembly, and perhaps further split into `mlkem/asm/clean` and `mlkem/asm/opt`, where `clean` are the handwritten versions and `opt` will be the micro-optimized ones once we have them.

WDYT

cothan: I agree.

cothan: Done in https://github.com/pq-code-package/mlkem-c-aarch64/pull/47/commits/f5d283fbe620b53af0e4d8528160b83ebd88404d

mlkem/rej_uniform_asm.s
cothan: So I check, register from x8 to x17 can be used freely by callee as temporary register. Register x19-x28 must be reserved, I tried not to use x19-x28 so I think the current approach has no problem.

hanno-becker: You might need some more registers when scheduling the code through SLOTHY, but we can add the necessary prologue/epilogue then.

For now, can you add a comment / warning please?

cothan: Comment added in https://github.com/pq-code-package/mlkem-c-aarch64/pull/47/commits/f4ead9afa2b892eed146a027a4308c97e4771b9b. 
Thanks for the suggestion @hanno-becker !
I think at kernel level (if this work is adopted to kernel), people would use intra-procedure-call scratch registers `x16, x17`, thus I stick to strict ARM64 calling convention. `x8, x16, x17, x29, x30` registers are preserved.

mlkem/asm/clean/rej_uniform_asm.s
```diff
@@ -0,0 +1,369 @@
...
+    aligned_len                 .req x14
+    unaligned_len               .req x15
+
+    cmp     copylen, #0                     // If copylen is less than or equal to 0
+    b.le    end                             // then we exit the function
+    cmp     copylen, #32                    // Check if size is less than 32 bytes
+    b.lo    small_copy                      // Branch to small copy if size is less than 32
+
+    // Align destination address
+    mov     aligned_addr, dst
```
hanno-becker, line 35: Why don't we just make it a precondition that `dst` is suitably aligned?

mlkem/asm/clean/rej_uniform_asm.s
```diff
@@ -0,0 +1,369 @@
+// SPDX-License-Identifier: Apache-2.0
+
+/*************************************************
+* Name:        simd_memcpy
```
hanno-becker, line 4: I am skeptical that we should write this by hand -- compilers will utilize have a highly-optimized `memcpy` already. What's your motivation here?

mlkem/asm/clean/rej_uniform_asm.s
```diff
@@ -0,0 +1,369 @@
...
+    kyber_q                     .req v30
+    bits                        .req v31
+    bits_q                      .req q31
+
+
+    /* 
+     * Preserve registers and restore at the end of function.
+     * Stack alignment must be 16 bytes. 
+     */
+    sub     sp, sp, #0x270
```
hanno-becker, line 188: Can you leave a comment describing the usage and layout of the stack?

mlkem/asm/clean/rej_uniform_asm.s
```diff
@@ -0,0 +1,369 @@
...
+* Arguments:   - int16_t *r:          pointer to output buffer
+*              - unsigned int len:    requested number of 16-bit integers
+*                                     (uniform mod q)
+*              - const uint8_t *buf:  pointer to input buffer
+*                                     (assumed to be uniform random bytes)
+*              - unsigned int buflen: length of input buffer in bytes
+*              - unsigned int *consumed: the length of consumed buffer
+* Returns number of sampled 16-bit integers (at most len)
+**************************************************/
+.macro div48    dst, src
```
hanno-becker, line 91: This is not a secret division, is it? Can we not just use `div`?

mlkem/asm/clean/rej_uniform_asm.s
cothan: I couldn't make it compile and link to external memcpy. 
I tried `extern memcpy`, still, the compile won't make it compile. I stuck for a while and decided to write my version. 
Since the data is small, ~600 bytes, an indepdent memcpy could save some code space. 
Is there a way to ask compiler `extern memcpy`?

mlkem/asm/clean/rej_uniform_asm.s
cothan: Correct, it's not secret division, using `div` instruction is not optimized in this case, since we divide by constant. It's faster to multiple with inverse, isn't it?

mlkem/asm/clean/rej_uniform_asm.s
cothan: Sure, will do now.

mlkem/asm/clean/rej_uniform_asm.s
hanno-becker: I haven't yet read the core assembly below in detail but am surprised that we need a `memcpy` at all, actually.

cothan: Honestly, I don't want to write my own version of `memcpy`. I spent so long and I can't figure out why `extern memcpy` or `.extern memcpy` doesn't compile. If I find a way to link `memcpy` I would definitely remove my own.

mlkem/asm/clean/rej_uniform_asm.s
cothan: added in https://github.com/pq-code-package/mlkem-c-aarch64/pull/47/commits/908164c2534b4d6dc24e43cd1ad8cd6fe4ffcbd2

mlkem/asm/clean/rej_uniform_asm.s
hanno-becker: I stand corrected, compilers also don't use `div 48` here.

Alternatively, though, one can just use `sub _, _, #48` in the loop, rather than having an explicit iteration counter. But SLOTHY would not yet support that, so let's leave it.

hanno-becker: Compilers agree with you that the approach above is faster. Still, this is a one-off, so I don't think this is worth optimizing, given that it comes with some cognitive and verification overhead?

cothan: Okay, I'll use `div` instruction directly. Update soon.

cothan: done in https://github.com/pq-code-package/mlkem-c-aarch64/pull/47/commits/b42c7a17e06fd23922bdb6a3ac518cfe517a6e59

mlkem/asm/clean/rej_uniform_asm.s
cothan: okay, I figure it out, turn out on macos the symbol for `memcpy` is `_memcpy`, and linux is `memcpy`. 
I added a macro for to switch. Now it's worked! the `simd_memcpy` is removed. 
Commit history is cleaned.

cothan:
Okay, after I removed the `simd_memcpy` and replace with `memcpy`. It works on Mac. 
I can't wrap my head why the `arm64` version land on the `#ifdef __APPLE_` branch and take on `_memcpy` while on Linux. 
@.@ I need to investigate why the compiler on `arm64` land on `__APPLE__` branch.

cothan:
Okay, I know why. The assembly extension must be capital. `.S` instead of `.s`. :(( https://stackoverflow.com/a/30327226

cothan:
My M1 benchmark:

**Before:**

| Parameter           | MLKEM512 | MLKEM768 | MLKEM1024 |
|---------------------|----------|----------|-----------|
| keypair cycles      | 60678    | 103056   | 161138    |
| encaps cycles       | 75188    | 122977   | 184479    |
| decaps cycles       | 93538    | 146287   | 214105    |

**After:**

| Parameter           | MLKEM512 | MLKEM768 | MLKEM1024 |
|---------------------|----------|----------|-----------|
| keypair cycles      | 59284    | 99135    | 153797    |
| encaps cycles       | 73790    | 118616   | 176734    |
| decaps cycles       | 91968    | 141692   | 206045    |

cothan:
The number is much bigger (probably same ratio) compare on RPi 4. 

This branch: 
https://github.com/pq-code-package/mlkem-c-aarch64/actions/runs/9718493064?pr=47

Main branch: 
https://github.com/pq-code-package/mlkem-c-aarch64/actions/runs/9714693885

hanno-becker (changes requested):
@cothan @mkannwischer I would like all assembly in MLKEM-C-AArch64 be guarded by at least one compile-time flag, so we can easily switch it off, for example to test whether functional issues are related to ASM or not.

@cothan Could you consider where such a flag should live, introduce it, and guard the newly introduced ASM of this PR accordingly?

hanno-becker:
Apologies @cothan, I've just discussed this with @mkannwischer again. We believe that this work is low-priority and would rather slow us down on the path towards an initial release, seeing that we strive for functional verification of all assembly. We therefore feel that the PR should be put on hold until after the first release.

I'm sorry @cothan that we did not sort this out earlier.

cothan:
No problem. Thanks for letting me know.

mkannwischer:
Turning this into a draft for now.

hanno-becker:
@cothan I feel that now would be a good time to revive and update this pull request, rejection sampling being the last performance-relevant function that is not yet written in assembly. I don't know if/when we will manage to verify it, but if not, we may hide it behind an appropriate flag if needed.

Do you have time to work on this?

cothan:
@hanno-becker , yup, I have time to work on this PR, I would like to use this PR to try out SLOTHY as well.

hanno-becker:
@cothan Nice! Let's start with the clean version in this PR, and do an optimized one as a follow-up. You will need to add quite a few instructions to SLOTHY.

PR #51, opened 2024-06-09
Add NEON `poly_frommsg` and `poly_tomsg`

cothan:
Signed-off-by: Duc Tri Nguyen <dnguye69@gmu.edu>

Based on #47. Please review #47 first. 

There is a constant time problem with `poly_tomsg` and `poly_frommsg`. Thus, we either have to match the C implementation with the updated C code or write our own the vectorize constant time version. 

This PR implement vectorize constant time of `poly_frommsg` and `poly_tomsg`.


hanno-becker:
@cothan I generally think that only the most performance or security sensitive routines should be replaced by assembly -- it is an overhead in terms of work, verification and maintenance.

Have you measured the relative performance of `poly_frommsg` and `poly_tomsg` in the overall context of MLKEM?

As for porting https://github.com/pq-crystals/kyber/commit/9b8d30698a3e7449aeb34e62339d4176f11e3c6c, I'd be more inclined to have a small library of constant-time stubs like `cmov_int16` so we can be sure that the compiler does not fiddle with it.

@mkannwischer What are your thoughts?

cothan:
@hanno-becker , I have not measured the performance. I think `poly_frommsg` and `poly_tomsg` are insignificant in terms of performance. But at least we don't have to rely on compiler optimization. 
This implementation does not vectorize `cmov_int16`. Although I use a vectorize comparison instruction. I check, they have fixed latency (3 on Cortex A-72) and good throughput (2 on Cortex-A72).

mkannwischer:
> @mkannwischer What are your thoughts?

I agree. We should only introduce assembly when there is a good reason to do so. If the reason is protection against timing attacks, then we should use very tiny functions that we can all manually inspect to make sure they are correct. 
If the reason is performance, we should first perform some benchmarks. I don't recall `poly_frommsg` and `poly_tomsg` to be very expensive functions. If we decide to introduce assembly, we need to keep in mind that this will come with significant verification effort down the line.

I support adding a `cmov_int16` function in assembly for implementing `poly_frommsg`
For `poly_tomsg`, there now is robust C code that does not have the KyberSlash vulnerability. That can probably stay in C.

cothan:
I have implemented a constant-time version of `poly_frommsg` and `poly_tomsg`. According to ARM documentation, all the instructions used have fixed latency, which suggests they are constant time.

Regarding correctness, I'm unsure if CBMC can verify ASM. Essentially, the results can be checked by brute-forcing the entire `int16_t` space and comparing it with the C version. `poly_frommsg` maps `1 bit` to `int16_t`, and `poly_tomsg` maps `int16_t` to `1 bit`. The mapping is one-to-one, so brute-force checking is straightforward.

Reverting to `cmove_int6` is acceptable, but like other methods, it relies on the compiler, hoping it performs correctly. I don't know, fighting with compiler is the fight perhaps we never win. I believe developing our own version is a better approach.

hanno-becker:
> Reverting to cmove_int6 is acceptable, but like other methods, it relies on the compiler, hoping it performs correctly. I don't know, fighting with compiler is the fight perhaps we never win. I believe developing our own version is a better approach.

No, the idea is to provide C interfaces to critical stubs like `cmove_int16`, but then provide dedicated ASM for them to avoid any doubt about overly clever compilers.

I don't see a benefit of adding ASM for all of `poly_tomsg/poly_frommsg`.

PR #53, opened 2024-06-10
Introduce common coeff_signed_to_unsigned function in poly.c

rod-chapman:
This PR introduces a common "coeff_signed_to_unsigned()" function in poly.c, its contracts and verification thereof using CBMC. This function reduces repetition of the same idiom in poly.c and simplifies the calling code in poly_compress() and other functions.

Additionally, this PR adds contract-based proof of scalar_compress_q_16 and scalar_compress_q_32


mlkem/poly.c
```diff
@@ -82,14 +81,43 @@ __CPROVER_ensures(__CPROVER_return_value < 32)
...
+ *
+ * Arguments: c: signed coefficient to be converted
+ ************************************************************/
+uint16_t coeff_signed_to_unsigned (int16_t c)
+{
+    int32_t r = (int32_t) c;
+
+    // Add Q if r is negative
+    int32_t factor = (r < 0); // 1 if r < 0; 0 if r >= 0
+    r = r + (factor * KYBER_Q);
```
hanno-becker, line 112: I think it's not unlikely that a compiler would introduce a branch here. One way to thwart this is by calculating `r` and `r + KYBER_q` and having a `safe_cond_select()` stub which, based on either (a) a mask that's 0 or -1, or (b) an int that's 0 or 1, picks one of the two arguments. This function would itself be a C stub that's ideally replaced by ASM.

hanno-becker:
@rod-chapman You need DCO's for the CI to be happy.

rod-chapman:
What's a DCO?

mlkem/poly.c
rod-chapman: Agreed - we need a translation unit of safe constant-time functions like cond_select() and others. Disagree about the third parameter - it's a Boolean not a bleedin' Integer!

cothan: The multiplication `factor*KYBER_Q` you want will disappear when compiler optimize it. 
https://godbolt.org/z/xaY39sdWr 
So the compiler see our intentions and convert to:
`r += (c >> 15) & KYBER_Q` or `r += (c >> 31) & KYBER_Q`.
On other platform like AVR it's a branch condition. 
I think the new style and the old style are equivalent in the eye of compiler.

cothan:
Hi @rod-chapman , Thanks for the PR. the DCO is agreement to contribute, so it needs your signature. 
You need to sign off your commit `git commit -s -m "my commit"` with your GPG key. 
When you click to the Details it shows what you need to do: https://github.com/pq-code-package/mlkem-c-aarch64/pull/53/checks?check_run_id=26040102532

mlkem/poly.c
rod-chapman: The semantics of the >> operator on a signed, negative integer is implementation-defined (see WG14 N2310 6.5.7(5) for example), so I prefer not to use it. Strictly speaking, the use of >> here defies verification with CBMC.

hanno-becker:
@rod-chapman Yes, all good.

hanno-becker:
@rod-chapman Can you rebase your PR on `main`?

hanno-becker:
@rod-chapman Linting fails (check the formatting of `poly.[ch]`), as well as CBMC.

hanno-becker:
@rod-chapman CBMC seems to be failing, can you have a look? Also, can you rebase on `main` and cleanup the history?

cothan:
@rod-chapman , can you do the honor, test the "update with rebase" button?

rod-chapman:
Seemed to work OK!

hanno-becker:
It looks like this PR needs manual rebase after https://github.com/pq-code-package/mlkem-c-aarch64/pull/67 was merged separately.

rod-chapman:
I think this branch and PR are dead and subsumed by #67. Work on other functions can begin afresh.

hanno-becker:
This PR did also introduce a loop-invariant based proof of `poly_compress`, which I'd like to resurrect. However, after cleaning up the PR and rebasing onto `main`, CBMC fails on `poly_compress`, and I cannot seem to find legible descriptions of what is actually going wrong:
```
[15/17] poly_compress: checking safety properties                                                                                                                                                                     FAILED: ~/repos/mlkem-c-aarch64/cbmc/proofs/poly_compress/logs/result.xml ~/repos/mlkem-c-aarch64/cbmc/proofs/output/litani/runs/9c45a16a-0a7b-4254-b0a2-33390d2d7cc3/status/e87ddcc8-996c-4c98-9240-d6af84d2cd85.json
```

@rod-chapman Where would I find human-readable descriptions of the failure here? Can you reproduce the issue locally?

PR #54, opened 2024-06-12
Add CBMC as dependency to nix

mkannwischer:
Right now we are assuming that the user has installed:
 - CBMC: https://www.cprover.org/cbmc/
 - https://github.com/model-checking/cbmc-viewer
 - Litani: https://github.com/awslabs/aws-build-accumulator

This PR moves that into the nix dependencies, so the developer can just type `nix develop` to install it. 
It also allows us to make sure alle developers are running the same versions of these tools (and we can install updates when needed).


litani.nix
```diff
@@ -0,0 +1,35 @@
+# SPDX-License-Identifier: Apache-2.0
+
+{ lib
+, stdenvNoCC
+, fetchFromGitHub
+, targets ? [ ]
```
potsrevennil, line 6: this should be removed

litani.nix
```diff
@@ -0,0 +1,35 @@
...
+    repo = "aws-build-accumulator";
+    rev = "8002c240ef4f424039ed3cc32e076c0234d01768";
+    sha256 = "sha256-UwF/B6lpsjpQn8SW+tCfOXTp14pNBr2sRGujJH3iPLk=";
+  };
+  dontConfigure = true;
+  installPhase = ''
+    mkdir -p $out/litani/
+    cp  litani $out/litani/
+    cp -r lib $out/litani/
+    cp -r templates $out/litani/
```
potsrevennil, line 23: Not sure why symblink is needed here, I would suggest to do this instead, maybe we can get rid of unnecessary symblink
```
mkdir -p $out/bin
cp litani $out/bin
cp -r lib $out/
cp -r templates $out/
```

litani.nix
```diff
@@ -0,0 +1,35 @@
+# SPDX-License-Identifier: Apache-2.0
+
+{ lib
```
potsrevennil, line 3: This is not needed as well

flake.nix
```diff
@@ -18,11 +18,26 @@
...
+            };
+          });
           core = builtins.attrValues
             {
+              litani = litani; # 1.29.0
+              cbmc-viewer = cbmc-viewer; # 3.8
+              astyle = astyle;
+
               inherit (pkgs)
+                cbmc# 5.91.1
```
hanno-becker, line 37: I think we should do the work now to build `cbmc` from source, so we have an easier time updating it. E.g. `brew install cbmc` already gives you 5.95.

hanno-becker (changes requested):
I think we should build CBMC from source as well to have an easier time updating the version as the CBMC responds to issues we find (Rod has already reported a few things I think).

flake.nix
potsrevennil: After updating to nix channel 24.05, the version of cbmc should be `5.95.1` as well, so the comment here also be updated

mkannwischer: The version in nix is actually `5.95.1` - sorry about that. I corrected it now. 

Do you think we should still build from source?

mkannwischer:
> I think we should build CBMC from source as well to have an easier time updating the version as the CBMC responds to issues we find (Rod has already reported a few things I think).

@potsrevennil points out that this is going to be easy by using `pkgs.cbmc.overrideAttrs` with a new version or commit hash from the repo. You can ping us once you actually need a newer version and we can change it. We won't need to write our own build script. 

We've done something similar for astyle as the most recent version is buggy: https://github.com/mkannwischer/mlkem-c-aarch64/blob/d9f9300c52f1e38aea1a172916edf80aa9ca0eca/flake.nix#L23

hanno-becker:
> @potsrevennil points out that this is going to be easy by using pkgs.cbmc.overrideAttrs with a new version or commit hash from the repo. You can ping us once you actually need a newer version and we can change it. We won't need to write our own build script.

Sounds great. Can you still set this up with 5.95.1 (obtained from the repo) for now?

mkannwischer:
It is `5.95.1` (actually it always was - the comment was just wrong)

hanno-becker:
@mkannwischer Yes, I understand, but I was wondering if you could still set things up in such a way that one can specify version or hash from the repository.

hanno-becker (approved):

PR #56, opened 2024-06-13
Build CBMC from source and add to nix CI cache

mkannwischer:
This PR is building CBMC from sources. Right now it's pinned to the `cbmc-5.95.1` tag (https://github.com/diffblue/cbmc/commit/731338d5d82ac86fc447015e0bd24cdf7a74c442), but we can update it to other commits when needed. 

The build requires around 10 minutes on my machine and in CI.

To make this feasible in the CI, the dependency builds are cached in the Github Action Cache using https://github.com/DeterminateSystems/magic-nix-cache. This means it should only take 10 minutes once, and then consecutive builds should be as fast as loading it from the package manager. 
I tested here by re-running the CI jobs and it seems to work fine.
If we see any CI performance degradation in other PRs, I'll need to look more into the configuration of https://github.com/DeterminateSystems/magic-nix-cache

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,8 +31,8 @@
             version = "731338d5d82ac86fc447015e0bd24cdf7a74c442";
             src = pkgs.fetchFromGitHub {
               owner = "diffblue";
-              repo = "cbmc";
-              rev = "731338d5d82ac86fc447015e0bd24cdf7a74c442";
+              repo = old.pname;
+              rev = "${version}";
               hash = "sha256-fDLSo5EeHyPTliAqFp+5mfaB0iZXIMXeMyF21fjl5k4=";
             };
           });
--- a/flake.nix
+++ b/flake.nix
@@ -27,14 +27,24 @@
               hash = "sha256-eKYQq9OelOD5E+nuXNoehbtizWM1U97LngDT2SAQGc4=";
             };
           });
+          cbmc = pkgs.cbmc.overrideAttrs (old: rec {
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -27,14 +27,24 @@
               hash = "sha256-eKYQq9OelOD5E+nuXNoehbtizWM1U97LngDT2SAQGc4=";
             };
           });
+          cbmc = pkgs.cbmc.overrideAttrs (old: rec {
+            version = "731338d5d82ac86fc447015e0bd24cdf7a74c442";
+            src = pkgs.fetchFromGitHub {
+              owner = "diffblue";
+              repo = "cbmc";
```
potsrevennil, line 34: I think you can change this to
```
repo = old.pname;
```

potsrevennil (commented):
The rest of the part looks good to me, this is nice!

flake.nix
```diff
@@ -27,14 +27,24 @@
...
+              rev = "731338d5d82ac86fc447015e0bd24cdf7a74c442";
```
potsrevennil, line 35: I think you can also reference version by ${version} here

flake.nix
mkannwischer: Done

flake.nix
mkannwischer: Changed. Thanks

potsrevennil (approved):

hanno-becker:
@mkannwischer Did you re-run the CI to see if it speeds up after initial build?

hanno-becker (approved):

PR #57, opened 2024-06-13
Use ubuntu-latest for lint job instead

potsrevennil:

I think it makes more sense to run a normal job like linting on standard GitHub-hosted runner instead of a runner for special purpose like `pqcp-arm64`.


mkannwischer (approved):
Thanks @potsrevennil. 
It makes sense to run as many of the CI tasks on the public Github runners as possible, so that they can run in forks (that don't have access to the pqcp runnes) prior to opening a PR.

PR #60, opened 2024-06-13
Emulate arm64 binary on x86_64 machine

potsrevennil:

I added `scripts/tests` (should be able to be executed by simply typing `tests` in nix shell) for functional, kat and nistkat tests.

The script currently automatically detects the host system and decides to run natively on ARM64 machines or emulate on QEMU for x86_64 machines. We might want to add the capability to manually specify whether to run tests natively or on QEMU in the future. However, I don't think we need this functionality for now.

Usage of the script is as follow:

```
> tests --help
Usage: tests [OPTIONS] COMMAND [ARGS]...

Options:
  --help  Show this message and exit.

Commands:
  func     Run the functional tests for all parameter sets
  kat      Run the kat tests for all parameter sets
  nistkat  Run the nistkat tests for all parameter sets
  run      Run the specified binary file

> tests nistkat --help
Usage: tests nistkat [OPTIONS]

Options:
  -v, --verbose  Show verbose output or not
  --help         Show this message and exit.
```

I have also added `black` as the Python formatter and updated the `format` and `lint` scripts to utilize it for checking the format of Python scripts. I don't have a strong opinion on the choice of Python formatter, so if there are any better suggestions, I am open to using them.

I made CI continues even when some part of the tests failed as @hanno-becker suggested in https://github.com/pq-code-package/mlkem-c-aarch64/issues/48 (sorry about that I somehow missed the message last week).

- **emulate arm64 binary on x86_64 machine**
- **Output binaries to bin/ directory**
- **add nistkat checksum back**
- **use black for python formatting**
- **add tests script for running a specify binary**
- **factor out the common part of compiling and running a binary**
- **add test command for func, nistkat and kat tests**
- **compare tests result to the META file**
- **add github step summary to tests script if the env var exists**
- **update ci to use the tests script**
- **cleanup**
- **only add -static if is linux x86_64**
- **still run kat/nistkat tests even if previous tests failed**


cothan (approved):

.gitignore
```diff
@@ -4,12 +4,4 @@
...
-test/test_kyber512
-test/test_kyber768
-test/test_kyber1024
-test/gen_KAT512
-test/gen_KAT768
-test/gen_KAT1024
-test/gen_NISTKAT512
-test/gen_NISTKAT768
-test/gen_NISTKAT1024
+bin/
```
hanno-becker, line 7: Any reason in particular why this was renamed? I think `test` is more suitable than `bin`.

hanno-becker (changes requested):
I understand that we want developers to use `nix`, but I think a bare `make` should work out of the box on standard environments -- just to satisfy people who are curious about the repo and want to give it a very quick try.

Can we keep `make` working outside of the `nix` environment @potsrevennil ?

potsrevennil:
@hanno-becker I'm not sure if I understand you correctly

`nix` is only used for specifying dependencies here, so `make` can definitely work outside of the `nix` shell. However, users will need to install all the dependencies themselves, and how they do this is out of scope for this PR, I think. For different users and different machines, there might be many ways to install all the dependencies, and nix is provided especially for those who want to give a quick try. If someone wants to do this without `nix`, there's not much we can do, but as long as they have `qemu` and `gcc` pre-installed, building with make should work.

To be more clear, there are two external dependencies for `make`: `qemu` and `gcc`. I assumed these dependencies should be available in the user's shell environment, though the shell does not need to be a `nix` shell. Users can make these dependencies available in any way they like if they are not using `nix`, or overwrite the Makefile variables via command line arguments.

.gitignore
potsrevennil: Mainly because I personally don't like having those binaries scattered in a folder alongside some C files. I prefer to keep those binaries under a separate directory. 

But I don't have a strong prefer over the folder name, or the folder location. I think I just followed some folder structure as `pqm4`.
And if we prefer to keep things under `test`, then I would prefer to keep them under `test/bin`.

hanno-becker:
@potsrevennil Ok, I'll have another look. When I try `make` outside of the `nix` shell, I get 
```
% make
make: *** No rule to make target `bin/test_kyber768', needed by `mlkem'.  Stop.
```

.gitignore
hanno-becker: `test/bin` sounds good to me.

potsrevennil:
Oh, this is not what I intended...

It seems like a incompatibility issue of different version of `make`, which version of `make` are you using ?

I tested on my macOS, under x86_64 environment I got `GNU Make 3.81 ... built for i386-apple-darwin11.3.0`, and it result in the same issue as your,

While under `aarch64` environment of my macOS and `nix` shell, I have `GNU Make 4.4.1 Built for aarch64-apple-darwin23.0.0` and `GNU Make 4.4.1 Built for aarch64-apple-darwin23.4.0` respectively, which worked just fine.

I pushed a new version should resolve this issue, but maybe we would also want to fix the version of `make` that we are using ?

hanno-becker:
@potsrevennil Can we remove the macros, even if it means a bit of repetition between the makefile targets?

potsrevennil:
@hanno-becker Done, please have a look again, thanks

scripts/tests
```diff
@@ -0,0 +1,240 @@
...
+
+
+def sha256sum(result):
+    m = hashlib.sha256()
+    m.update(result)
+    return m.hexdigest()
+
+
+def base_run(bin, verbose):
+    if platform.system() == "Linux" and platform.machine() == "x86_64":
```
hanno-becker, line 29: @potsrevennil The qemu-based testing is supposed to work even on AArch64 systems, right?

Can you perhaps add a `--force-qemu` or similar argument to `tests` so we can test with qemu on those systems, too?

hanno-becker (commented):
Minor comment regarding qemu on AArch64 systems.

scripts/tests
potsrevennil: I expect qemu-based testing to work on `Linux AArch64` systems.
 
They will not work on `macOS` at least for now, as `QEMU` doesn't provide `qemu-aarch64` binary for `macOS`. There is `qemu-system-aarch64` available on macOS, but I don't whether macOS can utilize this to emulate or not.

Therefore, I'll make the `--force-qemu` flag be blocked on `macOS` for now.

hanno-becker: Ok, thanks for the clarification!

hanno-becker (approved):

scripts/tests
```diff
@@ -0,0 +1,261 @@
...
+    return m.hexdigest()
+
+
+def base_run(bin, force_qemu, verbose):
+    if force_qemu or (platform.system() == "Linux" and platform.machine() == "x86_64"):
+        logging.debug(f"Emulating {bin} with QEMU")
+
+        p = subprocess.run(
+            ["make", "CROSS_PREFIX=aarch64-none-linux-gnu-", f"{bin}"],
+            stdout=subprocess.DEVNULL if not verbose else None,
```
hanno-becker, line 34: One more thing: Can we please keep the output of `make` in the CI? If something goes wrong, it's good to know what exact commands were attempted.

hanno-becker (changes requested):
Mostly LGTM, but can we keep the `make` output in the CI logs?

scripts/tests
potsrevennil: oh, yes good suggestion!

hanno-becker:
@potsrevennil 

Currently, I see the following in the CI log:
```
DEBUG > Emulating test/bin/gen_KAT512 with QEMU
  CC      
mlkem/poly.c: In function ‘pqcrystals_kyber512_ref_scalar_compress_q_16’:
mlkem/poly.c:27: warning: ignoring ‘#pragma CPROVER check’ [-Wunknown-pragmas]
   27 | #pragma CPROVER check push
```

Is there a way to show the full command line?

potsrevennil:
Ah, yes! Let me see how to do it

scripts/tests
```diff
@@ -46,10 +49,14 @@ def base_run(bin, force_qemu, verbose):
     else:
         logging.debug(f"Running {bin} natively")
 
+        args = ["make", f"{bin}"]
+        logging.info(' '.join(args))
+
         p = subprocess.run(
-            ["make", f"{bin}"],
+            args,
             stdout=subprocess.DEVNULL if not verbose else None,
```
hanno-becker, line 57: Is `verbose` set in the CI?

potsrevennil: yes, it is set in this commit https://github.com/pq-code-package/mlkem-c-aarch64/pull/60/commits/8048dd7d917cb82ec6a14b5dc6f84192710c66bf

hanno-becker: Hm ok -- why is it then that the CC command line is missing from 
```
INFO  > make CROSS_PREFIX=aarch64-none-linux-gnu- test/bin/gen_KAT1024
  CC      
mlkem/poly.c: In function ‘pqcrystals_kyber1024_ref_scalar_compress_q_16’:
mlkem/poly.c:27: warning: ignoring ‘#pragma CPROVER check’ [-Wunknown-pragmas]
   27 | #pragma CPROVER check push
      | 
mlkem/poly.c:28: warning: ignoring ‘#pragma CPROVER check’ [-Wunknown-pragmas]
   28 | #pragma CPROVER check disable "unsigned-overflow"
      | 
mlkem/poly.c:30: warning: ignoring ‘#pragma CPROVER check’ [-Wunknown-pragmas]
   30 | #pragma CPROVER check pop
      | 
mlkem/poly.c: In function ‘pqcrystals_kyber1024_ref_scalar_compress_q_32’:
mlkem/poly.c:[64](https://github.com/pq-code-package/mlkem-c-aarch64/actions/runs/9543519253/job/26300276213?pr=60#step:5:65): warning: ignoring ‘#pragma CPROVER check’ [-Wunknown-pragmas]
   64 | #pragma CPROVER check push
      | 
mlkem/poly.c:[65](https://github.com/pq-code-package/mlkem-c-aarch64/actions/runs/9543519253/job/26300276213?pr=60#step:5:66): warning: ignoring ‘#pragma CPROVER check’ [-Wunknown-pragmas]
   65 | #pragma CPROVER check disable "unsigned-overflow"
      | 
mlkem/poly.c:[67](https://github.com/pq-code-package/mlkem-c-aarch64/actions/runs/9543519253/job/26300276213?pr=60#step:5:68): warning: ignoring ‘#pragma CPROVER check’ [-Wunknown-pragmas]
   67 | #pragma CPROVER check pop
```

potsrevennil: Ah, thanks for the catch! It's because the extra quoting that I forgot to remove while removing the makefile macros.

hanno-becker: Thank you @potsrevennil, I can see the command now:
```
INFO  > make CROSS_PREFIX=aarch64-none-linux-gnu- test/bin/gen_KAT512
  CC      test/bin/gen_KAT512
mlkem/poly.c: In function ‘pqcrystals_kyber512_ref_scalar_compress_q_16’:
mlkem/poly.c:27: warning: ignoring ‘#pragma CPROVER check’ [-Wunknown-pragmas]
   27 | #pragma CPROVER check push
      | 
mlkem/poly.c:28: warning: ignoring ‘#pragma CPROVER check’ [-Wunknown-pragmas]
   28 | #pragma CPROVER check disable "unsigned-overflow"
      | 
mlkem/poly.c:30: warning: ignoring ‘#pragma CPROVER check’ [-Wunknown-pragmas]
   30 | #pragma CPROVER check pop
      | 
mlkem/poly.c: In function ‘pqcrystals_kyber5[12](https://github.com/pq-code-package/mlkem-c-aarch64/actions/runs/9543644196/job/26300636989?pr=60#step:5:13)_ref_scalar_compress_q_32’:
mlkem/poly.c:64: warning: ignoring ‘#pragma CPROVER check’ [-Wunknown-pragmas]
   64 | #pragma CPROVER check push
      | 
mlkem/poly.c:65: warning: ignoring ‘#pragma CPROVER check’ [-Wunknown-pragmas]
   65 | #pragma CPROVER check disable "unsigned-overflow"
      | 
mlkem/poly.c:67: warning: ignoring ‘#pragma CPROVER check’ [-Wunknown-pragmas]
   67 | #pragma CPROVER check pop
      |
```

Can one even go further and show the exact commands that were run?

potsrevennil: The exact command that were run is printed at the first line starting with the `INFO > ` prompt. Is this what you want ?

hanno-becker: No, I mean the commands that `make` runs

potsrevennil: ok, I removed some silent quotes, now it will not suppress the commands that `make` run

hanno-becker: Thank you!

hanno-becker (approved):
Thanks a lot @potsrevennil, LGTM

PR #62, opened 2024-06-13
Add Keccak X4 interface

### LINKED ISSUE #35: Add internal interface to 4-way Keccak and implement C-fallback
Partly implements: #33 

Acceptance criteria:
- [ ] The reference implementation is modified to leverage an internal C interface to a 4-way parallel Keccak permutation.
- [ ] The interface is implemented by fallback to the standard 1-way Keccak permutation in C.

Adding AArch64 assembly for 4-way parallel Keccak is _not_ part of this issue.

cothan:
# Add Keccak X4 interface. 

Potential optimization: 

- **Multiple Pass Squeezing**: Currently, if `rej_uniform` fails to sample the full vector size, it continues by squeezing a single Keccak lane. By rewriting this logic to append single Keccak calls into a single Keccak X-way call, CPU cycles can be saved. For simplicity, it is currently implemented as a single pass.

### Next Steps:

- Add Keccak_X4 assembly to replace the `keccak_absorb_x4keccak_squeezeblocks_x4` in `fips202x.c`.

Fixes #35

### DIFF
--- a/fips202/fips202x4.c
+++ b/fips202/fips202x4.c
@@ -1,0 +1,201 @@
+// SPDX-License-Identifier: Apache-2.0
+#include <string.h>
+#include "fips202x4.h"
+#include "fips202.h"
+#include "keccakf1600.h"
+
+#define KECCAK_CTX 25
+
+static void keccak_absorb_x4(uint64_t *s, uint32_t r,
+                             const uint8_t *in0,
+                             const uint8_t *in1,
+                             const uint8_t *in2,
+                             const uint8_t *in3,
+                             size_t inlen,
+                             uint8_t p)
+{
+
+    while (inlen >= r)
+    {
+
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 0, in0, 0, r);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 1, in1, 0, r);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 2, in2, 0, r);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 3, in3, 0, r);
+
+        KeccakF1600_StatePermute(s + KECCAK_CTX * 0);
+        KeccakF1600_StatePermute(s + KECCAK_CTX * 1);
+        KeccakF1600_StatePermute(s + KECCAK_CTX * 2);
+        KeccakF1600_StatePermute(s + KECCAK_CTX * 3);
+
+        in0 += r;
+        in1 += r;
+        in2 += r;
+        in3 += r;
+        inlen -= r;
+    }
+
+    if (inlen > 0)
+    {
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 0, in0, 0, inlen);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 1, in1, 0, inlen);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 2, in2, 0, inlen);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 3, in3, 0, inlen);
+    }
+
+    if (inlen == r - 1)
+    {
+        p |= 128;
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 0, &p, inlen, 1);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 1, &p, inlen, 1);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 2, &p, inlen, 1);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 3, &p, inlen, 1);
+    }
+    else
+    {
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 0, &p, inlen, 1);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 1, &p, inlen, 1);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 2, &p, inlen, 1);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 3, &p, inlen, 1);
+        p = 128;
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 0, &p, r - 1, 1);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 1, &p, r - 1, 1);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 2, &p, r - 1, 1);
+        KeccakF1600_StateXORBytes(s + KECCAK_CTX * 3, &p, r - 1, 1);
+    }
+}
+
+static void keccak_squeezeblocks_x4(uint8_t *out0,
+                                    uint8_t *out1,
+                                    uint8_t *out2,
+                                    uint8_t *out3,
+                                    size_t nblocks,
+                                    uint64_t *s,
+                                    uint32_t r)
+{
+
+    while (nblocks > 0)
+    {
+        KeccakF1600_StatePermute(s + KECCAK_CTX * 0);
+        KeccakF1600_StatePermute(s + KECCAK_CTX * 1);
+        KeccakF1600_StatePermute(s + KECCAK_CTX * 2);
+        KeccakF1600_StatePermute(s + KECCAK_CTX * 3);
+
+        KeccakF1600_StateExtractBytes(s + KECCAK_CTX * 0, out0, 0, r);
+        KeccakF1600_StateExtractBytes(s + KECCAK_CTX * 1, out1, 0, r);
+        KeccakF1600_StateExtractBytes(s + KECCAK_CTX * 2, out2, 0, r);
+        KeccakF1600_StateExtractBytes(s + KECCAK_CTX * 3, out3, 0, r);
+
+        out0 += r;
+        out1 += r;
+        out2 += r;
+        out3 += r;
+        nblocks--;
+    }
+}
+
+uint64_t *keccakx_get_lane_state(keccakx4_state *state, size_t index)
+{
+    if (index >= KECCAK_WAY)
+    {
+        return NULL;
+    }
+
+    return state->ctx + index * KECCAK_CTX;
+}
+
+void shake128x4_absorb(keccakx4_state *state,
+                       const uint8_t *in0,
+                       const uint8_t *in1,
+                       const uint8_t *in2,
+                       const uint8_t *in3,
+                       size_t inlen)
+{
+    memset(state->ctx, 0, sizeof(state->ctx));
+
+    keccak_absorb_x4(state->ctx, SHAKE128_RATE, in0, in1, in2, in3, inlen, 0x1F);
+}
+
+void shake256x4_absorb(keccakx4_state *state,
+                       const uint8_t *in0,
+                       const uint8_t *in1,
+                       const uint8_t *in2,
+                       const uint8_t *in3,
+                       size_t inlen)
+{
+    memset(state->ctx, 0, sizeof(state->ctx));
+
+    keccak_absorb_x4(state->ctx, SHAKE256_RATE, in0, in1, in2, in3, inlen, 0x1F);
+}
+
+
+void shake128x4_squeezeblocks(uint8_t *out0,
+                              uint8_t *out1,
+                              uint8_t *out2,
+                              uint8_t *out3,
+                              size_t nblocks,
+                              keccakx4_state *state)
+{
+    keccak_squeezeblocks_x4(out0, out1, out2, out3, nblocks, state->ctx, SHAKE128_RATE);
+}
+
+void shake256x4_squeezeblocks(uint8_t *out0,
+                              uint8_t *out1,
+                              uint8_t *out2,
+                              uint8_t *out3,
+                              size_t nblocks,
+                              keccakx4_state *state)
+{
+    keccak_squeezeblocks_x4(out0, out1, out2, out3, nblocks, state->ctx, SHAKE256_RATE);
+}
+
+void shake256x4_squeezeblocks_single(uint8_t *out,
+                                     size_t nblocks,
+                                     size_t index,
+                                     keccakx4_state *state)
+{
+    uint64_t *ctx = keccakx_get_lane_state(state, index);
+    while (nblocks > 0)
+    {
+        KeccakF1600_StatePermute(ctx);
+        KeccakF1600_StateExtractBytes(ctx, out, 0, SHAKE128_RATE);
+        out += SHAKE128_RATE;
+        nblocks--;
+    }
+}
+
+void shake256x4(uint8_t *out0,
+                uint8_t *out1,
+                uint8_t *out2,
+                uint8_t *out3,
+                size_t outlen,
+                uint8_t *in0,
+                uint8_t *in1,
+                uint8_t *in2,
+                uint8_t *in3,
+                size_t inlen)
+{
+    keccakx4_state statex;
+    size_t nblocks = outlen/SHAKE256_RATE;
+    uint8_t tmp[KECCAK_WAY][SHAKE256_RATE];
+
+    shake256x4_absorb(&statex, in0, in1, in2, in3, inlen);
+
+    shake256x4_squeezeblocks(out0, out1, out2, out3, nblocks, &statex);
+
+    out0 += nblocks * SHAKE256_RATE;
+    out1 += nblocks * SHAKE256_RATE;
+    out2 += nblocks * SHAKE256_RATE;
+    out3 += nblocks * SHAKE256_RATE;
+
+    outlen -= nblocks * SHAKE256_RATE;
+
+    if (outlen)
+    {
+        shake256x4_squeezeblocks(tmp[0], tmp[1], tmp[2], tmp[3], 1, &statex);
+        memcpy(out0, tmp[0], outlen);
+        memcpy(out1, tmp[1], outlen);
+        memcpy(out2, tmp[2], outlen);
+        memcpy(out3, tmp[3], outlen);
+    }
+}
--- a/fips202/fips202x4.h
+++ b/fips202/fips202x4.h
@@ -1,0 +1,66 @@
+// SPDX-License-Identifier: Apache-2.0
+#ifndef FIPS_202X4_H
+#define FIPS_202X4_H
+
+#ifndef KECCAK_WAY
+#define KECCAK_WAY 4
+#endif
+
+#include <stdint.h>
+
+typedef struct
+{
+    uint64_t ctx[25 * KECCAK_WAY];
+} keccakx4_state;
+
+uint64_t *keccakx_get_lane_state(keccakx4_state *state, size_t index);
+
+void shake128x4_absorb(keccakx4_state *state,
+                       const uint8_t *in0,
+                       const uint8_t *in1,
+                       const uint8_t *in2,
+                       const uint8_t *in3,
+                       size_t inlen);
+
+void shake256x4_absorb(keccakx4_state *state,
+                       const uint8_t *in0,
+                       const uint8_t *in1,
+                       const uint8_t *in2,
+                       const uint8_t *in3,
+                       size_t inlen);
+
+
+void shake128x4_squeezeblocks(uint8_t *out0,
+                              uint8_t *out1,
+                              uint8_t *out2,
+                              uint8_t *out3,
+                              size_t nblocks,
+                              keccakx4_state *state);
+
+void shake256x4_squeezeblocks(uint8_t *out0,
+                              uint8_t *out1,
+                              uint8_t *out2,
+                              uint8_t *out3,
+                              size_t nblocks,
+                              keccakx4_state *state);
+
+/*
+ * Squeezes a single lane in Keccak 4-way
+ */
+void shake256x4_squeezeblocks_single(uint8_t *out,
+                                     size_t nblocks,
+                                     size_t index,
+                                     keccakx4_state *state);
+
+void shake256x4(uint8_t *out0,
+                uint8_t *out1,
+                uint8_t *out2,
+                uint8_t *out3,
+                size_t outlen,
+                uint8_t *in0,
+                uint8_t *in1,
+                uint8_t *in2,
+                uint8_t *in3,
+                size_t inlen);
+
+#endif
--- a/mlkem/indcpa.c
+++ b/mlkem/indcpa.c
@@ -9,6 +9,8 @@
 #include "ntt.h"
 #include "symmetric.h"
 #include "randombytes.h"
+#include "fips202x4.h"
+#include "fips202.h"
 
 /*************************************************
 * Name:        pack_pk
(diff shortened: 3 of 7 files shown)


mlkem/indcpa.c
```diff
@@ -155,34 +157,69 @@ static unsigned int rej_uniform(int16_t *r,
 #define GEN_MATRIX_NBLOCKS ((12*KYBER_N/8*(1 << 12)/KYBER_Q + XOF_BLOCKBYTES)/XOF_BLOCKBYTES)
 // Not static for benchmarking
 void gen_matrix(polyvec *a, const uint8_t seed[KYBER_SYMBYTES], int transposed) {
-    unsigned int ctr, i, j, k;
-    unsigned int buflen, off;
-    uint8_t buf[GEN_MATRIX_NBLOCKS * XOF_BLOCKBYTES + 2];
+    unsigned int ctr, i, j;
+    unsigned int buflen;
+    uint8_t bufx[KECCAK_WAY][GEN_MATRIX_NBLOCKS *XOF_BLOCKBYTES + 2];
```
cothan, line 162: I have no idea why the buffer need to `+2`. Does anyone know why?

mlkem/indcpa.c
```diff
@@ -155,34 +157,69 @@ static unsigned int rej_uniform(int16_t *r,
...
+    uint8_t *buf = NULL;
+    int16_t *vec = NULL;
     xof_state state;
 
-    for (i = 0; i < KYBER_K; i++) {
-        for (j = 0; j < KYBER_K; j++) {
+    keccakx kx;
+    uint8_t seedxy[KECCAK_WAY][KYBER_SYMBYTES + 2];
```
cothan, line 168: The size of `KYBER_SYMBTES + 2` is `34`, which creates unaligned load from stack for each lane. 
I suggest make it to `KYBER_SYMBYTES + 16`. Wdyt?

mlkem/indcpa.c
```diff
@@ -155,34 +157,69 @@ static unsigned int rej_uniform(int16_t *r,
...
             }
 
-            xof_squeezeblocks(buf, GEN_MATRIX_NBLOCKS, &state);
-            buflen = GEN_MATRIX_NBLOCKS * XOF_BLOCKBYTES;
-            ctr = rej_uniform(a[i].vec[j].coeffs, KYBER_N, buf, buflen);
-
-            while (ctr < KYBER_N) {
-                off = buflen % 3;
-                for (k = 0; k < off; k++) {
-                    buf[k] = buf[buflen - off + k];
```
cothan, line 178: offset related are completely removed. This line can't be reached. See 
https://github.com/pq-crystals/kyber/issues/74

fips202/fips202x.h
```diff
@@ -0,0 +1,115 @@
...
+#define FIPS_202X_H
+
+#ifndef KECCAK_WAY
+#define KECCAK_WAY 7
+#endif
+
+#include <stdint.h>
+#include <stddef.h>
+
+typedef struct
```
hanno-becker, line 12: Why do we need this?

fips202/fips202x.h
```diff
@@ -0,0 +1,114 @@
+#ifndef FIPS_202X_H
+#define FIPS_202X_H
+
+#ifndef KECCAK_WAY
+#define KECCAK_WAY 7
```
hanno-becker, line 6: Why 7?

fips202/fips202x.h
cothan: the compiler warning that it doesn't understand `uint64_t` (added `stdint`) and `size_t` (added `stddef`).

fips202/fips202x.h
cothan: Ah, I tested with different number of KECCAK_WAY and 7 is the last number in the code. It works with any KECCAK_WAY.

mlkem/indcpa.c
```diff
@@ -155,34 +157,69 @@ static unsigned int rej_uniform(int16_t *r,
...
+        shake128x_squeezeblocks(&kx, GEN_MATRIX_NBLOCKS);
+
+        for (j = 0; j < KECCAK_WAY; j++) {
+            vec = a[ (i + j) / KYBER_K].vec[ (i + j) % KYBER_K].coeffs;
+            buf = bufx[j];
+            buflen = GEN_MATRIX_NBLOCKS *XOF_BLOCKBYTES;
+            ctr = rej_uniform(vec, KYBER_N, buf, buflen);
+
+            while (ctr < KYBER_N ) {
+                uint64_t *ctx = keccakx_get_lane_state(&kx, j);
```
cothan, line 195: switch to single keccak here. We can collect all single keccak and write a 2nd pass, it's probably faster. However, it will make the code logic a little complicated.

fips202/fips202x.h
cothan: set to 4.

hanno-becker:
@cothan Can you fix the CI and mark the PR as ready-for-review when you think it's in a good shape?

mlkem/indcpa.c
hanno-becker: Good spot

cothan:
Sure, @hanno-becker , I was waiting for a few PR to land on main first. 
PR is marked as ready.

mlkem/indcpa.c
cothan: Track link: https://github.com/pq-crystals/kyber/issues/79

fips202/fips202x.c
mkannwischer, line 1: Plaese rename this to fips202x4

mlkem/indcpa.c
```diff
@@ -166,41 +168,85 @@ static unsigned int rej_uniform(int16_t *r,
 // Not static for benchmarking
 void gen_matrix(polyvec *a, const uint8_t seed[KYBER_SYMBYTES], int transposed)
```
mkannwischer, line 169: I do not quite understand why this function looks so different from https://github.com/pq-crystals/kyber/blob/main/avx2/indcpa.c? 
Shouldn't it be essentially the same (just ripping out the AVX2 instrinsics)?

I guess this is mostly because this part of the code handles arbitrary `KECCAK_WAY`. Cannot comment on if it results in the same calls to keccakx4 and keccakx1 as the AVX2 version? 

For which values for `KECCAK_WAY` should this work and for which did you test it?

mlkem/indcpa.c
mkannwischer: Sure, but please add a comment.

mlkem/indcpa.c
mkannwischer: Indeed seems to be an artifact from the Kyber90s version. Please delete.
Please also fix the spaces around `*` - it triggers my OCD.

fips202/fips202.h
```diff
@@ -35,6 +35,23 @@ typedef struct
...
+ * @param h Pointer to the output buffer where the squeezed data will be stored.
+ * @param nblocks Number of blocks to squeeze from the state. Each block is r bytes long, where r is the rate
+ *        specified when initializing the state.
+ * @param s Pointer to the Keccak state array. This array must contain at least 25 elements of type uint64_t.
+ * @param r Rate in bytes. Specifies the rate parameter of the Keccak function, which determines how much
+ *        data is processed per permutation.
+ */
+void keccak_squeezeblocks(uint8_t *h, size_t nblocks, uint64_t *s, uint32_t r);
+
+
```
mkannwischer, line 54: Why did you have to expose this function? I don't think it should be exposed. 
You should only be using the `shake128_squeezeblocks` and `shake256_squeezeblocks` functions

mkannwischer (changes requested):
I added a few comments and questions.

For the noise vectors, you can also use the Keccakx4 interface. 
Please have a look here and add: https://github.com/pq-crystals/kyber/blob/main/avx2/poly.c

mlkem/indcpa.c
cothan: Yeah, my homebrew astyle reformat the `*`. I have to manually add the space. Update soon.

mlkem/indcpa.c
cothan: The `KYBER_SYMBYTES` is `32 bytes`. So the function signature in `avx2` hardcoded the number. 
I don't like to have `#ifdef KYBER_K` to handle different value of `KYBER_K`, so our function has 2 for loops to handle all the cases. I tested with arbitrary value of `KECCAK_WAY = 1...8`, it always work. 
The compiler will optimize the bound of each for loop, so the overhead is very very mall.

mlkem/indcpa.c
cothan: Comment added.

fips202/fips202.h
cothan: I have to expose this function to switch to single lane Keccak. 
As in `gen_matrix`, when the `rej_sampling` fails to fullfill a vector, additional for a single lane is executed.
So a single lane state data is feed into `keccak_squeezeblocks` function. 
I'll see if I can wrap this function in another way.

cothan: okay, I add a function call `shake256x1_squeezeblocks` replace the direct call to `keccak_squeezeblocks`.

fips202/fips202x.c
cothan: done

mlkem/indcpa.c
cothan: updated.

mlkem/polyvec.c
```diff
@@ -273,3 +275,49 @@ void polyvec_add(polyvec *r, const polyvec *a, const polyvec *b)
...
+*              - uint8_t nonce: one-byte input nonce
+**************************************************/
+void polyvec_getnoise_eta1(polyvec *r, const uint8_t seed[KYBER_SYMBYTES], uint8_t nonce)
+{
+    uint8_t buf[KYBER_ETA1 * KYBER_N / 4];
+    unsigned int i = 0;
+    for (i = 0; i < KYBER_K; i++)
+    {
+        prf(buf, sizeof(buf), seed, nonce++);
+        poly_cbd_eta1(&r->vec[i], buf);
```
cothan, line 298: When `KYBER_K = 4 == KECCAK_WAY`, we can vectorize the whole function. `prf` is convert to `prfx4` and unroll `poly_cbd_eta1_x4`. 

This seem inefficient when `KYBER_K < KECCAK_WAY`, so I decided to unroll `poly_getnoise_eta1` in `indcpa.c` completely like `avx2`.

mlkem/indcpa.c
```diff
@@ -288,15 +349,35 @@ void indcpa_enc(uint8_t c[KYBER_INDCPA_BYTES],
...
-        poly_getnoise_eta2(ep.vec + i, coins, nonce++);
-    }
+    #if KYBER_K == 2
+    poly_getnoise_eta1(&sp.vec[0], coins, nonce++);
+    poly_getnoise_eta1(&sp.vec[1], coins, nonce++);
+
+    poly_getnoise_eta2(&ep.vec[0], coins, nonce++);
+    poly_getnoise_eta2(&ep.vec[1], coins, nonce++);
+
+    poly_getnoise_eta2(&epp, coins, nonce++);
```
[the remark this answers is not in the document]
cothan: When `KECCAK = 4`, the total shake256 here is 5, we can do KECCAK_WAY = 4, then a single keccak follow. 
Or we can have `KECCAK_WAY = 3`, `KECCAK_WAY = 2` here.

mlkem/indcpa.c
```diff
@@ -288,15 +349,35 @@ void indcpa_enc(uint8_t c[KYBER_INDCPA_BYTES],
...
+    #elif KYBER_K == 3
+    poly_getnoise_eta1(&sp.vec[0], coins, nonce++);
+    poly_getnoise_eta1(&sp.vec[1], coins, nonce++);
+    poly_getnoise_eta1(&sp.vec[2], coins, nonce++);
+
+    poly_getnoise_eta2(&ep.vec[0], coins, nonce++);
+    poly_getnoise_eta2(&ep.vec[1], coins, nonce++);
+    poly_getnoise_eta2(&ep.vec[2], coins, nonce++);
+
+    poly_getnoise_eta2(&epp, coins, nonce++);
```
cothan, line 369: When `KECCAK_WAY = 4`, the total shake256 here is 7, we can add a dummy lane, or support additional `KECCAK_WAY=3` along with `KECCAK_WAY=4`.

mlkem/indcpa.c
```diff
@@ -288,15 +349,35 @@ void indcpa_enc(uint8_t c[KYBER_INDCPA_BYTES],
...
+    #elif KYBER_K == 4
+    polyvec_getnoise_eta1(&sp, coins, nonce);
+    nonce += KYBER_K;
+
+    polyvec_getnoise_eta2(&ep, coins, nonce);
+    nonce += KYBER_K;
+
     poly_getnoise_eta2(&epp, coins, nonce++);
```
cothan, line 377: The total of SHAKE256 calls here are 9. So we can have `KECCAK_WAY=4, 5` to efficiently compute keccak here.

mlkem/indcpa.c
cothan: I analyze `getnoise_eta1` and `getnoise_eta2` below, it's seem like we can maximize the performance by having more than one KECCAK_WAY.

fips202/fips202x4.h
```diff
@@ -0,0 +1,66 @@
...
+                             uint8_t *out1,
+                             uint8_t *out2,
+                             uint8_t *out3,
+                             size_t nblocks,
+                             keccakx4_state *state);
+
+/*
+ * Squeezes a single lane in Keccak 4-way
+ */
+int shake256x1_squeezeblocks(uint8_t *out,
```
mkannwischer, line 50: I propose to rename this to `shake256x4_squeezeblocks_single`. 
Also I don't think referring to this as a `lane` is correct, because a lane in Keccak is well defined. 
I'd just call it `single Keccak state`

fips202/fips202x4.h
```diff
@@ -0,0 +1,66 @@
+// SPDX-License-Identifier: Apache-2.0
+#ifndef FIPS_202X_H
+#define FIPS_202X_H
```
mkannwischer, line 3: X -> X4

mlkem/poly.c
```diff
@@ -338,6 +339,39 @@ void poly_getnoise_eta1(poly *r, const uint8_t seed[KYBER_SYMBYTES], uint8_t non
...
+{
+    uint8_t buf[KECCAK_WAY][KYBER_ETA1 *KYBER_N / 4];
+    prf(buf[0], sizeof(buf[0]), seed, nonce0);
+    prf(buf[1], sizeof(buf[1]), seed, nonce1);
+    prf(buf[2], sizeof(buf[2]), seed, nonce2);
+    prf(buf[3], sizeof(buf[3]), seed, nonce3);
+    poly_cbd_eta1(r0, buf[0]);
+    poly_cbd_eta1(r1, buf[1]);
+    poly_cbd_eta1(r2, buf[2]);
+    poly_cbd_eta1(r3, buf[3]);
```
mkannwischer, line 372: These four calls to `prf` should be replaced with a single call to `shake256x4`?

mlkem/poly.c
```diff
@@ -357,6 +391,73 @@ void poly_getnoise_eta2(poly *r, const uint8_t seed[KYBER_SYMBYTES], uint8_t non
...
+                           uint8_t nonce0,
+                           uint8_t nonce1,
+                           uint8_t nonce2,
+                           uint8_t nonce3)
+{
+    uint8_t buf[KECCAK_WAY][KYBER_ETA2 * KYBER_N / 4];
+    prf(buf[0], sizeof(buf[0]), seed, nonce0);
+    prf(buf[1], sizeof(buf[1]), seed, nonce1);
+    prf(buf[2], sizeof(buf[2]), seed, nonce2);
+    prf(buf[3], sizeof(buf[3]), seed, nonce3);
```
mkannwischer, line 420: These four calls to `prf` should be replaced with a single call to `shake256x4`?

mlkem/poly.c
```diff
@@ -357,6 +391,73 @@ void poly_getnoise_eta2(poly *r, const uint8_t seed[KYBER_SYMBYTES], uint8_t non
...
+                              uint8_t nonce1,
+                              uint8_t nonce2,
+                              uint8_t nonce3)
+{
+    uint8_t buf1[KECCAK_WAY/2][KYBER_ETA1 * KYBER_N / 4];
+    uint8_t buf2[KECCAK_WAY/2][KYBER_ETA2 * KYBER_N / 4];
+    prf(buf1[0], sizeof(buf1[0]), seed, nonce0);
+    prf(buf1[1], sizeof(buf1[1]), seed, nonce1);
+    prf(buf2[0], sizeof(buf2[0]), seed, nonce2);
+    prf(buf2[1], sizeof(buf2[1]), seed, nonce3);
```
mkannwischer, line 454: These four calls to `prf` should be replaced with a single call to `shake256x4`?

cothan: updated in https://github.com/pq-code-package/mlkem-c-aarch64/pull/62/commits/6c716d18f1aada08abcd445c905fb77937dce1d5

mlkem/poly.c
cothan: fixed in https://github.com/pq-code-package/mlkem-c-aarch64/pull/62/commits/6c716d18f1aada08abcd445c905fb77937dce1d5

fips202/fips202x4.h
cothan: That's good idea. update in https://github.com/pq-code-package/mlkem-c-aarch64/pull/62/commits/6c716d18f1aada08abcd445c905fb77937dce1d5

mlkem/indcpa.c
cothan: After careful thoughts, I think it's not worth it to introduce additional KECCAK_WAY.

mkannwischer (approved):
Thanks @cothan. I'm happy with this now. 
@hanno-becker?

fips202/fips202x4.c
```diff
@@ -0,0 +1,242 @@
...
+}
+
+int shake128x4_absorb(keccakx4_state *state,
+                      const uint8_t *in0,
+                      const uint8_t *in1,
+                      const uint8_t *in2,
+                      const uint8_t *in3,
+                      size_t inlen)
+{
+    if (state == NULL || in0 == NULL || in1 == NULL || in2 == NULL || in3 == NULL)
```
hanno-becker, line 114: We should be uniform regarding the checking of parameters. All others functions seem to assume that pointers are valid. Why are we checking them here?

mlkem/indcpa.c
```diff
@@ -166,41 +168,86 @@ static unsigned int rej_uniform(int16_t *r,
...
+    uint8_t bufx[KECCAK_WAY][GEN_MATRIX_NBLOCKS * XOF_BLOCKBYTES];
+    uint8_t *buf = NULL;
+    int16_t *vec = NULL;
+    uint8_t x, y;
     xof_state state;
 
-    for (i = 0; i < KYBER_K; i++)
+    keccakx4_state statex;
+    // The input data to each Keccak lane.
+    // Original size; KYBER_SYMBYTES + 2, we add padding to make align load/store.
```
hanno-becker, line 181: What is the exact intent here? Adding `16` does not ensure alignment, which is indeed not guaranteed for an array for `uint8_t`. You would need an alignment directive here.

mlkem/indcpa.c
```diff
@@ -166,41 +168,86 @@ static unsigned int rej_uniform(int16_t *r,
...
-                xof_absorb(&state, seed, j, i);
+                seedxy[j][KYBER_SYMBYTES + 0] = y;
+                seedxy[j][KYBER_SYMBYTES + 1] = x;
             }
+        }
 
-            xof_squeezeblocks(buf, GEN_MATRIX_NBLOCKS, &state);
-            buflen = GEN_MATRIX_NBLOCKS * XOF_BLOCKBYTES;
-            ctr = rej_uniform(a[i].vec[j].coeffs, KYBER_N, buf, buflen);
+        shake128x4_absorb(&statex, seedxy[0], seedxy[1], seedxy[2], seedxy[3], KYBER_SYMBYTES + 2);
```
hanno-becker, line 203: Did you get to the bottom of why `+2` is needed here?

mlkem/indcpa.c
```diff
@@ -166,41 +168,86 @@ static unsigned int rej_uniform(int16_t *r,
...
+    // The input data to each Keccak lane.
+    // Original size; KYBER_SYMBYTES + 2, we add padding to make align load/store.
+    uint8_t seedxy[KECCAK_WAY][KYBER_SYMBYTES + 16];
+
+    for (i = 0; i < (KYBER_K * KYBER_K / KECCAK_WAY) * KECCAK_WAY; i += KECCAK_WAY)
     {
-        for (j = 0; j < KYBER_K; j++)
+        for (j = 0; j < KECCAK_WAY; j++)
         {
+            memcpy(seedxy[j], seed, KYBER_SYMBYTES);
```
hanno-becker, line 188: This can be hoisted out of the loop?

mlkem/poly.c
```diff
@@ -367,6 +369,46 @@ void poly_getnoise_eta1(poly *r, const uint8_t seed[KYBER_SYMBYTES], uint8_t non
...
+    uint8_t extkey[KECCAK_WAY][KYBER_SYMBYTES + 1];
+    memcpy(extkey[0], seed, KYBER_SYMBYTES);
+    memcpy(extkey[1], seed, KYBER_SYMBYTES);
+    memcpy(extkey[2], seed, KYBER_SYMBYTES);
+    memcpy(extkey[3], seed, KYBER_SYMBYTES);
+    extkey[0][KYBER_SYMBYTES] = nonce0;
+    extkey[1][KYBER_SYMBYTES] = nonce1;
+    extkey[2][KYBER_SYMBYTES] = nonce2;
+    extkey[3][KYBER_SYMBYTES] = nonce3;
+    shake256x4(buf[0], buf[1], buf[2], buf[3], KYBER_ETA1 * KYBER_N / 4,
```
hanno-becker, line 404: There is a fair amount of unnecessary copying here which we should remove at some point, e.g. by specializing `shake256x4` to the relevant case where all input buffers are equal except for the last byte.

hanno-becker (commented):
Thank you @cothan.

I think we may need to tweak the interfaces further to reduce the amount of copying. However, this can be a follow-up.

The batched interfaces does input validation not present in the non-batched interface. I think it should be uniform, and would opt to remove the validation in the batched functions for now.

Otherwise, `gen_matrix()` can be slightly improved by hoisting out the prepraration of the bulk of the `seed` array.

mlkem/poly.c
cothan: Good idea, I add #88 so we don't forget in the future.

mlkem/indcpa.c
cothan: You're right, I overlook this line. I moved it out of the loop.

fips202/fips202x4.c
cothan: At first, I thought the Keccak interface will be reused elsewhere and for other projects. So there are the checks. 
I removed all the pointers check in the latest commit.

hanno-becker (approved):

PR #69, opened 2024-06-19
Add benchmark workflow (RPi4 / Arm Cortex-A72)

potsrevennil:

This PR depends on #68

I Added a benchmark workflow. Note that the workflow will only run on `main` branch or PR labeled as `benchmark`. The `tests` script is also refactored a bit in order to show the benchmark summary on GitHub action.

I refactored the `nix` shell specification in order to speed up ci, and also
disabled ci jobs running on self-host runner when pushed to fork repositories by the way.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   bench:
     runs-on: self-hosted-rpi4
-    if: github.repository_owner == 'pq-code-package'
+    if: github.repository_owner == 'pq-code-package' && (contains(github.event.pull_request.labels.*.name, 'benchmark') || github.ref == 'refs/heads/main')
     steps:
       - uses: actions/checkout@v4
       - name: Setup nix
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -26,6 +26,9 @@ jobs:
               - $(astyle --version)
               - $(${{ matrix.cross_prefix }}gcc --version | grep -m1 "")
               - $(bash --version | grep -m1 "")
+
+              ## CPU Info
+              $(cat /proc/cpuinfo)
             EOF
       - name: Run benchmark
         shell: nix develop .#ci -c bash -e {0}
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,0 +1,51 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Bench
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+jobs:
+  bench:
+    runs-on: self-hosted-rpi4
+    if: github.repository_owner == 'pq-code-package'
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup nix
+        uses: ./.github/actions/setup-nix
+        with:
+          devShell: ci
+          script: |
+            ARCH=$(uname -m)
+            cat >> $GITHUB_STEP_SUMMARY <<-EOF
+              ## Setup
+              Architecture: $ARCH
+              - $(uname -a)
+              - $(nix --version)
+              - $(astyle --version)
+              - $(${{ matrix.cross_prefix }}gcc --version | grep -m1 "")
+              - $(bash --version | grep -m1 "")
+            EOF
+      - name: Run functional tests
+        id: func_test
+        shell: nix develop .#ci -c bash -e {0}
+        run: |
+          tests func -v
+      - name: Run KAT tests
+        id: kat_test
+        if: |
+          success()
+          || steps.func_test.conclusion == 'failure'
+        shell: nix develop .#ci -c bash -e {0}
+        run: |
+          tests kat -v
+      - name: Run Nistkat tests
+        id: nistkat_test
+        if: |
+          success()
+          || steps.func_test.conclusion == 'failure'
+          || steps.kat_test.conclusion == 'failure'
+        shell: nix develop .#ci -c bash -e {0}
+        run: |
+          tests nistkat -v
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -33,4 +33,4 @@ jobs:
       - name: Run benchmark
         shell: nix develop .#ci -c bash -e {0}
         run: |
-          tests bench -c PMU -v
+          tests bench -c PMU --cpu cortex-a72 -v
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -27,25 +27,7 @@ jobs:
               - $(${{ matrix.cross_prefix }}gcc --version | grep -m1 "")
               - $(bash --version | grep -m1 "")
             EOF
-      - name: Run functional tests
-        id: func_test
+      - name: Run benchmark
         shell: nix develop .#ci -c bash -e {0}
         run: |
-          tests func -v
-      - name: Run KAT tests
-        id: kat_test
-        if: |
-          success()
-          || steps.func_test.conclusion == 'failure'
-        shell: nix develop .#ci -c bash -e {0}
-        run: |
-          tests kat -v
-      - name: Run Nistkat tests
-        id: nistkat_test
-        if: |
-          success()
-          || steps.func_test.conclusion == 'failure'
-          || steps.kat_test.conclusion == 'failure'
-        shell: nix develop .#ci -c bash -e {0}
-        run: |
-          tests nistkat -v
+          tests bench -c PMU -v
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -33,4 +33,4 @@ jobs:
       - name: Run benchmark
         shell: nix develop .#ci -c bash -e {0}
         run: |
-          tests bench -c PMU --cpu cortex-a72 -v
+          tests bench -c PMU --cflags -mcpu=cortex-a72 -v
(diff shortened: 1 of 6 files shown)


.github/workflows/bench.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          devShell: ci
+          script: |
+            ARCH=$(uname -m)
+            cat >> $GITHUB_STEP_SUMMARY <<-EOF
+              ## Setup
+              Architecture: $ARCH
+              - $(uname -a)
+              - $(nix --version)
+              - $(astyle --version)
```
mkannwischer, line 26: Can you add a `cat /proc/cpuinfo` here, please?

.github/workflows/bench.yml
```diff
@@ -0,0 +1,33 @@
...
+              - $(${{ matrix.cross_prefix }}gcc --version | grep -m1 "")
+              - $(bash --version | grep -m1 "")
+            EOF
+      - name: Run benchmark
+        shell: nix develop .#ci -c bash -e {0}
+        run: |
+          tests bench -v
```
mkannwischer, line 33: You'll need to change this to `tests bench -c PMU -v` here after my recent changes.

Makefile
```diff
@@ -9,23 +9,23 @@ INCLUDE_FIPS202 = -I fips202
 INCLUDE_MLKEM = -I mlkem
 INCLUDE_RANDOM = -I randombytes
 INCLUDE_NISTRANDOM = -I test/nistrng
-CFLAGS += -Wall -Wextra -Wpedantic -Werror -Wmissing-prototypes -Wredundant-decls \
+override CFLAGS += -Wall -Wextra -Wpedantic -Werror -Wmissing-prototypes -Wredundant-decls \
```
hanno-becker, line 12: What does that do?

scripts/tests
```diff
@@ -279,21 +300,24 @@ def kat(force_qemu, verbose):
...
     results = test_schemes(
         "benchmark",
         lambda scheme: scheme.name.replace("MLKEM", "test/bin/bench_kyber"),
         lambda _: True,
         lambda _: True,
+        process_bench,
         force_qemu,
         verbose,
-        cycles=cycles,
+        [f"CYCLES={cycles}", f"CFLAGS=-mcpu={cpu}"],
```
hanno-becker, line 319: Should `-mcpu` not be omitted if the `--cpu` argument isn't given?

hanno-becker (changes requested):
I'd like to make the `--cpu` to `tests bench` optional and omit `-march` if `-mcpu` is specified.

scripts/tests
potsrevennil: Oh yes! I was trying if this work, and forgot to handle it properly. Thanks for the catch

hanno-becker: We might also want to omit `-march` when `-mcpu` is specified, as otherwise there may be conflicts. I was just trying this with `-mcpu=neoverse-n1` and got
```
cc1: error: switch ‘-mcpu=neoverse-n1’ conflicts with ‘-march=armv8-a’ switch [-Werror]
```

Makefile
potsrevennil: So we think it might be nice to be able to add some CFLAGS from command line arguments, so that things like `mcpu`, `werror`, ... could be configurable without adding a bunch of `Makefile` variables.

But it seems like we have to add this override to every `CFLAGS += ...` so that `CFLAGS` in the Makefile will not be overwritten by the `CFLAGS` from the command line arguments.

However, I don't know if there is any better solution for this.

hanno-becker: @potsrevennil I don't yet follow, isn't that what the `+=` would take care of? The initial value of `CFLAGS` would be amended to, not overwritten.

If it's an issue of priority, could one just make sure to append to the _left_ rather than the _right_, so command line `CFLAGS` take priority over defaults?

potsrevennil: According to GNU Make reference manual
> If a variable has been set with a command argument (see [Overriding Variables](https://www.gnu.org/software/make/manual/make.html#Overriding)), then ordinary assignments in the makefile are ignored. If you want to set the variable in the makefile even though it was set with a command argument, you can use an override directive,

Unfortunately, only `+=` won't be able to take care of this.

So I'm afraid that this is the only way that we can do except for setting a bunch of `Makefile` variables, if we want to allow user for more flexibility (setting `CFLAGS` from cli).

scripts/tests
potsrevennil: @hanno-becker Where is the `-march` coming from ? We currently don't have this flag set in our `Makefile` right ?

potsrevennil: btw, on which platform you got this error ? 
I didn't got any error for setting `-mcpu=neoverse-n1` when compiling on my mac

Makefile
hanno-becker: @potsrevennil This may depend on how the command line arguments are specified?

With this dummy Makefile
```
TEST ?=
TEST += beef

all:
	@echo $(TEST)
```
I get:
```
beckphan@884d7ce1daa2 ~ % make -f test_makefile
beef
beckphan@884d7ce1daa2 ~ % TEST=dead make -f test_makefile
dead beef
beckphan@884d7ce1daa2 ~ % make TEST=dead -f test_makefile
dead
```

scripts/tests
hanno-becker: I ran it on a Neoverse N1

hanno-becker: > Where is the -march coming from ? We currently don't have this flag set in our Makefile right ?

Good question, I just assumed we'd set this somewhere in the `Makefile`. Let me check.

Makefile
potsrevennil: Oh yes, you're right.
I assumed that we wanted the command line arguments specified as `make TEST=dead ...`.
If that is not the case, then we can indeed remove override directive

scripts/tests
hanno-becker: This is a bit confusing `-march` is indeed not specified anywhere:
```
gcc -mcpu=neoverse-n1 -Wall -Wextra -Wpedantic -Werror -Wmissing-prototypes -Wredundant-decls -Wshadow -Wpointer-arith -Wno-unknown-pragmas -O3 -fomit-frame-pointer -pedantic -I mlkem -I fips202 -I randombytes -DKYBER_K=2 mlkem/kem.c mlkem/indcpa.c mlkem/polyvec.c mlkem/poly.c mlkem/ntt.c mlkem/cbd.c mlkem/reduce.c mlkem/verify.c fips202/keccakf1600.c fips202/fips202.c mlkem/symmetric-shake.c randombytes/randombytes.c test/hal.c test/bench_kyber.c -o test/bin/bench_kyber512
cc1: error: switch ‘-mcpu=neoverse-n1’ conflicts with ‘-march=armv8-a’ switch [-Werror]
cc1: all warnings being treated as errors
```

Interestingly, I only get this in the nix shell, not with the native gcc.

potsrevennil: What is the gcc version of both environment ?

potsrevennil: Ok, I also got the same issue as you on the rpi4 (same machine as the selfhost runner)

The `gcc` version in `nix` shell is
```
gcc (GCC) 13.2.0
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE
```

While the native `gcc` is
```
gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

mkannwischer: I guess there is no way around also passing `-march=armv8.2-a` in this case. 
Maybe it's better to just let the user pass in ARCH_FLAGS.

hanno-becker: @mkannwischer I find this very odd -- if you don't specify `-march` on the command line, `gcc` shouldn't complain about a conflict. Unless in the Nix shell it's somehow defined via an alias..

hanno-becker: For the moment, we can either (a) replace `-mcpu` by `-mtune`, which does not interfere with `-march`. IIRC, `-mcpu` is a combination of `-march` (derived from the CPU) and `-mtune`. (b) Add an explicit `--arch` parameter similar to `tests`. (c) Allow the user to specify `ARCH_FLAGS`.

I can live with all of them, but prefer (b) or (c).

potsrevennil: I don't think it's nix shell causing the problem. I've tried to use the same gcc version as nix shell in the native shell environment, and it result in the same error. And I've checked the env vars of nix shell, there is no CFLAGS and ARCH_FLAGS preexists.

I vote for (c), and I pushed a version where each command of `tests` script can set both extra cflags and arch flags.

hanno-becker (changes requested):
Thank you @potsrevennil, I confirm that `--arch-flags` works.

May I ask for one more usability improvement: For `perf`, I have to run the binaries with `sudo`, but somehow `sudo ./scripts/test ...` doesn't work (and is also not ideal, because the build itself really needs not sudo rights).

Can you add an option to the test script which builds only, and prints the names of the files that have been produced?

scripts/tests
hanno-becker: Great, thank you @potsrevennil

potsrevennil:
@hanno-becker Yeah, I agree with you that it indeed makes more sense in some cases to separate building and running the binary. There're also cases that we would like to run the binaries in a different machine.

However, I'm thinking this might a bit out of scope for this PR, maybe we could open an issue to address this more, and make it a separate PR. What do you think ?

hanno-becker:
@potsrevennil Ok!

PR #76, opened 2024-06-21
Remove mention of 'clean' code from README

### LINKED ISSUE #63: Regarding the definition of "Assurance"
Dear MLKEM-C-AARCH64 maintainers,

I think the definition of "Assurance" doesn't reflect the state-of-the-art development in the literature.
"Clean" is not a necessity for a program to be amenable to verification. There are continuous efforts to formally verify high-performance, highly complicated (extensively shuffled), manually written, and highly scalable (in the case of power-of-two-size NTT/iNTT) assembly code.

In the second item of "tensions",
it says "Optimal code is complex (e.g. relying on handwritten assembly), impeding maintainenance and amenability for audit or verification." (maintainenance -> maintenance). This was already resolved for some of the frequently used operations, such as Keccak permutation by https://eprint.iacr.org/2023/1861, and NTT by https://tches.iacr.org/index.php/TCHES/article/view/9838.

Recently, the latest manually shuffled NTT/iNTT for Kyber (https://github.com/neon-ntt/neon-ntt/blob/master/kyber768/ntt/__asm_NTT.S) has been formally verified with the tool CryptoLine. See https://github.com/fmlab-iis/cryptoline/commit/1e0173c2a3e662961f71f707c6623c3abc19515c. 
Readability for the general audience is not a requirement for formal verification (I'm sure that the resulting programs are readable for the authors if they put effort into structuring the extensively optimized assembly programs). This is one of the goals of CryptoLine: formally verifying assembly programs optimized "in the wild." 
As for maintenance, you don't have to worry about that. I'm the author of the latest NTT/iNTT for Kyber and I only spent two afternoons on shuffling (the resulting programs perform comparably to the programs by SLOTHY on Cortex-A72 and Apple M1 Pro). 

I believe users of this repository should be informed about the above, so they don't have to worry about something that is already addressed.

Vincent

hanno-becker:
The README previously suggested that 'clean' code is a necessity for amenability to verification. This was not intentional, and as noted in #63, not true.

This commit removes the mentioning of 'clean' code from the README. The primary qualities that we seek from 'cleanliness' are amenability for manual audit and adaptability. The former was already mentioned explicitly in the README, while the latter is added in this commit.

Closes #63

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,13 +5,14 @@
 ## Goals of mlkem-c-aarch64
 
 The primary goals of this project are as follows:
-- _Assurance:_ Clean code that's extensively tested and amenable for audit and verification
-- _Ease of use:_ Permissive licensing, modularity, few dependencies
+- _Assurance:_ Offer code that's extensively tested and amenable for audit and verification
+- _Ease of use:_ Permissive licensing, modularity, adaptability, few dependencies
 - _Performance:_ Competitive performance for most Armv8-A/Armv9-A platforms
 
 There are tensions between these goals:
 - Optimal code is target-specific, but a large variety of CPU-specific implementations makes a library harder to both use and maintain.
-- Optimal code is complex (e.g. relying on handwritten assembly), impeding maintainenance and amenability for audit or verification.
+- Optimal code is complex (e.g. relying on handwritten assembly and extensive interleaving), impeding maintainenance and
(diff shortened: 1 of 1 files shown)


mkannwischer (approved):
I agree that this is the better wording.

PR #77, opened 2024-06-25
Make the bench workflow trigger more naturally

potsrevennil:

Follow the [slothy](https://github.com/slothy-optimizer/slothy/blob/main/.github/workflows/test_basic.yaml#L8) approach as @hanno-becker and @mkannwischer suggested.


mkannwischer (approved):
Thanks @potsrevennil. 
I tested it and it works as intended: When you add the benchmark label it benchmarks once. If you want to benchmark again, you need to remove the label and add it again.

PR #78, opened 2024-06-26
Add github benchmark action

mkannwischer:


scripts/tests
```diff
@@ -345,7 +345,13 @@ def kat(force_qemu, verbose, cflags, arch_flags):
     default="NO",
     help="Method for counting clock cycles. PMU requires (user-space) access to the Arm Performance Monitor Unit (PMU). PERF requires a kernel with perf support.",
 )
-def bench(force_qemu, verbose, cycles, cflags, arch_flags):
+@click.option(
+    "-r",
+    "--resultfile",
+    nargs=1,
+    help="Path to result file in json format",
+)
```
potsrevennil, line 353: I'm thinking maybe it will be better to name the flag as `output` or `output-file`, and use `-o` for the short alias ?

mkannwischer:
After lots of experimenting, I figured out that we cannot test this in a pull request. 

The maximum permission of the GITHUB_TOKEN in pull requests from forks on `contents` in the repo is `read`, but we need `write` to be able to create commits in the `gh-pages`. In the end we want to only run it on the main branch, so that's not a problem in the end. 

See:  https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

I have performed quite some tests in https://github.com/mkannwischer/bench-test today, and I believe that what we have here right now should work.

scripts/tests
```diff
@@ -360,6 +366,28 @@ def bench(force_qemu, verbose, cycles, cflags, arch_flags):
...
+                d = {
+                    k: int(v)
+                    for k, v in (l.decode().split("=") for l in r.splitlines()[:3])
+                }
+                for primitive in ["keypair", "encaps", "decaps"]:
+                    v.append(
+                        {
+                            "name": f"{schemeStr} {primitive}",
+                            "unit": "cycles",
+                            "value": d[f"{primitive} cycles"],
```
hanno-becker, line 386: Nit, why append "cycles" if we've specified the unit before?

scripts/tests
```diff
@@ -360,6 +366,28 @@ def bench(force_qemu, verbose, cycles, cflags, arch_flags):
...
+        import json
+
+        with open(output, "w") as f:
+            v = []
+            for scheme in results:
+                schemeStr = str(scheme)
+                r = results[scheme]
+                d = {
+                    k: int(v)
+                    for k, v in (l.decode().split("=") for l in r.splitlines()[:3])
```
hanno-becker, line 379: A short comment indicating what the expected format here is would be useful.

hanno-becker (commented):
Minor comments, otherwise LGTM

scripts/tests
mkannwischer: It does not append 'cycles'. That's the key for looking it up from the output.

hanno-becker: ah, got it

scripts/tests
mkannwischer: I added a comment.

hanno-becker (approved):

PR #79, opened 2024-06-26
Makefile keep track of variables

potsrevennil:

This PR resolves the issue mentioned in https://github.com/pq-code-package/mlkem-c-aarch64/issues/75

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 .vscode
 .idea
 test/bin/
+test/obj/
--- a/Makefile
+++ b/Makefile
@@ -1,32 +1,14 @@
 # SPDX-License-Identifier: Apache-2.0
 
-Q ?= @
-QEMU = qemu-aarch64
-CROSS_PREFIX ?=
-CC := $(CROSS_PREFIX)gcc
 
-INCLUDE_FIPS202 = -I fips202
-INCLUDE_MLKEM = -I mlkem
+.PHONY: all mlkem kat nistkat clean
+
+all: mlkem bench kat nistkat
+
+include mk/config.mk
+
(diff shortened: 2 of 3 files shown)


mkannwischer (changes requested):
Yes, @potsrevennil. 

I verified that now the Makefile triggers a re-build when a different `CYCLES` flag is passed. 

Maybe also keep track of the CFLAGS that are potentially passed from the outside?

potsrevennil:
We cannot keep track of CFLAGS for now, since we chose to modify CFLAGS with environment variables (see discussion in https://github.com/pq-code-package/mlkem-c-aarch64/pull/69), while the current approach can only keep track of makefile variables.

Since the purpose of keep tracking of CFLAGS is that we want to avoid `make clean` while compiling for different architectures. There might be two ways of doing this:

1. Add a makefile variable like EXTRACFLAGS, and set `-mcpu=...` via this EXTRACFLAGS, and keep track of it is a way. 
2. Add a makefile variable like PLATFORM,  and set `mcpu`, `march` flags accordingly, like [mlkem-c-embedded](https://github.com/pq-code-package/mlkem-c-embedded/blob/e1ab38d94529eb6a43705cf7dd938879b177ea07/mk/config.mk#L37-L60) have done

Maybe we can discuss this further in https://github.com/pq-code-package/mlkem-c-aarch64/issues/75, and open up a separate PR for solving this.

mkannwischer (approved):
Okay, understood. 
Let's merge this one first, and think about the `CFLAGS` separately. 
For now users using custom CFLAGS need to run `make clean` in between, but I guess that is fine for now.

PR #82, opened 2024-06-28
Add support for static linking of glibc on aarch64 and x86_64 linux 

potsrevennil:

In the `nix` shell, we specify that `gcc` is downloaded from arm gnu official website for `x86_64` linux, and use the nixpkgs `gcc` for `aarch64` linux. 

However glibc only comes as a dynamic library in both cases, making compiling with `-static` result in error: 
```
ld: cannot find -lc: No such file or directory
```

Now I added glibc static library as a dependency, and configured it into one of the gcc default search path. Additional context, some package manager like `pacman` in `archlinux` packed the gcc with glibc, so that compilation with `-static` works automatically in this case.

@hanno-becker @mkannwischer Please have a look and see if the static compilation works, and the original usage won't break. The static compilation can be tested such as:
```
CFLAGS="-mcpu=cortex-a55 -march=armv8.2-a -static" make bench CYCLES=PERF # aarch64
CFLAGS="-mcpu=cortex-a55 -march=armv8.2-a -static" CROSS_PREFIX=aarch64-none-linux-gnu- make bench CYCLES=PERF # x86_64
```

I've also cleanup the nix file a bit, hope that it can be more readable and easier to maintain.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -63,7 +63,7 @@
               aarch64-gcc =
                 if pkgs.stdenv.isx86_64
                 then [ cross-gcc ]
-                else [ pkgs.glibc pkgs.glibc.static ]; # I assume this is used in `mkShell`, therefore for aarch64 platforms gcc is already available in PATH
+                else [ (pkgs.gcc13.override { propagateDoc = true; isGNU = true; }) pkgs.glibc pkgs.glibc.static ];
             in
             pkgs.lib.optionals pkgs.stdenv.isLinux aarch64-gcc ++
             builtins.attrValues {
@@ -86,7 +86,7 @@
             });
         in
         {
-          devShells.default = wrapShell pkgs.mkShell {
+          devShells.default = wrapShell pkgs.mkShellNoCC {
             packages = core ++ linters ++ cbmcpkg ++
               builtins.attrValues {
                 inherit (pkgs)
@@ -95,8 +95,8 @@
               };
           };
 
-          devShells.ci = wrapShell pkgs.mkShell { packages = core; };
-          devShells.ci-cbmc = wrapShell pkgs.mkShell { packages = (core ++ cbmcpkg); };
+          devShells.ci = wrapShell pkgs.mkShellNoCC { packages = core; };
+          devShells.ci-cbmc = wrapShell pkgs.mkShellNoCC { packages = (core ++ cbmcpkg); };
           devShells.ci-linter = wrapShell pkgs.mkShellNoCC { packages = linters; };
         };
       flake = {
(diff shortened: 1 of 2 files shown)


mkannwischer:
It's likely worth to mention that static compilation is a prerequisite for benchmarking on some platforms. 
For example, Cortex-A55 (A78/X1) is a Snapdragon 888 which is running Android. Running statically linked binaries is easy, but currently we don't have a way to run dynamically linked binaries there. 
Hence, this problem came up and turned out much more annoying than we anticipated it to be.

potsrevennil:
@mkannwischer Thanks for the additional information!

mkannwischer:
I tested this PR on my RPi4 (AArch64) and my x86 machine. 
On x86 I hadd to add `CROSS_PREFIX=aarch64-linux-gnu-`, but then it works fine. 
The resulting benchmarking binaries work fine on the A55. 

I also checked that this does not break the native and qemu tests.

mkannwischer (approved):
The code changes also look good to me. Thanks @potsrevennil!

flake.nix
```diff
@@ -55,8 +50,23 @@
...
+                cc = callPackage ./arm-gnu-gcc.nix { };
+                bintools = with pkgsCross.aarch64-multiplatform; wrapBintoolsWith {
+                  bintools = binutils-unwrapped;
+                  libc = glibc.static;
+                };
+              };
+              aarch64-gcc =
+                if pkgs.stdenv.isx86_64
+                then [ cross-gcc ]
+                else [ pkgs.glibc pkgs.glibc.static ]; # I assume this is used in `mkShell`, therefore for aarch64 platforms gcc is already available in PATH
```
mkannwischer, line 66:
```suggestion
                else [ pkgs.glibc pkgs.glibc.static ]; # assumed to be used in `mkShell`, therefore for aarch64 platforms gcc is already available in PATH
```

flake.nix
```diff
@@ -66,49 +76,28 @@
               inherit (pkgs.python3Packages)
                 python
                 click;
-            }
-          ++ {
-            "x86_64-linux" = [ (pkgs.callPackage ./arm-gnu-gcc.nix { }) ];
-            "aarch64-linux" = [ (pkgs.gcc13.override { propagateDoc = true; isGNU = true; }) ];
-            "aarch64-darwin" = [ ];
```
mkannwischer, line 73: I'm confused by this. 
on aarch64-linux, didn't we install gcc from nix before? Now we do not anymore?

potsrevennil: Before I used `pkgs.mkShellNoCC` to build to nix shell, which explicitly excluded nix CC in the shell. But now I changed it to `pkgs.mkShell` which packed gcc and configured it in nix shell by default, so that I don't have to do the similar trick as x86_64 for aarch64.

potsrevennil: Oh, turns out I just have to add the specific gcc into the dependency then we're all fine

potsrevennil: Thanks for the catch, it will be less confusing to add the specific gcc version in the shell

mkannwischer: Thanks!

PR #84, opened 2024-06-28
Benchmarking on A55

potsrevennil:

Mainly added a flag `-w`/`--exec-wrapper` for `bench` command allowing user to pass in there custom wrapper for running the benchmark binary.
```
Usage: tests bench [OPTIONS]

Options:
...
  -w, --exec-wrapper TEXT         Run the benchmark binary with the user-
                                  customized wrapper.
```

This is due to for some target machines, we might not be able to compile direcly on the machine, so we would need to cross compile, and then run the binary remotely. See https://github.com/pq-code-package/mlkem-c-aarch64/pull/82#issuecomment-2196212811

The current github bench ci looks ugly, will fix it after getting everything working.


mkannwischer:
This is now functional as far as I can tell. The CI yml needs some clean up, but I'm afraid that has to wait until next week.

scripts/tests
```diff
@@ -25,90 +25,85 @@ def sha256sum(result):
...
+    )
 
-        if result.returncode != 0:
-            logging.error(
-                f"Running {bin} natively failed: {result.returncode} {result.stderr.decode()}"
-            )
-            sys.exit(1)
+    if result.returncode != 0:
+        logging.error(
+            f"Running {bin} failed: {result.returncode} {result.stderr.decode()}"
```
hanno-becker, line 104: Maybe log `cmd` here as well?

scripts/tests
```diff
@@ -25,90 +25,85 @@ def sha256sum(result):
...
+    cmd = [f"./{bin}"]
+    if force_qemu or (platform.system() == "Linux" and platform.machine() == "x86_64"):
+        logging.info(f"Emulating {bin} with QEMU")
+        cmd = ["qemu-aarch64"] + cmd
+    else:
+        if exec_wrapper:
+            logging.info(f"Running {bin} with customized wrapper.")
+            cmd = [f"{exec_wrapper}"] + cmd
+
         if run_as_root:
```
hanno-becker, line 86: Should this be hoisted out to just before the `subprocess.run`?

scripts/tests
```diff
@@ -25,90 +25,85 @@ def sha256sum(result):
...
             logging.info(
-                "Running benchmarks as root -- you may need to enter your root password."
+                f"Running {bin} as root -- you may need to enter your root password."
             )
             cmd = ["sudo"] + cmd
 
-        if mac_taskpolicy is not None:
+        if platform.system() == "Darwin" and mac_taskpolicy is not None:
```
hanno-becker, line 92: Can we subsume `mac_taskpolicy` under `exec_wrapper`? That is, if `mac_taskpolicy` is defined, set `exec_wrapper` accordingly. If both `mac_taskpolicy` and `exec_wrapper` are defined, fail. Document accordingly that `mac_taskpolicy` is just an abbreviation for `exec_wrapper {taskpolicy ...}`.

scripts/tests
```diff
@@ -388,17 +384,31 @@ def kat(force_qemu, verbose, cflags, arch_flags):
     type=bool,
     help="Benchmarking binary is run with sudo.",
 )
+@click.option(
+    "-w",
+    "--exec-wrapper",
```
hanno-becker, line 389: This seems to work only for individual commands, not for e.g. `--exec-wrapper "bash my_script.sh"` or `--exec-wrapper "my_wrapper arg0 arg1"`. It would be nice if it could work roughly like `xargs`.

hanno-becker (changes requested):
Left minor suggestions for improvement -- otherwise looks good to me. I tested `--exec-wrapper` with a dummy wrapper script.

scripts/tests
```diff
@@ -416,7 +416,7 @@ def bench(
         else:
             exec_wrapper = f"taskpolicy -c {mac_taskpolicy}"
 
-    if " " in exec_wrapper:
+    if exec_wrapper and " " in exec_wrapper:
         exec_wrapper = exec_wrapper.split(" ")
```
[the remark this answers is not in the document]
mkannwischer: yes, maybe that's a little ugly. let me always turn it into a list.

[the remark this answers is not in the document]
mkannwischer: better?

scripts/tests
```diff
@@ -416,8 +413,11 @@ def bench(
         else:
             exec_wrapper = f"taskpolicy -c {mac_taskpolicy}"
 
-    if exec_wrapper and " " in exec_wrapper:
-        exec_wrapper = exec_wrapper.split(" ")
+    if exec_wrapper:
+        if " " in exec_wrapper:
+            exec_wrapper = exec_wrapper.split(" ")
+        else:
+            exec_wrapper = [exec_wrapper]
```
hanno-becker, line 420: Nit: I think you can unconditionally do `exec_wrapper.split(" ")` as that will return a singleton list in case `" "` does not occur in `exec_wrapper.`

mkannwischer: true. Thanks

mkannwischer:
I think this is ready now. 
@hanno-becker and other comments?

potsrevennil:
Why is there a duplicated `tests` script in the root directory ?

mkannwischer:
> Why is there a duplicated `tests` script in the root directory ?

uff. good catch. that was stupid. fixed.

scripts/tests
```diff
@@ -388,20 +378,44 @@ def kat(force_qemu, verbose, cflags, arch_flags):
...
 
+    if mac_taskpolicy:
+        if exec_wrapper:
+            logging.error(f"cannot set both --mac-taskpolicy and --exec-wrapper")
+            sys.exit(1)
+        else:
+            exec_wrapper = f"taskpolicy -c {mac_taskpolicy}"
+
+    if exec_wrapper:
+        exec_wrapper = exec_wrapper.split(" ")
```
potsrevennil, line 417: I'm thinking, probably it's cleaner to move this into `base_run` function, near L87

mkannwischer: indeed. I moved it there.

potsrevennil:
@mkannwischer Thanks! everything looks good to me!

hanno-becker (approved):

PR #67, opened 2024-06-18
Add coeff_signed_to_unsigned() functions and its proof

rod-chapman:
This PR adds _only_ the coeff_signed_to_unsigned() function in poly.c and its proof.

This was previously added in support of the proof of poly_compress(), but that proof is blocked pending CBMC bug fixes.

This PR adds coeff_signed_to_unsigned() in isolation so that proofs of other functions may proceed, such as poly_tobytes()

To run proofs, cd to "cbmc/proofs/coeff_signed_to_unsigned", then either

1. "make" (runs with litani and cbmc-viewer)
or
2. "make -f desktop.mk" (run with command-line CBMC tools only - best for inspecting details results.)

PR #85, opened 2024-06-28
Add z3_4_12 to standard NIX config for all platforms.

rod-chapman:
Adds Z3_4_12 to standard NIX config, ensuring that "cbmc --smt2" will work on all platforms and CI runners.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -71,6 +71,7 @@
                 yq
                 ninja# 1.11.1
                 qemu# 8.2.4
+                z3_4_12
                 cadical;
 
               inherit (pkgs.python3Packages)


hanno-becker (dismissed):
This should be in the `cbmcpkg` config, so it's not installed in builds which do not use CBMC.

rod-chapman:
I don't understand NIX language... where should it go?

mkannwischer:
Apologies - sometimes nix is not very intuitive. 
I was too lazy earlier to work it out correctly, but @hanno-becker is right. I have moved it to the `cbmcpkg`

mkannwischer (approved):
LGTM. Required for #67

PR #89, opened 2024-07-22
Refactor nix configuration on ci and the aarch64 gcc dependency on x86_64 machines 

potsrevennil:
#### 1. Utilization of `pkgCross.aarch-multiplatform`:
- The `pkgCross.aarch-multiplatform` package in `Nix` is now used to set up `GCC` targeting the `AArch64` platform. This change enhances clarity and simplicity.
- The cumbersome `arm-gnu-gcc.nix` file can thus be removed.
- Note: `AArch64` `macOS` will continue using the native `Clang` compiler as per the original setup.

#### 2. Cross Prefix Modification:
- The cross prefix for GCC in `pkgCross.aarch-multiplatform` has changed from `aarch64-none-linux-gnu-` to `aarch64-unknown-linux-gnu-`.
- To accommodate this, the test script has been updated to allow users to specify the cross prefix they are using.

#### 3. CI Configuration Refactor:

- The CI configuration has been streamlined by utilizing GitHub defaults for setting up the Nix shell, thereby avoiding multiple shell declarations.


cothan (approved):

scripts/tests
```diff
@@ -239,6 +241,14 @@ _shared_options = [
...
         help="Show verbose output or not",
     ),
+    click.option(
+        "-cp",
+        "--cross-prefix",
+        show_default=True,
+        default="aarch64-unknown-linux-gnu-",
+        nargs=1,
+        help="Cross prefix for compilation on non-aarch64 machines",
+    ),
```
mkannwischer, line 251: How do you handle that on MacOS you don't want to default to `aarch64-unknown-linux-gnu-`?

mkannwischer:
I tested that I can cross compile on x86 and the produced binaries still run on A55. Functionally this is looking good for me. If you fix the one comment I had, we can get this merged. 
Thanks @potsrevennil!

scripts/tests
potsrevennil: Thanks for the catch!
I think it'd be better to set the options default to empty string for those who don't need this option, for example MacOS, to make it less confusing.

mkannwischer (approved):
Thanks @potsrevennil! This is much cleaner now! 

I tested that everything still works on my laptop (cross-compiled + qemu) and on my RPi (natively).

PR #90, opened 2024-07-30
Update CBMC to v6.1.1

hanno-becker:

PR #93, opened 2024-08-09
Fix CBMC build on Linux and run CBMC in CI

mkannwischer:
#90 breaks the cbmc build on my linux machine. 
This PR
-  fixes the cbmc build (by re-applying an updated version of the patch from  https://github.com/NixOS/nixpkgs/blob/nixos-24.05/pkgs/applications/science/logic/cbmc/0001-Do-not-download-sources-in-cmake.patch)
 - additionally runs the CBMC proofs in CI; so that mistakes like this will be caught in the future.


mkannwischer:
This is likely due to https://github.com/NixOS/nixpkgs/blob/nixos-24.05/pkgs/applications/science/logic/cbmc/0001-Do-not-download-sources-in-cmake.patch not being applied (thanks @potsrevennil).

mkannwischer:
This SHOULD NOT work on MacOS either, as the build should not have network access. I am rather confused.

potsrevennil:
Seems like I was wrong before where I suggested to remove the patches in https://github.com/pq-code-package/mlkem-c-aarch64/pull/90, as I didn't look into what the patch is doing and thought that it is no longer needed. 

Turns out that it's just that the line numbers of git diff in the patch needed to update accordingly. So we will need to include our own patch into our repository to solve this (I have a working version now, I can submit an update afterwards)

But I'm also confused by why macOS can work before... . Without the patch, cbmc will download cadical in build phase, which is not allowed in nix, therefore it doesn't make sense that macOS worked before...

potsrevennil:
Since adding a patch file is required, for clarity, I moved all the cbmc related nix specification into cbmc folder, and used them via an entry file `cbmc/default.nix` following the nix convention.

For the patch file `0001-Do-not-download-sources-in-cmake.patch`, I retained the name from nixpkgs for consistency. When ever we're going to update cbmc version in the future, we should check if this patch still applied and changed it accordingly as well.

flake.nix
```diff
@@ -18,24 +18,7 @@
...
-                };
-                patches = [ ];
-              }); # 6.1.1
-              litani = pkgs.callPackage ./litani.nix { }; # 1.29.0
-              cbmc-viewer = pkgs.callPackage ./cbmc-viewer.nix { }; # 3.8
-
-              inherit (pkgs)
-                z3_4_12; # 4.12.5
-            };
+          cbmcpkg = pkgs.callPackage ./cbmc { };
```
mkannwischer, line 21:
```suggestion
          cbmcpkg = pkgs.callPackage ./cbmc { }; # 6.1.1
```

mkannwischer (commented):
Maybe add the version of cbmc as a comment?

hanno-becker (approved):

PR #98, opened 2024-08-26
Add basic ntt

potsrevennil:
depends on #100


hanno-becker:
@potsrevennil Can you hoist the build system refactoring into a separate PR?

potsrevennil:
@hanno-becker Thanks for reminding me. Yes, I'm also consider doing it. But I'm packed by other stuff now, will get back to this late this week or next week.

cbmc/litani.nix
```diff
@@ -4,7 +4,7 @@
 , fetchFromGitHub
 }:
 
-stdenvNoCC.mkDerivation rec {
+stdenvNoCC.mkDerivation {
```
hanno-becker, line 7: What's this about?

hanno-becker:
@potsrevennil I took the liberty to manually rebase this on top of `main`, as there were conflicts.

hanno-becker:
@potsrevennil 

I would like the configuration of `OPT` to be as simple as possible. As a user, I don't want to pass a complicated argument as `NTT123_4567`, but just say whether ASM should be enabled or not. Internally, or for more experienced users, we can also expose a fine-grained option, but I think this should be secondary.

Even when `OPT` is set, the AArch64 ASM should only be used when we are actually running on an AArch64 system. There should be a system header detecting this once and for all, exporting something like `USE_AARCH64_ASM` as the combination of running on an AArch64 system and `OPT` being set. Then, the actual source should cleanly rely on that flag only.

For benchmarking, I would suggest that we set `OPT=1` by default, and add new `-ref` benchmarks which force `OPT=0` and thereby fall-back to the reference implementation.


@mkannwischer You may want to chime in here.

hanno-becker:
Closing in favour of a PR on the main repo, to allow benchmarking.

cbmc/litani.nix
potsrevennil: `rec` stands for recursive attribute set, which allows the attributes in the attribute set to refer to each other. 
Since in the attribute set, we are not self referencing, it's actually redundant, and my linter was complaining, so I removed it by the way.

PR #109, opened 2024-09-08
Basic ntt

hanno-becker:
Moving #98 to main repo to enable benchmarking workflow.


hanno-becker:
@potsrevennil

I would like the configuration of OPT to be as simple as possible. As a user, I don't want to pass a complicated argument as NTT123_4567, but just say whether ASM should be enabled or not. Internally, or for more experienced users, we can also expose a fine-grained option, but I think this should be secondary.

Even when OPT is set, the AArch64 ASM should only be used when we are actually running on an AArch64 system. There should be a system header detecting this once and for all, exporting something like USE_AARCH64_ASM as the combination of running on an AArch64 system and OPT being set. Then, the actual source should cleanly rely on that flag only.

For benchmarking, I would suggest that we set OPT=1 by default, and add new -ref benchmarks which force OPT=0 and thereby fall-back to the reference implementation.

@mkannwischer You may want to chime in here.

potsrevennil (approved):

PR #110, opened 2024-09-08
Hoist benchmarking steps into reusable Github action

hanno-becker:


.github/actions/bench/action.yml
```diff
@@ -0,0 +1,57 @@
...
+      - NO
+      - PERF
+      - PMU
+      - M1
+  cflags:
+    description: CFLAGS to pass to compilation
+    required: true
+  archflags:
+    description: ARCHFLAGS to pass to compilation
+    required: true
```
potsrevennil, line 25: maybe we don't need the `required: true` ?

.github/workflows/bench.yml
```diff
@@ -16,13 +16,22 @@ jobs:
...
+          archflags: -mcpu=cortex-a72
+          cflags:
+          bench_extra_args:
         - system: rpi5
           name: Arm Cortex-A76 (Raspberry Pi 5) benchmarks
-          cmd: tests bench -c PERF --arch-flags "-mcpu=cortex-a76 -march=armv8.2-a" -v --output output.json
+          bench_pmu: PERF
+          archflags: "-mcpu=cortex-a76 -march=armv8.2-a"
+          cflags:
+          bench_extra_args:
```
potsrevennil, line 28: This can be removed if the default values are empty string for the two fields

hanno-becker: We are referring to those values below, though. I find it cleaner to set them to '' instead of relying on `${{ matrix.target.XXX }}` to default to the empty string if `XXX` is not set.

.github/actions/bench/action.yml
hanno-becker: Ok

potsrevennil (approved):

PR #113, opened 2024-09-09
Add cpucap.h header and detect AArch64 systems

hanno-becker:
Also, allow `FORCE_AARCH64` to double-check that a system is recognized as AArch64. Use this in all AArch64-based CI builds.


potsrevennil (approved):
Using external and excluded it is nice! 
And I confirmed that it worked

PR #115, opened 2024-09-09
Fix variable detection of recursive make

### LINKED ISSUE #114: The `--cflags` parameter in the `tests` script does not work
I observe the following behaviour:

```
> nix develop --experimental-features 'nix-command flakes' .#ci
$ tests nistkat
... (* OK *)
$ tests nistkat --cflags "-DSOMETHING_IRRELEVANT" 
...
INFO  > Nistkat Test
INFO  > CFLAGS=-DSOMETHING_IRRELEVANT make CROSS_PREFIX=  nistkat
test/gen_NISTKAT.c:33:5: error: call to undeclared function 'nist_kat_init'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    nist_kat_init(entropy_input, NULL, 256);
    ^
test/gen_NISTKAT.c:62:9: error: call to undeclared function 'nist_kat_init'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        nist_kat_init(seed, NULL, 256);
        ^
2 errors generated.
make[1]: *** [mk/rules.mk:22: test/build/mlkem512/test/gen_NISTKAT.c.o] Error 1
```

Digging a bit, it looks like the default `CFLAGS` is overwritten, not amended.

potsrevennil:
Fix  #114

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,32 +1,28 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
-.PHONY: all mlkem kat nistkat clean
-
-all: mlkem bench kat nistkat
+.PHONY: mlkem kat nistkat clean
 
 include mk/config.mk
+-include mk/$(MAKECMDGOALS).mk
 include mk/crypto.mk
 include mk/schemes.mk
-include mk/hal.mk
 include mk/rules.mk
-
-MAKEFLAGS = --no-print-directory
 
 mlkem: \
   $(MLKEM512_DIR)/bin/test_kyber512 \
   $(MLKEM768_DIR)/bin/test_kyber768 \
-  $(MLKEM1024_DIR)/bin/test_kyber1024 \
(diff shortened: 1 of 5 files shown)


hanno-becker:
@potsrevennil I still get the same error when trying `tests nistkat --cflags "-DSOMETHING_IRRELEVANT"`

hanno-becker:
I don't think this is specific to RNG, but to the entire CFLAGS from `config.mk` being overwritten.

potsrevennil:
Thanks for the check, did a more thoroughly check. Seems like there are some other minor issues  caused by recursive make happening as well. 
For example, CFLAGS is in fact re-appended everytime when make recursively. There are some caveats in recursive make, so in general I should try to avoid using it when possible. 😭

hanno-becker:
@potsrevennil It all seems to work now, thank you 🙏

hanno-becker (approved):

PR #127, opened 2024-09-12
Add native x86_64 test to CI

hanno-becker:


hanno-becker:
This works by virtue of x86_64 GH runners having gcc pre-installed. nix should be updated to include it explicitly for x86_64. This can be a follow-up PR, cc @potsrevennil.

.github/workflows/bench_ec2_any.yml
```diff
@@ -27,22 +27,21 @@ on:
         default:
       archflags:
         description: Custom ARCH flags for compilation
-        default: -mcpu=cortex-a76 -march=armv8.2-a
+        default: ''
```
hanno-becker, line 30: The default was problematic: If you trigger a run where no cflags are needed and you thus clear them, the default would be picked. There's no distinction between "no cflags" and "empty cflags".

scripts/tests
```diff
@@ -113,10 +113,11 @@ def base_run(
     verbose,
     run_as_root=False,
     exec_wrapper=None,
+    cross_prefix="",
 ):
     """Run the binary in all different ways"""
     cmd = [f"./{bin}"]
-    if force_qemu or (platform.system() == "Linux" and platform.machine() == "x86_64"):
+    if force_qemu or cross_prefix.strip() != "":
```
hanno-becker, line 120: This was needed to stop `tests` using `qemu` unconditionally on x86_64. Now it's only using it if cross compilation is used.

potsrevennil (approved):
All looks good to me!

PR #128, opened 2024-09-13
x86 64 gcc

potsrevennil:
In order to support native and cross compiled gcc on x86_64 linux, I added some extra nix shell, since mixing different version of gcc in a same shell, and correctly linking against standard library like glibc is a nightmare...

Therefore there are some changes for ci workflows accordingly, including ec2 related workflows. 

I figured maybe I should reopen the pr from branch in main repo to properly test the ec2 related workflows ?

### DIFF
--- a/.github/actions/bench/action.yml
+++ b/.github/actions/bench/action.yml
@@ -29,11 +29,11 @@ inputs:
   gh_token:
     description: GitHub access token
     required: true
-  use-nix:
-    description: Whether to run in the default Nix environment
-    default: "true"
+  nix-shell:
+    description: Run in the specified Nix environment if exists
+    default: "ci"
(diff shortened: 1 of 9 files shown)


.github/actions/bench/action.yml
```diff
@@ -29,11 +29,11 @@ inputs:
   gh_token:
     description: GitHub access token
     required: true
-  use-nix:
-    description: Whether to run in the default Nix environment
-    default: "true"
+  nix-shell:
+    description: Run in the specified Nix environment if exists
+    default: "ci"
```
hanno-becker, line 34: It's not possible to unset this, is it? If you pass `""`, the variable will default to `"ci"`, which is not intended.

PR #136, opened 2024-09-18
Add ec2 workflow for ci

potsrevennil:


.github/workflows/bench.yml
```diff
@@ -45,3 +46,38 @@ jobs:
...
+            cflags: -DFORCE_AARCH64
+    uses: ./.github/workflows/bench_ec2_reusable.yml
+    if: github.repository_owner == 'pq-code-package' && (github.event.label.name == 'benchmark' || github.ref == 'refs/heads/main')
+    with:
+      ec2_instance_type: ${{ matrix.target.ec2_instance_type }}
+      ec2_ami: ${{ matrix.target.ec2_ami }}
+      archflags: ${{ matrix.target.archflags }}
+      cflags: ${{ matrix.target.cflags }}
+      opt: ${{ matrix.opt.value }}
+      store_results: ${{ matrix.opt.value }} # Only store optimized results
```
mkannwischer, line 81: store results should only be true if it's the main branch, right?

mkannwischer (dismissed):

.github/workflows/bench.yml
potsrevennil: This is related to https://github.com/pq-code-package/mlkem-c-aarch64/pull/126/files

It was due to that we didn't disambiguate opt vs. non-opt in the stats pushed to github pages.

mkannwischer: that's okay, but it should still only store it on the main branch
currently it creates a lot of commit for G2 and G3 on the PRs - that is inconsistent with the other benchmarks.

potsrevennil: oh yes! your're right

mkannwischer:
Sorry you will have to rebase this after #139  got merged.

.github/workflows/bench.yml
hanno-becker: @potsrevennil Can you open a separate PR for this?

potsrevennil: ok

hanno-becker (approved):

PR #138, opened 2024-09-18
Fix nix installer cache issue

potsrevennil:
There is a cache issue in the Nix installer action, and since we don't really need to cache the nix installer, and there is no config for turning this off, I figured it'd be better to simply install the nix installer manually.

### DIFF
--- a/.github/actions/bench/action.yml
+++ b/.github/actions/bench/action.yml
@@ -32,9 +32,6 @@ inputs:
   nix-shell:
     description: Run in the specified Nix environment if exists
     default: "ci"
-  nix-not-installed:
-    description: Indicates whether to install Nix if it is not already present on the system.
-    default: 'true'
   custom_shell:
     description: The shell to use. Only relevant if no nix-shell specified
     default: "bash"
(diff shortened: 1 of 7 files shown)


potsrevennil:
Somehow, for our rpi4/rpi5 self-hosted runners, nix is pre-installed, but it cannot be found github default shell.
If I uninstalled nix, then the workflow will pass on the first try, then fail on the second.

Which means the startup files of teh default github shell didn't set the PATH variable correctly. (But I cannot wrap my head out which startup file I should modify...)

If I tryied to source `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh`, then rpi4 can read the `PATH` properly, but rpi5 still couldn't...

Fortunately workaround by referencing `nix` via its absolute path can solve the issue

potsrevennil:
Then there's still some weird error happening

Magic cache will sometimes encounter this retry timeout error (setting `continue-on-error` for this step can avoid the jobs failed at least)
https://github.com/pq-code-package/mlkem-c-aarch64/actions/runs/10926593405/job/30330814152

Then this is the weirdest error...
on macOS shared runner, when installing nix-installer, it said that there's already pre-exists configuration ...
Only encounter this error once, still need to keep an eye on this issue though...
https://github.com/potsrevennil/mlkem-c-aarch64/actions/runs/10926439327/job/30330297755

hanno-becker (approved):

PR #139, opened 2024-09-19
Add hybrid Neon/Neon Keccak-f1600, don't use scalar ASM on Cortex-A72

hanno-becker:


fips202/asm/aarch64/keccak_f1600_x2_v8a_v84a_hyrid.S
mkannwischer, line 1: There is a typo in the file name here

hanno-becker: Fixed

hanno-becker:
Benchmarks suggest that something isn't yet working as intended. Both Gv2 and Gv3 get slower.

mkannwischer (approved):
Thanks! This looks great and works on my machine. 

We somehow need to add to the CI output which implementations is actually used, but we can do that separately. I'll open an issue for that.

PR #143, opened 2024-09-19
Add experimental SLOTHY-optimized NTT

hanno-becker:
This needs further refinement, but it's a start.


mlkem/poly.c
```diff
@@ -612,6 +614,7 @@ void poly_tomont(poly *r)
 *
 * Arguments:   - poly *r: pointer to input/output polynomial
 **************************************************/
+#if !defined(MLKEM_USE_AARCH64_ASM)
```
mkannwischer, line 617: Please move this pre-processor conditional inside of `poly_reduce` - just like for `ntt()`

mkannwischer (approved):
Thanks this looks great - I tested this on my x86 and it works fine.
I also looked at the outputs of the Raspberry Pis in CI and it looks good.

One small nit, then you can go ahead and merge it.

PR #144, opened 2024-09-20
Add cache for scaled polys/polyvecs in NTT domain

hanno-becker:
Asymmetric multiplication [1] speeds up matrix-vector multiplication in NTT-domain via additional caching on the polyvec argument.

This commit provides a C implementation of this strategy and paves the way for clean integration of ASM.

It introduces types poly_mulcache and polyvec_mulcache for precomputed data used by poly/polyvec base multiplication. Those types are meant to be opaque to the caller, an exposed only to allow their construction on the stack.

The mulcache for a poly/polyvec can be computed via `poly[vec]_mulcache_compute()`. It is then passed to base multiplication routines as an additional argument.

Apart from caching repeated computations, another benefit of asymmetric multiplication is that it allows lazy Montgomery reduction during base multiplication, and even matrix-vector multiplication. This will be fully leveraged on the ASM level later on. On the C-level, we merely use lazy reduction for a small speedup of the cached basemul.

[1]: https://eprint.iacr.org/2021/986
     Neon NTT: Faster Dilithium, Kyber, and Saber on Cortex-A72 and Apple M1

[//]: # (SPDX-License-Identifier: CC-BY-4.0)

PR #146, opened 2024-09-20
Add AArch64 ASM for matrix-vector base multiplication

hanno-becker:
Based on ~#144~ ~#143~ and https://github.com/pq-code-package/mlkem-c-aarch64/pull/151.

Fixes https://github.com/pq-code-package/mlkem-c-aarch64/issues/152


mlkem/asm/aarch64/optimize.sh
```diff
@@ -6,20 +6,92 @@ set -e
...
 
 slothy-cli Arm_AArch64 $TARGET \
-  poly_clean.S -o poly_opt.S \
+  poly_opt.S -o poly_opt.S \
   -r poly_reduce_asm_clean,poly_reduce_asm_opt \
   -l loop_start \
   -c sw_pipelining.enabled=true \
   -c inputs_are_outputs \
-  -c reserved_regs="[x0--30,v10--v31,sp]" \
+  -c reserved_regs="[x0--30,v8--v15,sp]" \
```
mkannwischer, line 63: Why is this `x0--30`, but `v8-v15`? Looks inconsitent. Is this a SLOTHY bug?

mlkem/asm/aarch64/optimize.sh
mkannwischer, line 1: It does not look like a good idea to do everything in a single script. 
You probably don't want to re-run all optimization for every change. 
At least it should be possible to parameterize it to only run a single optimization.

mlkem/asm/asm.h
```diff
@@ -16,18 +16,60 @@ void intt_asm_opt(int16_t *);
 void poly_reduce_asm_clean(int16_t *);
 void poly_reduce_asm_opt(int16_t *);
 
+void poly_tomont_asm_clean(int16_t *);
+void poly_tomont_asm_opt(int16_t *);
+
+void poly_mulcache_compute_asm_clean(int16_t *, int16_t *, int16_t *, int16_t *);
+void poly_mulcache_compute_asm_opt(int16_t *, int16_t *, int16_t *, int16_t *);
+
+void poly_tobytes_asm_clean(uint8_t *r, const int16_t *a);
```
mkannwischer, line 25: What's the reasonf or not having an opt version of `poly_tobytes_asm_clean`?

mlkem/asm/asm.h
```diff
@@ -16,18 +16,60 @@ void intt_asm_opt(int16_t *);
...
+void polyvec_basemul_acc_montgomery_cached_asm_k4_clean(
+    int16_t *r, const int16_t *a, const int16_t *b, const int16_t *b_cache);
+
+void polyvec_basemul_acc_montgomery_cached_asm_k2_opt(
+    int16_t *r, const int16_t *a, const int16_t *b, const int16_t *b_cache);
+void polyvec_basemul_acc_montgomery_cached_asm_k3_opt(
+    int16_t *r, const int16_t *a, const int16_t *b, const int16_t *b_cache);
+void polyvec_basemul_acc_montgomery_cached_asm_k4_opt(
+    int16_t *r, const int16_t *a, const int16_t *b, const int16_t *b_cache);
+
```
mkannwischer, line 40: Since you are using pre-processor conditionals to only include the corresponding parameter set in `polyvec_opt.S`, I don't think it is necessary to include the parameter set in the function name, is it? 
I'd prefer to keep it simpler here.

mlkem/poly.c
```diff
@@ -272,6 +272,7 @@ void poly_decompress(poly *r, const uint8_t a[KYBER_POLYCOMPRESSEDBYTES])
 *              - r: pointer to output byte array
 *                   (of KYBER_POLYBYTES bytes)
 **************************************************/
+#if !defined(MLKEM_USE_AARCH64_ASM)
```
mkannwischer, line 275: Move the pre-procesor conditional inside of `poly_tobytes`

mlkem/poly.c
```diff
@@ -621,6 +627,7 @@ void poly_basemul_montgomery_cached(poly *r, const poly *a, const poly *b, const
 *
 * Arguments:   - poly *r: pointer to input/output polynomial
 **************************************************/
+#if !defined(MLKEM_USE_AARCH64_ASM)
```
mkannwischer, line 630: same here

mlkem/poly.c
```diff
@@ -698,6 +711,7 @@ void poly_sub(poly *r, const poly *a, const poly *b)
 * Arguments: - poly_mulcache *x: pointer to output cache.
 *            - const poly *a: pointer to input polynomial
 **************************************************/
+#if !defined(MLKEM_USE_AARCH64_ASM)
```
mkannwischer, line 714: same here. Let's do this consistently.

mlkem/poly.c
```diff
@@ -707,3 +721,38 @@ void poly_mulcache_compute(poly_mulcache *x, const poly *a)
...
+    1461, -1461, -642, 642, 939, -939, -1021, 1021, -892, 892, -941, 941, 733, -733,
+    -992, 992, 268, -268, 641, -641, 1584, -1584, -1031, 1031, -1292, 1292, -109, 109,
+    375, -375, -780, 780, -1239, 1239, 1645, -1645, 1063, -1063, 319, -319, -556, 556,
+    757, -757, -1230, 1230, 561, -561, -863, 863, -735, 735, -525, 525, 1092, -1092,
+    403, -403, 1026, -1026, 1143, -1143, -1179, 1179, -554, 554, 886, -886, -1607, 1607,
+    1212, -1212, -1455, 1455, 1029, -1029, -1219, 1219, -394, 394, 885, -885, -1175, 1175
+};
+
+const int16_t zetas_mulcache_twisted_asm[256] =
+{
```
mkannwischer, line 739: Can these arrays be `static const`?

mlkem/polyvec.c
```diff
@@ -237,6 +239,9 @@ void polyvec_basemul_acc_montgomery_cached(poly *r, const polyvec *a, const poly
     }
 
     poly_reduce(r);
+    #else /* MLKEM_USE_AARCH64_ASM */
+    polyvec_basemul_acc_montgomery_cached_asm((int16_t *) r, (int16_t *) a, (int16_t *) b, (int16_t *) b_cache);
```
mkannwischer, line 243: Do we prefer `(int16_t *) r` or `r->coeffs`? Let's do one of the two consistently. I have a slight preference for the latter. It's unlikely that the `poly` struct will change, but who knows.

mkannwischer (changes requested):
A few small comments and questions. 
Rest looks great to me.

mlkem/asm/aarch64/optimize.sh
hanno-becker: Can this be a follow-up? For the moment I don't care too much about convenience in this script. When we add the flexibility to choose the uArch, it's a good time to think a bit more about the design.

mlkem/asm/asm.h
hanno-becker: Some instructions are not yet supported by SLOTHY's uArch model. Let's address this in a follow-up.

mlkem/asm/asm.h
hanno-becker: Currently, the name is needed to disambiguate the functions to SLOTHY.

mlkem/poly.c
hanno-becker: Yes, will fix.

mlkem/polyvec.c
hanno-becker: Ok

mlkem/asm/aarch64/optimize.sh
hanno-becker: I've made it more consistent, but the intention is to forbid the use of callee-saved registers in clean ASM which does not stash them. And the decision of when to stash in clean ASM is based on whether it seems likely that optimized code would need to use them.

mlkem/poly.c
hanno-becker: I actually think it's cleaner this way. Also, for functions like the basemul ASM which need extra static variables, you end up with multiple if-def blocks if you mandate one within the function-body itself.

I'm happy to use the above style (single if-def around function def) everywhere if you're happy with it.

mlkem/poly.c
hanno-becker: As above.

mlkem/poly.c
hanno-becker: I've done that for now, have a look.

mlkem/asm/aarch64/optimize.sh
mkannwischer: Yes, follow-up PR is fine. It makes sense to give it more thought when we optimize for multiple difference uArchs.

mlkem/poly.c
mkannwischer: Yes, I'm fine with that.

mkannwischer (approved):
Thanks for making the adjustments! 
Now it looks good to me!

PR #151, opened 2024-09-21
Force alignment of structures

### LINKED ISSUE #145: Consider alignment of typedefs and struct fields
Measurable performance gains can be achieved by aligning large data items (e.g. polynomial arrays) on the boundary of a CPU's L1 D-Cache line, which is typically 64 bytes (Graviton 2) or other power of two.

Consider defining macro PREFERRED_DCACHE_ALIGNMENT with a sensible default value, but possibly overridden for particular micro-architectures.

use C11's <stdalign.h> definitions that define the "alignas()" macro for objects and struct fields.

hanno-becker:
Unaligned structures can incur significant performance penalties during memory operations.

Fixes #145 

### DIFF
--- a/mlkem/common.h
+++ b/mlkem/common.h
@@ -1,0 +1,7 @@
+// SPDX-License-Identifier: Apache-2.0
+#ifndef COMMON_H
+#define COMMON_H
+
+#define ALIGN(x) __attribute__((aligned(x)))
+
+#endif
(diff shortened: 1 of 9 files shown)


mkannwischer (commented):
Did you exclude the buffers in `kem.c` and `poly.c` on purpose or should those also be aligned?

Besides that this looks good to me and works well. Thanks!

hanno-becker:
@mkannwischer Added alignment constraints in `poly.c` and `kem.c`.

mkannwischer (approved):
You missed `poly_getnoise_eta1`, but that one is not used anymore anywhere anyway. I'll remove that in a separate PR.

PR #154, opened 2024-09-23
Run CBMC on EC2 runner in CI and guard by label

hanno-becker:
Fixes https://github.com/pq-code-package/mlkem-c-aarch64/issues/117

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ on:
     branches: ["*"]
   pull_request:
     branches: ["main"]
+    types: [ "labeled" ]
 jobs:
   build_kat:
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -96,17 +97,26 @@ jobs:
...
+            ec2_ami: ubuntu-latest (custom AMI)
+            ec2_ami_id: ami-08ddb0acd99dc3d33
+            cbmc: 'true'
     name: ${{ matrix.target.name }}
     permissions:
       contents: 'read'
       id-token: 'write'
     uses: ./.github/workflows/ci_ec2_reusable.yml
-    if: github.repository_owner == 'pq-code-package' || github.ref == 'refs/heads/main'
+    if: github.repository_owner == 'pq-code-package' && (github.event.label.name == 'ci' || github.ref == 'refs/heads/main')
```
mkannwischer, line 115: I don't think this corresponds to what the title of the PR says. 
This actually disables all CI for PRs on `AMD EPYC 4th gen (t3a)` and `Intel Xeon 4th gen (t3)` unless you add the `ci` label. Was that intentional? 
If it was not you will have to change it such that only the cbmc part depends on the `ci` label.

mkannwischer (changes requested):
I am a little confused - see my comment.

.github/workflows/ci.yml
hanno-becker: Fair point. I don't think we want EC2 runners to be triggered by remote PRs, which is what was the case previously and is fixed here. I will amend the PR title and commit message.

hanno-becker: I don't know how to run the EC2 CI _always_ (without labels) on PRs from the main repository.

mkannwischer (approved):
Okay, makes sense. Thanks.

PR #159, opened 2024-09-24
DRAFT: Bounds checks

hanno-becker:

### DIFF
--- a/mlkem/debug.c
+++ b/mlkem/debug.c
@@ -1,0 +1,43 @@
+
+#include "debug.h"
+
+#if defined(MLKEM_DEBUG)
+
+static char debug_buf[256];
+
+void mlkem_debug_check_bounds(const char *file,
+                             int line,
+                             const char *description,
+                             const int16_t *ptr, unsigned len,
+                             int16_t lower_bound_inclusive,
+                             int16_t upper_bound_inclusive)
+{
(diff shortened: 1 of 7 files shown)


mlkem/debug.c
mkannwischer, line 1: Can we move this to a separate directory? 
I'd prefer not mix this in with the mlkem implementation.

hanno-becker: Good idea!

rod-chapman: Why are we doing this? Isn't all this runtime bounds checking rendered redundant by proof of type safety? I am all for a "belt and braces" approach but these feel like repeating ourselves - we're going to end up with preconditions that sat
 "x must be in a .. b"
immediately followed by a runtime assertion that says 
 "x must be in a .. b"

What is the proposed discipline here?

hanno-becker: > I am all for a "belt and braces" approach but these feel like repeating ourselves 

Isn't this what "belt and braces" is about? Yes, _once_ we prove the respective conditions in CBMC and/or on the ASM, they are redundant. But we are expecting to work with a variety of (ASM) implementations in the future, and debug-assertions help us catch bounds violations early in the process.

hanno-becker:
Has been integrated into another PR.

PR #161, opened 2024-09-25
Prove poly decompress

rod-chapman:
This PR updates the CBMC proofs of poly_decompress() to use unbounded and contract-based verification.

All tests pass.

"format" run with no changes.

### DIFF
--- a/mlkem/poly.c
+++ b/mlkem/poly.c
@@ -220,10 +220,8 @@ void poly_compress(uint8_t r[KYBER_POLYCOMPRESSEDBYTES], const poly *a)
 **************************************************/
 void poly_decompress(poly *r, const uint8_t a[KYBER_POLYCOMPRESSEDBYTES])
 {
-    unsigned int i;
-
     #if (KYBER_POLYCOMPRESSEDBYTES == 128)
-    for (i = 0; i < KYBER_N / 2; i++)
+    for (int i = 0; i < KYBER_N / 2; i++)
         __CPROVER_assigns(i, __CPROVER_object_whole(r))
         __CPROVER_loop_invariant(i >= 0)
         __CPROVER_loop_invariant(i <= KYBER_N / 2)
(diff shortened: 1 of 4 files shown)


mlkem/poly.c
```diff
@@ -224,34 +224,68 @@ void poly_decompress(poly *r, const uint8_t a[KYBER_POLYCOMPRESSEDBYTES])
...
+
+
+
     #elif (KYBER_POLYCOMPRESSEDBYTES == 160)
-    unsigned int j;
-    uint8_t t[8];
     for (i = 0; i < KYBER_N / 8; i++)
+        __CPROVER_assigns(i, __CPROVER_object_whole(r))
+        __CPROVER_loop_invariant(i >= 0)
+        __CPROVER_loop_invariant(i <= KYBER_N / 8)
```
cothan, line 246: Should it be at line 246: `i < KYBER_N/8`?

cothan (changes requested):

mlkem/poly.c
```diff
@@ -224,34 +224,68 @@ void poly_decompress(poly *r, const uint8_t a[KYBER_POLYCOMPRESSEDBYTES])
...
 
-        for (j = 0; j < 8; j++)
-        {
-            // REF-CHANGE: Truncation happened before
-            r->coeffs[8 * i + j] = ((uint32_t) t[j] * KYBER_Q + 16) >> 5;
-        }
+        // Decompress 5 8-bit bytes (so 40 bits) into
+        // 8 5-bit values stored in t[]
+        t = 0x1F &  (a[offset + 0] >> 0);
+        r->coeffs[8 * i]     = ((uint32_t) t * KYBER_Q + 16) >> 5;
```
hanno-becker, line 267: What are the obstacles to leaving the code as-is? You say it eases the proof -- but can you get the proof through without making changes?

mlkem/poly.h
```diff
@@ -81,7 +81,15 @@ __CPROVER_assigns(__CPROVER_object_whole(r));
 
 
 #define poly_decompress KYBER_NAMESPACE(poly_decompress)
-void poly_decompress(poly *r, const uint8_t a[KYBER_POLYCOMPRESSEDBYTES]);
+void poly_decompress(poly *r, const uint8_t a[KYBER_POLYCOMPRESSEDBYTES])
+/* *INDENT-OFF* */
+__CPROVER_requires(a != NULL)
+__CPROVER_requires(__CPROVER_is_fresh(a, KYBER_POLYCOMPRESSEDBYTES))
```
hanno-becker, line 87: What does `__CPROVER_is_fresh` mean?

mlkem/poly.c
rod-chapman: Afraid not... CBMC implicitly places the invariant just _before_ the final loop exit test, so I can indeed be equal to KYBER_N/8 just before the final loop exit.

mlkem/poly.c
rod-chapman: The original code wasn't very pretty to start with.. it assigns 8 elements to a local array, and only then copies that array into r->coeffs in a loop. I will try to restore and prove the original to see how it comes out.

mlkem/poly.h
rod-chapman: See https://diffblue.github.io/cbmc/contracts-memory-predicates.html
it basically means "This pointer has been allocated and is pointing at N bytes of memory"

rod-chapman:
I have committed it with the inner loop restored, and its loop invariant. See what you think.

mlkem/poly.c
hanno-becker: Thank you!

PR #165, opened 2024-09-26
Inline scalar [de]compression functions

hanno-becker:
Based on ~#162~ #168  

This PR inlines scalar (de)compression functions for better performance.

The change _should_ be agnostic to proofs. Unfortunately, however, CBMC does not seem to be able to handle function contract for inlined functions, so instead scalar functions are inlined to their call-sites again.


cothan:
Looks good to me, can you fix the CI?

hanno-becker:
Something's tripping up CBMC -- not sure yet what's going on...

mkannwischer (approved):
Thanks @hanno-becker. This shaves off a few tousand cycles: 
E.g., 

A55: ML-KEM-768
encaps 139192 -> 137294
decaps 156231 -> 152146

A72: ML-KEM-768
encaps 123494 -> 122634
decaps 135417 -> 133497

Unfortunately, the CBMC proofs take MUCH longer now. Before
` KYBER_K=2 ./run-cbmc-proofs.py --summarize;`
took 30 seconds for me, now I'm at 15 minutes. Do you also see that behavior?
Can you please open an issue for that?

hanno-becker:
@mkannwischer Thanks for catching this -- this is a problem.

I am quite surprised by this -- my understanding is that CBMC already inlined the functions `proof_compress()`, and it's only for `poly_decompress()` that we stopped using a function contract with the inlining.

hanno-becker:
@rod-chapman Any insights?

hanno-becker:
@mkannwischer The CI took merely 5min, which is surprising.

hanno-becker:
Let's merge this and sort out the CBMC performance later. It's good enough for the CI at least.

PR #166, opened 2024-09-26
Manually break down ld2/st2 into uOps

hanno-becker:
Most microarchitectures do this anyway, and by manually breaking it down in assembly, SLOTHY has more flexibility rearranging things.


cothan (approved):
The IPC number increases look good to me. Bravo Slothy!

PR #170, opened 2024-09-29
Add binary for benchmarking MLKEM components

hanno-becker:
Until now, `tests bench` would only give performance numbers for the MLKEM toplevel API. While most interesting to the user, the developer needs more insight into where cycles are spent.

This commit introdues `tests bench --components`, backed by a new benchmarking binary `bench_components_kyber{512,768,1024}`, benchmarking various performance critical components of MLKEM.

For now, there is one binary per security level, despite all benchmarks except for the base multiplication being the same for all levels. This is mostly to fit easily into the existing makefile framework, and will likely need refining.


mkannwischer (approved):
Thanks @hanno-becker. This is great! Very nice to see both the clean and the opt implementations next to each other. 

I looked at the output for all the CI runnners and tested it on my x86 (which is meaningless, but there at least it does not break anything). 

Regarding the split between security levels: I'd keep that as is. It may make sense to add some point benchmark something like matrix-vector multiplications for which it makes sense to keep the separation between security levels.

PR #172, opened 2024-10-01
tests: Fix bug in export of benchmarking results

hanno-becker:
Symptom: `scripts/tests` does no longer write benchmarking outputs, which leads to subsequent invocations of the `benchmark-action` failing. This only happens _after_ merge to `main`, as the call to `benchmark-action` is skipped otherwise.

Bug: A previous commit replaced an `if output` check by `if output is True` in the wrong assumption that `output` is a boolean. Instead, it's `None` or a string, so the previous version of the check is necessary.

This commit reverts the respective change.

### DIFF
--- a/scripts/tests
+++ b/scripts/tests
@@ -493,7 +493,7 @@ def bench(
         ],
     )
 
-    if output is True and components is False:
+    if output is not None and components is False:
         import json
 
         with open(output, "w") as f:


mkannwischer (approved):
Oh, right. Damn. 
Before this was `if output:` which is bad practice. 

The commit that changes this is https://github.com/pq-code-package/mlkem-c-aarch64/commit/b3d28218ab9ebc3414d85a4a5bf58a44db6b50dd

PR #178, opened 2024-10-01
Fix CI workflow trigger

potsrevennil:
Run the CI workflow automatically as before, 
enable re-running it by adding a label, 
and ensure the CI ec2 workflow only runs if the PR is created from the main repository.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,7 @@
 
 name: CI
 on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


mkannwischer:
Why do we need a label for re-running CI? We can just re-run in the Github Actions UI, can't we?

hanno-becker:
We need it for fork PRs, and we may as well use it for main PRs, too, to simplify the restart.

PR #182, opened 2024-10-02
Add asm for rejection sampling

hanno-becker:


mlkem/asm/aarch64/rej_uniform_asm_clean.S
```diff
@@ -0,0 +1,585 @@
...
+
+    ASM_LOAD(bit_table, bit_table_data)
+    ASM_LOAD(table_idx, table_data)
+
+    /* Vectorize code start */
+
+    ldr         bits_q, [bit_table]
+    movz        tmp, #3329
+    dup         kyber_q.8h, tmp
+    round_48x   bound, buflen
```
mkannwischer, line 405: Can it ever happen that bound is zero here,  i.e., buflen < 48?

mlkem/asm/aarch64/rej_uniform_asm_clean.S
```diff
@@ -0,0 +1,585 @@
...
+#if defined(MLKEM_USE_AARCH64_ASM)
+
+// Needed to provide ASM_LOAD directive
+#include "common.i"
+
+/* Return multiple of 48: (n/48) * 48 */
+.macro round_48x    dst, src
+    mov     w8, #48
+    udiv    \dst, \src, w8
+    mul     \dst, \dst, w8
```
mkannwischer, line 29: I'm not a fan of this at all.

mlkem/asm/aarch64/rej_uniform_asm_clean.S
```diff
@@ -0,0 +1,585 @@
...
+*
+* Description: Full SIMD lane, run rejection sampling on uniform random bytes to generate
+*              uniform random integers mod q
+*
+* Arguments:   - int16_t *r:          pointer to output buffer
+*              - unsigned int len:    requested number of 16-bit integers
+*                                     (uniform mod q)
+*              - const uint8_t *buf:  pointer to input buffer
+*                                     (assumed to be uniform random bytes)
+*              - unsigned int buflen: length of input buffer in bytes
```
mkannwischer, line 14: There seem to be some assumptions about the buflen here, right? 
Could you document those please? 
Also if the assumption is that buflen is divisible by 24 and is at least 48, then maybe it's better to pass it in as a number of 24-byte blocks? That would make it harder to misused this code and make it a little bit easier to read.

mlkem/rej_uniform.c
```diff
@@ -0,0 +1,92 @@
...
+* Arguments:   - int16_t *r:          pointer to output buffer
+*              - unsigned int len:    requested number of 16-bit integers
+*                                     (uniform mod q)
+*              - const uint8_t *buf:  pointer to input buffer
+*                                     (assumed to be uniform random bytes)
+*              - unsigned int buflen: length of input buffer in bytes
+*
+* Returns number of sampled 16-bit integers (at most len)
+**************************************************/
+static unsigned int rej_uniform_scalar(int16_t *r,
```
mkannwischer, line 22: Do you think it makes sense to keep this C code closer to what the assembly is doing? That would make it easier to follow what's happening in the asm.

mkannwischer (changes requested):
Thanks @hanno-becker and @cothan!
Performance-wise this is looking quite amazing. 
Looking at the Cortex-A72 and ML-KEM-768:
keygen: 104714 -> 91488
encaps: 116588 -> 103253
decaps:  127050 -> 114224

I think it is worthwhile to spend at least a few words on describing what the strategy is for vectorizing rejection sampling here. As it reads right now, it's impossible to understand what is going on. 

Also, can you please add this to the components benchmarks?

mlkem/asm/aarch64/rej_uniform_asm_clean.S
cothan: This is the division by constant, where is `n` is the `buflen` input.

mlkem/asm/aarch64/rej_uniform_asm_clean.S
cothan: yes it can, the ASM version is batch sampling, which consumes 48 (loop48) or 24 (loop24) bytes at a time. When the `buflen` input is larger than 48, it goes to loop48, if it less than 48, it goes to loop24, if it less than 24, it goes to the scalar version.

hanno-becker: But the 48-loop is never skipped?

mlkem/rej_uniform.c
cothan: I had the intrinsic version, however, I think we should keep the scalar version as-is, and either ASM or instrinsic, preferably ASM version.

mlkem/asm/aarch64/rej_uniform_asm_clean.S
cothan: I forget to add the check, it should check first then execute. I'm adding it now.

hanno-becker: @cothan I'm going over everything right now and fixing it, no need to take action from your end.

mlkem/rej_uniform.c
hanno-becker: I agree, we want to stay close to the reference implementation.

mlkem/asm/aarch64/rej_uniform_asm_clean.S
```diff
@@ -284,15 +285,46 @@ table_data:
...
+.macro push_stack
+        sub sp, sp, #STACK_SIZE
+.endm
+
+.macro pop_stack
+        add sp, sp, #STACK_SIZE
+.endm
+
+.macro save_gprs
+    str     x8,       [sp, #(STACK_OFFSET_GPRS + 0)]
```
hanno-becker, line 315: @cothan It looks like we only need to stash `x30` -- `x8`, `x16` and `x17` don't have to be kep, and `x29` is not changed?

hanno-becker:
The code makes the assumption that the output buffer has length `< KYBER_N`. This is true for all call-sites, but still should be documented. For a larger buffer, the stack will overflow I think.

mlkem/asm/aarch64/rej_uniform_asm_clean.S
mkannwischer: Sure, everything is public in the rejection sampling, but this is not the fastest way to compute division by 48.
Additionally, the current code cannot handle anything that's not a multiple of 24, so there is really no reason to do that division here.

mlkem/asm/aarch64/rej_uniform_asm_clean.S
cothan: I don't remember much, yes, if the x29 isn't changed then feel free to adjust the file as you see fit.

mlkem/asm/aarch64/rej_uniform_asm_clean.S
hanno-becker: I removed this in the latest commit

mlkem/asm/aarch64/rej_uniform_asm_clean.S
hanno-becker: I rewrote things so that the buffer length can be anything (hopefully...)

mlkem/asm/aarch64/rej_uniform_asm_clean.S
```diff
@@ -0,0 +1,644 @@
...
+#include "config.h"
+#include "params.h"
+#if defined(MLKEM_USE_AARCH64_ASM)
+
+// Needed to provide ASM_LOAD directive
+#include "common.i"
+
+.data
+table_data:
+    .byte -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1 // 0
```
mkannwischer, line 28: Please briefly explain what this table is for. Maybe add the code that generated this table?

mlkem/asm/aarch64/rej_uniform_asm_clean.S
```diff
@@ -0,0 +1,644 @@
...
+.align 4
+.global     rej_uniform_asm_clean
+.global     _rej_uniform_asm_clean
+rej_uniform_asm_clean:
+_rej_uniform_asm_clean:
+    push_stack
+    save_gprs
+
+    // TODO: This routine seems to write the results on the stack first, which is statically sized.
+    //       It therefore only works for input buffers of bounded size, which needs to be documented.
```
mkannwischer: You can remove this now since you have added the check, right?

mlkem/asm/aarch64/rej_uniform_asm_clean.S
```diff
@@ -0,0 +1,644 @@
...
+    .byte 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15           // 255
+
+bit_table_data:
+    .short 0x1, 0x2, 0x4, 0x8, 0x10, 0x20, 0x40, 0x80
+
+// Maximum size of int16 output array
+#define REJ_UNIFORM_ASM_MAX_OUTPUT KYBER_N
+
+#define STACK_LOCS 2
+#define STACK_SIZE_TMP_OUTPUT (2*REJ_UNIFORM_ASM_MAX_OUTPUT + 64) // We overwrite by max. 64 bytes in loop48
```
mkannwischer, line 292: That comment is not quite correct, is it? 
The worse case is that only 1 coefficient it still needed and 48 bytes turn into 32 16-bit integers to be stored in which case we overrun by 62 bytes?

mkannwischer (approved):
Thanks @hanno-becker for refactoring this. 
Some small nits left, rest looks good!

mlkem/asm/aarch64/rej_uniform_asm_clean.S
hanno-becker: This is explained in the code

mlkem/asm/aarch64/rej_uniform_asm_clean.S
hanno-becker: Yes

mkannwischer:
Looked at the performance again after the recent changes: 

Cortex-A72 and ML-KEM-768:
keygen: 104714 -> 91488 -> 90270 
encaps: 116588 -> 103253  -> 102599
decaps: 127050 -> 114224 -> 113429 

It got even faster :)

mlkem/asm/aarch64/rej_uniform_asm_clean.S
hanno-becker: Yes, but for alignment we wouldn't want to use only 62 bytes here.

mlkem/asm/aarch64/rej_uniform_asm_clean.S
hanno-becker: Removed

Issue #155, opened 2024-09-23
Skipped CI jobs have wrong name


mkannwischer:
![image](https://github.com/user-attachments/assets/654386ab-db83-474a-9207-a20c651fd30b)

See, e.g., https://github.com/pq-code-package/mlkem-c-aarch64/pull/144
https://github.com/pq-code-package/mlkem-c-aarch64/actions/runs/10988042686/job/30503873249?pr=144

potsrevennil:
There is a discussion on github community here
https://github.com/orgs/community/discussions/13261

It seems like it's currently not possible to get reid of this issue

PR #187, opened 2024-10-03
Fix CI skipped job

potsrevennil:
Seemed to found a way to fix https://github.com/pq-code-package/mlkem-c-aarch64/issues/155 by using matrix exclude

### DIFF
--- a/.github/workflows/ci_ec2_reusable.yml
+++ b/.github/workflows/ci_ec2_reusable.yml
@@ -104,15 +104,15 @@ jobs:
       matrix:
         tests:
         - name: Functional tests
-          action: ./.github/actions/functest
+          action: functest
           nix-shell: ci
           run: ${{ inputs.functest }}
         - name: Linting
-          action: ./.github/actions/lint
+          action: lint
           nix-shell: ci-linter
           run: ${{ inputs.lint }}
         - name: CBMC
-          action: ./.github/actions/cbmc
+          action: cbmc
           nix-shell: ci-cbmc
           run: ${{ inputs.cbmc }}
         exclude:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_ec2_reusable.yml
```diff
@@ -95,51 +95,63 @@ jobs:
...
+          - {tests: {
+              name: CBMC,
+              action: cbmc,
+              nix-shell: ci-cbmc,
+              run: false,
+            }}
     steps:
       - uses: actions/checkout@v4
-      - uses: ./.github/actions/cbmc
+      - uses: jenseng/dynamic-uses@v1
```
hanno-becker, line 139: Why is this needed?

potsrevennil: as I pointed out in #155, currently with the `if` statement, github will still show the skipped jobs of a workflow in UI.

While with the `matrix` `exclude`, we can hide the skipped jobs from UI, but since the jobs we are running (functest/lint/cbmc) use different action, we'd need to dynamic load them.

And currently github doesn't support expression evaluation in `uses`, so the only way would be to write an action.yml on the fly as a middleware to dynamic pointing to use the specify action, which is what `jenseng/dynamic-uses` have done.

hanno-becker: Ok!

potsrevennil: btw, I haven't yet confirmed, but I think that when using matrix, if the whole matrix is excluded, there'd still be a dummy job executing (where all the matrix fields are evaluated to empty string).

In this case, I think we'd at least run one of functest/lint/cbmc job, so it'd be fine, but it's not the case for ec2 related job (which we'd like to exclude the jobs if triggered from fork). 

Therefore I didn't apply this to ec2 related jobs

hanno-becker (approved):
It's sad that such hard-to-read trickery is needed, but well... thanks Han for looking into it!

PR #192, opened 2024-10-04
Start adding AVX2 native backend

hanno-becker:
This PR commences the construction of an x86_64 native backend based on the AVX2 code from the Kyber reference implementation.

As a start, only rejection sampling is added.


mlkem/native/x86_64/profiles/ref.h
mkannwischer, line 1: ref is the wrong name for this profile. Can we call it default? 
I assume we will only have one profile for AVX2 for the foreseeable future?

hanno-becker: I called it `ref` for "reference implementation" (of Kyber).

mkannwischer: But it is not the reference implementation? It's the official AVX2 implementation.
The reference implementation is C only. 
The reason they put that in the same repo is because NIST wantes to see both ref and x86 for the submission.

hanno-becker: Ok, changed

mlkem/native/x86_64/README.md
```diff
@@ -0,0 +1,4 @@
+[//]: # (SPDX-License-Identifier: CC-BY-4.0)
+
+This directory contains the native x86_64 arithmetic backend for ML-KEM provided by the [AVX2
+branch](https://github.com/pq-crystals/kyber/tree/main/avx2) of the Kyber reference implementation.
```
mkannwischer, line 4:
```suggestion
This directory contains the native x86_64 arithmetic backend for ML-KEM provided by the official [AVX2
implementation](https://github.com/pq-crystals/kyber/tree/main/avx2) of the Kyber team.
```

mlkem/rej_uniform.c
```diff
@@ -70,7 +70,7 @@ unsigned int rej_uniform(int16_t *r, unsigned int len, const uint8_t *buf,
   // Sample from large buffer with full lane as much as possible.
   ret = rej_uniform_native(r, len, buf, buflen);
   if (ret != -1)
-      return (unsigned) ret;
+    return (unsigned)ret;
```
mkannwischer, line 73: I am confused by the formatting changes. Shouldn't this already be enforced by CI before?

mkannwischer (approved):
Thanks @hanno-becker.
This is a good start regarding adding AVX2 support. 

The benchmarking setup is not ideal - we should disable hyperthreading and TurboBoost to get performance metrics that we can compare to the literature. 
But that can be done in a separate PR. I'll open an issue. 

I am a bit annoyed that I have to run 
```
tests nistkat -cp '' --cflags '-mavx2 -mbmi2 -mpopcnt'
```
on my x86 rather than just `tests nistkat`. I'll also open an issue for that.

mlkem/rej_uniform.c
hanno-becker: I'll double-check

PR #195, opened 2024-10-04
Simplify native interface for rej_uniform

hanno-becker:
Previously, the native interface for rejection sampling was

```
static inline unsigned int rej_uniform_native
    (int16_t *r, unsigned int len,
     const uint8_t *buf,
     unsigned int *buf_consumed,
     unsigned int buflen)
```

Here, `*buf_consumed` would contain the number of bytes processed from the input buffer, and the return value would be the number of valid coefficients sampled.

This generality is not needed -- iInstead, there are only two cases to consider: First, the native call does not support the input arguments, in which case no data is being processed, and one should fallback to the generic C implementation. Second, the call satisfies the same semantics as the C call, that is, it returns the number of sampled coefficients, and guarantees that the entire input buffer was processed if that number is smaller than `len`.

This restricted interface can be more succinctly captured via

```
static inline int rej_uniform_native
    (int16_t *r, unsigned int len,
     const uint8_t *buf, unsigned int buflen)
```

where a return value of -1 captures that the parameters were unsupported.

This commit modifies the native interface accordingly, and adjusts the AArch64 implementation: Specifically, the AArch64 ASM implementation only supports input buffers of length KYBER_N, which can be quickly detected in the recently introduced wrapper round the ASM routine, rather than in assembly (that wrapper didn't exist when the ASM routine was introduced).

### DIFF
--- a/mlkem/native/aarch64/arith_native_aarch64.h
+++ b/mlkem/native/aarch64/arith_native_aarch64.h
@@ -13,9 +13,8 @@ void ntt_asm_opt(int16_t *);
 void intt_asm_clean(int16_t *);
 void intt_asm_opt(int16_t *);
 
-unsigned int rej_uniform_asm_clean(int16_t *r, unsigned int len,
+unsigned int rej_uniform_asm_clean(int16_t *r,
                                    const uint8_t *buf,
-                                   unsigned int *buf_consumed,
                                    unsigned int buflen);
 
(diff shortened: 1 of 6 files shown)


mkannwischer (approved):
Thanks @hanno-becker. 

This indeed looks cleaner! Works fine on my x86. 
Performance seems negligibly better. 
LGTM

PR #203, opened 2024-10-06
Add AVX2 [inv]NTT from official Kyber repository

hanno-becker:
Based on #201 

The AVX2 Kyber NTT produces non-reduced outputs in non-standard order, for better performance. Neither is currently supported by the C<->Native interface.

To accommodate, we temporarily reduce and reorder the output of the AVX2 NTT to match the current C<->Native interface. This will likely need changing later.

Likewise, the inverse NTT operates on a non-standard order that is currently not supported by our C<->Native interface. To accommodate, we shuffle the data in the expected order first before calling the AVX2 iNTT. This may need revisiting later for better performance.

### DIFF
--- a/mk/schemes.mk
+++ b/mk/schemes.mk
@@ -5,7 +5,7 @@ ifeq ($(OPT),1)
 	CPPFLAGS += -DMLKEM_USE_NATIVE
 endif
 
-CPPFLAGS += -Imlkem -Imlkem/sys -Imlkem/native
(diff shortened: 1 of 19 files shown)


cothan (approved):

mkannwischer (approved):
The changes in our sources look fine. 
The assembly looks like a literal copy from the Kyber repo. 
It could use some additional documentation and some clean-up, but I'm fine with keeping it as is.

PR #207, opened 2024-10-07
Cross compilation and Emulation for x86_64 on aarch64 machines

potsrevennil:
- nix shells are merged, now both the `develop` and `ci` shells are equipped with both `x86_64-unknown-linux-gnu-gcc` and `aarch64-unknown-linux-gnu-gcc`, and they are properly linked to both dynamic and static linked `glibc`, so `x86_64-linux-*` prefix for x86_64 machines is no longer needed.

- Merge `force-qemu` flag into `cross-prefix` flag for determining wether to emulate on qemu or not

- Made `--opt/--no-opt` boolean flag (default to --opt), so there is no need to pass argument to the opt flag (e.g. `--opt=true`), only either `--opt` or `--no-opt` is needed.

- Add `--run/--no-run`, `--compile/--no-compile` flags for convenience, and also allow speeding up CI (also for cleaner logging) by splitting compilation and execution.

- Add makefile variable `AUTO` for controlling whether to auto set system specific preprocessor.

- Introduce a `tests all` command that runs multiple tests in parallel to accelerate CI, while also providing granular control over which tests to execute.

- Refactor the GitHub summary for improved readability.

- Update CI triggers to prevent duplicate workflows. It will only run on pushes to main or PRs to main, with newer pushes canceling previous runs for the same GitHub ref

- We currently run native tests on macOS, and native/cross tests on x86_64 and aarch64 Linux. Cross tests are slow, so I’ve split them between GitHub runners and EC2 instances. I don’t have a strong preference on which runner handles which test, as the workflows are flexible and can be easily adjusted as needed.


.github/workflows/ci.yml
```diff
@@ -21,33 +21,52 @@ jobs:
...
            cross-prefix: ''
-           nix-shell: 'ci'
+         - runner: ubuntu-latest
+           name: 'ubuntu-latest (x86_64, cross)'
+           arch: 'x86_64'
+           cross-prefix: 'aarch64-unknown-linux-gnu-'
+         - runner: ubuntu-latest
+           name: 'ubuntu-latest (x86_64, native)'
+           arch: 'x86_64'
+           cross-prefix: ''
```
potsrevennil, line 45: @hanno-becker do you think this is nice to have ? Which allows forks to run ci for linux.

hanno-becker:
@potsrevennil Is this ready for review? Could you squash / reorganize the history?

.github/actions/bench/action.yml
```diff
@@ -61,8 +61,21 @@ runs:
     - name: Run benchmark
       shell: ${{ env.SHELL }}
       run: |
-        tests bench -c ${{ inputs.perf }} --cross-prefix="${{ inputs.cross_prefix }}" --cflags="${{ inputs.cflags }}" --arch-flags="${{ inputs.archflags }}" --opt="${{ inputs.opt }}" -v --output=output.json ${{ inputs.bench_extra_args }}
-        tests bench --components -c ${{ inputs.perf }} --cross-prefix="${{ inputs.cross_prefix }}" --cflags="${{ inputs.cflags }}" --arch-flags="${{ inputs.archflags }}" --opt="${{ inputs.opt }}" -v --output=output.json ${{ inputs.bench_extra_args }}
+        catch() {
+          local line_no="$1"
+          echo "::error title=Failed at $line_no: $?"
+        }
+        trap 'catch $LINENO' ERR
```
hanno-becker, line 68: Sorry, what is this doing?

potsrevennil: Ah, I thought I can make the error annotation better, but turns out this didn't work out as I expected, I'll remove it for now

hanno-becker (approved):
Looks great, tested native and cross compilation on both aarch64 and x86_64 systems. Thanks @potsrevennil !

PR #208, opened 2024-10-07
Document bounds and remove redundant reductions

hanno-becker:
This PR documents bounds on arithmetic data during processing through NTT, basemul, and invNTT, and uses that knowledge to remove redundant reductions, including all uses of `poly_reduce()` except those used before/after serialization/deserialization.

The primary purpose of the removal of those interim `poly_reduce()` invocations is not to improve performance, but to strive for a unified semantics of `poly_reduce()` across native implementations: The C/AArch64 semantics of `poly_reduce()` is to give signed canonical representatives, while the AVX2 implementation gives _unsigned_ almost-canonical representatives in `[0,...,q]` (inclusive upper bound!). This distinction then ramifies into serialization routines. _Conceptually_, however, one always wants an _unsigned canonical_ interface to `poly_reduce()` when being used in the context of [de]serialization. On the other hand, prior to this PR, `poly_reduce()` was also used to reduce the size of coefficients, for which canonicity and sign are not relevant, and it would be an unnecessary performance drag to e.g. use an unsigned canonical reduction. With the removal of `poly_reduce()` for interim reductions, we can switch to the unified semantics of `poly_reduce()` as producing _unsigned canonical_ data in a follow-up PR.


mlkem/native/aarch64/intt_clean.S
```diff
@@ -29,7 +29,15 @@
 // Needed to provide ASM_LOAD directive
 #include "common.i"
 
+// Bounds:
+// In: |src| < q * C
+// Out: |dst| < q * (0.0508 * C + 1/2)
```
rod-chapman, line 34: I like this, but where does the 0.0508 come from?

mlkem/native/aarch64/intt_clean.S
```diff
@@ -226,11 +234,41 @@ _intt_asm_clean:
...
+
+        ASM_LOAD(xtmp, ninv_addr)
+        ld1r {ninv.8h}, [xtmp]
+        ASM_LOAD(xtmp, ninv_tw_addr)
+        ld1r {ninv_tw.8h}, [xtmp]
+
+        mov inp, in
+        mov count, #8
+
+scale_start:
```
rod-chapman, line 248: What's going on here? This doesn't seem to be much to do with "documenting the bounds"

mlkem/native/aarch64/intt_clean.S
```diff
@@ -254,15 +300,21 @@ layer4567_start:
         // Layer 5
         gs_butterfly data0, data1, root0, 2, 3
         gs_butterfly data2, data3, root0, 4, 5
+        // Max bound: 8q
 
+        // Not all of those reductions are needed, but the bounds tracking
```
rod-chapman, line 305: OK.. this is good.

mlkem/native/aarch64/intt_clean.S
```diff
@@ -314,16 +358,13 @@ layer123_start:
...
+
         str q_data4, [in, #(4*(512/8))]
         str q_data5, [in, #(5*(512/8))]
         str q_data6, [in, #(6*(512/8))]
         str q_data7, [in, #(7*(512/8))]
 
-        // Scale half the coeffs by 1/n; for the other half, the scaling has
-        // been merged into the multiplication with the twiddle factor on the
-        // last layer.
-        mul_ninv data0, data1, data2, data3, data0, data1, data2, data3
```
rod-chapman, line 325: So you've removed this code, and adjusted the twiddle factors back to their original values?

mlkem/native/aarch64/ntt_clean.S
```diff
@@ -29,10 +29,20 @@
 // Needed to provide ASM_LOAD directive
 #include "common.i"
 
+// Bounds:
+// In: |src| < q * C
+// Out: |dst| < q * (0.0508 * C + 1/2)
```
rod-chapman, line 34: As above - the 0.0508 needs a bit of explanation.

mlkem/native/aarch64/ntt_clean.S
```diff
@@ -204,6 +214,20 @@ _ntt_asm_clean:
         load_roots_123
 
         .p2align 2
+
+        // Bounds reasoning:
+        // - There are 7 layers
+        // - When passing from layer N to layer N+1, each layer-N value
+        //   is modified through the addition/subtraction of a Montgomery
+        //   product of a twiddle of absolute value < q/2 and a layer-N value.
+        // - Recalling that |fqmul(a,t)| < q * (0.0254*C + 1/2) for
```
rod-chapman, line 223: You lost me at this point. What is C? You're using "^" to mean "power" not "exclusive or", right? That's confusing for C programmers.

mlkem/native/aarch64/ntt_clean.S
```diff
@@ -204,6 +214,20 @@ _ntt_asm_clean:
...
+        //   |a| < C*q and |t|<q/2, we see that the coefficients of layer N
+        //   (starting with layer 0 = input data) are bound by q * f^N(1), where
```
rod-chapman, line 225: What does f^N(1) mean? I can't follow that notation?

mlkem/native/aarch64/ntt_clean.S
```diff
@@ -204,6 +214,20 @@ _ntt_asm_clean:
...
+        //   f(C) = 1/2 + 1.0508*C.
+        //   For N=7, we get the bound of f^7(1) * q < 18295.
```
rod-chapman, line 227: Isn't it easier to just show that for N=7, the upper bound is 7Q + 1664? That's a much simpler explanation, and still fits in 16 bits.

mlkem/native/aarch64/ntt_clean.S
```diff
@@ -268,11 +292,6 @@ layer4567_start:
         ct_butterfly_v data0, data1, root1, root1_tw
         ct_butterfly_v data2, data3, root2, root2_tw
 
-        barrett_reduce data0
```
rod-chapman, line 271: Err.. why have these been removed? Why is that OK?

rod-chapman (changes requested):
See comments inline

mlkem/native/aarch64/ntt_clean.S
hanno-becker: That's what the PR is about!

mlkem/debug/debug.h
```diff
@@ -0,0 +1,71 @@
...
+
+void mlkem_debug_check_bounds(const char *file, int line,
+                              const char *description, const int16_t *ptr,
+                              unsigned len, int16_t lower_bound_inclusive,
+                              int16_t upper_bound_inclusive);
+
+void mlkem_debug_print_error(const char *file, int line, const char *msg);
+
+#define BOUND(ptr, len, abs_bound, msg)                                   \
+  do {                                                                    \
```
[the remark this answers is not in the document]
hanno-becker: That's a common idiom in C

mlkem/debug/debug.h
```diff
@@ -0,0 +1,71 @@
...
+    mlkem_debug_check_bounds(__FILE__, __LINE__, (msg), (int16_t *)(ptr), \
+                             (len), -((abs_bound)-1), ((abs_bound)-1));   \
+  } while (0)
+
+#define POLY_BOUND_MSG(ptr, abs_bound, msg) \
```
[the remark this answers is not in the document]
hanno-becker: `poly`

mlkem/debug/debug.h
```diff
@@ -0,0 +1,71 @@
+// SPDX-License-Identifier: Apache-2.0
+#ifndef MLKEM_DEBUG_H
+#define MLKEM_DEBUG_H
+
+#if defined(MLKEM_DEBUG)
+#include <stdio.h>
+#include <stdlib.h>
+
+void mlkem_debug_check_bounds(const char *file, int line,
```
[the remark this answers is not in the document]
hanno-becker: It's only for the macros, but it still needs a declaration in the header.

mlkem/debug/debug.h
```diff
@@ -0,0 +1,71 @@
...
+  do {                                                                   \
+    for (unsigned _debug_polyvec_bound_idx = 0;                          \
+         _debug_polyvec_bound_idx < KYBER_K; _debug_polyvec_bound_idx++) \
+      POLY_BOUND_MSG(&(ptr)->vec[_debug_polyvec_bound_idx], (abs_bound), \
+                     "polyvec bound for " #ptr ".vec[i]");               \
+  } while (0)
+
+// Following AWS-LC to define a C99-compliant static assert
+#define MLKEM_CONCAT(left, right) left##right
+#define MLKEM_STATIC_ASSERT_DEFINE(cond, msg)                            \
```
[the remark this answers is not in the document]
hanno-becker: This is a common way to encode a static assert in C99 -- the implementation details are irrelevant, but here we create an anonymous struct with bitfield length 1 or -1 depending on the condition. If the condition fails, a bitfield of size -1 leads to a compile error.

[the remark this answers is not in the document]
hanno-becker: https://stackoverflow.com/questions/3385515/static-assert-in-c

mlkem/indcpa.c
```diff
@@ -76,6 +84,7 @@ static void pack_sk(uint8_t r[KYBER_INDCPA_SECRETKEYBYTES], polyvec *sk) {
 static void unpack_sk(polyvec *sk,
                       const uint8_t packedsk[KYBER_INDCPA_SECRETKEYBYTES]) {
   polyvec_frombytes(sk, packedsk);
+  polyvec_reduce(sk);
```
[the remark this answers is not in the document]
hanno-becker: See the commit message

mlkem/indcpa.c
```diff
@@ -245,6 +254,9 @@ void gen_matrix(polyvec *a, const uint8_t seed[KYBER_SYMBYTES],
  *              - const uint8_t *coins: pointer to input randomness
  *                             (of length KYBER_SYMBYTES bytes)
  **************************************************/
+
+STATIC_ASSERT(NTT_BOUND + KYBER_Q < INT16_MAX, indcpa_enc_bound_0)
```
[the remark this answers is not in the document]
hanno-becker: Yes

mlkem/native/aarch64/intt_clean.S
hanno-becker: Please see the commit messages -- we're hoisting the scaling in the invNTT from end to start, as a means to simultaneously scale and reduce the size of the representatives.

mlkem/native/aarch64/intt_clean.S
hanno-becker: The code moved to the top.

mlkem/native/aarch64/ntt_clean.S
hanno-becker: This is explained in `reduce.c` and empirically checked in `test_bounds.py` -- I'll add a reference

mlkem/debug/debug.h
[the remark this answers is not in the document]
hanno-becker: https://stackoverflow.com/questions/1067226/c-multi-line-macro-do-while0-vs-scope-block

mlkem/indcpa.c
```diff
@@ -289,8 +301,10 @@ void indcpa_keypair_derand(uint8_t pk[KYBER_INDCPA_PUBLICKEYBYTES],
     poly_tomont(&pkpv.vec[i]);
   }
 
+  // Arithmetic cannot overflow, see static assertion at the top
   polyvec_add(&pkpv, &pkpv, &e);
   polyvec_reduce(&pkpv);
+  polyvec_reduce(&skpv);
```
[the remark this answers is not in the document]
hanno-becker: This is reduction prior to serialization. Previously this was done at the end of the fwd NTT, but that reduction is generally not needed and has been removed.

mlkem/native/aarch64/ntt_clean.S
hanno-becker: `f(f(...(f(1))...)`

mlkem/native/aarch64/ntt_clean.S
hanno-becker: Yes, meant it as "power". Maybe "**" will be more universally understood?

mlkem/debug/debug.h
```diff
@@ -0,0 +1,71 @@
...
+#if defined(MLKEM_DEBUG)
+#include <stdio.h>
+#include <stdlib.h>
+
+void mlkem_debug_check_bounds(const char *file, int line,
+                              const char *description, const int16_t *ptr,
+                              unsigned len, int16_t lower_bound_inclusive,
+                              int16_t upper_bound_inclusive);
+
+void mlkem_debug_print_error(const char *file, int line, const char *msg);
```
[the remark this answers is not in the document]
hanno-becker: It goes to `stderr` -- will document.

mlkem/native/aarch64/ntt_clean.S
rod-chapman: Possibly, but ** gets mangled because it means something special in MarkDown format.

mlkem/native/aarch64/ntt_clean.S
rod-chapman: Thanks

mlkem/debug/debug.h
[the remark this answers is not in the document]
rod-chapman: OK.. please add that as a comment. Imagine someone else is reading this, 5 years from now...

mlkem/debug/debug.h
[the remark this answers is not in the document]
rod-chapman: Accepted

mlkem/debug/debug.h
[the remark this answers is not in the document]
rod-chapman: It's still ugly. But I guess there's no other way.

mlkem/native/aarch64/ntt_clean.S
rod-chapman: Thanks

mlkem/debug/debug.c
```diff
@@ -0,0 +1,39 @@
...
+
+void mlkem_debug_check_bounds(const char *file, int line,
+                              const char *description, const int16_t *ptr,
+                              unsigned len, int16_t lower_bound_inclusive,
+                              int16_t upper_bound_inclusive) {
+  int err = 0;
+  unsigned i;
+  for (i = 0; i < len; i++) {
+    int16_t val = ptr[i];
+    if (val < lower_bound_inclusive || val > upper_bound_inclusive) {
```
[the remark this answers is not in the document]
rod-chapman: OK - got it.

mlkem/native/aarch64/ntt_clean.S
rod-chapman: OK... so we're documenting the bounds _and_ removing redundant reductions as a side-effect of that.

mlkem/native/aarch64/ntt_clean.S
hanno-becker: Fair point. There are two parts here:
- Show that the arithmetic _within_ the NTT does not overflow.
- Exhibit a specific bound on the output of the NTT that is sufficient to show that _subsequent_ arithmetic does not overflow.

It seems to me that (a) basemul implementations currently need _no_ bound, (b) keygen merely needs `NTT_BOUND + KYBER_Q < INT16_MAX` to handle the addition https://github.com/pq-code-package/mlkem-c-aarch64/blob/main/mlkem/indcpa.c#L292. That would actually be supported by a 8*Q bound. Still, I don't think it hurts to document a tighter bound in case changes in the subsequent arithmetic needs evaluation.

mlkem/native/aarch64/ntt_clean.S
hanno-becker: Ok, I'll leave it as is for now.

mlkem/native/aarch64/ntt_clean.S
hanno-becker: I'll spell it out

mlkem/native/aarch64/ntt_clean.S
rod-chapman: You're right... if the inputs are bounded by +/- Q, then you must be within 8Q after 7 layers. I appreciate that your bound is tighter, but 8Q < 32767 is all we need, right?

rod-chapman: That's what my proofs in SPARK and C show anyway...

mlkem/native/aarch64/ntt_clean.S
rod-chapman: Lovely

mlkem/native/aarch64/ntt_clean.S
hanno-becker: No, it's note quite enough, because there's further arithmetic following the NTT: Specifically, there's https://github.com/pq-code-package/mlkem-c-aarch64/blob/main/mlkem/indcpa.c#L292 adding a value up to size Q. So there one needs 9Q < INT16_MAX.

mlkem/debug/debug.h
```diff
@@ -0,0 +1,71 @@
...
+
+#define BOUND(ptr, len, abs_bound, msg)                                   \
```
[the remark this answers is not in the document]
hanno-becker: Done

mlkem/native/aarch64/ntt_clean.S
hanno-becker: Done

mlkem/native/aarch64/ntt_clean.S
rod-chapman: Ah yes.

mlkem/ntt.c
```diff
@@ -57,9 +58,20 @@ const int16_t zetas[128] = {
...
+ *              a polynomial in place.
+ *
+ *              The input is assumed to be in normal order and
+ *              coefficient-wise bound by KYBER_Q in absolute value.
+ *
+ *              The output polynomail is in bitreversed order, and
+ *              coefficient-wise bound by NTT_BOUND in absolute value.
+ *
+ *              Values of NTT_BOUND:
+ *              - C: 16546
```
rod-chapman, line 70: I'm not sure I like this, since it complicates the C/Asm boundary with potentially different contracts for each back-end. Can't we reduce here so that _all_ backends always return reduced coefficients that are always in the same range?

mlkem/ntt.c
```diff
@@ -95,37 +117,54 @@ void poly_ntt(poly *p) { ntt_native(p); }
...
- *order
+ *              order
+ *
+ *              The input is assumed to be in bitreversed order.
+ *
+ *              The output polynomail is in normal order, and
+ *              coefficient-wise bound by INVNTT_BOUND in absolute value.
+ *
+ *              Values of INVNTT_BOUND:
+ *              - C: 3/4 q
```
rod-chapman, line 128: As above - I think the abstraction should be the same for all backends.

mlkem/ntt.c
```diff
@@ -95,37 +117,54 @@ void poly_ntt(poly *p) { ntt_native(p); }
...
 #if !defined(MLKEM_USE_NATIVE_INTT)
 // REF-CHANGE: Removed indirection poly_invntt_tomont -> invntt()
+// REF-CHANGE: Moved scalar multiplication with f ahead of the core invNTT
 void poly_invntt_tomont(poly *p) {
   unsigned int start, len, j, k;
   int16_t t, zeta;
   const int16_t f = 1441;  // mont^2/128
   int16_t *r = p->coeffs;
 
+  for (j = 0; j < 256; j++) {
```
rod-chapman, line 142: Can you justify why this still returns the right answer?

rod-chapman (commented):
More comments...

mlkem/ntt.c
hanno-becker: The modular arithmetic is unchanged, but we're moving the scaling step to the front so it acts as a reduction of the input data. Does that answer your question?

mlkem/ntt.c
hanno-becker: Yes, I want a _single_ "worst-case" contract between C and ASM that _all_ implementations satisfy. But the way it's currently phrased, that doesn't come across. I'll try again.

mlkem/ntt.c
hanno-becker: Done

mlkem/ntt.c
hanno-becker: Done

mlkem/ntt.c
```diff
@@ -57,9 +58,20 @@ const int16_t zetas[128] = {
...
  * Description: Computes negacyclic number-theoretic transform (NTT) of
- *              a polynomial in place;
- *              inputs assumed to be in normal order, output in bitreversed
- *order
+ *              a polynomial in place.
+ *
+ *              The input is assumed to be in normal order and
+ *              coefficient-wise bound by KYBER_Q in absolute value.
+ *
+ *              The output polynomail is in bitreversed order, and
```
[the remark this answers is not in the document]
hanno-becker: Done

mlkem/debug/debug.h
```diff
@@ -0,0 +1,103 @@
...
+ *              - len: Number of int16_t in ptr
+ *              - lower_bound_inclusive: Inclusive lower bound
+ *              - upper_bound_inclusive: Inclusive upper bound
+ **************************************************/
+void mlkem_debug_check_bounds(const char *file, int line,
+                              const char *description, const int16_t *ptr,
+                              unsigned len, int16_t lower_bound_inclusive,
+                              int16_t upper_bound_inclusive);
+
+/* Print error message alongside file and line information */
```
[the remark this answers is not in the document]
hanno-becker: Done

mlkem/debug/debug.h
```diff
@@ -0,0 +1,103 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+
+/*************************************************
+ * Name:        mlkem_debug_check_bounds
+ *
+ * Description: Check whether values in an array of int16_t
+ *              are within specified bounds.
+ *
+ *              Prints an erro message and calls exit(1) if not.
```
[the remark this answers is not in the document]
hanno-becker: Done

mlkem/ntt.c
rod-chapman: Yes... makes sense, but it's such an obvious departure from the reference implementation that an explanatory comment with worth-while.

hanno-becker: Oh, that's a very good point. We _do_ want those `REF-CHANGE` comments, at least in major deviations like this. I'll add it.

hanno-becker: Oh, actually there is already a comment at the top of the function

rod-chapman: Noted

mlkem/ntt.c
```diff
@@ -134,18 +197,30 @@ void poly_invntt_tomont(poly *p) { intt_native(p); }
...
+  BOUND(a, 2, KYBER_Q, "basemul input bound");
+
   int32_t t0, t1;
+
   t0 = (int32_t)a[1] * b_cached;
   t0 += (int32_t)a[0] * b[0];
   t1 = (int32_t)a[0] * b[1];
   t1 += (int32_t)a[1] * b[0];
+
+  // |ti| < 2 * q * 2^15
```
rod-chapman, line 220: The PQCrystals/Kyber ref/reduce.c code says that the input to montgomery_reduce() must in in
"{-q2^15 .. q2^15 - 1}" but your new code in reduce.c says no constraint on the argument at all.
If the former case is right, then t1 < 2q2^15 doesn't work?

mlkem/reduce.c
```diff
@@ -9,14 +9,40 @@
  * Description: Montgomery reduction; given a 32-bit integer a, computes
  *              16-bit integer congruent to a * R^-1 mod q, where R=2^16
  *
- * Arguments:   - int32_t a: input integer to be reduced;
- *                           has to be in {-q2^15,...,q2^15-1}
```
rod-chapman, line 13: Is this really no longer required? See comment in ntt.c. The code appears identical to the Kyber ref, so why can this
constraint be removed?

mlkem/reduce.c
```diff
@@ -9,14 +9,40 @@
...
+  //
+  // |t| <= |a|/2^16 + |t|*q/2^16
+  //      < q * C / 2^16 + q/2
+  //      = q (C/2^16 + 1/2)
+  //
+  // - Case |a| < (q/2) * C * q
+  //
+  // Replace C -> C * q in the above and estimate
+  // q / 2^17 < 0.0254.
+
```
rod-chapman, line 45: Can you add two BOUND() macros here to check the pre-condition and the bound of the result?

rod-chapman (commented):
See comments on basemul_cached() and montgomery_reduce(). More to come...

mlkem/reduce.c
rod-chapman: I really need _weakest_ precondition on montgomery_reduce(), and the post-condition that expresses a bound on the output if that pre-condition it true, both expressed using simple static constants (i.e. not complex expressions involving real numbers...) please.

mlkem/ntt.c
hanno-becker: But, please feel free to double-check.

mlkem/reduce.c
hanno-becker: > Is this really no longer required? See comment in ntt.c. The code appears identical to the Kyber ref, so why can this constraint be removed?

I don't see a need for the constraint in the Kyber repo, and indeed it too strict for the application in other parts of the code. Please check my bounds.

hanno-becker: > I really need weakest precondition on montgomery_reduce(), and the post-condition that expresses a bound on the output if that pre-condition it true, both expressed using simple static constants (i.e. not complex expressions involving real numbers...) please.

This is really a parametric spec here. We'll need to see how to best express this in CBMC, but that's not for this PR to solve.

mlkem/reduce.c
hanno-becker: No, because everything is parametric, we don't know what `C` is. But that's what `test_bounds.py` is for.

mlkem/ntt.c
rod-chapman: I'd really like to see a CBMC proof of montgomery_reduce() to prove that it really does work for all int32_t.
I may already have that... let me see...

hanno-becker: I don't know if CBMC can do that, because it's parametric. I added `test_bounds.py` to brute force this.

rod-chapman: NOted

mlkem/reduce.c
rod-chapman: Noted

mlkem/reduce.c
rod-chapman: OK

rod-chapman (approved):
All good now

mlkem/reduce.c
```diff
@@ -43,8 +43,17 @@ int16_t montgomery_reduce(int32_t a) {
   // Replace C -> C * q in the above and estimate
   // q / 2^17 < 0.0254.
 
-  t = (int16_t)a * QINV;
-  t = (a - (int32_t)t * KYBER_Q) >> 16;
+  uint16_t u;
+  int16_t t;
```
rod-chapman, line 47: Redeclaration of t here fails to compile.

hanno-becker: oops.

rod-chapman: Fixed.

PR #215, opened 2024-10-12
Modify semantics of `poly_reduce()` to require unsigned canonical output

hanno-becker:
Previously, the post-condition for `poly_reduce()` was that its
output coefficients are _signed_ canonical, an unsuitable choice
in two ways:

- `poly_reduce()` is only used in the context of (de)serialization,
  where we're dealing with _unsigned_ canonical data. This mean that
  even though we were calling `poly_reduce()` ahead of `poly_tobytes()`,
  the latter did still need to do a conditional addition.
- The AVX2 implenentation of `poly_reduce()` does not produce signed
  canonical representatives, but 'almost' unsigned canonical ones in
  [0,1,...,KYBER_Q] (inclusive upper bound!) that are then subject to
  a conditional _subtraction_ during deserialization.

This commit is an attempt to clear up the situation by requiring
that `poly_reduce()` produce _unsigned canonical_ representatives,
and adjusting the C and native AArch64 implementations as well as the
serialization routines accordingly.

Finally, AVX2 assembly for poly_reduce() is added from https://github.com/pq-crystals/kyber/tree/main/avx2,
adding a conditional subtraction to achieve unsigned canonical output data.


mlkem/debug/debug.h
```diff
@@ -50,11 +63,26 @@ void mlkem_debug_print_error(const char *file, int line, const char *msg);
...
+         msg)
+
 /* Check absolute bounds on coefficients in polynomial
  * ptr: poly* of poly_mulcache* pointer to polynomial (cache) to check
  * abs_bound: Exclusive upper bound on absolute value to check */
 #define POLY_BOUND(ptr, abs_bound) \
-  POLY_BOUND_MSG((ptr), (abs_bound), "poly bound for " #ptr)
+  POLY_BOUND_MSG((ptr), (abs_bound), "poly absolute bound for " #ptr)
+
+/* Check absolute bounds on coefficients in polynomial
```
mkannwischer, line 80:
```suggestion
/* Check unsigned bounds on coefficients in polynomial
```
??

mlkem/poly.c
```diff
@@ -512,20 +513,30 @@ void poly_tomont(poly *r) {
...
  **************************************************/
 #if !defined(MLKEM_USE_NATIVE_POLY_REDUCE)
 void poly_reduce(poly *r) {
   unsigned int i;
   for (i = 0; i < KYBER_N; i++) {
-    r->coeffs[i] = barrett_reduce(r->coeffs[i]);
+    // Barrett reduction, giving signed canonical representative
+    int16_t t = barrett_reduce(r->coeffs[i]);
+    // Conditional addition to get unsigned canonical representative
+    r->coeffs[i] = scalar_signed_to_unsigned_q_16(t);
```
mkannwischer, line 535: Isn't it easier to introduce an unsigned barrett reduction?

mkannwischer (changes requested):
Small nits.

I find it a bit confusing that some of the bound functions are inclusive of the upper bounds while other are exclusive. Is there any reason for that?

Rest looks good. Thanks @hanno-becker.

mlkem/debug/debug.h
hanno-becker: yep

mlkem/poly.c
hanno-becker: you still need a conditional subtraction then, like in the avx2 code

mlkem/debug/debug.h
hanno-becker: fixed

mlkem/poly.c
mkannwischer: okay, then nvm.

hanno-becker:
> I find it a bit confusing that some of the bound functions are inclusive of the upper bounds while other are exclusive. Is there any reason for that?

Fair point. I moved to exclusive bounds everywhere.

mkannwischer (approved):
LGTM now.

PR #216, opened 2024-10-12
Add AVX2 assembly for base multiplication

hanno-becker:
Add AVX2 assembly for base multiplication from official Kyber repository.

Since we don't yet integrate all AVX2 assembly, we need to conjugate the basemul assembly with transformation to/from standard coefficient order and the custom order used by the AVX2 code.

PR #217, opened 2024-10-12
fix ci event trigger for labelling

potsrevennil:
The GH action behavior isn't ideal...

Since we need to have the event trigger set as
```
  pull_request:
    branches: ["main"]
    types: [ "labeled", "synchronize" ]
```

concurrency group would need the adjustment to avoid the previous in-progress jobs being cancelled
```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event.action != 'labeled' }}
```

While setting
```
if: ${{ github.event.action != 'labeled' || github.event.label.name == 'ci' }}
```
could avoiding the job be re-run upon labeled event,

once a label is set, the UI would show that all the newly triggered job is skipped, thus can be merged, while the previous triggered jobs might failed or still in-progress. 
Anyway we can still check all the status at the `checks` tab.

IDK if this is better than the original behavior though 😞

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,10 +11,11 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event.action != 'labeled' }}
 
 jobs:
   build_kat:
+    name: Functional tests (${{ matrix.target.name }})
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


hanno-becker:
Cherry-picking this in #216

hanno-becker:
I can't confirm that this works as intended. Instead, after cherry-picking in #216, if I set the `bench` label, CI jobs will be terminated but _not_ be restarted, and vice-versa. I didn't manage to run both sets of workflows anymore at all.

PR #218, opened 2024-10-12
Fix tests and GH cache issue

potsrevennil:
Currently since we're equipping our shells with both x64_64 and aarch64 gcc and glibc, which results in quite a large cache size, if we were to upload to and fetch the GH cache, we would easily run into GH cache throttling and rate limiting issues.

Therefore, allowing the nix-setup interface to control the use of the GH cache and disabling it in certain scenarios would be better.

### DIFF
--- a/.github/workflows/bench_ec2_reusable.yml
+++ b/.github/workflows/bench_ec2_reusable.yml
@@ -36,10 +36,10 @@ on:
         type: boolean
         description: Indicates if results should be pushed to github pages
         default: false
-      always_terminate:
+      verbose:
+        description: Determine for the log verbosity
         type: boolean
-        description: Indicates if instance should always be terminated, even on failure
-        default: true
+        default: false
       bench_extra_args:
         type: string
         description: Additional command line to be appended to `bench` script
@@ -106,7 +106,9 @@ jobs:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/bench
         with:
-          nix-shell: bench
+          nix-shell: ci
+          nix-cache: false
+          nix-verbose: ${{ inputs.verbose }}
           name: ${{ inputs.name }}
           cflags: ${{ inputs.cflags }}
           archflags: ${{ inputs.archflags }}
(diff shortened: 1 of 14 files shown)


.github/workflows/bench_ec2_reusable.yml
```diff
@@ -106,7 +106,9 @@ jobs:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/bench
         with:
-          nix-shell: bench
+          nix-shell: ci
+          nix-cache: false
+          nix-verbose: ${{ inputs.always_terminate }}
```
hanno-becker, line 111: Why's that correlation between `nix-verbose` and `always_terminate`?

potsrevennil: There's inputs limit in GH dispatch workflow, so I can not add more inputs for bench_ec2_any now.
And I figure, if there is anything wrong, and we might set always terminate to false so that we can examinate what's wrong in ec2. 
Therefore it make sense to set nix-verbose to True to log more as well
Does that also make sense to you ?

hanno-becker: Doesn't the present version do the opposite?

hanno-becker: I'd say throw out `always_terminate` for the time being -- I've only once used it so far, and it's rare that you cannot reproduce issues by just firing up an instance and doing what was done in CI.

flake.nix
```diff
@@ -96,7 +82,7 @@
           };
 
           devShells.bench = wrapShell pkgs.mkShellNoCC { packages = core; };
```
hanno-becker, line 84: Is this redundant now?

hanno-becker (changes requested):
Two minor changes, otherwise lgtm

.github/workflows/bench_ec2_reusable.yml
potsrevennil: oh yes, you're right, I mess up with the value..

scripts/tests
```diff
@@ -236,11 +236,15 @@ class State(object):
...
         self,
-        bin,
+        bin: str,
         run_as_root=False,
         exec_wrapper=None,
-    ):
+    ) -> bytes:
         """Run the binary in all different ways"""
+        if not os.path.isfile(bin):
+            logging.error(f"{bin} not exists")
```
hanno-becker, line 245:
```suggestion
            logging.error(f"{bin} does not exist")
```

hanno-becker:
@potsrevennil Seems fine, but can you squash the history and write a commit message summarizing the change please?

hanno-becker (approved):

PR #219, opened 2024-10-12
AVX2: Switch to custom NTT coefficient ordering 

hanno-becker:
The AVX2 assembly routines use a custom coefficient ordering in NTT domain, but until now, we've always conjugated the routines with permutations to keep them in standard order. This commit removes those conjugations, switching to using the custom order throughout.


mkannwischer (approved):
LGTM. Thanks @hanno-becker

mkannwischer:
This gives a small bump in performance. 
ML-KEM-768 encaps on Intel Xeon 4th gen (c7i): 68089 -> 65593

PR #220, opened 2024-10-13
Add AVX2 Keccak-x4 from XKCP

hanno-becker:
Add AVX2 code for Keccak-f1600-x4 from XKCP / Kyber official repository.

Based on #219


mkannwischer (approved):
Performance improvement is neat. 
ML-KEM-768 encaps on Intel Xeon 4th gen (c7i) 65581 -> 49919

PR #221, opened 2024-10-13
Smaller performance improvements of AVX2 backend

hanno-becker:
Previously, the AVX2 NTT and invNTT assembly was suffixed by a call to `poly_reduce()` to ensure polynomial coefficients sufficiently small that the subsequent arithmetic would not overflow.

If we trust the bounds for the output of the NTT and invNTT [provided by the Kyber repository](https://github.com/pq-crystals/kyber/blob/main/avx2/poly.c#L387), those reductions are not necessary: The bounds are already below the contractual bounds for the C<->native interface on the basis of which correctness of the remaining code can be argued.

This PR therefore removes the reductions after the AVX2 NTT and invNTT.

It also modifies the vector-vector native code to use AVX2 intrinsics for the accumulation, and adds an AVX2 version for `poly_tomont()`.


hanno-becker:
We should double-check and document the bounds for the AVX2 NTT and invNTT bounds ourselves. Opened #222

potsrevennil (approved):

PR #225, opened 2024-10-14
Set C standard to C99

hanno-becker:
MLKEM-C-AArch64 strives to be compatible with a wide range of (old) compilers. Force restriction to C99 to facilitate this.


hanno-becker:
@potsrevennil @mkannwischer I'm not sure how to appease the CI here. I can't seem to figure out how to set the right `#define`s / include the right headers, so that `syscall()` won't lead to an 'no function prototype' warning -> error.

mkannwischer (approved):
LGTM.

PR #228, opened 2024-10-14
Pin clang-tools version to 17.0.6 in nix flake

potsrevennil:
Somehow nix clang-tools version control is off between linux and macOS, so we need to pin the clang-tools version for now.

According to the `clang-tools` source code, the version is determined by the parameter `clang-unwrapped`, which is why I do nix override `clang-unwrapped` instead of `version`.

By a quick check, using `llvmPackages_18` would results in quite a large size for the nix derivation, I'm not sure the reason of this, but anyway, sticking to `llvmPacakges_17` for now should be fine

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -7,11 +7,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1709336216,
-        "narHash": "sha256-Dt/wOWeW6Sqm11Yh+2+t0dfEWxoMxGBvv3JpIocFl9E=",
+        "lastModified": 1727826117,
+        "narHash": "sha256-K5ZLCyfO/Zj9mPFldf3iwS6oZStJcU4tSpiXTMYaaL0=",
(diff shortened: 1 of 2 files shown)


potsrevennil:
@hanno-becker Mind trying out to see if this solves the clang-format issue on your M1 ?

hanno-becker:
Tested locally, LGTM

flake.lock
```diff
@@ -22,11 +22,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1718086528,
-        "narHash": "sha256-hoB7B7oPgypePz16cKWawPfhVvMSXj4G/qLsfFuhFjw=",
+        "lastModified": 1728740863,
```
hanno-becker, line 25: Whats this change to `flake.lock` about?

hanno-becker (approved):

PR #229, opened 2024-10-14
Fix CI Pull_Request labeled event trigger

potsrevennil:


hanno-becker (changes requested):
This does not seem to work yet. Setting `benchmark` will kill previous `ci` jobs, and also not restart them.

potsrevennil:
@hanno-becker What do you mean by kill the previous CI jobs ?
If what you mean is that the CI jobs in the UI got removed, then this is as expected. 

You can check that in the Checks tab, there's still previous CI workflow running without being canceled, and there is another CI workflow below waiting. 

As soon as the previous CI workflow is done, it would start running and got cancelled immediately (due to the condition I set), thus the UI will show up the CI jobs with the skip status

As what I said in #217, this doesn't seem to be ideal, and I'm not sure if this is what we want

hanno-becker:
> If what you mean is that the CI jobs in the UI got removed, then this is as expected.
> You can check that in the Checks tab, there's still previous CI workflow running without being canceled, and there is another CI workflow below waiting.

If I understand correctly, then that's not acceptable for us. The UI should be an accurate reflection of the jobs that run for the given PR. 

The intended behaviour is this:
* When a PR is opened, CI job runs, but bench jobs don't.
* If `bench` label is set, benchmark jobs are run. CI jobs are unaffected.
* If `ci` label is set, CI is re-triggered. Bench jobs are unaffected.
* The UI should always reflect the state of the CI and Bench jobs.
* Desirable addition: Old CI / Bench jobs for the same PR are cancelled when new ones are triggered.

potsrevennil:
Then I would say probably it's not doable as far as I know

To trigger the CI/bench workflow when the ci/bench label is set, the following configuration is needed:
```
on:
  pull_request:
    branches: ["main"]
    types: [ "labeled" ]
```

However, since the pull_request labeled event in GitHub Actions doesn’t allow filtering by specific label values, the workflow will be triggered whenever any label is applied, not just ci/bench.

If the workflow is triggered, the UI status will be updated based on the newly triggered workflow. If we skip the workflow, the UI will reflect a “skipped” status, replacing the current CI state. To ensure the UI properly reflects the state of the CI, we cannot skip workflows, meaning they will have to rerun. This approach conflicts with the intention to leave certain jobs unaffected, as mentioned in the 2nd and 3rd points.

hanno-becker:
@potsrevennil and I agreed to work around this issue by removing the label-trigger for the CI workflow.

hanno-becker (approved):

PR #232, opened 2024-10-14
CBMC proofs for type-safety of reduce.c

rod-chapman:
This PR adds CBMC type-safety proof for the reduce.c module,
Specifically, montgomery_reduce() and barrett_reduce()

These functions depend on implementation-defined behaviour, such as casting int32_t to int16_t, and shift-right on
a signed (possibly negative) integer, so proofs of full correctness have not been attempted at this stage.

### DIFF
--- a/mlkem/reduce.c
+++ b/mlkem/reduce.c
@@ -87,6 +87,12 @@ int16_t montgomery_reduce(int32_t a) {
   return (int16_t)r;
 }
 
+// To divide by KYBER_Q using Barrett multiplication, the "magic number"
+// multiplier is round_to_nearest(2**26/KYBER_Q)
+#define BPOWER 26
+static const int32_t barrett_multiplier =
+    ((1 << BPOWER) + KYBER_Q / 2) / KYBER_Q;
+
 /*************************************************
  * Name:        barrett_reduce
  *
@@ -99,20 +105,15 @@ int16_t montgomery_reduce(int32_t a) {
  * Returns:     integer in {-(q-1)/2,...,(q-1)/2} congruent to a modulo q.
  **************************************************/
 int16_t barrett_reduce(int16_t a) {
-  int16_t t;
-  const int16_t v = ((1 << 26) + KYBER_Q / 2) / KYBER_Q;
+  // Compute round_to_nearest(a/KYBER_Q) using the multiplier
+  // above and shift by BPOWER places.
+  //
+  // PORTABILITY: Right-shift on a signed integer is, strictly-speaking,
+  // implementation-defined for negative left argument. Here,
+  // we assume it's sign-preserving "arithmetic" shift right. (C99 6.5.7 (5))
+  const int32_t t = (barrett_multiplier * a + (1 << (BPOWER - 1))) >> BPOWER;
 
-  t = ((int32_t)v * a + (1 << 25)) >> 26;
-
-  // CBMC is over-zealous in checking integer conversions, even
-  // if they are well-defined, so we disable this check here.
-#ifdef CBMC
-#pragma CPROVER check push
-#pragma CPROVER check disable "conversion"
-#endif
-  t *= KYBER_Q;
-  return a - t;
-#ifdef CBMC
-#pragma CPROVER check pop
-#endif
+  // t is in -10 .. +10, so we need 32-bit math to
+  // evaluate t * KYBER_Q and the subsequent subtraction
+  return (int16_t)(a - t * KYBER_Q);
 }
--- a/mlkem/reduce.c
+++ b/mlkem/reduce.c
@@ -3,6 +3,35 @@
 #include <stdint.h>
 #include "params.h"
 
+// QINV == -3327 converted to uint16_t == -3327 + 65536 == 62209
+static const uint32_t QINV = 62209;  // q^-1 mod 2^16
+
+/*************************************************
+ * Name:        cast_uint16_to_int16
+ *
+ * Description: Cast uint16 value to int16
+ *
+ * Returns:
+ *   input x in     0 .. 32767: returns value unchanged
+ *   input x in 32768 .. 65535: returns (x - 65536)
+ **************************************************/
+#ifdef CBMC
+#pragma CPROVER check push
+#pragma CPROVER check disable "conversion"
+#endif
+static inline int16_t cast_uint16_to_int16(uint16_t x) {
+  // PORTABILITY: This relies on uint16_t -> int16_t
+  // being implemented as the inverse of int16_t -> uint16_t,
+  // which is implementation-defined (C99 6.3.1.3 (3))
+  //
+  // CBMC (correctly) fails to prove this conversion is OK,
+  // so we have to suppress that check here
+  return (int16_t)x;
+}
+#ifdef CBMC
+#pragma CPROVER check pop
+#endif
+
 /*************************************************
  * Name:        montgomery_reduce
  *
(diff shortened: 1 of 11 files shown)


hanno-becker (changes requested):
`montgomery_reduce()` needs a non-trivial precondition to ensure absence of arithmetic overflow. Something like `|a| < 2^31 - KYBER_Q * 2^15`

rod-chapman:
Pre-condition on montgomery_reduce() strengthened appropriately and proofs updated...

mlkem/reduce.c
```diff
@@ -41,26 +41,48 @@ int16_t montgomery_reduce(int32_t a) {
   // Replace C -> C * q in the above and estimate
   // q / 2^17 < 0.0254.
 
-  uint16_t u;
+  // QINV == -3327 converted to uint16_t == -3327 + 65536 == 62209
+  const uint16_t UQINV = 62209;
```
hanno-becker, line 45: I find the new version of the code very unpleasant to read, because of (a) the pragmas, (b) the trivial assertions, but also (c) the added verbosity. (c) may come down to taste, but I personally don't think it helps to write
```c
  int32_t adjustment = (int32_t)t * KYBER_Q;
  int32_t r = a - adjustment;
```
as opposed to
```c
t = (a - (int32_t)t * KYBER_Q) >> 16;
```
Also, I think `u16/u32` as variable names is confusing given their proximitiy to `uint16/32` type names.

Which changes are absolutely necessary here to get the CBMC proof through?

rod-chapman: Noted. I added the "line by line" approach to make sure I could develop the proof one step at a time. I will seek to get it back to be more readable, and use better variable names.

rod-chapman: Updated code and contracts. See commit below.

mlkem/reduce.c
```diff
@@ -41,26 +41,41 @@ int16_t montgomery_reduce(int32_t a) {
   // Replace C -> C * q in the above and estimate
   // q / 2^17 < 0.0254.
 
-  uint16_t u;
-  int16_t t;
-  // Compute a*q^{-1} mod 2^16 in unsigned representatives
-  u = (uint16_t)a * QINV;
+  // QINV == -3327 converted to uint16_t == -3327 + 65536 == 62209
+  const uint32_t UQINV = 62209;
```
hanno-becker, line 45: `QINV` is already treated as an unsigned value, so perhaps we should just set `#define QINV 62209`?

mlkem/reduce.c
```diff
@@ -41,26 +41,41 @@ int16_t montgomery_reduce(int32_t a) {
...
-  // By construction, the LHS is divisible by 2^16
-  t = (a - (int32_t)t * KYBER_Q) >> 16;
-  return t;
+  // which is implementation-defined (C99 6.3.1.3 (3))
+  //
+  // CBMC (rightly) objects to this implementationd-defined cast,
+  // so that particular proof obligation is suppressed here.
+#pragma CPROVER check push
+#pragma CPROVER check disable "conversion"
+  const int16_t t = (int16_t)a_inverted;
```
hanno-becker, line 60: Can we hide this by adding a `static inline cast_uint16_int16`?

mlkem/reduce.c
```diff
@@ -41,26 +41,41 @@ int16_t montgomery_reduce(int32_t a) {
...
-  uint16_t u;
-  int16_t t;
-  // Compute a*q^{-1} mod 2^16 in unsigned representatives
-  u = (uint16_t)a * QINV;
+  // QINV == -3327 converted to uint16_t == -3327 + 65536 == 62209
+  const uint32_t UQINV = 62209;
+
+  //  Compute a*q^{-1} mod 2^16 in unsigned representatives
+  const uint16_t a_inverted =
+      (uint16_t)((((uint32_t)(a & 0xFFFF)) * UQINV) & 0xFFFF);
```
hanno-becker, line 49: Are all those annotations really necessary? It seems like CBMC would complain about _any_ situation where you multiply two `uint16_t` into a `uint16_t` because the multiplication happens promoted, and then gets truncated?

hanno-becker (changes requested):
Thanks Rod, it's looking better, but I would like to explore further how we can keep the code more readable. See comment.

mlkem/reduce.c
rod-chapman: My general approach is to always use a typed constant where possible rather than a #define so they get type-checked properly. It also shows conscious intent on the part of the programmer.

mlkem/reduce.c
rod-chapman: Possibly... let me try that.

mlkem/reduce.c
hanno-becker: That's too big a change from the ref implementation IMO. But it is true that `QINV` is never used as a signed integer, so its definition as `#define QINV -3327` should be changed to `#define QINV 62209`

mlkem/reduce.c
hanno-becker: In this instance, I think it is OK to have the reader stumble, because it actually a very important point. The routine would still be functionally correct with `t` as `uint16`, but the bounds would be _twice as bad_.

rod-chapman: Yes... this works well. The non-portable code is now isolated in a single function, and montgomery_reduce() is much clearer as a result. Proofs and tests all OK.

mlkem/reduce.c
rod-chapman: They "& 0xFFFF" masks are required to convince CBMC that the "*" operator doesn't overflow. Removing either causes multiple proof failures.

hanno-becker: So _any_ uint16 x uint16 -> uint16 multiplication in needs this contortion to go through CBMC?

mlkem/reduce.c
```diff
@@ -3,6 +3,29 @@
...
+ *   input x in 32768 .. 65535: returns (x - 65536)
+ **************************************************/
+static inline int16_t cast_uint16_to_int16(uint16_t x) {
+  // PORTABILITY: This relies on uint16_t -> int16_t
+  // being implemented as the inverse of int16_t -> uint16_t,
+  // which is implementation-defined (C99 6.3.1.3 (3))
+  //
+  // CBMC (correctly) fails to prove this conversion is OK,
+  // so we have to suppress that check here
+#pragma CPROVER check push
```
hanno-becker, line 23: Can you hoist those `#pragma` out of the function and bracket them with `#if defined(CBMC)`? This way, we hopefully don't get issues with compilation in non-CBMC mode because of unknown pragmas.

rod-chapman: Good idea

mlkem/reduce.c
```diff
@@ -41,36 +70,53 @@ int16_t montgomery_reduce(int32_t a) {
...
+  const int16_t t =
+      ((int32_t)barrett_multiplier * a + (1 << (BPOWER - 1))) >> BPOWER;
 
-  t = ((int32_t)v * a + (1 << 25)) >> 26;
-  t *= KYBER_Q;
-  return a - t;
+  // t is in -10 .. +10, so we need 32-bit math to
+  // evaluate t * KYBER_Q and the subsequent subtraction from a
+  const int32_t adjustment = (int32_t)t * KYBER_Q;
+  const int32_t result = (int32_t)a - adjustment;
```
hanno-becker, line 120: If you make `t` an `int32_t`, you can just write `result = a - t * KYBER_Q`, no?

mlkem/reduce.c
```diff
@@ -41,36 +70,51 @@ int16_t montgomery_reduce(int32_t a) {
...
+  // we assume it's sign-preserving "arithmetic" shift right. (C99 6.5.7 (5))
+  const int32_t t = (barrett_multiplier * a + (1 << (BPOWER - 1))) >> BPOWER;
 
-  t = ((int32_t)v * a + (1 << 25)) >> 26;
-  t *= KYBER_Q;
-  return a - t;
+  // t is in -10 .. +10, so we need 32-bit math to
+  // evaluate t * KYBER_Q and the subsequent subtraction
+  const int32_t result = (int32_t)a - t * KYBER_Q;
+  return (int16_t)result;
```
hanno-becker, line 119:
```suggestion
  return (a - t * KYBER_Q);
```

hanno-becker (approved):
LGTM

PR #233, opened 2024-10-14
Nix cache enhancement in CI and ping clang-tools version to `17.0.6`

potsrevennil:
- Re-ping clang-tools version to `17.0.6` as #228 is reverted in #230
- Solve nix cache issue on CI
  - `magic-nix-cache` somehow doesn't work on ec2 instances
  - `magic-nix-cache` runs into GH rate limit easily
  - replacing `magic-nix-cache` to `cache-nix-action` which is a wrapper over gH official `actions/cache`.
  - for adopting `cache-nix-action`, GH shared runner needed to install nix in single user mode
  - now nix cache is split by os/arch/nix shell/nix install mode/hash of nix files, for avoiding cache pollution
  - cache is sharable among branches, since we suffix cache by hash of nix files, it should be safe to share among branches
  
- ci nix shells are adjusted accordingly
  - ci: native shells
  - ci-cross: native shells + cross compiled gcc and glibc
  - ci-cbmc: ci shell + cbmc
  - ci-cbmc-cross: ci-cross shell + cbmc
  - default, ci-linter: stays as it was, not changed

- Since ci and ci-linter shells don't require long build time, everything is cached in public cache server, jobs that uses these shells should disable cache function, to avoid reaching GH cache storage limit (10 GB)
 
- Currently built caches are:
  - `Linux-ARM64-ci-cross-...`: built and used by GH shared arm64 runners
  - `Linux-X64-ci-cross-...`: built and used by GH shared x86_64 runners
  - `Linux-X64-ci-cross-multi-...`: built and used by x86_64 ec2's
  - `Linux-ARM64-ci-cbmc-cross-multi-`: built and used by arm64 ec2's which runs cbmc proofs

### DIFF
--- a/.github/actions/bench/action.yml
+++ b/.github/actions/bench/action.yml
@@ -34,7 +34,7 @@ inputs:
     default: "ci"
   nix-cache:
     description: Determine whether to enable nix cache
-    default: 'true'
+    default: 'false'
   nix-verbose:
     description: Determine wether to suppress nix log or not
     default: 'false'
--- a/.github/actions/bench/action.yml
+++ b/.github/actions/bench/action.yml
@@ -52,6 +52,7 @@ runs:
         nix-shell: ${{ inputs.nix-shell }}
         nix-cache: ${{ inputs.nix-cache }}
         nix-verbose: ${{ inputs.nix-verbose }}
+        gh_token: ${{ inputs.gh_token }}
         custom_shell: ${{ inputs.custom_shell }}
         script: |
           ARCH=$(uname -m)
--- a/.github/actions/bench/action.yml
+++ b/.github/actions/bench/action.yml
@@ -31,7 +31,7 @@ inputs:
     required: true
   nix-shell:
     description: Run in the specified Nix environment if exists
-    default: "bench"
+    default: "ci"
   nix-cache:
     description: Determine whether to enable nix cache
     default: 'true'
--- a/.github/actions/setup-nix/action.yml
+++ b/.github/actions/setup-nix/action.yml
@@ -23,6 +23,9 @@ inputs:
     description: Determine whether to purge cache with primary key or not
     required: false
     default: 'true'
+  gh_token:
+    description: Github access token to use
+    required: true
 
 runs:
   using: composite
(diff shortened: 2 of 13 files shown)


hanno-becker (changes requested):
The change looks innocent, but the CI runtime for the EC2 x86_64 runners seems to drop from ~7 to ~22 min, which I would very much like to avoid if possible.

potsrevennil:
Yes indeed, I will investigate this further, so please don't merge this PR yet!

.github/actions/setup-nix/action.yml
```diff
@@ -20,46 +20,73 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Nix install mode
+      shell: bash
+      run: |
+        if [[ ${{ runner.os }} != 'Linux' || $USER == 'root' ]]; then
```
hanno-becker, line 36: I think the logic here is not as described in the commit message. The commit message says a Linux root would have single-user mode, but this gives it multi-user mode.

.github/actions/bench/action.yml
```diff
@@ -52,6 +52,7 @@ runs:
         nix-shell: ${{ inputs.nix-shell }}
         nix-cache: ${{ inputs.nix-cache }}
         nix-verbose: ${{ inputs.nix-verbose }}
+        gh_token: ${{ inputs.gh_token }}
```
hanno-becker, line 55: What's the relevance of those tokens if they are not used for anything?

hanno-becker (approved):
I must admit that the details of this are out of my depths, and I have to trust you on that @potsrevennil. What I can see though is that everything is _much_ faster now, which is amazing. Makes our development easier and reduces contention and cost on EC2. Bravo!

hanno-becker (commented):
On second thought, I am not comfortable enough with my lack of understanding of the commit which propogates GH access tokens everywhere. Let's discuss this further when you have time again @potsrevennil. I opened https://github.com/pq-code-package/mlkem-c-aarch64/pull/253 to explore if the work is still similarly impactful without token propagation.

hanno-becker:
Ok, I found https://github.com/NixOS/nix/issues/4653 which explained that authenticated requests have higher rate limit, and that putting the token in `nix.conf` is the recommended way to proceed.

hanno-becker (approved):

PR #235, opened 2024-10-14
Add proof artefacts for poly_tobytes()

rod-chapman:
This PR adds contracts and proof artefacts for poly_tobytes()

Code updated to add contracts, and make type conversions more explicit for CBMC.

### DIFF
--- a/mlkem/poly.c
+++ b/mlkem/poly.c
@@ -202,8 +202,7 @@ void poly_tobytes(uint8_t r[KYBER_POLYBYTES], const poly *a) {
   for (unsigned int i = 0; i < KYBER_N / 2; i++)
       // clang-format off
   ASSIGNS(i, OBJECT_WHOLE(r))
-  INVARIANT(i >= 0)
-  INVARIANT(i <= KYBER_N / 2)
+  INVARIANT(i >= 0 && i <= KYBER_N / 2)
   DECREASES(KYBER_N / 2 - i)
     // clang-format on
     {
@@ -221,10 +220,10 @@ void poly_tobytes(uint8_t r[KYBER_POLYBYTES], const poly *a) {
       // Most significant bits 8 - 11 of t0 become the least significant
       // nibble of the second byte. The least significant 4 bits
       // of t1 become the upper nibble of the second byte.
-      r[3 * i + 1] = (uint8_t)(((t0 >> 8) & 0x0F) | ((t1 << 4) & 0xF0));
+      r[3 * i + 1] = (uint8_t)((t0 >> 8) | ((t1 << 4) & 0xF0));
 
       // Bits 4 - 11 of t1 become the third byte.
-      r[3 * i + 2] = (uint8_t)((t1 >> 4) & 0xFF);
+      r[3 * i + 2] = (uint8_t)(t1 >> 4);
     }
 }
 #else  /* MLKEM_USE_NATIVE_POLY_TOBYTES */
--- a/mlkem/poly.c
+++ b/mlkem/poly.c
@@ -215,15 +215,15 @@ void poly_tobytes(uint8_t r[KYBER_POLYBYTES], const poly *a) {
       // 3 bytes, as follows.
 
       // Least significant bits 0 - 7 of t0.
-      r[3 * i + 0] = (uint8_t)(t0 & 0xFF);
+      r[3 * i + 0] = t0 & 0xFF;
 
       // Most significant bits 8 - 11 of t0 become the least significant
       // nibble of the second byte. The least significant 4 bits
       // of t1 become the upper nibble of the second byte.
-      r[3 * i + 1] = (uint8_t)((t0 >> 8) | ((t1 << 4) & 0xF0));
+      r[3 * i + 1] = (t0 >> 8) | ((t1 << 4) & 0xF0);
 
       // Bits 4 - 11 of t1 become the third byte.
-      r[3 * i + 2] = (uint8_t)(t1 >> 4);
+      r[3 * i + 2] = t1 >> 4;
     }
 }
 #else  /* MLKEM_USE_NATIVE_POLY_TOBYTES */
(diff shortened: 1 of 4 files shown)


mlkem/poly.c
```diff
@@ -198,17 +198,34 @@ void poly_decompress(poly *r, const uint8_t a[KYBER_POLYCOMPRESSEDBYTES]) {
...
+      // significant data, so these can be packed into 24 bits or exactly
+      // 3 bytes, as follows.
+
+      // Least significant bits 0 - 7 of t0.
+      r[3 * i + 0] = (uint8_t)(t0 & 0xFF);
+
+      // Most significant bits 8 - 11 of t0 become the least significant
+      // nibble of the second byte. The least significant 4 bits
+      // of t1 become the upper nibble of the second byte.
+      r[3 * i + 1] = (uint8_t)(((t0 >> 8) & 0x0F) | ((t1 << 4) & 0xF0));
```
hanno-becker, line 224: The trunction `(t0 >> 8) & 0x0F` has no effect, does it? Should CBMC not be able to reason about this based on the precondition that `t0 < KYBER_Q`?

mlkem/poly.c
```diff
@@ -198,17 +198,34 @@ void poly_decompress(poly *r, const uint8_t a[KYBER_POLYCOMPRESSEDBYTES]) {
...
+
+      // Bits 4 - 11 of t1 become the third byte.
+      r[3 * i + 2] = (uint8_t)((t1 >> 4) & 0xFF);
```
hanno-becker, line 227: The trunction `(t1 >> 4) & 0xFF` has no effect, does it? Should CBMC not be able to reason about this based on the precondition that `t1 < KYBER_Q`?

mlkem/poly.c
```diff
@@ -198,17 +198,34 @@ void poly_decompress(poly *r, const uint8_t a[KYBER_POLYCOMPRESSEDBYTES]) {
...
-    t1 = a->coeffs[2 * i + 1];
-    // REF-CHANGE: Precondition change, we assume unsigned canonical data
-    r[3 * i + 0] = (t0 >> 0);
-    r[3 * i + 1] = (t0 >> 8) | (t1 << 4);
-    r[3 * i + 2] = (t1 >> 4);
-  }
+  for (unsigned int i = 0; i < KYBER_N / 2; i++)
+      // clang-format off
+  ASSIGNS(i, OBJECT_WHOLE(r))
+  INVARIANT(i >= 0)
```
hanno-becker, line 205: Are all those annotations necessary, or can CBMC figure some of this out itself? I'd like to reduce the clutter if possible.

mlkem/poly.c
rod-chapman: You're right - in this one case it can conclude that t >> 8 is OK without the mask. I will remove it.

mlkem/poly.c
rod-chapman: Correct again. I will remove the mask.

mlkem/poly.c
rod-chapman: Afraid not - CBMC knows that i can be assigned to in the body of the loop, so does not infer its range automatically at the moment. The range constraints on i can be reduced to a conjunction on a single line though, so I will do that.

hanno-becker: How about introducing an abbreviation like `LOOP(i, lower_bound, upper_bound)` -- still unsatisfactory because it's the same information that's in the `for`, but less text than `ASSIGNS + INVARIANT + INVARIANT`

rod-chapman: Not sure. I think we can only have ONE ASSIGNS clause, and it has to be "just right" for each loop in particular, including other variables that get changed by the loop body.

hanno-becker: Can you double-check?

rod-chapman: Yes.. _exactly_ one ASSIGNS contract must be given.

mlkem/poly.c
```diff
@@ -198,17 +198,33 @@ void poly_decompress(poly *r, const uint8_t a[KYBER_POLYCOMPRESSEDBYTES]) {
...
+      const uint16_t t0 = a->coeffs[2 * i];
+      const uint16_t t1 = a->coeffs[2 * i + 1];
+      // REF-CHANGE: Precondition change, we assume unsigned canonical data
+
+      // t0 and t1 are both < KYBER_Q, so contain at most 12 bits each of
+      // significant data, so these can be packed into 24 bits or exactly
+      // 3 bytes, as follows.
+
+      // Least significant bits 0 - 7 of t0.
+      r[3 * i + 0] = (uint8_t)(t0 & 0xFF);
```
hanno-becker, line 218: The `(uint8_t)( ... )` are not needed in any of the 3 assignments, I think.

hanno-becker (changes requested):
Almost there, one minor simplification pending

mlkem/poly.c
rod-chapman: Correct. I have removed them and pushed. proofs and tests all OK.

cothan:
I suggest update our current Makefile:

from:
```bash
CFLAGS += \
	$(ARCH_FLAGS) \
	-Wall \
	-Wextra \
	-Wpedantic \
	-Werror \
	-Wmissing-prototypes \
	-Wshadow \
	-Wpointer-arith \
	-Wno-unknown-pragmas \
	-O3 \
	-fomit-frame-pointer \
        -std=c99 \
	-pedantic \
	-MMD \
```

to support a few more conversion warnings flags, the commit https://github.com/pq-code-package/mlkem-c-aarch64/pull/235/commits/376127c0aa66a6fb06fb68dea4ff1694b2c40541 may introduce warning in compiler: 

```bash
CFLAGS += -Wall -Werror -Wextra -Wpedantic -Wmissing-prototypes -Wredundant-decls \
  -Wshadow -Wpointer-arith -Wno-unknown-pragmas -pedantic -Wimplicit-int-conversion \
  -O3 -g3 -fomit-frame-pointer -Wsign-conversion -Wshorten-64-to-32 \
```

hanno-becker:
@cothan Good point, though we should do that for the whole code base. Can you open an issue?

cothan:
@hanno-becker , I'll make a PR after this, let's merge this PR first.

PR #236, opened 2024-10-14
Remove explicit zipping/unzipping of Keccak-x4 state in AVX2 FIPS202 backend

hanno-becker:
The FIPS202 C<->native interface operates on de-interleaved batched Keccak states. That is, the Keccak-x4 state is the concatenation, not the interleaving, of four individual Keccak states.

In contrast, SIMD implementations of batched Keccak naturally operate on interleaved states. This is in particular true for the AVX2 implementation from XKCP which we are presently using. To remedy, the native interface implementation conjugates the XKCP core with zipping/unzipping functions for the (de)interleaving of Keccak states.

This commit slightly improves performance by absorbing the zip/unzip of Keccak-x4 states into the XKCP implementation of Keccak-x4. Specifically, the initial memory->register loads are replaced by gather loads, and the final register->memory stores are replaced by scatter stores. Note that while AVX2 has a gather load, it does not have a scatter store (that's introduced in AVX512), so one has to manually transpose the data.

### DIFF
--- a/fips202/native/x86_64/profiles/xkcp.h
+++ b/fips202/native/x86_64/profiles/xkcp.h
@@ -11,36 +11,9 @@
 
 #if defined(MLKEM_USE_NATIVE_X86_64) && defined(SYS_X86_64_AVX2)
 
-#include <string.h>
-
-static void keccakx4_zip(uint64_t *state) {
-  uint64_t tmp[4 * 25] ALIGN;
-  for (unsigned i = 0; i < 4; i++)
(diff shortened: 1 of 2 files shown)


mkannwischer (approved):
Looks good - thanks @hanno-becker!
I reviewed the performance change for 	ML-KEM-768 encaps: 

Intel Xeon 4th gen (c7i):  49192 -> 49010

My machine (Raptor Lake): 48276 -> 47623

Note that the AVX2 code in the Kyber repo only takes 38680 cycles so there is still a small gap.

PR #237, opened 2024-10-15
Add 'out of the box' make to CI

hanno-becker:
This PR adds another job to the CI workflow which runs out of the box `make quickcheck`on aarch64-ubuntu
and x86_64-ubuntu. With the exception of `lint` which is always run, other jobs are only run if this succeeds.


mkannwischer (approved):
LGTM and works the way it's supposed both on my machine as well as in CI.

PR #245, opened 2024-10-15
Allow overwriting default values CC, CPP, AR, LD

hanno-becker:
This commit changes the Makefiles so that the developer can write

```
CC=XXX make ...
```

to overwrite the default compiler choice `gcc` with `XXX`.

This is useful for quick exploration of different compiler choices.


potsrevennil (approved):
Thanks for the work!!

PR #246, opened 2024-10-16
Add first draft of "Proof Guide" document for review.

rod-chapman:
This PR adds a "Cookbook" for how to develop and run CBMC proofs for the C code in this repo.

The guide uses the proof of poly_tobytes() as an example, so PR #235 should be reviewed and merged first.


cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,391 @@
+# CBMC Proof Guide and Cookbook for MLKEM-C
```
hanno-becker, line 1: Maybe put this into `cbmc/` as `README.md`? Or at least reference it from there? Otherwise people might not find it, while I imagine a few people will curiously open `cbmc/` and stumble into whatever README is there.

hanno-becker:
I think this should also be reviewed by someone who does _not_ know CBMC. @mkannwischer What about you? I imagine you don't have time until November?

cbmc/proofs/proof_guide.md
rod-chapman: Well... I put it in cbmc/proofs because it's a proof-related artefact, but yes - let's link to it from the main README. I will take a look.

rod-chapman: I have added comments and links in the top-level README and in cbmc/proofs/README.md
See what you think.

README.md
```diff
@@ -66,6 +65,14 @@ After running `nix develop` you should automatically have a number of support sc
 - [`format`](scripts/format) formats all files. The format is enforced by our CI, so you should run this script prior to committing.
 - [`tests`](scripts/tests) run functional, kat tests natively or emulate them using `QEMU`. For information on how to use the script, please refer to the `--help` option.
 
+## Formal Verification
+
+We are currently aiming to prove that the C code is memory-safe, type-safe, and free of undefined behaviour, using the [CBMC](https://github.com/diffblue/cbmc) verification tool.
+The nix environment automatically installs CBMC and related tools.
```
hanno-becker, line 71:
```
... when using the `ci-cbmc` shell.
nix develop --experimental-features 'nix-command flakes' .#ci-linter
```

rod-chapman: I just do 
```
nix develop --extra-experimental-features flakes --extra-experimental-features nix-command
```
from the root directory of the repo, and everything works... is that not right?

cbmc/proofs/README.md
```diff
@@ -26,6 +27,10 @@ If `GITHUB_STEP_SUMMARY` is set, the proof summary will be appended to it.
 
 # Covered functions
 
-The following functions are currently covered:
+Each proved function has a eponymous sub-directory of its own. The shell command
```
hanno-becker, line 30:
```suggestion
Each proved function has an eponymous sub-directory of its own. The shell command
```

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
+[//]: # (SPDX-License-Identifier: CC-BY-4.0)
+
+# CBMC Proof Guide and Cookbook for MLKEM-C
+
+This doc acts as a guide to developing proofs of C code using CBMC. It concentrates
+on the use of _contracts_ to achieve _unbounded_ and _modular_ proofs of type-safety
+and correctness properties.
+
+This document uses the abbreviated forms of the CBMC contracts defined by macros in the
+cbmc.h header file in the MLKEM-C sources.
```
hanno-becker, line 10:
```suggestion
[cbmc.h](../../mlkem/cbmc.h) header file in the MLKEM-C sources.
```

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+INVARIANT(i >= 0)
+INVARIANT(i <= C)
+DECREASES(C - i)
+// clang-format on
+{
+    S;
+}
+```
+
+The `i <= C` in the invariant is NOT a typo. CBMC places the invariant just _after_ the loop counter has been incremented, but just _before_ the loop exit test, so it is possible for `i == C` at the invariant on the final iteration of the loop.
```
hanno-becker, line 49: I still find this confusing. Why is `INVARIANT(i >= 0)` not `INVARIANT(i > 0)`, then? 

In my mind, the invariant captures what remains true while the loop runs -- and that, in this instance, is `i >= 0` and `i < C`. So I share the confusion that you try to resolve here :-)

rod-chapman: For whatever reason, CBMC models such for loops:
```
for (int i = 0; i < C; i++)
INVARIANT(P)
{
  S;
}
```
as
```
int i = 0;
do
{
  INVARIANT(P);
  if !(i < C) break;
  S;
  i++;
} while(1);
```
Putting the exit test _after_ the Invariant means that it _is_ possible for i == C just before the loop exits.

hanno-becker: Thanks Rod, that's helpful

hanno-becker: You may want to add this here, as I won't be the only one confused by this.

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+5. Supply loop-invariants (if requied) and other interior contracts
+6. Prove it!
+
+These steps are expanded on in the following sub-sections
+
+### Populate a proof directory
+
+For MLKEM-C-AArch64, proof directories lie below `cbmc/proofs`
+
+Create a new sub-directory in there, where the name of the directory is the name of the function _without_ the `$(KYBER_NAMESPACE)` prefix.
```
hanno-becker, line 198: KYBER -> MLKEM

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+That directory needs to contain 3 files.
+
+* cbmc-proof.txt
+* Makefile
+* XXX_harness.c
+
+where "XXX" is the name of the function being proved - same as the directory name.
+
+I suggest that you copy these files from an existing proof directory and modify the latter two. The `cbmc-proof.txt` file is just a marker that says "this directory contains a CBMC proof" to the tools, so no modification is required.
```
hanno-becker, line 208: Consider adding a template directory?

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+### Update Makefile
+
+The `Makefile` sets options and targets for this proof. Let's imagine that the function we want to prove is called `XXX` (without the `$(KYBER_NAMESPACE)` prefix).
```
hanno-becker, line 212: Namespace changed to MLKEM

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+Edit the Makefile and update the definition of the following variables:
+
+* HARNESS_FILE - should be `XXX_harness`
+* PROOF_UID - should be `XXX`
+* PROJECT_SOURCES - should the files containing the source code of XXX
+* CHECK_FUNCTION_CONTRACTS - set to the `XXX`, but _with_ the `$(KYBER_NAMESPACE)` prefix if required
```
hanno-becker, line 219: Namespace changed

hanno-becker: And in multiple places below as well, no longer commenting on each of them.

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+The file `XXX_harness.c` should declare a single function called `XXX_harness()` that calls `XXX` exactly one, with appropriately typed parameters. Using contracts, this harness function should NOT need to contain any CBMC `ASSUME` or `ASSERT` statements at all.
+
+# Supply top-level contracts
+
+Now for the tricky bit. Does `XXX` need any top-level pre- and post-condition contracts, other that those inferred by its type-signature?
+
+A common case is where a function has an array parameter, where the numeric range of the array's elements are NOT just the range of the underlying predefined type.
+
+Secondly, a parameter of a predefined integer type might have to be constrained to a small range.
```
hanno-becker, line 237: `s/Secondly/Similarly`

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+Check for any BOUND macros in the body that give range constraints on the parmeters, and make sure that your contracts are identical or consistent with those. Also read the comments for XXX to make sure they agree with your contracts.
```
hanno-becker, line 239: `s/parmeters/parameters`

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+The order of the top-level contracts for a function is always:
```
hanno-becker, line 241: Emphasize this a bit more? This is unexpected and an easy source of mistakes.

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+```
+t1 XXX(params...)
+    // clang-format off
+REQUIRES()
+ASSIGNS()
+ENSURES();
+// clang-format on
+```
+with a final semi-colon on the end of the last one.
```
hanno-becker, line 250: Again, this is quite unexpected, so maybe worth emphasizing more.

cbmc/proofs/proof_guide.md
rod-chapman: If we add a template directory under "proofs", then the run-cbmc-proofs.py script will spot it and try to run the Makefile in there... this might cause a spurious failure.

cbmc/proofs/proof_guide.md
rod-chapman: Changed throughout

cbmc/proofs/proof_guide.md
rod-chapman: Done

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+### Update harness function
+
+The file `XXX_harness.c` should declare a single function called `XXX_harness()` that calls `XXX` exactly one, with appropriately typed parameters. Using contracts, this harness function should NOT need to contain any CBMC `ASSUME` or `ASSERT` statements at all.
+
+# Supply top-level contracts
+
+Now for the tricky bit. Does `XXX` need any top-level pre- and post-condition contracts, other that those inferred by its type-signature?
+
+A common case is where a function has an array parameter, where the numeric range of the array's elements are NOT just the range of the underlying predefined type.
```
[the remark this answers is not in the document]
rod-chapman: Done and added emphasis in italics.

cbmc/proofs/proof_guide.md
rod-chapman: Done

cbmc/proofs/proof_guide.md
rod-chapman: Added emphasis.

cbmc/proofs/proof_guide.md
rod-chapman: Added extra sentence to emphasize this rule.

cbmc/proofs/proof_guide.md
rod-chapman: Changed throughout.

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+The `Makefile` sets options and targets for this proof. Let's imagine that the function we want to prove is called `XXX` (without the `$(MLKEM_NAMESPACE)` prefix).
+
+Edit the Makefile and update the definition of the following variables:
+
+* HARNESS_FILE - should be `XXX_harness`
+* PROOF_UID - should be `XXX`
+* PROJECT_SOURCES - should the files containing the source code of XXX
+* CHECK_FUNCTION_CONTRACTS - set to the `XXX`, but _with_ the `$(MLKEM_NAMESPACE)` prefix if required
+* USE_FUNCTION_CONTRACTS - a list of functions that `XXX` calls where you want CBMC to use the contracts of the called function for proof, rather than 'inlining' the called function for proof. Include the `$(MLKEM_NAMESPACE)` prefix if required
+* CBMCFLAGS - additional flags to pass to the final run of CBMC. This is normally set to `--smt2` which tells cbmc to run Z3 as its underlying solver. Can also be set to `--bitwuzla` which is sometimes better at generaing counter-examples when Z3 fails.
```
hanno-becker, line 221: You told me about some issue with quantified variables generated by Z3 that CBMC can't handle, leading to odd errors -- and the resulting recommendation to use bitwuzla for initial proofs, and switch to Z3 in the end. Would you mind adding something along those lines to the doc, too?

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+### Iterating over an array for a for loop
+
+A common pattern - doing something to every element of an array. An example would be setting every element of a byte-array to 0x00 given a pointer to the first element and a length. Initially, we want to prove type safety of this function, so we won't even bother with a post-condition. The function specification might look like this:
+
+```
+void zero_array_ts (uint8_t *dst, int len)
+    // clang-format off
+REQUIRES(IS_FRESH(dst, len))
+ASSIGNS(OBJECT_WHOLE(dst));
```
mkannwischer, line 59: Could you add some words about what `IS_FRESH` and `OBJECT_WHOLE` are doing, please?

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+    }
+}
+```
+
+Rules of thumb:
+1. Don't overload your program variables with quantified variables inside your FORALL contracts. It get confusing if you do.
+2. The type of the quanitified variable is _signed_ "int". This is important.
+3. The path in the program from the loop-invariant, through the final loop exit test, to the implicit `return` statement is "nothing" or a "null statement". We need to prove that (on the final iteration), the loop invariant AND the loop exit condidtion imply the post-condition. Imagine having to do that if there's some really complex code _after_ the loop body.
+
+This pattern also brings up another important trick - the use of "null ranges" in FORALL expressions. Consider the loop invariant above. We need to prove that it's true on the first entry to the loop (i.e. when i == 0).
```
mkannwischer, line 123: I think "we need to prove" is a bit confusing, because if I understand correctly CBMC does this by itself if the annotations are all correct, right? Maybe rephrase to emphasize that this is what is happening under the hood - which I guess is useful to understand when debugging.

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+The `==>` is a logical implication, and we know that `False ==> ANYTHING` is True, so all is well.
+
+This comes up a lot. You often end up reasoning about such "slices" of arrays, where one or more of the slices has "null range" at either the loop entry, or at the loop exit point, and therefore that particular quantifier "disappears". Several examples of this trick can be found on the MLKEM codebase.
+
+This also explains why we prefer to use _signed_ integers for loop counters and quantified variables - to allow "0 - 1" to evaluate to -1.  If you use an unsigned type, then "0 - 1" evaluated to something like UINT_MAX, and all hell breaks loose.
+
+## Invariant && loop-exit ==> post-condition
+
+Another important sanity check. If there are no statements following the loop body, then the loop invariant AND the loop exit condition should imply the post-condition. For the example above, we need to prove:
```
mkannwischer, line 145: same here: "we need to prove"

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+* USE_FUNCTION_CONTRACTS - a list of functions that `XXX` calls where you want CBMC to use the contracts of the called function for proof, rather than 'inlining' the called function for proof. Include the `$(MLKEM_NAMESPACE)` prefix if required
+* CBMCFLAGS - additional flags to pass to the final run of CBMC. This is normally set to `--smt2` which tells cbmc to run Z3 as its underlying solver. Can also be set to `--bitwuzla` which is sometimes better at generaing counter-examples when Z3 fails.
+* FUNCTION_NAME - set to `XXX` with the `$(MLKEM_NAMESPACE)` prefix if required
+* CBMC_OBJECT_BITS. Normally set to 8, but might need to be increased if CBMC runs out of memory for this proof.
+
+For documentation of these (and the other) options, see the `cbmc/proofs/Makefile.common` file.
+
+### Update harness function
+
+The file `XXX_harness.c` should declare a single function called `XXX_harness()` that calls `XXX` exactly one, with appropriately typed parameters. Using contracts, this harness function should not need to contain any CBMC `ASSUME` or `ASSERT` statements at all.
```
mkannwischer, line 229:
```suggestion
The file `XXX_harness.c` should declare a single function called `XXX_harness()` that calls `XXX` exactly once, with appropriately typed parameters. Using contracts, this harness function should not need to contain any CBMC `ASSUME` or `ASSERT` statements at all.
```

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+Proof of a single function can be run from the proof directory for that function with `make result`.
+
+This produces `logs/result.txt` in plaintext format.
+
+Before pushing a new proof for a new function, make sure that _all_ proofs run OK from the `cbmc/proofs` directory with
+
+```
+export MLKEM_K=3 # or 2 or 4...
+./run-cbmc-proofs.py --summarize -j8
```
mkannwischer, line 268:
```suggestion
./run-cbmc-proofs.py --summarize -j$(nproc)
```

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+```
+
+Change the "8" to match the number of CPU cores that you have available.
```
mkannwischer, line 271:
```suggestion
```

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+ *                   (of MLKEM_POLYBYTES bytes)
+```
+So we need to write a REQUIRES contract to constrain the ranges of the coefficients denoted by the parameter `a`. There is no constraint on the output byte array, other than it must be the right length, which is given by the function prototype.
+
+We can use the macros in `cbmc.h` to help, thus:
+
+```
+void poly_tobytes(uint8_t r[MLKEM_POLYBYTES], const poly *a)
+     // clang-format off
+REQUIRES(a != NULL && IS_FRESH(a, sizeof(poly)))
```
mkannwischer, line 327: In the previous example using `IS_FRESH` there was no `a != NULL`. Are both required or does `IS_FRESH` imply that the pointer is not null?

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+REQUIRES(ARRAY_IN_BOUNDS(int, k, 0, (MLKEM_N - 1), a->coeffs, 0, (MLKEM_Q - 1)))
```
mkannwischer, line 328: Can you add some words about what each argument of `ARRAY_IN_BOUNDS` is?
In particular, please mention that bounds are inclusive.

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+2. Loop counter variable `i` is in range 0 .. MLKEM_N / 2 at the point of the loop invariant (remember the pattern above).
+3. The loop terminates because the expression `MLKEM_N / 2 - i` decreases on every iteration.
+
+Therefore, we add:
+
+```
+  for (unsigned int i = 0; i < MLKEM_N / 2; i++)
+      // clang-format off
+  ASSIGNS(i, OBJECT_WHOLE(r))
+  INVARIANT(i >= 0)
```
mkannwischer, line 353: Isn't this sort of pointless as it's unsigned? Should we change the loop counter to `int` in those cases?

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+```
+** 0 of 228 failed (1 iterations)
+VERIFICATION SUCCESSFUL
+```
+
+We can also use the higher-level Python script to prove just that one function:
+
+```
+cd cbmc/proofs
+./run-cbmc-proofs.py --summarize -j8 -p poly_tobytes
```
mkannwischer, line 391:
```suggestion
MLKEM_K=3 ./run-cbmc-proofs.py --summarize -j$(nproc) -p poly_tobytes
```
MLKEM_K does not have a default value, so this will fail.

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+The `Makefile` sets options and targets for this proof. Let's imagine that the function we want to prove is called `XXX` (without the `$(MLKEM_NAMESPACE)` prefix).
+
+Edit the Makefile and update the definition of the following variables:
+
+* HARNESS_FILE - should be `XXX_harness`
+* PROOF_UID - should be `XXX`
+* PROJECT_SOURCES - should the files containing the source code of XXX
+* CHECK_FUNCTION_CONTRACTS - set to the `XXX`, but _with_ the `$(MLKEM_NAMESPACE)` prefix if required
+* USE_FUNCTION_CONTRACTS - a list of functions that `XXX` calls where you want CBMC to use the contracts of the called function for proof, rather than 'inlining' the called function for proof. Include the `$(MLKEM_NAMESPACE)` prefix if required
```
mkannwischer, line 220: Could you comment on in which cases I should use the contracts and in which cases I should use inlining? My understanding is that we should use the contracts as much as possible.

mkannwischer (changes requested):
Thanks @rod-chapman - this is excellent.
Let's merge this after some very minor edits. 

Sorry for the the long delay - I wanted to take the time and write my first proof at the same time. I just added a proof for poly_tomont in #307 .

cbmc/proofs/proof_guide.md
hanno-becker, line 229: Can we expand on this? I get the impression that the harness MUST NOT use any stack allocated structures, but literally just declare variables of the right types and pass them to the function under proof, right? See #325

cbmc/proofs/proof_guide.md
```diff
@@ -218,12 +218,28 @@ Edit the Makefile and update the definition of the following variables:
...
+* EXTERNAL_SAT_SOLVER - should _always_ be "nothing" to prevent CBMC selecting a SAT backend over the selected SMT backend.
 * CBMCFLAGS - additional flags to pass to the final run of CBMC. This is normally set to `--smt2` which tells cbmc to run Z3 as its underlying solver. Can also be set to `--bitwuzla` which is sometimes better at generaing counter-examples when Z3 fails.
 * FUNCTION_NAME - set to `XXX` with the `$(MLKEM_NAMESPACE)` prefix if required
 * CBMC_OBJECT_BITS. Normally set to 8, but might need to be increased if CBMC runs out of memory for this proof.
 
 For documentation of these (and the other) options, see the `cbmc/proofs/Makefile.common` file.
 
+#### Z3 or Bitwuzla?
+
+I have found that it's better to use Bitwuzla in the initial stages of developing and debugging a new proof.
```
hanno-becker, line 230: Can you remove 1st person everywhere, and say "We" instead?

cbmc/proofs/proof_guide.md
hanno-becker: See #340. @rod-chapman can you extend the guide accoringly?

cbmc/proofs/proof_guide.md
rod-chapman: Done

tautschnig:
High-level comment: should there be any links towards more-general CBMC training (like https://model-checking.github.io/cbmc-training/)?

cbmc/proofs/proof_guide.md
rod-chapman: Advice on Z3 vs Bitwuzla added...

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
+d[//]: # (SPDX-License-Identifier: CC-BY-4.0)
```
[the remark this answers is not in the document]
rod-chapman: Done

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+
+# CBMC Proof Guide and Cookbook for MLKEM-C
```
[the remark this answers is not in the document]
rod-chapman: Changed consistently to mlkem-c-aarch64 throughout.

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,399 @@
...
+This document uses the abbreviated forms of the CBMC contracts defined by macros in the
+[cbmc.h](../../mlkem/cbmc.h) header file in the MLKEM-C sources.
+
+## Common Proof Patterns
+
+### Loops (general advice)
+
+1. A function should contain at most one outermost loop statement. If the function you're trying to prove has more than one outermost loop, then re-factor it into two or more functions.
+
+2. The one outermost loop statement should be _final_ statement before the function returns. Don't have complicated code _after_ the loop body.
```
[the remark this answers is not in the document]
rod-chapman: Fixed.

cbmc/proofs/proof_guide.md
rod-chapman: Improved wording to "The proof needs to establish that..."

cbmc/proofs/proof_guide.md
rod-chapman: Added words to that effect.

cbmc/proofs/proof_guide.md
rod-chapman: Added explanation

cbmc/proofs/proof_guide.md
rod-chapman: Yes... it's pointless for an unsigned type, but just being explicit.

rod-chapman: Added a note that it can be left out.

cbmc/proofs/proof_guide.md
rod-chapman: updated as suggested.

cbmc/proofs/proof_guide.md
rod-chapman: Added a note on use of contracts vs inlining.

cbmc/proofs/proof_guide.md
rod-chapman: Done

rod-chapman:
Most of that existing training material talks about using CBMC in "bounded model checking" mode, rather than "deductive sound contract-based proof" mode. I think pushing readers towards the former will cause confusion...

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,452 @@
...
+Note that `USE_FUNCTION_CONTRACTS` is left empty since `poly_tobytes()` is a leaf function that does not call any other functions at all.
+
+### Update harness function
+
+`poly_tobytes()` has a simple API, requiring two parameters, so the harness function is:
+
+```
+void harness(void) {
+  poly a;
+  uint8_t r[MLKEM_POLYBYTES];
```
tautschnig, line 350: Both of these seem to contradict the guidance you gave earlier in that pointer-type parameters should not use address-of-stack-allocated-object.

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,452 @@
...
+
+```
+ * Arguments:   INPUT:
+ *              - a: const pointer to input polynomial,
+ *                with each coefficient in the range [0,1,..,Q-1]
+ *              OUTPUT
+ *              - r: pointer to output byte array
+ *                   (of MLKEM_POLYBYTES bytes)
+```
+So we need to write a REQUIRES contract to constrain the ranges of the coefficients denoted by the parameter `a`. There is no constraint on the output byte array, other than it must be the right length, which is given by the function prototype.
```
tautschnig, line 369: Well ... C will just decay that array-type parameter to a pointer, so you don't really get that kind of guarantee (though I don't really know what warnings compilers might produce today if you pass something of the wrong size).

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,452 @@
...
+REQUIRES(ARRAY_IN_BOUNDS(int, k, 0, (MLKEM_N - 1), a->coeffs, 0, (MLKEM_Q - 1)))
+ASSIGNS(OBJECT_WHOLE(r));
+// clang-format on
+```
+
+`ARRAY_IN_BOUNDS` is a macro that expands to a quantified expression that expresses that the elemtns of `a->coeffs` between index values `0` and `(MLKEM_N - 1)` (inclusive) are in the range `0` through `(MLKEM_Q - 1)` (also inclusive). See the macro definition in `mlkem/cbmc.h` for details.
+
+### Interior contracts and loop invariants
+
+`poly_tobytes` has a reasonable simple, single loop statement:
```
tautschnig, line 386:
```suggestion
`poly_tobytes` has a single loop statement:
```

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,452 @@
...
+
+```
+  for (unsigned int i = 0; i < MLKEM_N / 2; i++)
```
tautschnig, line 389: Earlier you said that loop indices should use _signed_ integers.

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,452 @@
...
+  INVARIANT(i >= 0)
+  INVARIANT(i <= MLKEM_N / 2)
+  DECREASES(MLKEM_N / 2 - i)
+    // clang-format on
+    { ... }
+```
+
+Note that the invariant `i >= 0` could be ommitted since `i` is of an `unsigned` integer type. It is given here for clarity only.
+
+Another small set of changes is required to make CBMC happy with the loop body. By default, CBMC is pedantic and warns about conversions that potential truncate values of lose information via an implicit type conversion.
```
tautschnig, line 413:
```suggestion
Another small set of changes is required to make CBMC happy with the loop body. By default, CBMC is pedantic and warns about conversions that truncate values or lose information via an implicit type conversion.
```

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,452 @@
...
+```
+which has an implicit conversion from `uint16_t` to `uint8_t`. This is well-defined in C, but CBMC issues a warning just in case. To make CBMC happy, we have to explicitly reduce the range of t0 with a bitwise mask, and use an explicit conversion, thus:
+```
+r[3 * i + 0] = (uint8_t)(t0 & 0xFF);
+```
+and so on for the other two statements in the loop body.
+
+### Prove it!
+
+With those changes, CBMC completes the proof:
```
tautschnig, line 427: Perhaps give an indication on how long this might take.

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,452 @@
...
+
+```
+void zero_array_ts (uint8_t *dst, int len)
+    // clang-format off
+REQUIRES(IS_FRESH(dst, len))
+ASSIGNS(OBJECT_WHOLE(dst));
+// clang-format on
+```
+
+The `IS_FRESH(dst,len)` in the precondition means that the pointer value `dst` is not `NULL` and is pointing to at least `len` bytes of data. The `ASSIGNS` contract (in this case) means that when the function returns, it promises to have updated the whole object pointed to by dst - in this case `len` bytes of data.
```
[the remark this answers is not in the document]
tautschnig: Also, this might be the place to say a bit more about `IS_FRESH` given all the out-of-band conversation we've had.

tautschnig (approved):

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,452 @@
...
+1. It is good practice to declare the counter variable locally, within the scope of the loop.
+2. The counter variable should be a constant within the loop body. In the example above, do _not_ modify `i` in the body of the loop.
+3. "int" is the best type for the counter variable. "unsigned" integer types complicate matters with their modulo N semantics. Avoid "size_t" since its large range (possibly unsigned 64 bit) can slow proofs down.
+
+CBMC requires basic assigns, loop-invariant, and decreases contracts _in exactly that order_. Note that the contracts appear _before_ the opening `{` of the loop body, so we also need to tell `clang-format` not to reformat these lines. The basic pattern is thus:
+
+```
+for (int i = 0; i < C; i++)
+    // clang-format off
+ASSIGNS(i) // plus whatever else S does
```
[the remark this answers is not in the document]
rod-chapman: Will discuss this idea with other team members.

cbmc/proofs/proof_guide.md
rod-chapman: Noted.

cbmc/proofs/proof_guide.md
rod-chapman: "... in about 10 seconds."

hanno-becker (changes requested):
@rod-chapman Can you wrap this up, by rebasing and cleaning up the history and addressing whatever comments, if any, are still outstanding? The PR has been open for too long, let's get it in.

rod-chapman:
Rebased and squashed. Ready to approve and merge.

hanno-becker (dismissed):
@rod-chapman Can you change all occurrences of `mlkem-c-aarch64` to `mlkem-native`?

cbmc/proofs/proof_guide.md
```diff
@@ -0,0 +1,452 @@
+[//]: # (SPDX-License-Identifier: CC-BY-4.0)
+
+# CBMC Proof Guide and Cookbook for mlkem-c-aarch64
```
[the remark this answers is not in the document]
rod-chapman: Done

mkannwischer (approved):
Thanks @rod-chapman  for writing this up and for taking our comments into account! 
LGTM now! 
If we later want to extend it, we can always have another PR.

PR #250, opened 2024-10-17
Add `tests func` to quickcheck CI

hanno-becker:
Previously, the quickcheck CI (which guards other CI jobs)
would only run `make quickcheck`, but not exercise the `tests`
script.

This commit adds an invocation of `tests all` to the quickcheck CI,
after setting up the necessary dependencies via the newly introduced
`setup-ubuntu` action.


mkannwischer (approved):
Thanks this looks good to me and works out of the box on my machine (and in CI). 
Please clean up the history before you merge ;-)

PR #262, opened 2024-10-28
Rename Kyber -> ML-KEM

### LINKED ISSUE #247: Rename Kyber -> ML-KEM everywhere

mkannwischer:
Resolves #247 
This PR renames Kyber to ML-KEM everywhere to make it a little less confusing. 

The only place I have not changed is the following as that changes testvectors: https://github.com/pq-code-package/mlkem-c-aarch64/blob/3f579b25f33f726e5119cd0afad3df840b6268a8/mlkem/kem.h#L14
Do we want to change that as well? Then we either need to update the testvectors, or somehow pass the `ALGNAME` from outside for that test.


mkannwischer:
I don't understand why this PR did not trigger CI.

hanno-becker:
I don't, either. Can you force-push a 0-diff to see if it's spurious?

hanno-becker (commented):
LGTM, reproduced this locally and confirmed that no unwanted instances of "Kyber" remain.

hanno-becker:
> I don't understand why this PR did not trigger CI.

#263

hanno-becker (approved):

PR #263, opened 2024-10-28
Run CI when PR is opened

hanno-becker:
Previously, CI would only be run when a PR is _modified_, but not when it is originally opened.

This commit fixes that, running CI also when a PR is opened.


mkannwischer (approved):
LGTM. Thanks for catching that!

PR #264, opened 2024-10-28
Add gcc 4.8, 4.9, 7 tests to CI

hanno-becker:
This PR adds nix shells for gcc 4.8, 4.9, 7, 11, and clang-18, and adds build+functest runs for them to CI.


mkannwischer (approved):
Looks good - thanks! 
I'm positivley surprised that this does not slow down CI too much.

PR #266, opened 2024-10-28
Add namespacing 

### LINKED ISSUE #255: Enforce consistent namespacing across library
Most yet not all functions of external linkage are namespaced, notable exceptions being assembly routines. All exported symbols should be namespaced appropriately.

Relates to and should likely be implemented after: https://github.com/pq-code-package/mlkem-c-aarch64/issues/247

mkannwischer:
Resolves #255 

Depends on #267.

This PR is adding namespacing and enforcing it in CI. 

All ML-KEM symbols are now namespaced with `PQCP_MLKEM_NATIVE_MLKEM{512,768,1024}`.
All FIPS202 symbols are namespaced with `PQCP_MLKEM_NATIVE_FIPS202_`.

Assuming libraries are going to link multiple parameter sets, we could consider using a shared namespace for the functions that are the same for all parameter sets (e.g., NTTs). Then you only need to link them once. But that would require more restructuring (e.g., `poly.c` contains both functions that are common and functions that are parameter-set-specific.).

### DIFF
--- a/.github/actions/bench/action.yml
+++ b/.github/actions/bench/action.yml
@@ -79,6 +79,10 @@ runs:
               --cflags="${{ inputs.cflags }}" --arch-flags="${{ inputs.archflags }}" \
               $([[ ${{ inputs.opt }} == "false" ]] && echo "--no-opt")  \
               -v --output=output.json ${{ inputs.bench_extra_args }}
+    - name: Check namespace
+      shell: ${{ env.SHELL }}
+      run: |
+        check-namespace
     - name: Store benchmark result
       if: ${{ inputs.store_results == 'true' }}
       uses: benchmark-action/github-action-benchmark@v1
--- a/scripts/ci/check-namespace
+++ b/scripts/ci/check-namespace
@@ -1,0 +1,71 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+
+# This scripts runs nm on the object files (excluding test objects) and checks that all exported
+# symbols are properly namespaced.
+# It assumes that object files are present under test/build/mlkem{512,768,1024} and
+# test/build/fips202.
+
+# The checked namespaces are
+# PQCP_MLKEM_NATIVE_FIPS202_ for FIPS202 code
+# PQCP_MLKEM_NATIVE_MLKEM512_ for MLKEM512 code
+# PQCP_MLKEM_NATIVE_MLKEM768_ for MLKEM768 code
+# PQCP_MLKEM_NATIVE_MLKEM1024_ for MLKEM1024 code
+
+import subprocess
+import os
+
+def check_file(file_path, namespace):
+    if file_path.endswith("debug.c.o"):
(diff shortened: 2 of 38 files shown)


.github/actions/bench/action.yml
```diff
@@ -79,6 +79,10 @@ runs:
               --cflags="${{ inputs.cflags }}" --arch-flags="${{ inputs.archflags }}" \
               $([[ ${{ inputs.opt }} == "false" ]] && echo "--no-opt")  \
               -v --output=output.json ${{ inputs.bench_extra_args }}
+    - name: Check namespace
```
hanno-becker, line 82: That doesn't seem to belong in the bench tests.

scripts/ci/check-namespace
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+import subprocess
```
hanno-becker, line 3: Can you add a brief comment on what this script is checking?

scripts/ci/check-namespace
```diff
@@ -0,0 +1,60 @@
...
+import os
+
+def check_file(file_path, namespce):
```
hanno-becker, line 6: Typo? `namespce` -- also, it's set in L13?

hanno-becker (changes requested):
Thanks @mkannwischer for getting this chore out of the way!

The `check-namespace` script seems to ignore the namespace parameter.

.github/actions/bench/action.yml
mkannwischer: Do we test all different variants of the asm (esp. Keccak) in the functional tests? 
I was afraid that we may not, so I added it here. It's basically free.

scripts/ci/check-namespace
mkannwischer: Good catch... damn.

.github/actions/bench/action.yml
hanno-becker: Okay.

scripts/ci/check-namespace
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+
+# This scripts runs nm on the object files (excluding test objects) and checks that all exported
+# symbold are properly namespaced.
```
hanno-becker, line 5:
```suggestion
# symbols are properly namespaced.
```

scripts/ci/check-namespace
```diff
@@ -0,0 +1,71 @@
...
+# It assumes that object files are present under test/build/mlkem{512,768,1024} and
+# test/build/fips202.
+
+# The checked name spaces are
```
hanno-becker, line 9:
```suggestion
# The checked namespaces are
```

hanno-becker (approved):
LGTM, thanks @mkannwischer for adding a script to double-check this. I tested locally that the script fails when adding a non-namespaces definition of external linkage.

PR #267, opened 2024-10-29
Fix CI: Re-enable functional tests

mkannwischer:
Somehow our functional tests are not doing what they are supposed to do

For example: https://github.com/pq-code-package/mlkem-c-aarch64/actions/runs/11566834855/job/32196177392?pr=266

This is running:
```
tests all  --cross-prefix="" --cflags="" --opt --no-func --no-kat --no-nistkat --compile --no-run -v
tests all  --cross-prefix="" --cflags="" --opt --no-func --no-kat --no-nistkat --no-compile --run -v
```
Which are both no-ops. I am not quite sure what these are for, but this is an attempt to fix it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,9 +78,9 @@ jobs:
         with:
           gh_token: ${{ secrets.GITHUB_TOKEN }}
           compile_mode: native
-          func: false
-          nistkat: false
-          kat: false
+          func: true
+          nistkat: true
+          kat: true
       - name: native tests (+debug)
         uses: ./.github/actions/multi-functest
         with:
@@ -96,8 +96,9 @@ jobs:
           gh_token: ${{ secrets.GITHUB_TOKEN }}
           compile_mode: cross
           opt: opt
-          func: false
-          nistkat: false
+          func: true
+          nistkat: true
+          kat: true
   compiler_tests:
     name: Compiler tests (${{ matrix.target.name }})
     strategy:


.github/workflows/ci.yml
```diff
@@ -78,9 +78,9 @@ jobs:
         with:
           gh_token: ${{ secrets.GITHUB_TOKEN }}
           compile_mode: native
-          func: false
-          nistkat: false
-          kat: false
+          func: true
+          nistkat: true
+          kat: true
```
hanno-becker, line 83: The expectation was that this would at least build the library, but it doesn't seem to be the case. The reason to this was to save work because the next job runs all tests with debug flag. 
```suggestion
          func: true
          nistkat: false
          kat: false
```

potsrevennil:
@mkannwischer
Seemed like I made a mistake

Originally, we want the `native build` step to only test the plain compilation, then we would like to compile and run the tests with the `debug` flag on.
But the emulation on aarch64 for tests, especially the kat test was taking too long, so I'd like to only run one of the test on GH shared runners, and split the others on ec2s.

However I forgot about the interface of the `tests` script, if `--no-xxx` is set, then the xxx test binary won't be compiled or run.

So maybe the easiest fix is to change the compilation step in `functest` action from 
```
 tests all  --cross-prefix="${{ env._CROSS_PREFIX }}" --cflags="${{ inputs.cflags }}" --${{ env.OPT }} --${{ env.FUNC }} --${{ env.KAT }} --${{ env.NISTKAT }} --compile --no-run -v
```
to
```
 tests all  --cross-prefix="${{ env._CROSS_PREFIX }}" --cflags="${{ inputs.cflags }}" --${{ env.OPT }} --compile --no-run -v
```
making it compiles all tests, and letting the flags for each tests only control the run of the tests instead of both run and compilation.

.github/workflows/ci.yml
hanno-becker: On second thought, the CI runtime seems ok even with your change, so let's keep that.

hanno-becker (approved):
As @potsrevennil mentions, this was originally to save CI costs,

PR #268, opened 2024-10-29
Expand CI

hanno-becker:
* Remove non-working `compile_mode: all` from `multi-functest`
* Extend EC2 CI to run full `{opt, non-opt} x {cross, non-opt}` matrix.
* Add Gv2 CI target
* Move CBMC into separate job


mkannwischer (approved):
Looks good to me. Thanks @hanno-becker.

Why can you not use `compile_mode: all` for the X86 EC2 instances?

PR #269, opened 2024-10-29
Remove unused XKCP symbols 

mkannwischer:
Depends on #266 #267. 

KeccakP-1600-times4-SIMD256.c is copied from XKCP and contains a lot of symbols that we are not actually using.
This commit trims it down to just KeccakP1600times4_PermuteAll_24rounds.


hanno-becker (approved):
LGTM

PR #289, opened 2024-10-31
Add verbal specification and runtime-checking for `fqmul()`

hanno-becker:
This is a preparatory PR for CBMC proofs. It reorders arguments to `fqmul()` so that the signed canonical twiddle is always in the second position, says as much in the textual description of `fqmul()`, and adds runtime assertions checking the input and output bounds during runtime (for debug builds).

### DIFF
--- a/mlkem/debug/debug.h
+++ b/mlkem/debug/debug.h
@@ -37,13 +37,13 @@ void mlkem_debug_assert(const char *file, int line, const char *description,
  *              - description: Textual description of check
  *              - ptr: Base of array to be checked
  *              - len: Number of int16_t in ptr
- *              - lower_bound_inclusive: Inclusive lower bound
- *              - upper_bound_inclusive: Inclusive upper bound
+ *              - lower_bound_exclusive: Exclusive lower bound
+ *              - upper_bound_exclusive: Exclusive upper bound
  **************************************************/
 void mlkem_debug_check_bounds(const char *file, int line,
                               const char *description, const int16_t *ptr,
-                              unsigned len, int lower_bound_inclusive,
-                              int upper_bound_inclusive);
+                              unsigned len, int lower_bound_exclusive,
+                              int upper_bound_exclusive);
 
 /* Print error message to stderr alongside file and line information */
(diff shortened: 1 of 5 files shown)


mlkem/debug/debug.h
```diff
@@ -127,6 +162,12 @@ void mlkem_debug_print_error(const char *file, int line, const char *msg);
 
 #else /* MLKEM_DEBUG */
 
+#define CASSERT(...) \
+  do {               \
+  } while (0)
```
mkannwischer, line 167: Should we maybe prefix all of those with `MLKEM_DEBUG_` or `DEBUG_` to mark that they all belong to the `debug.c`?

mlkem/reduce.h
```diff
@@ -29,17 +30,21 @@ ENSURES(RETURN_VALUE >= -HALF_Q && RETURN_VALUE <= HALF_Q);
  * Description: Multiplication followed by Montgomery reduction
  *
  * Arguments:   - int16_t a: first factor
- *              - int16_t b: second factor
+ *                  Can be any int16_t.
+ *              - int16_t b: second factor.
+ *                  Must be signed canonical (abs value <|q|/2)
```
mkannwischer, line 35: What do you mean by `|q|`? Also the code checkes <= q/2, not < q/2 (as q/2 is rounded down).

mkannwischer (approved):
Nice - very good catch with the swapped arguments.

Some small nits.

mlkem/reduce.h
hanno-becker: I always mean division in the rationals, and with that interpretation of q/2 the comment is correct. However, it's not obvious that it's meant this way, and one should avoid any confusion by using < (q+1)/2 explicitly.

hanno-becker: I will fix this in the follow-up PRs.

mlkem/debug/debug.h
hanno-becker: Yes, could do, but let's do it in a separate PR.

PR #292, opened 2024-10-31
Strengthen spec for `montgomery_reduce()`

### LINKED ISSUE #280: CBMC: Strengthen spec for `montgomery_reduce()`
**Issue:** The current contract for `montgomery_reduce()` only guarantees that the output is within int16 bounds. This is not sufficient to prove the correctness of the matrix-vector base multiplication, which needs to accumulate up to `MLKEM_K` many such reductions. Also, the precondition for `montgomery_reduce()` is different depending on whether it is used for `fqmul()` or `basemul_cached()`.

**Task:** Introduce _two_ contracts for `montgomery_reduce()`, one suitable for application in `fqmul()`, the other suitable for base multiplication. Guard them by preprocessor macros so that in any higher-level CBMC proof, we can choose which spec to pick (as far as I see, we never need both at the same time).

hanno-becker:
Based on: #289 

Fixes #280 

There are two uses of the generic Montgomery reduction `montgomery_reduce()` in the code base:
- For modular multiplication with twiddles in `fqmul()`
- For Montgomery reduction after accumulation in the base multiplication.

Those two uses require different specifications for `montgomery_reduce()`. However, there does not appear to be a mechanism in CBMC to register more than one contract. 

To remedy, this commit moves the previous `montgomery_reduce()` into a static inline `montgomery_reduce_generic()`, and uses it in the definitions of `montgomery_reduce()` and `fqmul()`. The new `montgomery_reduce()` then makes bounds assumptions according to the use in base multiplication, while the call to `montgomery_reduce_generic()` provides stronger input bounds. Accordingly, the post-condition of `montgomery_reduce()` only guarantees coefficients < 3/2q, while `fqmul()` ensures < q.

The CBMC contracts an proofs for `montgomery_reduce()` and `fqmul()` are adjusted accordingly.


mlkem/reduce.c
```diff
@@ -87,6 +88,24 @@ int16_t montgomery_reduce(int32_t a) {
...
+}
+
+int16_t fqmul(int16_t a, int16_t b) {
+  SCALAR_BOUND(b, HALF_Q, "fqmul input");
+
+  int16_t res = montgomery_reduce((int32_t)a * (int32_t)b);
+
+  SCALAR_BOUND(res, MLKEM_Q, "fqmul output");
+  return res;
+}
```
mkannwischer, line 107: Please add similar comments to this as for `montgomery_reduce_generic`. Move them from reduce.h?

mkannwischer (approved):
Semantically this looks good to me. Thanks for clearing this up. 

Some docstring could be moved for consistency.

rod-chapman:
Question: why do montgomery_reduce() and fqmul() have the MLKEM_NAMESPACE macro applied to them? Surely, these functions are common to all parameter sets, so no need to adjust their name according to the value of MLKEM_K?

hanno-becker:
@rod-chapman Everything's prefixed by the K-dependent namespace to e.g. allow linking together multiple compilations of the lib with varying security levels. It's TBD if/how we want to avoid duplication of symbols for functions that don't depend on K.

mlkem/reduce.c
hanno-becker: Hm, why? Documentation ultimately belongs in the header?

rod-chapman: Documentation that a client needs to know goes in the .h file. Documentation that an implementer/maintainer needs to know goes in the .c file. No repetition between the two, and no repetition between comments and contracts either.

rod-chapman:
To avoid symbol duplication, we might need to split the code into translation units that _do_ depend on K, and those that don't. The latter would would be compiled exactly once.

mlkem/reduce.c
hanno-becker: Some repetition is useful, I think, esp. between comments and contracts, given that not everyone knows how to read the latter.

hanno-becker:
@rod-chapman We could do that, but it's a large restructuring that moves us further apart from the reference implementation. So unless we have a customer telling us about for simultaneous need for multiple security levels and deduplicated symbols, I don't think we should spend any time on this.

mlkem/reduce.c
rod-chapman: The actual content of the contract is a C language expression, which I'd expect every C programmer to be able to read. OK... there's a bit of syntactic sugar and macros to pick up. "For all" remains the only really new concept. I can teach any SDE to read contracts in about half a day...

rod-chapman:
Agree on the above re: symbol duplication.

mlkem/reduce.c
hanno-becker: >  I can teach any SDE to read contracts in about half a day...

Sure, but I want any C developer to be able to understand the documentation, without prior teaching. I agree most people will read and interpret the contracts fine, but there's no harm in saying things in plain english as well.

PR #315, opened 2024-11-03
Hoist core of gen_matrix() into helper functions

hanno-becker:
Previously, the matrix generation was done in a single routine, gen_matrix.
This routine is among the most complex in MLKEM. In particular, verification
is likely going to be challenging with the monolithic structure and up to
three nested loops.

This commit aims to facilitate readability and verifiability by hoisting
the core of the matrix generation into separate helper functions:

- gen_matrix_entry(), generating a single matrix entry from a seed buffer
- gen_matrix_entry_x4(), generating 4 matrix entries from 4 seed buffers

Apart from splitting the code between gen_matrix and the new helper
functions, some more comments have been added.

### DIFF
--- a/mlkem/indcpa.c
+++ b/mlkem/indcpa.c
@@ -173,6 +173,7 @@ void gen_matrix(polyvec *a, const uint8_t seed[MLKEM_SYMBYTES],
         seedxy[j][MLKEM_SYMBYTES + 0] = y;
         seedxy[j][MLKEM_SYMBYTES + 1] = x;
       }
+      vec[j] = a[x].vec[y].coeffs;
     }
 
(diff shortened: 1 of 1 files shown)


mkannwischer (approved):
Thanks! This is looking good.
I checked a few performance results and it does not seem to impact performance

Do you want to remove the following TODO while you are already at it? I don't think this needs to be addressed, do you?
```
  // TODO: All loops in this function should be unrolled for decent
  // performance. 
 ```

hanno-becker:
@mkannwischer I prefer to keep it in for now -- I haven't measured yet what the impact of (not) unrolling is.

PR #321, opened 2024-11-03
CBMC: Add spec + proof for rej_uniform

### LINKED ISSUE #316: CBMC: State & prove specs for rej_uniform

hanno-becker:
Resolves #316

### DIFF
--- a/mlkem/rej_uniform.c
+++ b/mlkem/rej_uniform.c
@@ -12,12 +12,22 @@
  *
  * Arguments:   - int16_t *r:          pointer to output buffer
  *              - unsigned int len:    requested number of 16-bit integers
- *                                     (uniform mod q)
+ *                                     (uniform mod q).
  *              - const uint8_t *buf:  pointer to input buffer
  *                                     (assumed to be uniform random bytes)
  *              - unsigned int buflen: length of input buffer in bytes
+ *                                     Must be <= 4096.
+ *                                     Must be a multiple of 3.
  *
- * Returns number of sampled 16-bit integers (at most len)
+ * Note: Strictly speaking, only a few values of buflen near UINT_MAX need
+ * excluding. The limit of 4096 is somewhat arbitary but sufficient for all
+ * uses of this function.
+ *
+ * Returns the number of sampled 16-bit integers, at most len.
+ * If it is strictly less than len, all of the input buffers is
+ * guaranteed to have been consumed. If it is equal to len, no
+ * information is provided on how many bytes of the input buffer
+ * have been consumed.
  **************************************************/
 static unsigned int rej_uniform_scalar(int16_t *r, unsigned int len,
                                        const uint8_t *buf,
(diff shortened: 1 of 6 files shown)


mlkem/rej_uniform.c
```diff
@@ -12,10 +12,11 @@
  *
  * Arguments:   - int16_t *r:          pointer to output buffer
  *              - unsigned int len:    requested number of 16-bit integers
- *                                     (uniform mod q)
+ *                                     (uniform mod q) Must be <= 1024.
  *              - const uint8_t *buf:  pointer to input buffer
  *                                     (assumed to be uniform random bytes)
  *              - unsigned int buflen: length of input buffer in bytes
+ *                                     Must be <= 4096.
```
mkannwischer, line 19: I don't understand where this limitation is coming from? Is this a CBMC limitation? In that case I would not make it part of the docstring here as this is not a real limitation of the C code. 
You can instead add a not to the `REQUIRES` stmt to clarify why that pre-condition exists.

mlkem/rej_uniform.h
```diff
@@ -2,8 +2,35 @@
...
+ **************************************************/
 unsigned int rej_uniform(int16_t *r, unsigned int len, const uint8_t *buf,
-                         unsigned int buflen);
-
+                         unsigned int buflen)  // clang-format off
+REQUIRES(buflen <= 4096 && len <= 1024)
+REQUIRES(r != NULL && IS_FRESH(r, sizeof(int16_t) * len))
+REQUIRES(buf != NULL && IS_FRESH(buf, buflen))
+ASSIGNS(OBJECT_UPTO(r, len * sizeof(int16_t)))
+ENSURES(RETURN_VALUE <= len && \
```
mkannwischer, line 33:
```suggestion
ENSURES(RETURN_VALUE <= len && 
```

mkannwischer (changes requested):
Thanks @hanno-becker. 
Bounds do look good.
I have one questions regarding the input size limitation that probably requires the addition of a comment.

mlkem/rej_uniform.c
hanno-becker: This is not a CBMC limitation -- the `pos + 3` calculation can overflow if you have `len == UINT_MAX-2`. I mentioned that in the code.

I don't like getting the doc and spec get out of sync

hanno-becker: To be fair, the limitation is for `buflen` only, not `inlen`. 

If you prefer, I can state the spec as strong as possible and give the largest possible bound for `buflen`.

mlkem/rej_uniform.h
hanno-becker: done

mlkem/rej_uniform.c
hanno-becker: Ok, left a comment and removed the bound for `len`

mkannwischer: Thanks for adjusting this!

mkannwischer (approved):
Thanks for your edits! LGTM.

PR #306, opened 2024-11-02
CBMC: Add spec + proof for `polyvec_basemul_acc_montgomery_cached` + `poly_add`

hanno-becker:
[//]: # (SPDX-License-Identifier: CC-BY-4.0)

PR #340, opened 2024-11-04
CBMC: Pass uninitialized pointers in all verification harnesses

hanno-becker:
Previously, when a function under proof would assume a pointer to a structure, the verification harness would construct an instance of that structure on the stack and pass its address.

This approach is unsound, because it adds to the contractual assumptions of the function the specifics of the function invocation in the test harness. For example, a missing bounds constraint in the spec might go undetected just because the stack-allocated variable happens to satisfy it.

The most robust way to deal with this, so it seems, is to minimize the harnesses to merely pass uninitialized variables of the right type to the function under proof. In particular, where a pointer is expected, pass an uninitialized pointer, rather than the address of a stack allocated structure.

Also, remove redundant `x != NULL` preconditions when `IS_FRESH(x,...)`
is assumed. This is not only redundant, but also error prone because of
https://github.com/diffblue/cbmc/issues/8492.

This commit implements this change for all harnesses implemented so far.

### DIFF
--- a/cbmc/proofs/basemul_cached/basemul_cached_harness.c
+++ b/cbmc/proofs/basemul_cached/basemul_cached_harness.c
@@ -23,7 +23,7 @@
  *
  */
 void harness(void) {
-  int16_t a[2], b[2], r[2], b_cached;
+  int16_t *a, *b, *r, b_cached;
 
   basemul_cached(r, a, b, b_cached);
 }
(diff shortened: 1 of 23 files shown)


mkannwischer (approved):
Thanks @hanno-becker. This looks good to me. I double checked that you have not missed any instances. Please make sure to adjust #306 with the same changes before you merge it.

PR #342, opened 2024-11-05
CI: Run CBMC proofs for MLKEM=2,3,4 on separate EC2 instances

hanno-becker:
By now, the CBMC proofs take more than >40min. This commit
- Spawns separate instances for MLKEM_K=2,3,4 CBMC proofs
- Uses -j8 to enable multicore checking

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,7 +233,7 @@ jobs:
       lint: false
       verbose: true
     secrets: inherit
-  cbmc:
+  cbmc_k2:
     needs: quickcheck
     permissions:
       contents: 'read'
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -252,5 +252,47 @@ jobs:
       functest: true
       kattest: false
       nistkattest: false
-      cbmc: true
+      cbmc: 2
```
mkannwischer, line 255: This is rather ugly. Can you split this into a boolean and an optional parameter named `cbmc_mlkem_k` or something like that?

mkannwischer (changes requested):

.github/workflows/ci.yml
hanno-becker: I thought we had a limit of 10 parameters for the workflow, but apparently it's already beyond that, so nevermind. I added a separate parameter.

mkannwischer (approved):
This cuts down CI time from 43 minutes to 25 minutes. 
Puzzling that the K=3 proofs (25 mins) take much longeer than the K=2 (6 mins) and K=4 (11 mins) proofs.

PR #347, opened 2024-11-06
Fix tests all not working if tests not in PATH

### LINKED ISSUE #251: `tests all` does not work if `tests` not in PATH
If the parent directory of `tests` is not in PATH, `tests all` will fail trying to spawn parallel child processes for `tests {func, kat, nistkat}`.

potsrevennil:
originally, `tests all` called `tests func/kat/nistkat` internally, since we assumed the tests script is executed in nix shell.

Calling the `func/kat/nistkat` functions internally instead, could avoid the issue if tests script is not executed in nix shell, which means if `tests` is not in PATH.

However, the `func/kat/nistkat` functions are transferred to click command instead, therefore it's not directly callable. The easiest way is to simply factor out the underlying logic to a separate functions (e.g. `_func/_kat/_nistkat`) and call them instead.

- Fix #251

### DIFF
--- a/scripts/tests
+++ b/scripts/tests
@@ -3,6 +3,7 @@
 
 import platform
 import sys
+import io
 import os
 import subprocess
 import logging
@@ -13,6 +14,8 @@ import asyncio
 from functools import reduce
 from enum import IntEnum
 from typing import Optional, Callable, TypedDict
+from contextlib import redirect_stdout, redirect_stderr
+import copy
 
 
 def config_logger(verbose):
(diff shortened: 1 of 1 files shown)


scripts/tests
```diff
@@ -365,6 +368,58 @@ class State(object):
         return results
 
 
+"""
+Underlying functional tests
+
+"""
+
+
+def _func(state: State):
```
hanno-becker, line 377: It seems that `State` is a misnomer? An instance of `State` represents a particular build configuration, and offers commands to build and run tests under that configuration?

Could you use this PR to expand on the documentation please and rename `State` (or explain why this is the best choice of name)?

potsrevennil: The naming was a mistake like you mentioned... 😢 
I actually had it fixed in other commit, and planned to open a separate pr for this.
Are you ok to have this PR to focus on #251 first ?

Issue #349, opened 2024-11-06
Bump bitwuzla from 0.4 to 0.5 or 0.6


hanno-becker:
Currently, our nix setup builds bitwuzla 0.4.0, but there's already version 0.6.0 available. Also, https://mynixos.com/nixpkgs/package/bitwuzla suggests that we can at least bump to 0.5 without problems.

PR #350, opened 2024-11-06
Update bitwuzla -> 0.6.0 &  Z3 -> 4.13.3

potsrevennil:
- #349 (bitwuzla is bumped to 0.6.0)
- #351 
- flake lock is updated, we can update gcc to 14.1.0 whenever we want now
- nix flake is slightly refactored by using nix flake packages, it'd be easier to test some nix derivations
  - ex: `nix build .#cbmc` would built cbmc dependencies in to local directory `result`. We could thus check the binaries in `result/bin` to see if everything works out as expected

### DIFF
--- a/cbmc/default.nix
+++ b/cbmc/default.nix
@@ -2,7 +2,6 @@
 { cbmc
 , fetchFromGitHub
 , callPackage
-, z3_4_12
 , bitwuzla
 , ninja
 }:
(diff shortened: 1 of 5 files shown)


hanno-becker:
@potsrevennil Can we at least bump to 0.5?

hanno-becker (approved):
Thank you @potsrevennil, it's good to be up to date, though sadly it makes the CI slower than before.

PR #348, opened 2024-11-06
CBMC: Add spec+proof for `gen_matrix_entry()`

hanno-becker:
This was pretty hard. I did not manage to verify `gen_matrix_entry()` in acceptable time as it stood, because the reasoning about the pointer increment in the coefficient buffer somehow tripped up the SAT solvers. Instead, I had to modify the signature of `rej_uniform()` to always take the polynomial base -- luckily, this arguably makes the code cleaner anyway. With that signature change in place, things are quite fast. See commit messages for some detail.

Resolves #326

PR #353, opened 2024-11-06
Speed up CBMC proof of polyvec_tobytes()

hanno-becker:
This was running very slowly. This commit reduces the runtime from minutes to seconds by switching to smt2 and using a loop invariant.


hanno-becker:
Integrated into #348

PR #354, opened 2024-11-06
FIPS202: Add _ctx_release() functions to support 3rd party FIPS202 implementations using dynamically allocated state

mkannwischer:
Some consumers may want to provide their own Kecak implementations instead of using the ones provided by us. Such implementations may be using Keccak implementations using dynamic memory allocations, e.g., the Keccak from OpenSSL does that.
In that case, we need to explicitly free the state once it is no longer needed.

This PR adds corresponding functions called _ctx_release() and calls them in the appropriate places.
Since all of our Keccak implementations use a statically allocated state, the xxx_ctx_release() is implemented as a no-op here.

Naming of functions follows PQClean and liboqs.


mkannwischer:
I'll need to look at the `gen_matrix_entry` first - that one breaks with this change.

hanno-becker (approved):

PR #358, opened 2024-11-06
Proofs of poly_frommsg

### LINKED ISSUE #336: CBMC: State + prove spec for poly_frommsg()

rod-chapman:
- Fixes #336 

Adds proof of poly_frommsg() function. In particular:

1. Strengthen post-condition of cmov_int16() to use OBJECT_UPTO rather than OBJECT_WHOLE to allow a caller to modify one element of a larger array.
2. Add contracts to specification and body of poly_frommsg()
3. Add CBMC proof artefacts for this function.

### DIFF
--- a/mlkem/poly.c
+++ b/mlkem/poly.c
@@ -279,31 +279,18 @@ void poly_frommsg(poly *r, const uint8_t msg[MLKEM_INDCPA_MSGBYTES]) {
 #error "MLKEM_INDCPA_MSGBYTES must be equal to MLKEM_N/8 bytes!"
 #endif
 
-  const int num_blocks = MLKEM_N / 8;
-
-  for (int i = 0; i < num_blocks; i++)
-      // clang-format off
+  for (int i = 0; i < MLKEM_N / 8; i++)  // clang-format off
     ASSIGNS(i, OBJECT_WHOLE(r))
-    INVARIANT(i >= 0)
-    INVARIANT(i <= num_blocks)
-    INVARIANT(num_blocks == 32)
+    INVARIANT(i >= 0 && i <= MLKEM_N / 8)
     INVARIANT(ARRAY_IN_BOUNDS(int, k, 0, (8 * i - 1), r->coeffs, 0, (MLKEM_Q - 1)))
-    // clang-format on
-    {
-      for (int j = 0; j < 8; j++)
-          // clang-format off
+    {      // clang-format on
+      for (int j = 0; j < 8; j++)  // clang-format off
         ASSIGNS(j, OBJECT_WHOLE(r))
-        INVARIANT(i >= 0 && i <  num_blocks)
-        INVARIANT(j >= 0 && j <= 8)
+        INVARIANT(i >= 0 && i <  MLKEM_N / 8 && j >= 0 && j <= 8)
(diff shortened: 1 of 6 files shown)


mlkem/poly.c
```diff
@@ -274,28 +274,38 @@ void poly_frombytes(poly *r, const uint8_t a[MLKEM_POLYBYTES]) {
...
-    }
-  }
+  const int num_blocks = MLKEM_N / 8;
+
+  for (int i = 0; i < num_blocks; i++)
+      // clang-format off
+    ASSIGNS(i, OBJECT_WHOLE(r))
+    INVARIANT(i >= 0)
+    INVARIANT(i <= num_blocks)
+    INVARIANT(num_blocks == 32)
```
hanno-becker, line 289: This should not be needed?

mlkem/poly.c
```diff
@@ -274,28 +274,38 @@ void poly_frombytes(poly *r, const uint8_t a[MLKEM_POLYBYTES]) {
...
+        ASSIGNS(j, OBJECT_WHOLE(r))
+        INVARIANT(i >= 0 && i <  num_blocks)
+        INVARIANT(j >= 0 && j <= 8)
+        INVARIANT(ARRAY_IN_BOUNDS(int, k, 0, (8 * i + j - 1), r->coeffs, 0, (MLKEM_Q - 1)))
+        // clang-format on
+        {
+          r->coeffs[8 * i + j] = 0;
+          cmov_int16(&r->coeffs[8 * i + j], HALF_Q, (msg[i] >> j) & 1);
 
+          // Check that the value just assigned must be either HALF_Q or 0
```
hanno-becker, line 304: This can be removed? It is a verbatim restatement of the post condition of `cmov_int16`?

hanno-becker (changes requested):
LGTM, but can you trim down on unneeded invariants and assertions please?

mlkem/poly.c
rod-chapman: Removing that num_blocks == 32 invariants changes the time to prove this function (1 CPU on my laptop) from 3 min 47 seconds to 5 min 20 seconds. Worth removing or not?

mlkem/poly.c
rod-chapman: Yes... removing it does not seem to change proof performance, so it can go.

mlkem/poly.c
hanno-becker: Can we not use write `MLKEM_N / 8` everywhere? That's how it's done in the other loops.

hanno-becker: I took the liberty to change that.

mlkem/poly.c
hanno-becker: I took the liberty to change that.

hanno-becker (approved):

mlkem/poly.c
rod-chapman: OK. I personally find the repetition dubious, but happy to oblige. I guess the repetition is only risky if there's a chance the value might change, which is in this case is zero...

mlkem/poly.c
rod-chapman: OK

PR #370, opened 2024-11-09
CBMC: Proof of type safety and memory safety for toplevel MLKEM KeyGen (modulo NTT+SHA3)

hanno-becker:
This PR brings together the work of the past weeks into a top-level type+memory safety proof of `crypto_kem_keypair_derand()` -- with only NTT and SHA3 axiomatized, but all other lower level functions specified and proved.

A few changes had to be made to the existing lower level proofs -- see the individual commit messages for details.

### DIFF
--- a/mlkem/indcpa.c
+++ b/mlkem/indcpa.c
@@ -421,10 +421,15 @@ void indcpa_keypair_derand(uint8_t pk[MLKEM_INDCPA_PUBLICKEYBYTES],
   poly_getnoise_eta1_4x(skpv.vec + 0, skpv.vec + 1, e.vec + 0, e.vec + 1,
                         noiseseed, 0, 1, 2, 3);
 #elif MLKEM_K == 3
-  poly_getnoise_eta1_4x(skpv.vec + 0, skpv.vec + 1, skpv.vec + 2, e.vec + 0,
-                        noiseseed, 0, 1, 2, 3);
-  poly_getnoise_eta1_4x(e.vec + 1, e.vec + 2, pkpv.vec + 0, pkpv.vec + 1,
-                        noiseseed, 4, 5, 6, 7);
+  // Only the first three output buffers are needed.
+  // The laster parameter is a dummy that's overwritten later.
+  poly_getnoise_eta1_4x(skpv.vec + 0, skpv.vec + 1, skpv.vec + 2,
+                        pkpv.vec + 0 /* irrelevant */, noiseseed, 0, 1, 2,
+                        0xFF /* irrelevant */);
+  // Same here
+  poly_getnoise_eta1_4x(e.vec + 0, e.vec + 1, e.vec + 2,
+                        pkpv.vec + 0 /* irrelevant */, noiseseed, 3, 4, 5,
+                        0xFF /* irrelevant */);
 #elif MLKEM_K == 4
   poly_getnoise_eta1_4x(skpv.vec + 0, skpv.vec + 1, skpv.vec + 2, skpv.vec + 3,
                         noiseseed, 0, 1, 2, 3);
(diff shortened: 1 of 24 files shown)


mlkem/polyvec.h
```diff
@@ -216,4 +216,24 @@ ENSURES(FORALL(int, j, 0, MLKEM_K - 1,
 ASSIGNS(OBJECT_WHOLE(r));
 // clang-format on
 
+// clang-format off
```
mkannwischer, line 227: move this clang format please

mlkem/indcpa.c
```diff
@@ -352,6 +352,26 @@ void gen_matrix(polyvec *a, const uint8_t seed[MLKEM_SYMBYTES],
 #endif /* MLKEM_USE_NATIVE_NTT_CUSTOM_ORDER */
 }
 
+// clang-format off
+STATIC_TESTABLE
+void matvec_mul(polyvec *out, polyvec const *a, polyvec const *v,
+    polyvec_mulcache const *vc)
```
mkannwischer, line 358: 1) change to `const polyvec` rather than `polyvec const`?

2) Can this be a `const polyvec a[MLKEM_K]`?

mkannwischer: Please add a commen documenting the bounds.

mlkem/poly.h
```diff
@@ -570,6 +570,7 @@ ASSIGNS(OBJECT_WHOLE(r))
 ENSURES(ARRAY_IN_BOUNDS(int, k, 0, MLKEM_N - 1, r->coeffs, -3 * HALF_Q + 1, 3 * HALF_Q - 1));
 // clang-format on
 
+// clang-format off
```
mkannwischer, line 630: Can you explain why you moved the `//clang-format off`? Do you want to change it everywhere? 
If I see correctly then this basically means turning off clang-format for all headers files entirely.

mlkem/polyvec.h
```diff
@@ -88,8 +88,16 @@ ENSURES(FORALL(int, k0, 0, MLKEM_K - 1,
...
 #define polyvec_ntt MLKEM_NAMESPACE(polyvec_ntt)
-void polyvec_ntt(polyvec *r);
+void polyvec_ntt(polyvec *r)
+  REQUIRES(IS_FRESH(r, sizeof(polyvec)))
+  ASSIGNS(OBJECT_WHOLE(r))
+  ENSURES(FORALL(int, j, 0, MLKEM_K - 1,
+    ARRAY_IN_BOUNDS(int, k, 0, MLKEM_N - 1,
+                    r->vec[j].coeffs, -NTT_BOUND + 1, NTT_BOUND - 1)));
```
mkannwischer, line 98: maybe we stick with `-(NTT_BOUND - 1)` and `(NTT_BOUND  - 1)` as we have in all other places?

mlkem/indcpa.c
```diff
@@ -408,10 +408,10 @@ void indcpa_keypair_derand(uint8_t pk[MLKEM_INDCPA_PUBLICKEYBYTES],
...
                         noiseseed, 0, 1, 2, 3);
 #elif MLKEM_K == 3
-  poly_getnoise_eta1_4x(skpv.vec + 0, skpv.vec + 1, skpv.vec + 2, e.vec + 0,
-                        noiseseed, 0, 1, 2, 3);
-  poly_getnoise_eta1_4x(e.vec + 1, e.vec + 2, pkpv.vec + 0, pkpv.vec + 1,
-                        noiseseed, 4, 5, 6, 7);
+  poly_getnoise_eta1_4x(skpv.vec + 0, skpv.vec + 1, skpv.vec + 2, pkpv.vec + 0,
+                        noiseseed, 0, 1, 2, 6);
+  poly_getnoise_eta1_4x(e.vec + 0, e.vec + 1, e.vec + 2, pkpv.vec + 1,
+                        noiseseed, 3, 4, 5, 7);
```
mkannwischer, line 414: This is very confusing and I'm not a fan of changing the code like this. 

Is there really no way to make this simpler? 
Intuitively, I do not understand why `IS_FRESH` doesn't work for two polynomials belonging to the same polyvec - they are still disjoint chunks of memory.

mkannwischer (changes requested):
Nice @hanno-becker. Congratulations to the first proof of a high-level function! Looks like things are coming together. 

Some suggestions and questions.

mlkem/indcpa.c
hanno-becker: That's unfortunately not how it works. `IS_FRESH` generates the obligation that the pointers belong to different objects.

mlkem/indcpa.c
hanno-becker: 1. OK 2. OK 3. done

mlkem/poly.h
hanno-becker: I personally find this easier to read, but I agree that we ultimately disable clang format this way. Let's not try to solve this in this PR.

mlkem/polyvec.h
hanno-becker: I don't want to go over the code and check where we do it one way or the other -- it doesn't really matter.

PR #384, opened 2024-11-12
CBMC Proof of indcpa_dec()

rod-chapman:
Adds contracts and proof artefacts for

unpack_ciphertext() (local function within indcpa.c)
and
indcpa_dec()

All tests OK
All proofs OK
lint OK


hanno-becker (changes requested):
LGTM, but @rod-chapman can you fix the commit messages? The headline can only 72 characters long.

hanno-becker (approved):

Issue #391, opened 2024-11-13
CI timeouts due to EC2 runners sometimes not picking up the CI jobs


mkannwischer:
See for example: https://github.com/pq-code-package/mlkem-native/actions/runs/11810341034/job/32902191292

![image](https://github.com/user-attachments/assets/af6c149f-2274-4bad-ba79-3cd80179455e)

The EC2 instances seem to be started correctly, but then they do not get registered as action runners correctly. 
This may be some race condition in https://github.com/machulav/ec2-github-runner.

Unfortunately this seems to be happening a lot recently.

mkannwischer:
This is no resolved and we should probably revert #392. I don't have time to look into it still this week.

mkannwischer:
So what is happening is that Github runners are sometimes not successfully de-registered from Github (fixed by https://github.com/pq-code-package/mlkem-native/pull/414 hopefully). Last week I deleted close to 200 inactive runners from there. 

Since then the number of times the issue in the screenshot above occurred decreased dramatically. I remember it only happening 2 times since Saturday, while it was happening on every CI run before I deleted those.

Given that it is not complely gone, there must be another problem.

mkannwischer:
I've not seen this happen for a while now, but I'm not 100% sure it is fixed. 
Closing for now, but maybe we have to re-open it.

PR #392, opened 2024-11-13
Fix EC2 failures when Github runner name is already taken

mkannwischer:
I suspect that our CI failures may be due to the runner registration failing due to the name already being used. By default, it is using the hostname as the Github runner name which defaults to the private IP address on EC2 instances. When instances get terminated, it does not delete the runners in Github which means we now have a lot of offline runners listed.

This may be explaining why our CI sometimes timesouts due to the runner not being registered.

I patched ec2-github-runner to add the `--replace` flag when registering the runner.

This may be fixing #391 -


mkannwischer:
Damn - it still happens

hanno-becker (approved):
Thank you @mkannwischer for investigating. Let's merge this and monitor if it helps.

PR #398, opened 2024-11-13
Fix tests bench --opt=all and tests all --no-acvp

potsrevennil:

### DIFF
--- a/scripts/lib/mlkem_test.py
+++ b/scripts/lib/mlkem_test.py
@@ -546,7 +546,12 @@ class Tests:
             exit(1)
 
     def _run_bench(
-        self, t: Test_Implementations, opt: bool, run_as_root: bool, exec_wrapper: str
+        self,
+        t: Test_Implementations,
+        opt: bool,
+        run_as_root: bool,
+        exec_wrapper: str,
+        mac_taskpolicy,
     ) -> TypedDict:
         cmd_prefix = []
         if run_as_root:
@@ -554,6 +559,13 @@ class Tests:
                 f"Running {bin} as root -- you may need to enter your root password.",
             )
             cmd_prefix.append("sudo")
+
+        if mac_taskpolicy:
+            if exec_wrapper:
+                logging.error(f"cannot set both --mac-taskpolicy and --exec-wrapper")
+                sys.exit(1)
+            else:
+                exec_wrapper = f"taskpolicy -c {mac_taskpolicy}"
 
         if exec_wrapper:
             logging.info(f"Running with customized wrapper.")
@@ -578,13 +590,6 @@ class Tests:
         else:
             t = self._bench_components
             output = False
-
-        if mac_taskpolicy:
-            if exec_wrapper:
-                logging.error(f"cannot set both --mac-taskpolicy and --exec-wrapper")
-                sys.exit(1)
-            else:
-                exec_wrapper = f"taskpolicy -c {mac_taskpolicy}"
 
         # NOTE: We haven't yet decided how to output both opt/no-opt benchmark results
         if self.opt.lower() == "all":
--- a/scripts/lib/mlkem_test.py
+++ b/scripts/lib/mlkem_test.py
@@ -260,6 +260,25 @@ class Test_Implementations:
         else:
             return results
 
+    def test(
+        self,
+        opt: bool,
+        compile: bool,
+        run: bool,
+        extra_make_envs={},
+        extra_make_args=[],
+        actual_proc: Callable[[bytes], str] = None,
+        expect_proc: Callable[[SCHEME, str], tuple[bool, str]] = None,
+        cmd_prefix: [str] = [],
+        extra_args: [str] = [],
+    ) -> TypedDict:
+        if compile:
+            self.compile(opt, extra_make_envs, extra_make_args)
+        if run:
+            return self.run_schemes(
+                opt, actual_proc, expect_proc, cmd_prefix, extra_args
+            )
+
 
 """
 Underlying functional tests
@@ -283,12 +302,11 @@ class Tests:
         self._bench_components = Test_Implementations(
             TEST_TYPES.BENCH_COMPONENTS, copts
         )
-
         self.compile_mode = copts.compile_mode()
         self.compile = opts.compile
         self.run = opts.run
 
-    def _run_func(self, opt: bool):
+    def _test_func(self, opt: bool, compile: bool, run: bool) -> TypedDict:
         """Underlying function for functional test"""
 
         def expect(scheme: SCHEME, actual: str) -> tuple[bool, str]:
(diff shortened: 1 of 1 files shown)


mkannwischer (changes requested):
I don't have time to review another big refactor of the test scripts - can you please cut this down to just what is necessary to make it work? For example, just the first commit would be fine.

hanno-becker (changes requested):
Thank you @potsrevennil. As discussed offline, please don't mix bugfixes and refactoring anymore in the future. 

For the review of the bugfix, could you please describe in the commit message what the bug was in the first place (What did you do? What was the expected behaviour? What was the actual behaviour?) and brieflt how it is fixed?

hanno-becker (approved):

hanno-becker (changes requested):
I tried, but the 2nd commit message is not enough for me to understand what is going on. I will open a separate PR with just the bugfix.

potsrevennil:
@hanno-becker Done

hanno-becker:
#407 cherry-picks first commit.

potsrevennil:
I tried to make the git diff looked friendlier, hope that it'd be clear that this was just a simple rewrite, which is just merging the conditional `compile` and `run_schemes` local functions (ex: `_func` in `func`) into a general function `Test_Implementation.test` to avoid similar mistake fixed in the first commit

hanno-becker:
@potsrevennil I seem to remember there a different diff options locally at least... anyone in particular that will make it more obvious what is happening?

scripts/lib/mlkem_test.py
```diff
@@ -260,6 +260,25 @@ def run_schemes(
...
 
+    def test(
+        self,
+        opt: bool,
+        compile: bool,
+        run: bool,
+        extra_make_envs={},
+        extra_make_args=[],
+        actual_proc: Callable[[bytes], str] = None,
+        expect_proc: Callable[[SCHEME, str], tuple[bool, str]] = None,
```
hanno-becker, line 271: Can you document, here and in the other functions sharing this part of the signature, what the arguments mean? Esp. `actual_proc` and `expected_proc` are not self-explanatory.

scripts/lib/mlkem_test.py
```diff
@@ -260,6 +260,25 @@ def run_schemes(
...
+        cmd_prefix: [str] = [],
+        extra_args: [str] = [],
+    ) -> TypedDict:
```
hanno-becker, line 274: Can you add a comment saying what's in the returned dictionary?

scripts/lib/mlkem_test.py
```diff
@@ -272,23 +291,16 @@ def __init__(self, opts: Options):
         copts = CompileOptions(
             opts.cross_prefix, opts.cflags, opts.arch_flags, opts.auto, opts.verbose
         )
+
+        self._f = lambda t: Test_Implementations(t, copts)
```
hanno-becker, line 295: Is this change merely meant as a unification of L278-285? The behaviour is different because `_f` creates an object upon each invocation, while L278-L285 create objects only once.

scripts/lib/mlkem_test.py
```diff
@@ -307,32 +319,27 @@ def expect(scheme: SCHEME, actual: str) -> tuple[bool, str]:
                 f"Failed, expecting {expect}, but getting {actual}" if fail else "",
             )
 
-        return self._func.run_schemes(
+        return self._f(TEST_TYPES.MLKEM).test(
```
hanno-becker, line 322: From the commit message I was expecting to see pairs of compile+run being replaced by test, but here we _only_ call `run_schemes` before, while the new call to `test()` would additionally compile. Is that deliberate?

scripts/lib/mlkem_test.py
```diff
@@ -526,29 +525,32 @@ def init_results() -> TypedDict:
 
         return fail
 
+    def _acvp(self, opt: bool, compmile: bool, run: bool, acvp_dir: str) -> bool:
```
hanno-becker, line 528: Why is this handled differently from say `_func`? In `_func`, you just call `self._f(...).test()`, which bundles `compile` and `run` -- while here you still call both separately. Note also that L530 and L532 do not operate on the same test object -- they call `_f` twice.

hanno-becker (changes requested):
I don't think this change makes the test script easier to understand than before. On the contrary, we now have one more indirection, namely a function `self._f()` returning a test object, and as indicated in the comments, we're at times compiling and running in separate test objects which morally ought to be the same (and were, previously).

@potsrevennil Could you think about how we can remove indirections rather than adding more?

scripts/lib/mlkem_test.py
potsrevennil: your right, this is not ideal, and not necessary

scripts/lib/mlkem_test.py
```diff
@@ -272,23 +291,16 @@ def __init__(self, opts: Options):
...
-        self._bench_components = Test_Implementations(
-            TEST_TYPES.BENCH_COMPONENTS, copts
-        )
-
         self.compile_mode = copts.compile_mode()
         self.compile = opts.compile
         self.run = opts.run
 
-    def _run_func(self, opt: bool):
+    def _func(self, opt: bool, compile: bool, run: bool) -> TypedDict:
```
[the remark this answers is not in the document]
potsrevennil: We could, but in the functions like self.func, there's logger configuration, 
which means if I called `self.func` instead in `all`, then the logger would be configured multiple times.
and also in `self.func` if `--opt=all` it'd compile/run both the implementations, but in `all` we need to first compile/run one implementation then do the other due to the current build system restriction, therefore if without the indirection, there might required more changes in `all`
As we have discussed about build system change in the future (separate opt/non-opt build artifacts), once it's done, we could thus remove this annoying indirection by the way.

scripts/lib/mlkem_test.py
potsrevennil: because the way we run acvp is very different from the other tests, the other tests only run once for each parameter set, but acvp test is run multiple times for each parameter set with different test vectors.
it's too hard and not necessary to reuse the same function `test` for now.

potsrevennil:
@hanno-becker thanks for the comment!
I removed one indirection, I think the indirections should be the same as before now.
And as I commented, once we have a build system that can separate opt/non-opt built artifacts, we could further reduce indriections, to make the tests script more readible.

scripts/lib/mlkem_test.py
potsrevennil: originally, the paris of compile+run is in the local function `_func` in `func`
so the `self._func` is actually a replacement of the local function `_func` in `self.func`

so before the function calls was like
```
func -> local _func -> self._func.compile+ self._run_func -> self._func.run_schemes
all -> self._func.compile + self._run_func -> self._func.run_schemes
```
which is pretty hard to understand

it now became
```
func -> self._func -> self.__func.test -> self.__func.compile + self.__func.run_schemes
all -> self._func -> self.__func.test -> self.__func.compile + self.__func.run_schemes
```

hanno-becker:
@potsrevennil We now have `self.func`, `self._func` and `self.__func`. This really can't be it...

potsrevennil:
@hanno-becker sorry, the naming was pretty bad, I updated it now

mkannwischer:
It seems this is no longer strictly necessary after #407  got merged. 
Let's not do refactoring of the tests anymore unless there is a good reason. 
Closing this for now.

PR #404, opened 2024-11-13
CBMC proof of verify()

rod-chapman:
Adds proof of type-safety AND partial correctness
of the verify() function.

All tests OK
lint OK
All proofs OK, including calling unit crypto_kem_dec()


hanno-becker (approved):
This is great, thank you @rod-chapman

PR #408, opened 2024-11-14
Improve speed of proof of matvec_mul()

rod-chapman:
Adjust modelling of arrays in CBMC for proof of this function, as recommended by CBMC team (See [CBMC Issue 8505](https://github.com/diffblue/cbmc/issues/8505)).

Proof for this function reduces from 70 seconds to about 9 seconds on desktop.

All tests OK
All proofs OK
lint OK


hanno-becker (changes requested):
Those are impressive gains, @rod-chapman. Can you leave a comment in `matvec_mul/Makefile` with a vague description of what is happening, and open an issue so we don't forget to incorporate this into the proof guide once we have a better understanding of those options?

rod-chapman:
Comment added to Makefile

hanno-becker (approved):

PR #416, opened 2024-11-16
CI: Add test for non-standard values of GEN_MATRIX_NBLOCKS

### LINKED ISSUE #186: Add test for non-standard values of GEN_MATRIX_NBLOCKS
During public matrix generation, we first squeeze and rejection sample `GEN_MATRIX_NBLOCKS` blocks, and then one block a time. The logic for this latter stage is very unlikely to be exercised properly for the default value `GEN_MATRIX_NBLOCKS == 3` (that's the point of this value!).

**Task:** Only set `GEN_MATRIX_NBLOCKS` if it's not yet set, and add tests exercising smaller the smaller values `1,2`.

hanno-becker:
GEN_MATRIX_NBLOCKS is a heuristic value estimating the amount of rejection sampling input to sample 256 polynomial coefficients.

While this value is hardcoded to 3 by default, one should be able to set it to any positive value. Moreover, some code-paths are not as easily reached with the default value, and testing a lower value increases the chances of catching bugs in those.

This commit allows the user to specify GEN_MATRIX_NBLOCKS in the CFLAGS, and extends the CI to test the non-standard values 1,2, and 4.

* Fixes #186 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,7 +194,64 @@ jobs:
           nix-shell: ci-linter
           gh_token: ${{ secrets.GITHUB_TOKEN }}
           cross-prefix: "aarch64-unknown-linux-gnu-"
-
+  # The purpose of this job is to test non-default yet valid configurations
+  config_variations:
+    name: Non-standard configurations
+    needs: quickcheck
+    strategy:
+      fail-fast: false
+      matrix:
+        external:
+         - ${{ github.repository_owner != 'pq-code-package' }}
+        target:
+         - runner: pqcp-arm64
+           name: 'ubuntu-latest (aarch64)'
+         - runner: pqcp-x64
+           name: 'ubuntu-latest (x86_64)'
+        exclude:
+          - {external: true,
+             target: {
+               runner: pqcp-arm64,
+               name: 'ubuntu-latest (aarch64)',
+             }}
+          - {external: true,
+             target: {
+               runner: pqcp-x64,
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -194,7 +194,63 @@ jobs:
           nix-shell: ci-linter
           gh_token: ${{ secrets.GITHUB_TOKEN }}
           cross-prefix: "aarch64-unknown-linux-gnu-"
-
+  config_variations:
+    name: Non-standard configurations
```
mkannwischer, line 198: Please add a comment here explaining what's happening in this job?

mkannwischer (approved):
Looks good to me. Please add a comment in the CI config - I won't remember why we need this in a few months.

mlkem/indcpa.c
```diff
@@ -124,8 +124,10 @@ static void unpack_ciphertext(polyvec *b, poly *v,
   poly_decompress(v, c + MLKEM_POLYVECCOMPRESSEDBYTES);
 }
 
+#ifndef GEN_MATRIX_NBLOCKS
```
mkannwischer, line 127: Please namespace this. At least it should start with `MLKEM_`

hanno-becker: Ok

.github/workflows/ci.yml
hanno-becker: Done

hanno-becker: Ok

PR #420, opened 2024-11-17
Makefile: Compile mlkem/* in a single step

hanno-becker:


hanno-becker:
@potsrevennil This should not be merged (yet), but I'm curious about your thoughts on this crude hack. How would you have gone about this? How would you extend it to cover also FIPS202?

PR #423, opened 2024-11-18
Adjust makefile for support -flto in nix shell

potsrevennil:
Currently in nix shell, the build system won't properly linked if `-flto` is set.

This is due to that nix `gcc` comes with lto library, binutils like `ar` doesn't seem to
properly configured regarding this.

The easiest workaround to make `-flto` work is therefore to avoid building
libraries, let `gcc` take over everything.

Namespace checking needed to be updated as well, due to the new build system compiles fips202 artifact to `test/build/mlkemxxx/fips202` now.

### DIFF
--- a/mk/crypto.mk
+++ b/mk/crypto.mk
@@ -1,14 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
-LDLIBS += -L$(LIB_DIR)
-
-LIBDEPS += $(LIB_DIR)/libfips202.a
-LDLIBS += -lfips202
 CPPFLAGS += -Ifips202 -Ifips202/native
 
-FIPS202_SRCS = $(wildcard fips202/*.c)
+SOURCES += $(wildcard fips202/*.c)
 ifeq ($(OPT),1)
-	FIPS202_SRCS += $(wildcard fips202/native/aarch64/*.S) $(wildcard fips202/native/x86_64/xkcp/*.c)
-	CPPFLAGS += -DMLKEM_USE_NATIVE
+	SOURCES += $(wildcard fips202/native/aarch64/*.S) $(wildcard fips202/native/x86_64/xkcp/*.c)
 endif
-
-$(LIB_DIR)/libfips202.a: $(call OBJS, $(FIPS202_SRCS))
--- a/mk/bench.mk
+++ b/mk/bench.mk
@@ -1,5 +1,3 @@
 # SPDX-License-Identifier: Apache-2.0
-LIBDEPS += $(LIB_DIR)/libhal.a
-LDLIBS += -lhal
 CPPFLAGS += -Itest/hal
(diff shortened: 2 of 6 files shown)


mk/crypto.mk
```diff
@@ -1,14 +1,7 @@
...
-
-LIBDEPS += $(LIB_DIR)/libfips202.a
-LDLIBS += -lfips202
 CPPFLAGS += -Ifips202 -Ifips202/native
 
-FIPS202_SRCS = $(wildcard fips202/*.c)
+SOURCES += $(wildcard fips202/*.c)
 ifeq ($(OPT),1)
-	FIPS202_SRCS += $(wildcard fips202/native/aarch64/*.S) $(wildcard fips202/native/x86_64/xkcp/*.c)
-	CPPFLAGS += -DMLKEM_USE_NATIVE
```
hanno-becker, line 11: Why was this removed?

potsrevennil: I simply added the fips202 source code to the `SOURCES` variable, which is then used in `mk/schemes.mk`

And this is merely a quick workaround, I wanted to first see if things worked as expected, then cleanup afterwards

.github/actions/functest/action.yml
```diff
@@ -91,10 +91,10 @@ runs:
         shell: ${{ env.SHELL }}
         run: |
           tests all  --cross-prefix="${{ env._CROSS_PREFIX }}" --cflags="${{ inputs.cflags }}" --opt=${{ inputs.opt }} --${{ env.FUNC }} --${{ env.KAT }} --${{ env.NISTKAT }} --${{ env.ACVP }} -v
-      - name: Check namespacing ${{ inputs.mode }} ${{ inputs.opt }} tests (${{ env.FUNC }}, ${{ env.KAT }}, ${{ env.NISTKAT }})
-        shell: ${{ env.SHELL }}
-        run: |
-          check-namespace
+      # - name: Check namespacing ${{ inputs.mode }} ${{ inputs.opt }} tests (${{ env.FUNC }}, ${{ env.KAT }}, ${{ env.NISTKAT }})
```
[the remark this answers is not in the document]
mkannwischer: the namespacing test runs on the object files. With the changes proposed here, there are no object files anymore, so there is nothing to check. 
The `assert checked > 0` is there for exactly that reason: To fail in case it is actually a no-op. 
We'll have to reimplement the namespacing check.

mk/crypto.mk
hanno-becker: I mean why was the change to CPPFLAGS removed?

hanno-becker (approved):

mkannwischer (approved):
I cleaned up a few more make variables that are no longer needed now. Looks good to me now. Thanks @potsrevennil

PR #424, opened 2024-11-18
Benchmark Regression alert

potsrevennil:


.github/actions/bench/action.yml
```diff
@@ -92,3 +92,6 @@ runs:
         output-file-path: output.json
         github-token: ${{ inputs.gh_token }}
         auto-push: true
+        comment-on-alert: true
+        summary-always: true
```
hanno-becker, line 96: Hm wait, but when would this happen? We want it to happen in the PR, not after merge. And `store-results` will only be set after merge.

hanno-becker (dismissed):
It's not yet clear to me that this will trigger at PR-time not only post-merge. @potsrevennil can you explain?

.github/actions/bench/action.yml
potsrevennil: I mislooked this part, so according to the documentation, `comment-on-alert` seemed to only work after merge indeed
https://github.com/benchmark-action/github-action-benchmark?tab=readme-ov-file#future-work

But there's still be warning and summary in CI workflow
https://github.com/potsrevennil/mlkem-c-aarch64/actions/runs/11890509403

However I encounter permission error in my own testing prs, I'll need to figure this out first, turning this back to draft for now

mkannwischer:
Okay, we have to resolve the variance first before proceeding with the alerts. 
I opened an issue for that: https://github.com/pq-code-package/mlkem-native/issues/429

Turning this into a draft PR for now.

.github/actions/bench/action.yml
```diff
@@ -84,11 +87,20 @@ runs:
       run: |
         check-namespace
     - name: Store benchmark result
-      if: ${{ inputs.store_results == 'true' }}
       uses: benchmark-action/github-action-benchmark@v1
```
mkannwischer, line 90: can you pin this to a specific version please.

potsrevennil: I opened a separate PR to ping benchamrk and nix cache action now

mkannwischer (approved):
LGTM - thanks @potsrevennil ! I tested this 3 times now and it's stable.

PR #440, opened 2024-11-21
Add copyright header to all files and add linting 

mkannwischer:


hanno-becker (approved):
LGTM, tested locally. CBMC slowdown is unrelated and appears to be proof brittleness which I fixed by using SMT lib arrays.

PR #441, opened 2024-11-21
CI: restirct EC2 benchmarking to one instance

mkannwischer:
Experimenting with restricting the benchmarking jobs to only launch a single instance so we can potentially switch to beefier boxes.


mkannwischer:
What I am trying to achieve here does not seem possible as per https://github.com/orgs/community/discussions/41518. 
Too bad.

PR #446, opened 2024-11-22
CI: Benchmark opt and non-opt on the same EC2 instance

hanno-becker:
Previously, we were spinning up separate instances for benchmarking
optimized and non-optimized code. This is inefficient and incurs
unnecessary cost for very large instances (e.g. metal).

This commit changes the benchmarking CI to benchmark optimized
and non-optimized code on the same EC2 instance.


mkannwischer (approved):
LGTM. Thanks. 
That should save some money :)

PR #448, opened 2024-11-22
Properly set $CC in nix shell for different architecture

potsrevennil:
This PR properly set $CC in Nix shells when using multiple GCC versions (the last GCC in the package list overwrites $CC), thus avoiding cumbersome nix setup.

1. Properly Set $CC:
- Overridden in stdenv to ensure consistent behavior across platforms.
- Explicitly set via shellHook on Darwin, where GCC is not installed via Nix.

2. Refactor GCC Configuration:
- Introduced native-gcc to streamline platform-specific logic (null on Darwin, wraps pkgs elsewhere).
- Simplified code by merging default_gcc into core and removing redundant checks.

These changes eliminate the need for manual workarounds and simplify working with GCC toolchains in Nix, and make it easier to integrate additional toolchains:
- see #449 

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -19,38 +19,34 @@
       systems = [ "x86_64-linux" "aarch64-linux" "aarch64-darwin" "x86_64-darwin" ];
       perSystem = { config, pkgs, system, inputs', ... }:
         let
-          default_gcc = { cross ? true }:
+          glibc-join = p: p.buildPackages.symlinkJoin {
+            name = "glibc-join";
+            paths = [ p.glibc p.glibc.static ];
+          };
+
+          wrap-gcc = p: p.buildPackages.wrapCCWith {
+            cc = p.buildPackages.gcc13.cc;
+            bintools = p.buildPackages.wrapBintoolsWith {
+              bintools = p.buildPackages.binutils-unwrapped;
+              libc = glibc-join p;
+            };
+          };
+
+          native-gcc =
+            if pkgs.stdenv.isDarwin
+            then null
+            else wrap-gcc pkgs;
+
+          # cross is for determining whether to install the cross toolchain or not
+          core = { cross ? true }:
             let
-              glibc-join = p: p.buildPackages.symlinkJoin {
-                name = "glibc-join";
-                paths = [ p.glibc p.glibc.static ];
-              };
-
-              wrap-gcc = p: p.buildPackages.wrapCCWith {
-                cc = p.buildPackages.gcc13.cc;
-                bintools = p.buildPackages.wrapBintoolsWith {
-                  bintools = p.buildPackages.binutils-unwrapped;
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -19,38 +19,34 @@
...
+          core = { cross ? true }:
+            let
               x86_64-gcc = wrap-gcc pkgs.pkgsCross.gnu64;
               aarch64-gcc = wrap-gcc pkgs.pkgsCross.aarch64-multiplatform;
             in
-            if pkgs.stdenv.isDarwin
-            then [ ]
-            else if cross
-            then if pkgs.stdenv.isAarch64
+            if (cross && !pkgs.stdenv.isDarwin)
```
hanno-becker: This is because we only support cross compilation on non-Mac AArch64, right? We don't support Mac cross compilation for x86, nor AArch64 cross compilation on x86.

If that's true, can you add a comment to this effect?

Can you also comment on why in the cross case we list two compilers, and whether the order matters?

flake.nix
```diff
@@ -56,8 +54,13 @@
             mkShell (attrs // {
               shellHook = ''
                 export PATH=$PWD/scripts:$PWD/scripts/ci:$PATH
+              '' + pkgs.lib.optionalString pkgs.stdenv.isDarwin ''
+                export CC=gcc
```
hanno-becker, line 58: Should we not be more specific here? How do I know if `gcc` resolves to `x86_64-gcc` or `aarch64-gcc` in the cross case?

potsrevennil: the `export CC=gcc` only applied to darwin, and in darwin we don't have the cross compilation toolchains, so I think it's fine to do it here, although this is not ideal

hanno-becker: Ack

hanno-becker (changes requested):
Thanks @potsrevennil, looks good to me, and works locally on my M1. Also, thank you for the helpful commit messages.

Before merging, can you please also add a few comments in the flake itself? The line
```
pkgs.lib.optionals (cross && !pkgs.stdenv.isDarwin) [ x86_64-gcc aarch64-gcc ]
```
remains rather mysterious.

flake.nix
hanno-becker: That's understood and fine. Let's add a comment and merge.

potsrevennil: In the cross case, although listing the one that is not the native toolchain should be sufficient, but that would require adding some platform checks. I personally think that's ugly :)

And nix should evaluate correctly that one of the toolchains in the list is exactly the same as the one `native-gcc` specified, thus not introducing redundant native toolchain.

potsrevennil: @hanno-becker done

hanno-becker (approved):
Thanks for adding the documentation @potsrevennil

PR #449, opened 2024-11-22
Add riscv64 toolchain support

potsrevennil:
- ~Based on #448~ 
- Add riscv64 toolchain in nix shell
- Add riscv64 qemu support in the tests script
- Fixed an issue where test scripts could incorrectly execute binaries using a custom exec wrapper in QEMU when a cross prefix differs from the native toolchain and a custom exec wrapper is provided.
- With the fix, we could now also run `tests func/kat/nistkat/all` with a provided custom wrapper

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -42,6 +42,7 @@
             let
               x86_64-gcc = wrap-gcc pkgs.pkgsCross.gnu64;
               aarch64-gcc = wrap-gcc pkgs.pkgsCross.aarch64-multiplatform;
+              riscv64-gcc = wrap-gcc pkgs.pkgsCross.riscv64;
             in
             # NOTE:
               # - native toolchain should be equipped in the shell via `mkShellWithCC` (see `mkShell`)
@@ -53,6 +54,7 @@
             pkgs.lib.optionals (cross && !pkgs.stdenv.isDarwin) [
               (pkgs.lib.optional (! pkgs.stdenv.hostPlatform.isx86_64) x86_64-gcc)
               (pkgs.lib.optional (! pkgs.stdenv.hostPlatform.isAarch64) aarch64-gcc)
+              (pkgs.lib.optional (! pkgs.stdenv.hostPlatform.isRiscV64) riscv64-gcc)
               native-gcc
             ]
             ++ builtins.attrValues {
--- a/flake.nix
+++ b/flake.nix
@@ -49,9 +49,12 @@
               # - providing cross compilation toolchain (x86_64/aarch64-linux) for darwin can be cumbersome 
               #   and won't just work for now
               # - equip all toolchains if cross is explicitly set to true
-              # - one of the toolchains in the list is exactly the same as the toolchain in `native-gcc`, 
-              #   nix should be able to handle this properly, so it won't be an issue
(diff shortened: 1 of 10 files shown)


hanno-becker (changes requested):
@potsrevennil Thank you for making a start on this.

Can you please extend the PR so it actually exercises the R-V toolchain? E.g. expand CI to have build+test for R-V on AArch64, just as we have it for x86 on AArch64.

flake.nix
```diff
@@ -66,6 +69,12 @@
               # NOTE: we don't support nix gcc toolchains for darwin system, therefore explicitly setting CC is required
               pkgs.lib.optionalString pkgs.stdenv.isDarwin ''
                 export CC=gcc
+                export CXX=g++
+                for cmd in \
+                    ar as ld nm objcopy objdump readelf ranlib strip strings size windres
```
hanno-becker, line 76: Can you explain here and in the commit message what this is concretely needed for?

scripts/lib/mlkem_test.py
```diff
@@ -290,6 +278,30 @@ def __init__(self, opts: Options):
...
+            if opts.run_as_root:
+                logging.info(
+                    f"Running {bin} as root -- you may need to enter your root password.",
+                )
+                self.cmd_prefix.append("sudo")
+
+            if opts.exec_wrapper:
+                logging.info(f"Running with customized wrapper {opts.exec_wrapper}")
+                self.cmd_prefix = self.cmd_prefix + opts.exec_wrapper.split(" ")
+            elif opts.cross_prefix and platform.system() != "Darwin":
```
hanno-becker, line 293: We might want to fail cleanly in case `platform.system() == "Darwin"`?

hanno-becker (changes requested):
Thanks for the update @potsrevennil!

I left some comments. Also, can you add a R-V build+test to the CI?

flake.nix
potsrevennil: We're using CC and LD now, and we might use AS in the future for building library, and some other env var might also be useful while developing. And since in here https://github.com/pq-code-package/mlkem-native/blob/d297d8639c821540fd735aae3018455e435b9bf6/flake.nix#L30

nix did this underlying
```
for cmd in \
    ar as ld nm objcopy objdump readelf ranlib strip strings size windres
do
    if
        PATH=$_PATH type -p "@targetPrefix@${cmd}" > /dev/null
    then
        export "${cmd^^}${role_post}=@targetPrefix@${cmd}";
    fi
done
```
https://github.com/NixOS/nixpkgs/blob/0c5b4ecbed5b155b705336aa96d878e55acd8685/pkgs/build-support/bintools-wrapper/setup-hook.sh#L57-L65

Therefore for system that installed toolchain from nix, these environment variables will be set properly. I'd like this to be consistent between different systems, and since we don't install toolchain from nix on macOS, we'll have to explicitly do this instead.

hanno-becker (changes requested):
Thank you @potsrevennil, this is looking good!

Given that we have no native R-V builds in the CI (yet?), I'd prefer to have a top-level job for cross R-V build + test, visible in the CI summary. This way, it is immediately apparent that we do test with R-V. 

Can you arrange that?

hanno-becker (approved):
Thanks @potsrevennil !

PR #451, opened 2024-11-22
Remove loop ASSIGNS contracts now CBMC can infer most of them

### LINKED ISSUE #450: Consider dropping loop ASSIGNS contract now CBMC can infer it
CBMC 6.4.0 has gained the ability to automatically infer the ASSIGNS contract for simple loops.

We should evaluate this feature and decide if we can and should drop the ASSIGNS contract
in loop statement where it is possible to do so. This reduces burden on developers and maintainers,
and decreases clutter in the code.

Initial result: All but 1 of the ASSIGNS contracts in poly.c can be removed with no loss in
proof performance. The 1 still required is in poly_basemul_montgomery_cached(), which
has a more complex loop body that assigns to _slices_ of the target array, so not surprising
that CBMC is not able to infer the correct contract in this case.

rod-chapman:
Fixes #450 

This PR tracks efforts to remove the ASSIGNS contract from loop statement now that CBMC 6.4.0 can infer them automatically, most of the time.

As a proof of concept, all-but-one of the ASSIGNS contracts in poly.c have been removed.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -17,4 +17,4 @@ IncludeBlocks: Preserve
 
 # Designate CBMC contracts/macros that appear in .h files
 # as "attributes" so they don't get increasingly indented line after line
-AttributeMacros: ['REQUIRES', 'ENSURES', 'ASSIGNS']
+AttributeMacros: ['REQUIRES', 'ENSURES', 'ASSIGNS', 'INVARIANT']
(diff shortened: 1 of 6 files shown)


hanno-becker:
@rod-chapman Great!

In my mind, we should do this for all loops [where it is straightforward] in one PR. Do you have time to do this?

rod-chapman:
There are 23 ASSIGNS clauses that might be removed, so Yes...

hanno-becker (approved):
LGTM. Great to see the number of annotations go down

PR #452, opened 2024-11-25
Move verify.c functions to native backends

### LINKED ISSUE #317: x86: Implement verify + cmov + cmov_int16 in asm
Currently our https://github.com/pq-code-package/mlkem-c-aarch64/blob/main/mlkem/verify.c is based on the Kyber reference implementation. 
The Kyber AVX2 implementation implements this in intrinsics: https://github.com/pq-crystals/kyber/blob/main/avx2/verify.c 

Those functions should be implemented in assembly to reliably stop the compiler from messing with it.

mkannwischer:
Currently, mlkem-native has the `verify`, `cmov`, and `cmov_int16` functions in the verify.c function similar as the reference implementation. By putting them in a separate compilation unit, we stop the compiler from turning this code into variable time code.

While being portable, this has various downside:
- If LTO were to be used, this can lead to timing side-channels. We stop that by not using LTO for verify.c in our own build, but we have no control over how other people compile our code - this feels dangerous.
- If these functions are in a separte compilation unit, the compiler cannot inline them. This is very expensive particularly for `cmov_int16`. By writing that gadget in inline assembly, we can allow the compiler to inline these calls, but still guarantee that they execute in constant-time.

This commit prepares the switch to a native implementation of the verify.c functions in a architecture-specific header file. For now these are not implemented in inline assembly as I wanted to gather feedback on this first.

**Before merging this, we need to implement the functions in inline assembly as this code in it's current shape is very likely vulnerable to timing attacks**.  I wanted to gather some feedback on the proposed change first. 

- [X] Move verify.c functions to platform specific headers
- [x]  Implement AArch64 verify functions in constant-time inline assembly
- [ ]  Implement x86_64 verify functions in constant-time inline assembly

Once completed, this will resolve #317 and #318.


mlkem/verify.h
```diff
@@ -72,4 +78,17 @@ ASSIGNS(OBJECT_UPTO(r, sizeof(int16_t)))
 ENSURES(*r == (b ? v : OLD(*r)));
 // clang-format on
 
+#else /* MLKEM_USE_NATIVE */
```
hanno-becker, line 81: I was expecting this to parallel existing native functions. That is, use `#if defined(MLKEM_USE_NATIVE_VERIFY)` as a guard, and if set, define `static inline` functions (same name as in the non-native case) that forward to the native functions, e.g.
```c
static inline void cmov_int16(int16_t *r, const int16_t v, const uint16_t b) 
{
    return cmov_int16_native(r, v, b);
}
```

mkannwischer: done.

mkannwischer:
Handing over to @hanno-becker to add the inline assembly - I not too familar with that.

mlkem/verify.h
```diff
@@ -6,9 +6,28 @@
...
+#include "arith_native.h"
 #include "cbmc.h"
 #include "params.h"
 
+// If a native backend is used, we include the native implementations from the
+// backend as those are using inline assembly. This guarantees that these
+// gadgets are constant-time, but also allows the compiler to inline them.
+// Otherwise, we use the reference code (verify.c) in a separate compilation
+// unit.
+#if defined(MLKEM_USE_NATIVE_VERIFY)
```
hanno-becker, line 18: If `MLKEM_USE_NATIVE_VERIFY` is set, we don't need to do anything, do we? It would be the responsibility of the native header to define `static inline` verify functions.

mkannwischer: I'm not sure I follow. What do you want me to change?

hanno-becker: I'll do it

hanno-becker:
@mkannwischer I believe we can close this now, as its superseded by #460 ?

PR #453, opened 2024-11-25
Auto-generate zeta table for NTT/invNTT

hanno-becker:
This commit introduces scripts/autogenerate_files.py. This script auto-generates a new compilation unit `zetas.c` with the array of twiddle factors used by NTT and inverse NTT.

The linter is adjusted to always call `autogenerate_files.py`, and fail if autogenerated files are up to date; similar to formatting.

It is likely that further files will be added to the scope of `autogenerated_files.py` in the future.


mkannwischer (changes requested):
Thanks!

The lint script is not the right place for this. This is going to be very confusing should it ever fail.

Can we add another script to check this and call it in CI workflow that also calls linting.

hanno-becker:
@mkannwischer I don't follow. If linting fails because of bad formatting, one calls `format` to remedy. If linting fails because of not-up-to-date autogenerated files, one calls `autogenerate_files.py` to remedy. This seems consistent?

mkannwischer (approved):
Okay - if you phrase it this way it does make sense.

LGTM

PR #454, opened 2024-11-26
AArch64: Hoist zeta values for [inv]NTT in C file and auto-generate

hanno-becker:
This commit hoists the zeta arrays used in the AArch64 forward NTT into a separate compilation unit `aarch64_zetas.c`. Instead of manual address fiddling in assembly, the addresses of those arrays are passed directly as arguments into the assembly.

Moreover, the `autogenerate_files.py` script is extended to fully auto-generate the AArch64 zeta arrays.


mlkem/native/aarch64/arith_native_aarch64.h
```diff
@@ -9,17 +9,30 @@
...
+  MLKEM_NAMESPACE(aarch64_invntt_zetas_layer56)
+
+extern const int16_t aarch64_ntt_zetas_layer01234[];
+extern const int16_t aarch64_ntt_zetas_layer56[];
+extern const int16_t aarch64_invntt_zetas_layer01234[];
+extern const int16_t aarch64_invntt_zetas_layer56[];
+
 #define ntt_asm_clean MLKEM_NAMESPACE(ntt_asm_clean)
-void ntt_asm_clean(int16_t *);
+void ntt_asm_clean(int16_t *, const int16_t *, const int16_t *);
```
rod-chapman, line 32: Is it not possible to name these formal parameters?

mlkem/native/aarch64/aarch64_zetas.c
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) 2024 The mlkem-native project authors
+// SPDX-License-Identifier: Apache-2.0
+
+// WARNING: This file is auto-generated from scripts/autogenerate_files.py
```
rod-chapman, line 4: Of this file is auto-generated, why both to add it to the repo at all? Why not just re-generate it on demand as part of the build process? I guess that creates a dependence on Python?

hanno-becker: To make it easy to use

mkannwischer (approved):
LGTM. I tested that the lint script fails as expected if you mess with the autogenerated files. 
Thanks @hanno-becker!

PR #456, opened 2024-11-26
Add poly_getnoise_eta2_4x alias for poly_getnoise_eta1_4x

mkannwischer:
In encapsulation of ml-kem-768 and ml-kem-1024, we are using poly_getnoise_eta1_4x to sample both the secret and the error. This works because ETA1 = ETA2 = 2.
However, this is somewhat confusing when reading the code.

This commit uses poly_getnoise_eta2_4x instead and defines it as an alias for poly_getnoise_eta1_4x in case ETA1=ETA2. I think this the cleaner solution.


hanno-becker (approved):

hanno-becker:
@mkannwischer I agree, this is cleaner. Thank you

PR #458, opened 2024-11-28
CBMC: Make CBMC annotations more readable

hanno-becker:
Previously, CBMC annotations were uppercase macros such as INVARIANT, REQUIRES, ENSURES, ...

They had to be surrouneded by clang-format on/off annotations to avoid clang-format messing with them in a bad way.

This commit changes the syntax for CBMC contracts and loop invariants. Now, a function contract is enclosed in __contract__(...) and a loop contract is enclosed in __loop__(...). Moreover, the capitalization of the annotations has been changed to lowercase.

A hack for clang-format is included which makes it format __contract__ and __loop__ correctly, so we don't need the clang-format on/off annotations anymore, which improves readability.

The underscores in __loop__ and __contract__ are deliberate, since it is expected that the average C developer will recognize this as an attribute, and not wonder if it is executable code.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -17,6 +17,8 @@ IncludeBlocks: Preserve
 
 # Designate CBMC contracts/macros that appear in .h files
 # as "attributes" so they don't get increasingly indented line after line
-AttributeMacros: ['REQUIRES', 'ENSURES', 'ASSIGNS', 'INVARIANT']
-
 BreakBeforeBraces: Allman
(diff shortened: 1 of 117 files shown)


hanno-becker:
**Acknowledgement:** This PR was supported by `sed`.

mkannwischer (approved):
Thanks @hanno-becker. I think is is much easier to read and I'm very happy that we don't have to constantly turn off `clang-format`

PR #460, opened 2024-11-29
Harden constant time functions against inlining and optimizations

hanno-becker:
Harden constant time functions against inlining and optimizations

The functions in `verify.c` are suspectible to be compiled into variable-time
code when inlined. For this reason, we disable LTO for `verify.c` in our own
build, and explicitly warn about it in the README.md and `verify.c`.

Yet, with mlkem-native being intended as a code-package, there is risk that
users will copy-paste the code and compile it with LTO enabled, overlooking
the warning.

This commit is an attempt to improve robustness towards LTO by rewriting
verify.c so that it is safe even when inlined.

The main idea is to mask values which the compiler can detect to be 0 or 1,
to avoid branches being introduced for them. More precisely, we introduce
a "value barrier" which is a functionally a no-op, but is written so that
the compiler cannot reason that it is. Then, all 0/1 values are fed through
a value barrier before further processing.

We consider two approaches to implement a value barrier:
- When supported, an empty inline asm block which marks the target value
  as clobbered.
- XOR'ing with the value of a volatile global that's set to 0.

The first approach is rather cheap because it only prevents the compiler
from reasoning about the value of the variable past the barrier. The second
generates redundant loads which therefore incur a runtime cost. Yet, it
seems both more portable and more robust.


mlkem/verify.c
```diff
@@ -1,81 +1,5 @@
...
-  // b == 0 => mask = 0x0000
-  // b == 1 => mask = 0xFFFF
-  const uint16_t mask = -b;
-
-  // mask == 0x0000 => *r == (*r ^ 0x0000) == *r
-  // mask == 0xFFFF => *r == (*r ^ (*r ^ v)) == (*r ^ *r) ^ v == 0 ^ v == v
-  *r ^= mask & ((*r) ^ v);
-#pragma CPROVER check pop
-}
+volatile uint64_t ct_opt_blocker_u64;
```
mkannwischer, line 5: maybe explicitly set this to zero? I know it gets initialized to zero anyway, but I had to look that one up.

mlkem/verify.h
```diff
@@ -9,66 +9,227 @@
 #include "cbmc.h"
 #include "params.h"
 
-#define verify MLKEM_NAMESPACE(verify)
+#if !defined(CBMC) && !defined(_WIN32)
```
mkannwischer, line 12: Add a comment here in which case which approach is being used?

mlkem/verify.h
```diff
@@ -9,66 +9,227 @@
...
+  ensures((return_value == 0) == forall(int, i, 0, ((int)len - 1), (a[i] == b[i]))))
+{
+  uint8_t r = 0, s = 0;
+
+  // Switch to a _signed_ ilen value, so that our loop counter
+  // can also be signed, and thus (i - 1) in the loop invariant
+  // can yield -1 as required.
+  const int ilen = (int)len;
+
+  for (int i = 0; i < ilen; i++)
```
mkannwischer, line 188: In the following function you just used size_t for the loop counter as well. 
Did you try that here, too?

mkannwischer (changes requested):
Thanks @hanno-becker. The core of this PR is great and greatly improves my confidence in this being constant time with many compiler using many flags. A few small nits (mostly about comments). 

Could you also put a summary on the impact on performance at end of the commit message. 

What is left to do is to actually check that these are constant-time in many scenarios. I suggest using valgrind with the patch from https://kyberslash.cr.yp.to/papers.html for detecting secret dependent divisons (+ secret-dependent memory accesses, branches) - we should run that in CI with many compilers and flag combinations. I'll open an issue for that.

.github/actions/bench/action.yml
```diff
@@ -97,6 +97,7 @@ runs:
         comment-on-alert: true
         summary-always: true
         alert-threshold: ${{ inputs.alert_threshold }}
+        comment-always: true # TEST ONLY
```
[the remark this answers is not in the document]
hanno-becker: Actually, I think we should keep this in. When we run benchmarks, we want to know their results. The autogenerated comments provide that summary together with % deviations. I'll remove the comment.

cbmc/proofs/ct_cmask_neg_i16/ct_cmask_neg_i16_harness.c
```diff
@@ -7,8 +7,8 @@
  */
 
 /**
- * @file cmov_int16_harness.c
- * @brief Implements the proof harness for cmov_int16 function.
+ * @file ct_csel_int16_harness.c
+ * @brief Implements the proof harness for ct_csel_int16 function.
```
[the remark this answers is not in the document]
hanno-becker: Yes, I will change it

cbmc/proofs/ct_cmask_nonzero_u8/ct_cmask_nonzero_u8_harness.c
```diff
@@ -0,0 +1,30 @@
...
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT-0 AND Apache-2.0
+
+/*
+ * Insert copyright notice
+ */
+
+/**
+ * @file ct_csel_int16_harness.c
+ * @brief Implements the proof harness for ct_csel_int16 function.
```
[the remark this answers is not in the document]
hanno-becker: Yes, I will change it

mlkem/verify.h
hanno-becker: This is because we have a more complicated loop invariant which is easier to state with `int` than `size_t`.

mlkem/verify.h
hanno-becker: Yes

hanno-becker:
@mkannwischer Thanks for the review 🙏 I think I've addressed your comments.

.github/actions/bench/action.yml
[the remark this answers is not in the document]
mkannwischer: okay!

mlkem/verify.h
mkannwischer: okay.

mlkem/verify.h
mkannwischer: Thanks for that ellaborate comment - that's really useful!

mlkem/verify.c
mkannwischer: great!

mkannwischer (approved):
Thanks for the changes -- looks good to me now! 
CI benchmarking failed on c7i-metal due to vCPU limitations. I restarted it now. Should be fine in a few minutes.

PR #464, opened 2024-11-30
[TEST] Hardcode MLKEM_GEN_MATRIX_NBLOCKS to 1

hanno-becker:

### DIFF
--- a/.github/actions/bench/action.yml
+++ b/.github/actions/bench/action.yml
@@ -97,6 +97,7 @@ runs:
         comment-on-alert: true
         summary-always: true
         alert-threshold: ${{ inputs.alert_threshold }}
+        comment-always: true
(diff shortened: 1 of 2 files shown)


hanno-becker:
@mkannwischer I got curious about this seeing that the BoringSSL code effectively uses the value 1 here.

hanno-becker:
@mkannwischer I got curious about this seeing that the BoringSSL code effectively uses the value 1 here. I am a bit confused to see performance warnings for the opt-variants only, and virtually no change for the no-opt code.

hanno-becker:
Explanation: Values > 1 for MLKEM_GEN_MATRIX_NBLOCKS are only relevant if an optimized rejection sampling is used that benefits from larger buffer lengths.

PR #465, opened 2024-12-01
C NTT: Use Barrett instead of Montgomery multiplication

hanno-becker:
This commit changes the C implementation of NTT and inverse NTT to use Barrett multiplication (https://eprint.iacr.org/2021/986) for multiplications with twiddles.

Using Barrett multiplication requires precomputing twisted twiddles, which is done by adjusting the twiddle generation script autogenerate_files.py.


hanno-becker:
No merit.

PR #466, opened 2024-12-01
C NTT: Use (64-bit) Barrett instead of 32-bit Montgomery reduction

hanno-becker:


hanno-becker:
No merit.

PR #467, opened 2024-12-02
Achieve C90 compliance

mkannwischer:
This PR achieves C90 compliance of our code - with one exception: We require the `long long` datatype and disable the corresponding warning using `-Wno-long-long`. This is what other crypto libraries (e.g., OpenSSL) are also doing. 

Still working on making sure this runs well on all platforms. Once that works, I will switch out default builds back to C99, and add another CI workflow for testing C90 compliance.


mlkem/common.h
```diff
@@ -5,16 +5,36 @@
...
+
+/*
+ * C90 does not have the inline compiler directive yet.
+ * We don't use it in C90 builds.
+ * However, in that case the compiler warns about some inline functions in
+ * header files not being used in every compilation unit that includes that
+ * header. To work around it we silence that warning in that case using
+ * __attribute__((unused)).
+ */
+#if !defined(inline)
```
hanno-becker, line 18: Have you tested that this works, since `inline` is a keyword, not a macro (I think)?

We might want to test the C standard here instead?

fips202/fips202.c
```diff
@@ -1,5 +1,7 @@
-// Copyright (c) 2024 The mlkem-native project authors
-// SPDX-License-Identifier: CC0-1.0
+/*
+ * Copyright (c) 2024 The mlkem-native project authors
```
hanno-becker, line 2: I assume you did not do this by hand -- how did you automate the reformatting?

mlkem/verify.h
```diff
@@ -237,9 +240,9 @@ __contract__(
 {
   uint8_t r = 0, s = 0;
 
-  // Switch to a _signed_ ilen value, so that our loop counter
-  // can also be signed, and thus (i - 1) in the loop invariant
-  // can yield -1 as required.
+  /* Switch to a _signed_ ilen value, so that our loop counter */
+  /* can also be signed, and thus (i - 1) in the loop invariant */
```
hanno-becker, line 244: Ideally, those should be a single block comment.

mlkem/ntt.c
```diff
@@ -53,7 +53,8 @@ __contract__(
 {
   /* `bound` is a ghost variable only needed in the CBMC specification */
   ((void)bound);
```
hanno-becker: How is that compliant? Isn't this technically a statement?

mlkem/ntt.c
```diff
@@ -90,10 +91,11 @@ __contract__(
   ensures(array_abs_bound(r, 0, MLKEM_N - 1, (layer + 1) * MLKEM_Q - 1)))
 {
   /* `layer` is a ghost variable only needed in the CBMC specification */
+  int start;
```
hanno-becker: Misplaced declaration, should not be between comment and `((void) layer`

mlkem/poly.c
```diff
@@ -305,16 +311,17 @@ void poly_frombytes(poly *r, const uint8_t a[MLKEM_POLYBYTES])
...
 #endif
 
-  for (int i = 0; i < MLKEM_N / 8; i++)
+  for (i = 0; i < MLKEM_N / 8; i++)
   __loop__(
     invariant(i >= 0 && i <= MLKEM_N / 8)
     invariant(array_bound(r->coeffs, 0, (8 * i - 1), 0, (MLKEM_Q - 1))))
   {
-    for (int j = 0; j < 8; j++)
+    for (j = 0; j < 8; j++)
```
hanno-becker, line 324: Can we keep the declaration at the top of the _block_?

fips202/native/aarch64/profiles/cortex_a55.h
```diff
@@ -15,7 +15,7 @@
 // On Cortex-A55, we use lazy rotation assembly for Keccak-x1,
 // but no batched assembly implementation.
 #define MLKEM_USE_FIPS202_X1_NATIVE
-static inline void keccak_f1600_x1_native(uint64_t *state)
+static INLINE void keccak_f1600_x1_native(uint64_t *state)
```
hanno-becker, line 18: I wonder if instead we should have `STATIC_INLINE`? We already have `STATIC_INLINE_TESTABLE`

mlkem/poly.h
```diff
@@ -38,17 +38,6 @@ typedef struct
   int16_t coeffs[MLKEM_N >> 1];
 } poly_mulcache;
 
-#define scalar_compress_d1 MLKEM_NAMESPACE(scalar_compress_d1)
```
hanno-becker, line 39: There are probably more examples for this? Any particular reason why this change was made?

mlkem/polyvec.c
```diff
@@ -94,13 +94,14 @@ void polyvec_basemul_acc_montgomery_cached(poly *r, const polyvec *a,
                                            const polyvec *b,
                                            const polyvec_mulcache *b_cache)
 {
+  int i;
+  poly t;
+
+
```
hanno-becker, line 100: Nit: Remove double newline

test/nistrng/rng.c
```diff
@@ -49,7 +49,7 @@ static void nistkat_update(const unsigned char *provided_data,
 {
   int i;
   int len = AES256_KEYBYTES + AES_BLOCKBYTES;
-  uint8_t tmp[len];
```
hanno-becker, line 48: Oof, I didn't realize we did this -- we never should have. Variable-length stack arrays are dangerous.

hanno-becker (changes requested):
Overall this looks good -- it's surprising how few changes were needed after all.

Left a few nits

test/nistrng/rng.c
mkannwischer: note that this is code written by NIST.

mlkem/common.h
mkannwischer: This works in gcc - yes. But you are right this is not a portable solution - it did not work with `msvc` . I re-worked it now to be more portable.

fips202/fips202.c
mkannwischer: I replaced all copyright headers in a one-off replace. 
Then I replaced all one line comments `//` with `/* */` using a reg-ex. 
Then. I went through all other comments manually and replaced them manually depending on what makes sense. 

Finally (after your review), I went through all multi-line comments manually and replaced
```
/* XXX */ 
/* YYY */ 
```

with 
``` 
/* 
  * XXX
  * YYY
  */
  ```

File: mlkem/verify.h (hunk 3 above)
Comment (mkannwischer) [reply mlkem/verify.h]: resolved now.

File: mlkem/ntt.c (hunk 4 above)
Comment (mkannwischer) [reply mlkem/ntt.c]: resolved.

File: mlkem/ntt.c (hunk 5 above)
Comment (mkannwischer) [reply mlkem/ntt.c]: resolved

File: mlkem/poly.c (hunk 6 above)
Comment (mkannwischer) [reply mlkem/poly.c:347]: yes - I changed this now. Can you please double check as it's easy to miss.

File: mlkem/poly.h (hunk 8 above)
Comment (mkannwischer) [reply mlkem/poly.h:39]: I just removed it here because I saw it. 
I've dropped it here and added it to my todo list to remove it later for all static functions.

File: fips202/native/aarch64/profiles/cortex_a55.h (hunk 7 above)
Comment (mkannwischer) [reply fips202/native/aarch64/profiles/cortex_a55.h:20]: hmm. Can you elaborate why you think that is better?

File: mlkem/polyvec.c (hunk 9 above)
Comment (mkannwischer) [reply mlkem/polyvec.c]: done.

File: fips202/native/aarch64/profiles/cortex_a55.h (hunk 7 above)
Comment (hanno-becker) [reply fips202/native/aarch64/profiles/cortex_a55.h:20]: I find `STATIC_INLINE` + `STATIC_INLINE_TESTABLE` more uniform than `STATIC_INLINE_TESTABLE + static INLINE`. But this just taste, let's leave it.

**Comment by mkannwischer** [comment]
This is ready for review now and should be merged before we merge any other PRs as rebasing this is going to be quite painful.

File: fips202/keccakf1600.c [hunk 11]
```diff
@@ -24,9 +24,10 @@
 void KeccakF1600_StateExtractBytes(uint64_t *state, unsigned char *data,
                                    unsigned int offset, unsigned int length)
 {
+  unsigned int i;
```
hanno-becker, line 27: For `!SYS_LITTLE_ENDIAN`, there is now a double-declaration of `i`

fips202/keccakf1600.c
```diff
@@ -43,9 +44,10 @@ void KeccakF1600_StateExtractBytes(uint64_t *state, unsigned char *data,
 void KeccakF1600_StateXORBytes(uint64_t *state, const unsigned char *data,
                                unsigned int offset, unsigned int length)
 {
+  unsigned int i;
 #if defined(SYS_LITTLE_ENDIAN)
   uint8_t *state_ptr = (uint8_t *)state + offset;
-  for (unsigned int i = 0; i < length; i++)
+  for (i = 0; i < length; i++)
```
hanno-becker, line 49: same here

mlkem/indcpa.c
```diff
@@ -354,7 +354,7 @@ void gen_matrix(polyvec *a, const uint8_t seed[MLKEM_SYMBYTES], int transposed)
    */
   for (i = 0; i < MLKEM_K; i++)
   {
-    for (int j = 0; j < MLKEM_K; j++)
+    for (j = 0; j < MLKEM_K; j++)
```
hanno-becker, line 361: Change of type, but benign

fips202/keccakf1600.c
mkannwischer: resolved.

fips202/keccakf1600.c
mkannwischer: resolved.

.github/workflows/ci.yml
```diff
@@ -49,6 +49,39 @@ jobs:
...
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: make quickcheck
+        run: |
+          OPT=0 CPPFLAGS=-std=c90 make quickcheck >/dev/null
+          make clean            >/dev/null
+          OPT=1 CPPFLAGS=-std=c90 make quickcheck >/dev/null
+      - uses: ./.github/actions/setup-ubuntu
+      - name: tests func
+        run: |
+          CPPFLAGS="-std=c90" ./scripts/tests func
```
hanno-becker, line 81: If we use `tests`, we should use the `--cflags` argument

hanno-becker (changes requested):
LGTM except for a minor nit, and could you please also update the README to say "C90 implementation" and "deviate only to ..." -> "deviate primarily to ..."?

.github/workflows/ci.yml
hanno-becker: Ok, @mkannwischer explained that this does not work at the moment because `std=c90` would end up before `std=c99` on the cmd line, and `c99` would take precedence

mkannwischer:
Thank you @hanno-becker for your review(s)!
I've updated the README just now.

hanno-becker (approved):
Thanks a lot @mkannwischer !

PR #471, opened 2024-12-02
Add -march=native -mtune=native for benchmarking

mkannwischer:
On my machine adding `-march=native -mtune=native` improves performance. 
Since we know have a bot telling us performance changes, let's just try in CI as well.


mkannwischer:
no change - nevermind.

PR #476, opened 2024-12-03
CBMC: add `tests cbmc` command

mkannwischer:
Currently to run all CBMC proofs, we have to run
```
cd cbmc/proofs
MLKEM_K=2 python3 run-cbmc-proofs.py
MLKEM_K=3 python3 run-cbmc-proofs.py
MLKEM_K=4 python3 run-cbmc-proofs.py
```

This commit wraps all that into a single wrapper:
`tests cbmc`

This is not integrated into CI.

### DIFF
--- a/.github/actions/cbmc/action.yml
+++ b/.github/actions/cbmc/action.yml
@@ -46,7 +46,6 @@ runs:
       - name: Run CBMC proofs (MLKEM_K=${{ inputs.mlkem_k }})
         shell: ${{ env.SHELL }}
         run: |
-          cd cbmc/proofs;
           echo "::group::cbmc_${{ inputs.mlkem_k }}"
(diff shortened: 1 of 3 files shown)


hanno-becker (approved):
Tested locally, inside and outside nix shell.

hanno-becker (changes requested):
We should change the CI to use this command as well?

mkannwischer:
> We should change the CI to use this command as well?

Then we need to extend it to take a param which parameter set should be tested. Currently we don't have infrastructure for that in the test scripts.

mkannwischer:
> > We should change the CI to use this command as well?
> 
> Then we need to extend it to take a param which parameter set should be tested. Currently we don't have infrastructure for that in the test scripts.

I've added it just now. 
You can now run it as `tests cbmc --k <MLKEM_K>`

hanno-becker (approved):

PR #493, opened 2024-12-04
CBMC: Prove `mlkem_shake256_prf` and `mlkem_shake256_rkprf`

mkannwischer:
These two are still missing. As they are basically wrappers around the FIPS202 functions, these are straightforward given the FIPS202 contracts.

### DIFF
--- a/cbmc/proofs/mlkem_shake256_rkprf/Makefile
+++ b/cbmc/proofs/mlkem_shake256_rkprf/Makefile
@@ -1,0 +1,54 @@
+# SPDX-License-Identifier: Apache-2.0
+
+include ../Makefile_params.common
+
+HARNESS_ENTRY = harness
+HARNESS_FILE = mlkem_shake256_rkprf_harness
+
+# This should be a unique identifier for this proof, and will appear on the
+# Litani dashboard. It can be human-readable and contain spaces if you wish.
+PROOF_UID = mlkem_shake256_rkprf
+
+DEFINES +=
+INCLUDES +=
+
+REMOVE_FUNCTION_BODY +=
+UNWINDSET +=
+
+PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
+PROJECT_SOURCES += $(SRCDIR)/mlkem/symmetric-shake.c $(SRCDIR)/fips202/fips202.c
+
+CHECK_FUNCTION_CONTRACTS=$(MLKEM_NAMESPACE)mlkem_shake256_rkprf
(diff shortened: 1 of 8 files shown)


cbmc/proofs/mlkem_shake256_rkprf/Makefile
```diff
@@ -0,0 +1,54 @@
...
+PROOF_UID = mlkem_shake256_rkprf
+
+DEFINES +=
+INCLUDES +=
+
+REMOVE_FUNCTION_BODY +=
+UNWINDSET +=
+
+PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
+PROJECT_SOURCES += $(SRCDIR)/mlkem/symmetric-shake.c $(SRCDIR)/fips202/fips202.c
```
hanno-becker, line 19: Why do we need to import fips202.c?

mkannwischer: That's because it's calling `shake256_inc_ctx_release` which does not have a contract but is an empty function for us.

fips202/fips202x4.h
```diff
@@ -16,8 +16,15 @@
...
-                       const uint8_t *in3, size_t inlen);
-
+                       const uint8_t *in3, size_t inlen)
+__contract__(
+  requires(memory_no_alias(state, sizeof(keccakx4_state)))
+  requires(memory_no_alias(in0, inlen))
+  requires(memory_no_alias(in1, inlen))
+  requires(memory_no_alias(in2, inlen))
+  requires(memory_no_alias(in3, inlen))
+  assigns(object_whole(state))
```
hanno-becker, line 26: `memory_slice` instead of `object_whole`?

cbmc/proofs/mlkem_shake256_rkprf/mlkem_shake256_rkprf_harness.c
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2024 The mlkem-native project authors
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT-0
+
+/*
```
hanno-becker, line 5: I know we're not consistent, but I've turned to just removing all those boilerplate comments for new Makefiles. Can you remove them here too?

cbmc/proofs/mlkem_shake256_rkprf/Makefile
hanno-becker: Ok

fips202/fips202x4.h
hanno-becker: If it works for the caller, it should be OK

cbmc/proofs/mlkem_shake256_rkprf/mlkem_shake256_rkprf_harness.c
hanno-becker: Ok, we can do this another time.

hanno-becker (approved):

PR #499, opened 2024-12-04
Fix `tests bench` mac-taskpolicy handling in benchmarks

potsrevennil:
- #449 accidentally broke `tests bench` cli parsing on mac when `mac-taskpolicy` option is presented.
- `tests bench -c M1 --run-as-root` would still work as expected

### DIFF
--- a/scripts/lib/mlkem_test.py
+++ b/scripts/lib/mlkem_test.py
@@ -602,11 +602,7 @@ class Tests:
             output = False
 
         if mac_taskpolicy:
-            if exec_wrapper:
-                logging.error(f"cannot set both --mac-taskpolicy and --exec-wrapper")
-                sys.exit(1)
-            else:
-                exec_wrapper = f"taskpolicy -c {mac_taskpolicy}"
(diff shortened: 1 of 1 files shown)


mkannwischer (changes requested):
Thanks @potsrevennil.

Please clarify your commit message (also the first line is too long) and describe the bug this is fixing. 

My understanding is that `--mac-taskpolicy` didnt work at all anymore. 
This doesn't affect the (more common) benchmarks on the high-performance cores of the M1, as for that you would just run 
```
tests bench -c M1 --run-as-root
```

Corrrect?

potsrevennil:
> My understanding is that `--mac-taskpolicy` didnt work at all anymore. This doesn't affect the (more common) benchmarks on the high-performance cores of the M1, as for that you would just run
> 
> ```
> tests bench -c M1 --run-as-root
> ```
> 
> Corrrect?

Yes

mkannwischer (approved):
Thanks @potsrevennil for the minmal change! 

I have checked that
```
tests bench -c M1 --run-as-root --mac-taskpolicy background
```
works now. 

I also checked that
```
tests bench -c M1 --run-as-root
```

still works.

PR #503, opened 2024-12-09
Add riscv64 board (Bananapi bpi-f3) to benchmark workflow

potsrevennil:


mkannwischer (approved):
Thanks @potsrevennil - this works like a charm. 

I ran the BPi benchmarking CI 3 times and got the following cycle counts for ML-KEM 768 encaps: 826547, 826481, 826483. That seems reasonably stable for our benchmarking. Still a good idea to watch the benchmarks over the next few PRs that shouldn't change performance.

Issue #434, opened 2024-11-20
Move all docstrings to header files


mkannwischer:
Originally docstrings were only in the C sources, but not in the header files. 
As a part of doing the CBMC proofs, we pulled most of the docstrings to the headers files, but not all. 
Now it is rather inconsistent, see e.g., https://github.com/pq-code-package/mlkem-native/blob/main/mlkem/polyvec.c 
This is quite upseting for my OCD. 

@hanno-becker, do you agree that we move all docstrings to the header files? While we are at it should we enforce something else about the docstrings?

Issue #211, opened 2024-10-11
Appropriate comments in header and body files


rod-chapman:
For future consideration:

In the PQCrystals/Kyber reference code, there is a trend for all functions comments to appear in the body (*.c) files
for all translation units. This is counter to common practice, where what the user/client needs to know should be in the
header (.h) file, while additional information pertaining to the implementation should only appear in the body.

This Issue raised to track resolution for all units.

hanno-becker:
@rod-chapman Can you do this?

rod-chapman:
We should first agree a standard for this. Then do one of the lowest-level modules to see how it looks (e.g reduce.c) before moving on to others.

hanno-becker:
@rod-chapman I think you are right that it is uncommon to have function documentation in the .c file. The format of the documentation is a separate question. So I'd say let's move the documentation over to .h in a first step, and then discuss & rework what we want it to say exactly -- I assume there will be a lot of changes upcoming with CBMC.

mkannwischer:
Related to #434. 

We should also automatically generate documentation from these comments - I'll open a separate issue for that.

mkannwischer:
I guess we can close this one and add to #434 that we need to make sure all functions have documentation in the headers.

PR #504, opened 2024-12-09
Move all function signature documentation to header files

mkannwischer:
We have partially moved documentations describing function sigantures to header files while writing CBMC proofs. However, it's currently not consistent and there is still come documentation left in C sources.

This commit moves all those comments to the header files.

Resolves https://github.com/pq-code-package/mlkem-native/issues/434 Resolves https://github.com/pq-code-package/mlkem-native/issues/211

I also cleaned up the placement of the namespacing.


hanno-becker (approved):
LGTM

PR #505, opened 2024-12-10
Remove redundant reduction from `unpack_pk()` and `unpack_sk()`

### LINKED ISSUE #480: Remove redundant `poly_reduce()` from `unpack_pk()`
Resolve https://github.com/pq-code-package/mlkem-native/blob/41b272ffaca9314f1a392caf3f6169395b5e5e2d/mlkem/indcpa.c#L63-L68

hanno-becker:
Resolves #480

The specifications for the polynomial base multiplication currently require one input to be bound by MLKEM_Q in absolute value. To ensure this, the unpacking routines `unpack_pk()` and `unpack_sk()` currently explicitly reduce their output via `poly_reduce()`.

Functionally, this reduction is not needed:
- For the PK, the modulus-check already ensures that the unpacked PK will be unsigned canonical.
- For the SK, there is no requirement in the standard to check that the SK is reduced, or reduce it after unpacking.
- There is no risk of overflow in the base multiplication even if the inputs are bound by 4096 rather than MLKEM_Q.

This commit strengthens the specifications for all polynomial base multiplication routines
- basemul_cached
- poly_basemul_montgomery_cached
- polyvec_basemul_acc_montgomery_cached to work with the input bound 4096 instead of MLKEM_Q. Accordingly, the output bound needs weakening from 3/2 * Q to 2*Q; this could be much tighter if need be, but it's unnecessary.

The polynomial reductions in `check_pk` and `check_sk` can then be safely removed.

### DIFF
--- a/mlkem/polyvec.h
+++ b/mlkem/polyvec.h
@@ -93,7 +93,7 @@ __contract__(
  *
  * Arguments:   - const polyvec *a: pointer to output vector of polynomials
  *                 (of length MLKEM_POLYVECBYTES). Output will have coefficients
- *                 normalized to [0,..,q-1].
+ *                 normalized in [0..4095].
  *              - uint8_t *r: pointer to input byte array
  **************************************************/
 void polyvec_frombytes(polyvec *r, const uint8_t a[MLKEM_POLYVECBYTES])
@@ -102,7 +102,7 @@ __contract__(
   requires(memory_no_alias(a, MLKEM_POLYVECBYTES))
   assigns(object_whole(r))
   ensures(forall(int, k0, 0, MLKEM_K - 1,
-        array_bound(r->vec[k0].coeffs, 0, (MLKEM_N - 1), 0, 4095)))
+        array_bound(r->vec[k0].coeffs, 0, (MLKEM_N - 1), 0, UINT12_MAX)))
 );
 
 #define polyvec_ntt MLKEM_NAMESPACE(polyvec_ntt)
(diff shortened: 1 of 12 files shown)


mlkem/polyvec.h
```diff
@@ -171,9 +171,9 @@ __contract__(
   requires(memory_no_alias(r, sizeof(poly)))
   requires(memory_no_alias(a, sizeof(polyvec)))
   requires(memory_no_alias(b, sizeof(polyvec)))
-/* Input is coefficient-wise < q in absolute value */
+  /* Input is coefficient-wise < q in absolute value */
```
mkannwischer, line 174: remove this comment p

mlkem/reduce.h
```diff
@@ -20,16 +20,16 @@
...
- *              smaller than 3/2 q in absolute value.
+ *              smaller than 2 * q in absolute value.
  **************************************************/
 int16_t montgomery_reduce(int32_t a)
 __contract__(
-  requires(a > -(2 * MLKEM_Q * 32768))
-  requires(a <  (2 * MLKEM_Q * 32768))
-  ensures(return_value > -(3 * HALF_Q) && return_value < (3 * HALF_Q))
+  requires(a > -(2 * 4096 * 32768))
+  requires(a <  (2 * 4096 * 32768))
```
mkannwischer, line 31: Should this be `UINT12_MAX` = 4095?

hanno-becker: No, it's a `<`

mkannwischer (approved):
Thanks @hanno-becker - this is great and results in a surprisingly large performance gain on some platforms of up to 5%. 

One nit.

mlkem/polyvec.h
hanno-becker: Done

mkannwischer:
Thanks for the change!

PR #507, opened 2024-12-10
Ease integration into other libraries

mkannwischer:
This PR makes a series of changes easing integration. Those were brought up during the liboqs by @bhess. 

The changes are: 
- Include backend (C, AARCH64, X86_64) in namespace.
- ~Align FIPS202 + FIPS202x4 interface with liboqs.~ (see #508)
- Move definition of `KECCAK_WAY` to `params.h` so that it is set even if a different FIPS202 backend is used.
- ~Add liboqs META files~ 

Please review the individual commits.


hanno-becker (commented):
@mkannwischer Many thanks for the work!

I don't follow yet why we have added the `"inc"` qualification to the FIPS202 contexts -- after all, we _do_ only use them with (a) a single absorb step, (b) block-wise squeezing? 

I would rather drop the additional META files, at least so long as we _also_ retain the existing `META.yml`.

mkannwischer:
> @mkannwischer Many thanks for the work!
> 
> I don't follow yet why we have added the `"inc"` qualification to the FIPS202 contexts -- after all, we _do_ only use them with (a) a single absorb step, (b) block-wise squeezing?

Yes, this is quite confusing - we need to come up with a better solution for that. 
The reason for this is that liboqs _only_ supports a fully incremental API: https://github.com/open-quantum-safe/liboqs/blob/bbf1dbec293df3869c5c1c28b4b774b5db122ad7/src/common/sha3/sha3.h.
The `absorb_once` is also not existing there, but it's a simple wrapper of reset+aborb+finalize: https://github.com/open-quantum-safe/liboqs/blob/bbf1dbec293df3869c5c1c28b4b774b5db122ad7/src/common/pqclean_shims/fips202.c 

I can think of two options how we can resolve this cleaner:
- We switch to a fully incremental API. Since we already need this for `shake256`, this does not sound to far fetched. This would be easy for the `shake128` (as we already have the functions there, but somehwat annoying for `shake256x4` and `shake128x4` as those need to be reimplemented. 
- We leave the current API in place. And provide shims that map it onto the liboqs API. I think we should definitely change the `keccakx4_state` to `shake128x4ctx` and `shake256x4ctx` otherwise the mapping is going to be impossible. 

I prefer the latter option. Would you be okay with that?

> 
> I would rather drop the additional META files, at least so long as we _also_ retain the existing `META.yml`.

Agreed. If it is useful for oqs, we could extra fields to the existing META.yml such as the testvector hashes or the flags needed to compile a certain implementation.

hanno-becker (approved):

Issue #457, opened 2024-11-27
Demonstrate build with all 3 parameter sets in CI


rod-chapman:
Consumers of mlkem-native (e.g. other major cryptographic libraries) will likely want to include all three ML-KEM parameter sets for their users.

We therefore need to demonstrate that mlkem-native can be built 3 times with MLKEM_K=[2,3,4] and the resulting object files combined into a single library with no clashes.

On platforms that support it, this should be demonstrated for producing both a static- and a dynamically-linked library.

hanno-becker:
I agree.

@mkannwischer @potsrevennil What are your thoughts on changing the build so it generates a `libmlkem{512,768,1024}.a` in the interim, and we demonstrate in a separate test that all those link together into `libmlkem.a`?

rod-chapman:
I like that idea. Some consumers might want just one (small) library that implements just 1 parameter set. Some consumers might want all 3. We need to demonstrate that both options are possible and easy.

hanno-becker:
No, they are all namespaced. Doing things without duplication would be out of scope initially.

mkannwischer:
@rod-chapman - is this something you need relatively soon or just an idea that we may want to consider when we have time? 

> I agree.
> 
> @mkannwischer @potsrevennil What are your thoughts on changing the build so it generates a `libmlkem{512,768,1024}.a` in the interim, and we demonstrate in a separate test that all those link together into `libmlkem.a`?

I agree with that proposal - but for now this is just for our build system, not something that we ship in releases or want the consumer to use directly. If we actually want to ship a library that is going to be much more work - esp. w.r.t. Keccak implementation variants.

> No, they are all namespaced. Doing things without duplication would be out of scope initially.

I also strongly agree with this. If you want to do that, you will have to re-work a lot of code and move further away from the reference implementation which is not something we want to do unless we have to. 
If this is something that's important for one of our (potential) consumers, then we should discuss that separately.

hanno-becker:
> I agree with that proposal - but for now this is just for our build system, not something that we ship in releases or want the consumer to use directly. If we actually want to ship a library that is going to be much more work - esp. w.r.t. Keccak implementation variants.

I agree. But even if we just showcase the possibility in our own build system, that's useful.

potsrevennil:
> I agree.
> 
[...]

PR #509, opened 2024-12-11
Build static library for 3 parameter set

potsrevennil:
**Summary**:
- Resovles #457 

**Steps**:
- Since `$AR` doesn't care about duplicated symbols, the build system now links the static library after it is built for chekcing any duplicates. Additionally, a quick check for this has been added to CI.

- The build system is adjusted to build test binaries by linking with static library instead of directly with object files, since we'll need to test the built static library anyway. Therefore the build system would be simpler and easier to maintain if we do it this way.

- However, seems like `gcc` and `ar` are not happy with linking with static library if `-flto` is provided. `--plugin` would need to be properly set to the lto plugin location. Using `gcc-ar` as default instead of `ar` should resolve the issue, as the wrapper is already properly set with the configuration. 
  - Note that this is not the case for `ar` shipped with `apple clang` - 
  - Haven't checked with `llvm clang` yet, but there's also `llvm-ar` wrapper, which should make fixing any issues straightforward if they arise.

**Performed local tests**:
 - [x] `lint` passing
 - [x] `tests all` passing
 - [x] `tests bench` passing
 - [ ] `tests cbmc` passing

### DIFF
--- a/mk/rules.mk
+++ b/mk/rules.mk
@@ -20,6 +20,26 @@ $(BUILD_DIR)/%.a: $(CONFIG)
 	$(Q)rm -f $@
 	$(Q)$(CC_AR) rcs $@ $(filter %.o,$^)
 
+# NOTE:
+# 	$AR doesn't care about duplicated symbols, one can only find it out via actually linking.
+# 	The easiest one to do this that one can think of is to create a dummy C file with empty main function on the fly, pipe it to $CC and link with the built library
+	$(eval _LIB := $(subst $(BUILD_DIR)/lib,,$(@:%.a=%)))
+ifneq ($(findstring clang,$(shell $(CC) --version)),) # if CC is clang
+	$(Q)echo "int main() {return 0;}" \
+		| $(CC) -x c - -L$(BUILD_DIR) \
+		 -all_load -Wl,-undefined,dynamic_lookup -l$(_LIB) \
+		 -Imlkem $(wildcard test/notrandombytes/*.c)
+	$(Q)rm -f a.out
+else                                                  # if CC is not clang
+	$(Q)echo "int main() {return 0;}" \
+		| $(CC) -x c - -L$(BUILD_DIR) \
+		-Wl,--whole-archive,--unresolved-symbols=ignore-in-object-files -l$(_LIB) \
+		-Wl,--no-whole-archive \
+		-Imlkem $(wildcard test/notrandombytes/*.c)
+	$(Q)rm -f a.out
+endif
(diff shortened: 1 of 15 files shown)


mk/rules.mk
```diff
@@ -20,6 +20,24 @@ $(BUILD_DIR)/%.a: $(CONFIG)
 	$(Q)rm -f $@
 	$(Q)$(CC_AR) rcs $@ $(filter %.o,$^)
 
+# NOTE: Checking if there's duplicated symbols by actually linking
```
hanno-becker, line 23: This piece of Makefile is a bit crypto -- can you expand on what is happening here?

hanno-becker (changes requested):
@potsrevennil Thank you, this looks great!

I have one not-so-technical concern: In my mind, `libmlkemXXX.a` and `libfips202.a` are test-artifacts only, and _not_ meant to be used by consumers of mlkem-native. Their names, however, suggest the latter.

I think we should:
1. For now, give those libraries a name that makes it umambiguous that they're temporary -- such as `tmp_xxx.a`.
2. As a follow-up, build `libmlkem512.a, libmlkem768.a, libmlkem1024.a` and `libmlkem.a` as the combination of those temporary libs. I think that we should not offer a separate `libfips202.a`, since we're really only providing what we need for ML-KEM, and if someone wants to drop in their own FIPS-202 implementation, they have to re-build anyway (because the context structures may be different).

So, the only concrete change request I have is to rename the libs as `tmp_xxx.a`, and put them in some `tmp` directory where it's clear that they're not meant to be used.

mk/rules.mk
potsrevennil: So after discuss with @mkannwischer, I think we want to make sure that we properly namespace everything, which means there's no duplicated symbols in the built library.

However as I mentioned in the PR description, we cannot check this with `$AR`, we need to actually use it to find out if there's duplicated symbol.

Therefore the easiest way I can think of is to create a dummy c file on the fly, pipe it into `$CC` and link with the built library for checking the duplicated symbol.

Btw, We can have a separate make target for doing this. I added this here is because I thought that it'd be nice if the build system can complain about duplicated symbols immediately.

hanno-becker:
Oh, wow. This was not an issue with this PR, but with an old `mk/acvp.mk` that I had still lying around, and that was triggered by the line `-include mk/$(MAKECMDGOALS).mk` in the `Makefile`.

hanno-becker (approved):
Thanks @potsrevennil, LGTM

PR #511, opened 2024-12-12
reduce.c: Simplify bounds documentation and fix runtime assertion

hanno-becker:
The previous bounds documentation in reduce.h was difficult to follow
and lacked details regarding the transition between rational, integral,
and C computations.

This commit simplifies the documentation by providing a single generic
bound for `montgomery_reduce_generic()`, and deriving (only) the contractual
bounds for the specializations `montgomery_reduce()` and `fqmul()`.


mkannwischer (approved):
Thanks for fixing the bound and refining the comments.

PR #512, opened 2024-12-12
Inline reduce.c into reduce.h

hanno-becker:
Those functions are performance critical, and would otherwise only inlined into their call-sites in `ntt.c` and `poly.c` if LTO is used.


mkannwischer (approved):
Awesome! Thank you @hanno-becker. 

Thanks for addressing my comments, too.

PR #517, opened 2024-12-12
FIPS202: Remove redundant zeroization in shake128_absorb_once()

hanno-becker:
`shake128_absorb_once()` calls `keccak_absorb_once()`, which already includes zeroization of the state. The zeroization can therefore be dropped from `shake128_absorb_once()`.


mkannwischer (approved):
LGTM.
Sorry for missing that.

PR #522, opened 2024-12-13
CBMC: Add custom reordering loop into CBMC scope

### LINKED ISSUE #470: Include custom permutation logic into scope of CBMC
`indcpa.c` contains the following snippet at the end of matrix generation:
```c
#if defined(MLKEM_USE_NATIVE_NTT_CUSTOM_ORDER)
  // The public matrix is generated in NTT domain. If the native backend
  // uses a custom order in NTT domain, permute A accordingly.
  for (i = 0; i < MLKEM_K; i++)
  {
    for (int j = 0; j < MLKEM_K; j++)
    {
      poly_permute_bitrev_to_custom(&a[i].vec[j]);
    }
  }
#endif /* MLKEM_USE_NATIVE_NTT_CUSTOM_ORDER */
```

This is currently out of scope for CBMC.

Make the double-loop unconditional, and provide a dummy implementation of `poly_permute_bitrev_to_custom` if `MLKEM_USE_NATIVE_NTT_CUSTOM_ORDER` is not defined. Any reasonable compiler will then drop the entire double loop. `poly_permute_bitrev_to_custom` should also have a contract specified saying that it does not change the bounds (we don't need more than that, for now).

hanno-becker:
* Resolves #470 

`gen_matrix()` concludes with an entry-wise permutation of polynomial coefficients, if required by the native backend.

Previously, the loop over the matrix coefficients was conditional on a custom permutation being present, which would exclude it from the scope of CBMC.

With this commit, the loop is unconditional, and a contract used for the custom permutation. If no custom permutation is needed, it is defined as a no-op which the compiler will automate away.

### DIFF
--- a/cbmc/proofs/gen_matrix/Makefile
+++ b/cbmc/proofs/gen_matrix/Makefile
@@ -13,13 +13,13 @@ DEFINES +=
 INCLUDES +=
 
 REMOVE_FUNCTION_BODY +=
-UNWINDSET += $(MLKEM_NAMESPACE)gen_matrix.0:4 $(MLKEM_NAMESPACE)gen_matrix.1:4 $(MLKEM_NAMESPACE)gen_matrix.2:4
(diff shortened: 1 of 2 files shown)


mkannwischer (approved):
I am not quite sure if this is adding much value, but stating the pre- and post-conditions clearly is probably a good idea.

hanno-becker:
@mkannwischer We want to say that CBMC covers the whole C frontend. This gets tricky is the frontend has conditional code that's only verified when a backend is present.

PR #524, opened 2024-12-13
Build libraries for individual and all security levels

hanno-becker:
This PR changes the build so that we generate four libraries:
* `libmlkem.a`
* `libmlkem{512,768,1024}.a`

All of those libraries include the FIPS functionality, which is no longer built as a separate `.a` file.


potsrevennil (approved):

mkannwischer (approved):
LGTM.

PR #529, opened 2024-12-14
Add examples for how to use mlkem-native as a code-package

hanno-becker:
This PR adds `examples/mlkem_native_as_code_package` and `examples/bring_your_own_fips202` as minimal examples for how to use mlkem-native as-is (with fips202 and all backends) as a code package, or with a custom FIPS-202 implementation.


mkannwischer (approved):
LGTM. Thanks @hanno-becker.
CI is unhappy about the symlinks. Once that is fixed, we can merge this.

mkannwischer (approved):
Nice, thanks for adding the second example. 
Works for me out of the box and produces the same ss as the first example.

PR #535, opened 2024-12-16
Introduce global config and reorganize backends

hanno-becker:
This commit introduces a global configuration file `mlkem/config.h`
which should contain all user-configurable parameters. With this
commit, it contains:

- MLKEM_K
- MLKEM_NAMESPACE
- FIPS202_NAMESPACE
- MLKEM_USE_NATIVE
- MLKEM_NATIVE_ARITH_BACKEND
- MLKEM_NATIVE_FIPS202_BACKEND

The backends have been reorganized to follow a simpler file structure:

Every backend profile is identified by metadata file in the toplevel
directory of the backend. For example, `aarch64` has `opt.h` and `clean.h`.
Those metadata files so far only set the name of the backend, and point
to the actual implementation. The reason why the metadata file and the
implementation are kept separate is so that assembly files can include
the metadata file and know if they should be assembled: For example,
`aarch64/opt.h` sets `MLKEM_NATIVE_ARITH_BACKEND_AARCH64_OPT` which all
relevant files are guarded by; similar for clean. Previously, they were
all guarded more coarsely by `MLKEM_USE_NATIVE_AARCH64` or
`MLKEM_USE_NATIVE_X86_64` -- those have been removed.

The source code of the backends has been moved into `src` directories.

Ultimately, we may want to split `aarch64` into `aarch64_opt` and `aarch64_clean`,
so the distinction between profile and backend goes away, but this is not yet
attempted.

====

An example in `examples/` is added which demonstrates how to use a custom
backend and custom config.


.github/workflows/bench.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           name: Arm Cortex-A55 (Snapdragon 888) benchmarks
           bench_pmu: PERF
           archflags: "-mcpu=cortex-a55 -march=armv8.2-a"
-          cflags: "-flto -static -DFORCE_AARCH64 -DFIPS202_NATIVE_PROFILE=\"aarch64/profiles/cortex_a55.h\""
+          cflags: "-flto -static -DFORCE_AARCH64 -DMLKEM_NATIVE_FIPS202_BACKEND=\\\\\\\"aarch64/cortex_a55.h\\\\\\\""
```
hanno-becker, line 46: Not sure if I should cry or laugh. The hour of trial-and-error before I got this right certainly made me want to cry.

hanno-becker: This is so that `MLKEM_NATIVE_FIPS202_BACKEND` evaluates to a _string_ in C, and we can set `#define MLKEM_NATIVE_FIPS202_BACKEND "foo.h"` in the `config.h`. Previously, `FIPS202_NATIVE_PROFILE` played the role of `MLKEM_NATIVE_FIPS202_BACKEND` but was stringified within C, so one had to write `#define MLKEM_NATIVE_FIPS202_BACKEND foo.h`, which arguably is unnatural.

mkannwischer (approved):
LGTM. Thanks Hanno.

PR #538, opened 2024-12-17
README: Fix links after renaming of backend API

hanno-becker:


mkannwischer (changes requested):
Thanks for catching this!

There are more broken links, could you please fix them all?
Example: https://github.com/pq-code-package/mlkem-native/blob/f2d9f79bea4ab00a93a63e7af0d23d1a66316550/mlkem/native/aarch64/ntt_clean.S

hanno-becker:
I'll do this as part of https://github.com/pq-code-package/mlkem-native/pull/539

PR #542, opened 2024-12-17
[DRAFT] AArch64: Use transposed coefficient order in NTT domain

hanno-becker:
This PR is an experiment for using a transposed order of coefficients in NTT domain in the AArch64 backend.

The motivation to use such an ordering is the reduced shuffling cost in invNTT and NTT. The downside is that shuffling is introduced during key generation and serialization of polynomials.

At this point, there are 'clean' AArch64 assembly implementations for (a) `poly_tobytes`, (b) `poly_frombytes`, (c) `poly_transpose`. Initial benchmarks suggest that at least `poly_{to,from}bytes()` in clean AArch64 assembly has no performance benefit compared to C, so we may want to remove them. SLOTHY-optimizing them requires adding support for further instructions to SLOTHY first.


mkannwischer (commented):
This does not result in better performance on _any_ uArch. Why would you like to add this?

hanno-becker:
> This does not result in better performance on any uArch. Why would you like to add this?

@mkannwischer This is part of an exploration with @jargh on whether it's useful to use the transposed NTT order for AArch64 (as we do for x86_64). However, since `poly_{to,from}bytes()` work in NTT domain, they need adjusting, and therefore we currently require that native implementations be present.

hanno-becker:
@mkannwischer Looks like removing the existing (clean, admittedly) `poly_tobytes()` also has no bearing.

hanno-becker:
Agreed with @jargh that we are not pursuing this for now.

PR #547, opened 2024-12-18
AArch64: Remove ASM_LOAD directives 

### LINKED ISSUE #546: Remove all uses of the ASM_LOAD assembly macro
We currently load global symbol addresses in assembly using 
```C
#if __APPLE__
#define ASM_LOAD(dst, symbol)                                                  \
  adrp dst, symbol @PAGE %% add dst, dst, symbol @PAGEOFF
#else /* __APPLE__ */
#define ASM_LOAD(dst, symbol)                                                  \
  adrp dst, symbol;                                                            \
  add dst, dst, : lo12 : symbol;
#endif /* __APPLE__ */
```

Where possible, we should avoid this, and either use PC-relative addressing, or pass the target address directly as a parameter.

hanno-becker:
The ASM_LOAD macro loads the address of a symbol into a GPR. It uses platform-specific assembly, and is thus cumbersome to work with. It is also unnecessarily complicated if the goal is a load from a literal pool: In this case, one can use a PC-relative load directly, rather than using ASM_LOAD to load the address of the pool in a register first, and then load the actual data. Where ASM_LOAD(...) is used to establish a pointer to a global symbol, one can pass the address of that symbol from C instead.

This PR removes all uses of ASM_LOAD in this way.

**Please review commit by commit.**

* Resolves #546


mkannwischer (approved):
Thanks! loooks mostly good to me! Great that you auto generate these weird constant tables for the rejection sampling now. 

Small nits: 
 - [dd25efbda65c55c05088363ad4f3b9119955e5a0](https://github.com/pq-code-package/mlkem-native/pull/547/commits/dd25efbda65c55c05088363ad4f3b9119955e5a0) fixes a buf that was intorduced in the commit before (renaming dup8u to dup8h)
 - the `.req` is not aligned in keccak_f1600_x2_v84a_asm_clean.S
 - "morally" is maybe the wrong word here - but I am not sure about this one.

PR #549, opened 2024-12-19
Remove double-namespacing for assembly symbols

### LINKED ISSUE #545: Don't generate two symbols for every assembly function
We currently generate global symbols `_funcname` and `funcname` for every assembly function. This works, but is a bit untidy. We could instead have a different namespace macro for assembly files which, depending on the target platform, puts a `_` in front of the symbol name or not.

hanno-becker:
* Resolves #545 

We previously exported two versions of every assembly symbol, one prefixed with a leading underscore, one not. This was to accommodate both Apple and non-Apple platforms.

This commit removes duplicated assembly symbols by introducing separate namespace wrappers
- MLKEM_ASM_NAMESPACE, and
- FIPS202_ASM_NAMESPACE which wrap around the configurable {MLKEM,FIPS202}_NAMESPACE by adding a leading underscore for Apple systems (only).

### DIFF
--- a/examples/custom_backend/mlkem_native/custom_config.h
+++ b/examples/custom_backend/mlkem_native/custom_config.h
@@ -44,7 +44,6 @@
 #define CONC(a, b) __CONC(a, b)
 
 #define MLKEM_NAMESPACE(sym) CONC(CUSTOM_TINY_SHA3_, sym)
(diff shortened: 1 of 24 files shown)


mkannwischer (changes requested):
Can you also adjust the `custom_backend` example please? 
That one still has `_MLKEM_NAMESPACE` and `_FIPS202_NAMESPACE`

mkannwischer (approved):
Thanks. LGTM now.

PR #562, opened 2024-12-21
CI: Add Docker-based compatibility tests, add tests to `make quickcheck`

### LINKED ISSUE #560: Make test scripts usable in older versions of Python
The Python scripts in `scripts` require at least Python 3.10, and hence don't work out of the box on systems with older Python versions. 

The test lib should be rewritten to not use any newer Python features -- this should be as plain as possible to ensure maximum usability. 

It's TBD what version of Python we should require.

hanno-becker:
This PR adds Docker-based compatibility tests for various Ubuntu-20, Ubuntu-22, Debian, and Amazon Linux docker images; so far, AArch64 only.

A number of changes had to be made, esp. to the test scripts; see commits, paving the way for the removal of `tests`.

* Fix #560


mkannwischer (approved):
Thanks @hanno-becker - very nice step towards better compatibility. We should really eliminate the tests script in the near future.

I don't think it's a good idea to remove `-O3` from the default flags as someone may benchmark mlkem-native with default flags and draw wrong conclusions about performance. I instead changed it to explicitlity setting `-O0` for some of the CI jobs.
You may want to add that to more jobs. 

Rest looks very good to me if the goal is to eliminatethe test script soon, but then we should really finish it soon.

PR #563, opened 2024-12-21
CI: Add Docker-based compatibility tests 

mkannwischer:
This PR adds Docker-based compatibility tests for various Ubuntu-20 and Ubuntu-22  docker images; so far, AArch64 only.

A number of changes had to be made, esp. to the test scripts; see commits.

Fix https://github.com/pq-code-package/mlkem-native/issues/560 

(Hoisted out from #562)

### DIFF
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -13,11 +13,10 @@ CC  ?= gcc
 CPP ?= cpp
 AR  ?= ar
 # NOTE: gcc-ar is a wrapper around ar that ensures proper integration with GCC plugins,
-# 		such as lto. Using gcc-ar is preferred when creating or linking static libraries
-# 		if the binary is compiled with -flto.
-# 		However, this doesn't apply to darwin as it is using clang instead, and there's no
-# 		gcc-ar wrapper as well.
-CC_AR ?= $(if $(findstring Darwin,$(shell uname -s)),ar,gcc-ar)
+# 	such as lto. Using gcc-ar is preferred when creating or linking static libraries
+# 	if the binary is compiled with -flto. However, it is not universally present, so
+#       only use it if available.
+CC_AR ?= $(if $(shell which gcc-ar),gcc-ar,ar)
 
 CC  := $(CROSS_PREFIX)$(CC)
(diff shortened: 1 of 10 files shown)


scripts/lib/mlkem_test.py
```diff
@@ -355,7 +408,8 @@ def _func(opt: bool):
             exit(1)
 
     def _run_nistkat(self, opt: bool):
-        def expect_proc(scheme: SCHEME, actual: str) -> tuple[bool, str]:
+        def expect_proc(scheme: SCHEME, actual: str):
```
mkannwischer, line 411: In other places you removed the `:str` type information. 
Should that be removed in all places?

mk/config.mk
```diff
@@ -13,11 +13,10 @@ CC  ?= gcc
...
 # NOTE: gcc-ar is a wrapper around ar that ensures proper integration with GCC plugins,
-# 		such as lto. Using gcc-ar is preferred when creating or linking static libraries
-# 		if the binary is compiled with -flto.
-# 		However, this doesn't apply to darwin as it is using clang instead, and there's no
-# 		gcc-ar wrapper as well.
-CC_AR ?= $(if $(findstring Darwin,$(shell uname -s)),ar,gcc-ar)
+# 	such as lto. Using gcc-ar is preferred when creating or linking static libraries
+# 	if the binary is compiled with -flto. However, it is not universally present, so
+#       only use it if available.
+CC_AR ?= $(if $(shell which gcc-ar),gcc-ar,ar)
```
mkannwischer, line 19: Doesn't this break if you are using Darwin with gcc-ar installed, but you compile with clang?
I'm not very sure about this.

hanno-becker: I tried to fix this by looking at `cc --version` to see if it's clang or not, and if it's gcc, check for gcc-ar.

scripts/lib/mlkem_test.py
hanno-becker: Fixed (hopefully)

mkannwischer:
Included in #562

PR #566, opened 2024-12-22
Tests/CI: Reconcile `cflags` and `extraflags` arguments

hanno-becker:


mkannwischer (approved):
LGTM. 
This is much cleaner. Thanks @hanno-becker!

Issue #567, opened 2024-12-22
Rewrite `functest` workflow to use `make` only


hanno-becker:
- [x] The `functest` action is based on `make check_func / check_kat / check_nistkat / check_acvp`, not `tests`

I think this should be largely straightforward, but we need to add a makefile parameter for the exec prefix and set that to the right QEMU binary for cross compilation. The command line would then look something like `make CROSS_PREFIX=... EXEC_PREFIX=... checkall`. This logic is currently in `tests`.

One would also need to explicitly resolve `opt=all` in the workflow, calling `make` with `OPT=0` and `OPT=1`.

hanno-becker:
On second thought, I think the `tests` script is convenience worth having (however, it should be a shallow wrapper around `make`), so closing this for now.

PR #571, opened 2024-12-22
Test: Use `make` instead of `tests` in `functest` action

hanno-becker:
* Fix https://github.com/pq-code-package/mlkem-native/issues/567

This PR replaces calls to `tests` in the `functest` CI action by direct invocations of `make`. There is a bit of explicit repetition, which we may want to remove: IIRC, we always want to run all tests, so the splitting into `func/acvp/nistkat/kat` is redundant, and we could as well just call `make checkall`.

The PR does not yet remove the increasingly obsolete test-functionality of `tests`. Note that the benchmarking logic is more involved and may merit keeping `tests` (perhaps renamed to `bench`).

### DIFF
--- a/.github/actions/functest/action.yml
+++ b/.github/actions/functest/action.yml
@@ -25,6 +25,9 @@ inputs:
   cross_prefix:
     description: Binary prefix for cross compilation
     default: ""
+  exec_wrapper:
+    description: Binary wrapper for execution (e.g. QEMU)
+    default: ""
   opt:
     description: Whether to build opt/non-opt binaries or all (all | opt | no_opt)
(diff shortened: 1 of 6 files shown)


hanno-becker:
@mkannwischer This fails in case noopt is disabled because then you do a make clean before check namespace. I guess one should do a check namespace after opt and noopt separately, and suitably conditioned. Can you do that?

hanno-becker:
On second thought, I think the convenience of `tests` is worth having -- it _should_ be majorly simplified to turn into a rather shallow wrapper around invocations of `make`, but we likely want to keep it.

Issue #574, opened 2024-12-23
AVX2 backend not enabled by default


dkostic:
**Describe the bug**
Since this [commit](https://github.com/pq-code-package/mlkem-native/commit/b94f8ce2e47e26ae93a54f0c6a04e83ebbec6360#diff-c96841396bcc20102aa6e34c2da581c69d98ec0099f196e388f15e0b071b339a) the AVX2 backend doesn't compile by default. My platform is a c7i EC2 instance with Intel(R) Xeon(R) Platinum 8488C, running Ubuntu 22.04 with gcc 11.4.0.

**Platform**: x86_64 Linux

**Compiler**: gcc 11.4.0

**How to reproduce**
```
make clean; make -j
```
One of the easiest ways to confirm that the AVX2 backend is not enabled is to run any of the benchmarks and compare the numbers against the expected ones (https://pq-code-package.github.io/mlkem-native/dev/bench/). Otherwise, you can insert `#error fail_the_build_if_this_is_compiled` for example [here](https://github.com/pq-code-package/mlkem-native/blob/2a6999efbacaa1cfca6eff6fb947422dce954236/mlkem/native/x86_64/src/rej_uniform_avx2.c#L14) and  try to build the library which will succeed, meaning that the AVX2 backend is not compiled.

**Expected behavior**

**Additional context**

hanno-becker:
Thank you @dkostic for reporting this. We should have caught this in CI. Can you confirm whether the build fails with `FORCE_X86_64`?

hanno-becker:
@dkostic can you confirm that #578 fixed the issue for you?

dkostic:
Thanks for quick reaction! The fix in https://github.com/pq-code-package/mlkem-native/pull/578 works on my machine as well.

PR #575, opened 2024-12-23
Fix the AVX2 backend build

dkostic:
**Summary**:
For reasons that elude me, this fixed issue #574 on my machine.

**Steps**:

**Performed local tests**:
 - [ ] `lint` passing
 - [x] `tests all` passing
 - [ ] `tests bench` passing
 - [ ] `tests cbmc` passing

### DIFF
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -29,7 +29,7 @@ CC_AR  := $(CROSS_PREFIX)$(CC_AR)
 #################
 # Common config #
 #################
-CFLAGS := \
+CFLAGS += \
 	$(ARCH_FLAGS) \
 	-Wall \
 	-Wextra \
@@ -45,8 +45,7 @@ CFLAGS := \
 	-fomit-frame-pointer \
 	-std=c99 \
 	-pedantic \
-	-MMD \
-	$(CFLAGS)
+	-MMD
 
 ##################
 # Some Variables #


mk/config.mk
```diff
@@ -29,7 +29,7 @@ CC_AR  := $(CROSS_PREFIX)$(CC_AR)
 #################
 # Common config #
 #################
-CFLAGS := \
+CFLAGS += \
```
hanno-becker, line 32: Unfortunately we can't do that without reverting further changes: We want user specified CFLAGS do overwrite default CFLAGS, but with this version it's the other way around

hanno-becker:
The issue is lazy vs. eager evaluation of ARCH_FLAGS within CFLAGS: with := the variable is eagerly evaluated at time of definition, while with += it is lazily evaluated. This matters since auto.mk sets default ARCH_FLAGS only later.

I see two options:
1. Just move the inclusion of auto.mk upwards before the setting of CFLAGS. @dkostic can you try that please?
2. Make a copy of the original CFLAGS as (say) `ORIG_CFLAGS := $CFLAGS` and switch back to lazy evaluation for CFLAGS, but using ORIG_CFLAGS in the end.

If it works, 1. would be simpler. It depends on whether there are other places where ARCH_FLAGS is adjusted.

hanno-becker (changes requested):
Thanks a lot @dkostic for spotting this. `make` apparently eludes me, too.

See the comment for what happened. For the fix we can't use the proposed change as we want to give the original CFLAGS precedence. Can you try including auto.mk earlier? Could you please also adjust the CI tests to pass `FORCE_AARCH64` and `FORCE_X86_64` accordingly, so this would have been caught?

Before we can consider merging, we need to move this PR to a non-remote branch -- this is because of some peculiarities about our EC2 CI still... (sorry!) @mkannwischer can you either give @dkostic developer rights so he can push branches, or otherwise run CI yourself when things look final?

mkannwischer:
> Before we can consider merging, we need to move this PR to a non-remote branch -- this is because of some peculiarities about our EC2 CI still... (sorry!) @mkannwischer can you either give @dkostic developer rights so he can push branches, or otherwise run CI yourself when things look final?

Sorting out access during the holiday's is likely going to be impossible as it need approval in the TSC repository. Happy to merge this into a local branch first to run the full CI.

mkannwischer:
> The issue is lazy vs. eager evaluation of ARCH_FLAGS within CFLAGS: with := the variable is eagerly evaluated at time of definition, while with += it is lazily evaluated. This matters since auto.mk sets default ARCH_FLAGS only later based on your system.
> 
> Some options:
> 
> 1. Just move the inclusion of auto.mk upwards before the setting of CFLAGS.
> 2. Modify CFLAGS, not ARCH_FLAGS, in auto.mk. We still need to keep ARCH_FLAGS in the CFLAGS defaults though because some test scripts set it, too.
> 3. Make a copy of the original CFLAGS as (say) `ORIG_CFLAGS := $CFLAGS` and switch back to lazy evaluation for CFLAGS, but using ORIG_CFLAGS in the end.
> 
> If it works, 2. would be simplest. Can you try @dkostic ?

I implemented (2) in #578 - let's get that merged quick and discuss any further improvment separatelty.

hanno-becker:
Thanks @mkannwischer!

With this merged, how should this be detected in the future? Either of (a) not setting arch flags in CI, but relying on defaults, or (b) passing FORCE_xxx would have caught this -- we should probably do both. Note though that even with (a), what we benchmark is not the default build, because the latter disables LTO.

mkannwischer:
Superseded by #578

mkannwischer:
> Thanks @mkannwischer!
> 
> 
> 
> With this merged, how should this be detected in the future? Either of (a) not setting arch flags in CI, but relying on defaults, or (b) passing FORCE_xxx would have caught this -- we should probably do both. Note though that even with (a), what we benchmark is not the default build, because the latter disables LTO.

Let's do both.

Issue #477, opened 2024-12-03
AVX2: Use mulcache for base multiplication


hanno-becker:
The AVX2 backend should use a mulcache like the C and AArch64 backends. This essentially means moving https://github.com/pq-code-package/mlkem-native/blob/7294554d0c8b1f868716c3e88e75bf74bf649f42/mlkem/native/x86_64/basemul.S#L90-L105 into a new `poly_mulcache_compute_native` routine.

hanno-becker:
Done via #996

PR #579, opened 2024-12-24
AVX2 implementation of mulcache

dkostic:
* Fixes https://github.com/pq-code-package/mlkem-native/issues/477

**Summary**:
AVX2 implementation of mulcache

Slight performance improvement:
```
|--------------------------------------|
|  MLKEM 512 | Before | After | Improv |
|    keypair |  15101 | 14835 |  1.02x |
|     encaps |  19664 | 18631 |  1.05x |
|     decaps |  25824 | 24420 |  1.05x |
|--------------------------------------|
|  MLKEM 768 | Before | After | Improv |
|    keypair |  26187 | 25157 |  1.04x |
|     encaps |  28014 | 27248 |  1.03x |
|     decaps |  36989 | 35659 |  1.03x |
|--------------------------------------|
| MLKEM 1024 | Before | After | Improv |
|    keypair |  36014 | 35630 |  1.01x |
|     encaps |  39797 | 39347 |  1.00x |
|     decaps |  52139 | 51524 |  1.01x |
|--------------------------------------|
```
measured on Intel(R) Xeon(R) Platinum 8488C.

Addresses #477 .

**Steps**:

**Performed local tests**:
 - [ ] `lint` passing
 - [x] `tests all` passing
 - [ ] `tests bench` passing
 - [ ] `tests cbmc` passing

**Do you expect this change to impact performance**: Yes, see above.


mlkem/native/x86_64/src/basemul.c
```diff
@@ -8,15 +8,57 @@
...
 
 #include "consts.h"
+
 #include "poly.h"
 #include "polyvec.h"
 
 #include "arith_native_x86_64.h"
 
-static void poly_basemul_montgomery_avx2(poly *r, const poly *a, const poly *b)
+int16_t zetas_avx2[64] = {
```
hanno-becker, line 17: Can we autogenerate those constants via `autogenerate_files.py`?

mlkem/native/x86_64/src/basemul.S
```diff
@@ -52,14 +50,20 @@ vmovdqa		%ymm13,(%rsp)
...
 
 /* Compute low-parts of monomials in (a1+b1*X)*(c1+d1*X), */
 /* using Montgomery twists calculated before */
 vpmullw		%ymm7,%ymm11,%ymm6			# a1c1.lo
 vpmullw		%ymm8,%ymm11,%ymm11			# a1d1.lo
 vpmullw		%ymm7,%ymm12,%ymm7			# b1c1.lo
-vpmullw		%ymm8,%ymm12,%ymm12			# b1d1.lo
+
+/* Use cached (d * zeta) to compute the high part of (b*d) terms */
+vmovdqa		(32*\off+ 0)*2(%r8),%ymm8	# d0z
```
hanno-becker, line 61: Now that we use the mulcache, can we accumulate in 32bit and reduce once in the end? That's how it's done in C and AArch64 backend, and the mulcache enables it (otherwise there are two chained muls anyway).

mlkem/native/x86_64/src/arith_native_x86_64.h
```diff
@@ -37,7 +37,11 @@ void reduce_avx2(__m256i *r, const __m256i *qdata);
 
 #define basemul_avx2 MLKEM_NAMESPACE(basemul_avx2)
 void basemul_avx2(__m256i *r, const __m256i *a, const __m256i *b,
-                  const __m256i *qdata);
+                  const __m256i *qdata, const __m256i *bcache);
+
+#define poly_mulcache_compute_avx2 \
+  MLKEM_NAMESPACE(poly_mulcache_compute_avx2)
```
hanno-becker, line 43: @dkostic can you run `format` to appease the linter? You just need to install nix and run 
```
nix develop --experimental-features 'nix-command flakes'
```
Then `format` and `lint` are in your PATH and configured in the same way as in CI.

hanno-becker (changes requested):
Thanks a lot @dkostic for working that out! Up to 5% performance uplift is worth taking, esp. since it simplifies the code and makes it more uniform across backends.

Left some small comments. The only non trivial question is whether AVX2 allows to accumulate in 32bit, in which case we can defer the reduction to after accumulation, in line with the C and AArch64 backends, also further improving performance.

mlkem/native/x86_64/src/arith_native_x86_64.h
dkostic: I tried to do that but I get the following error:
```
➜  mlkem-native sudo nix develop --experimental-features 'nix-command flakes'

error: flake 'path:/home/ubuntu/workspace/mlkem-native' does not provide attribute 'devShells.x86_64-linux.devShell.x86_64-linux', 'packages.x86_64-linux.devShell.x86_64-linux', 'legacyPackages.x86_64-linux.devShell.x86_64-linux', 'devShell.x86_64-linux' or 'defaultPackage.x86_64-linux'
```

mlkem/native/x86_64/src/basemul.S
dkostic: I thought about this before opening the PR and I don't think that can be easily done with AVX2 because if you want to accumulate in 32bit then the multiplication also has to be in 32bit, so we'll basically have to load the coefficients into ymm's as 32-bit instead of 16-bit integers and work like that. It's definitely worth exploring so that's going to be my next mini project :)

mlkem/native/x86_64/src/basemul.c
dkostic: I tried to do that, but the script doesn't work on my machine (even without any changes to it). I get this error:
```
➜  mlkem-native ./scripts/autogenerate_files.py
Failed to auto-format autogenerated code (clang-format return code 1
```
even though I have `clang-format` installed and in the PATH.

mlkem/native/x86_64/src/arith_native_x86_64.h
hanno-becker: @potsrevennil does that ring a bell?

mlkem/native/x86_64/src/basemul.S
hanno-becker: Can one not load in 16bit, use separate hi/login bit multiply, zip to 32bit, accumulate in 32bit, and reduced? It certainly has more Overhead than AArch64 where one has a widening multiply to begin with. But perhaps pairs of lo/hi multiply are anyway fused? I don't know Intel

mlkem/native/x86_64/src/basemul.c
hanno-becker: @dkostic This is odd. We do test on a host of machines, but your's seems to hit a gap. 

Can you open issues for this and for your nix setup issue please, describing the exact setup and steps you took, and the machine you are working with?

mlkem/native/x86_64/src/arith_native_x86_64.h
hanno-becker: @dkostic Which version of nix are you running?

potsrevennil: I'd also suspect it's due to the nix version, maybe we should also add to the document about nix version > 2.18 ?

potsrevennil: I was wrong, I did some tests with nix 2.17.2 docker image, and everything works as usual. 
Maybe it's not related to nix version, or maybe we could support nix version > 2.17 (or even older).

hanno-becker: Thanks @potsrevennil !

@dkostic how did you install nix?

mlkem/native/x86_64/src/basemul.S
dkostic: yes, but that's the same as I described with the order of operations a bit different. My suggestion is to load the coefficients in 16bit and then zip to 32bit with a zero vector, then single mul and reduce. IIUC you are suggesting to change the order of mul and zip (and do mullo/mulhi). If we do it my way then we don't have to worry about pairs of mul lo/hi being fused or not.

mlkem/native/x86_64/src/basemul.c
dkostic: sure, I can do that

mlkem/native/x86_64/src/basemul.S
hanno-becker: Yes but a 32bit mul may be more expensive than 2x16bit muls. We'll see!

mlkem/native/x86_64/src/arith_native_x86_64.h
dkostic: `sudo apt-get install nix`

mlkem/native/x86_64/src/basemul.c
dkostic: opened #581 and #582

hanno-becker: Thanks! Looks like version issues, we need to document the requirements better.

For this PR, I suggest you install nix through the official script, which should unblock you.

mlkem/native/x86_64/src/basemul.S
dkostic: I don't think so, according to https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html both have latency of 5 cycles and throughput of 0.5 CPI on Intel architectures. Anyway, I can test both and see which one is better performing. But I'd rather do that in a new PR if that's ok with you?

mlkem/native/x86_64/src/basemul.c
dkostic: you mean this: https://nixos.org/download/? I did that, but the issue persist.

mlkem/native/x86_64/src/basemul.S
hanno-becker: Yes I think that's better - this PR is already an improvement and a good first step, and soon ready to go once we get past the tooling issues.

mlkem/native/x86_64/src/basemul.c
hanno-becker: Did you uninstall the apt version and double check `nix --version`? Can you update the issue accordingly?

It would be puzzling if not even the setup script worked, we test this in CI for c7i (see the nix-setup action for what exactly is run).

hanno-becker: Any update @dkostic?

dkostic: hey Hanno, I updated the PR. After successfully installing `nix` I could `format` and `lint` and also the autogenerate script. Let me know if you have any other feedback!

examples/custom_backend/mlkem_native/mlkem/poly.c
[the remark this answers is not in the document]
hanno-becker: The CI is unhappy with the sign-off in this commit

mlkem/native/x86_64/src/basemul.c
```diff
@@ -8,15 +8,55 @@
...
+#define QINV (-3327) /* q^-1 mod 2^16 */
+
+void poly_mulcache_compute_avx2(poly_mulcache *x, const poly *y)
+{
+  __m256i q, qinv, a0, a1, z, t0, t1, s0, s1, r0, r1;
+
+  q = _mm256_set1_epi16(MLKEM_Q);
+  qinv = _mm256_set1_epi16(QINV);
+
+  for (int j = 0; j < 4; j++)
```
hanno-becker, line 30: This fails the C90 build -- can you move the declaration of `int j` to the top of the function?

dkostic: fixed

hanno-becker:
@dkostic There is a segfault in the `clang-18` compiler test. Can you have a look?

examples/custom_backend/mlkem_native/mlkem/poly.c
```diff
@@ -1 +0,0 @@
-../../../../mlkem/poly.c
```
hanno-becker, line 1: I think this symlink should have been kept?

mlkem/native/x86_64/src/basemul.c
```diff
@@ -8,15 +8,56 @@
...
 #include "consts.h"
+
 #include "poly.h"
 #include "polyvec.h"
 
 #include "arith_native_x86_64.h"
 
-static void poly_basemul_montgomery_avx2(poly *r, const poly *a, const poly *b)
+static const int16_t zetas_avx2[64] = {
+#include "x86_64_zetas_mulcache.i"
```
hanno-becker, line 18: Nit: Unless there's a good reason, I'd prefer the autogenerated files to be `.h` files. The other autogenerated AVX2 file isn't because it's included in the middle of an array, but here you could just generate the whole array in `x64_64_zetas_mulcache.h`.

mlkem/native/x86_64/src/basemul.c
```diff
@@ -8,15 +8,56 @@
...
+};
+
+#define QINV (-3327) /* q^-1 mod 2^16 */
+
+void poly_mulcache_compute_avx2(poly_mulcache *x, const poly *y)
+{
+  int j;
```
hanno-becker, line 25: Since we aim to use assembly-level verification using HOL-Light, could we implement this in ASM rather than intrinsics, please? ISTM that this should essentilly be the code following`TODO: This could be precomputed in the mulcache` in (the old) `basemul.S`.

examples/custom_backend/mlkem_native/mlkem/poly.c
dkostic: done, must have been a mistake I did during a rebase.

mlkem/native/x86_64/src/basemul.c
dkostic: done

mlkem/native/x86_64/src/basemul.c
dkostic: sure, you want me to do it as part of this PR?

hanno-becker (changes requested):
The CI is still failing in a few places. Once https://github.com/pq-code-package/tsc/pull/125 is merged, you will be able to open a PR from the main repo, and run the full CI yourself.

mlkem/native/x86_64/src/basemul.c
hanno-becker: Preferably, yes.

examples/custom_backend/mlkem_native/mlkem/poly.c
hanno-becker: This was a bug in the formatter, now fixed (#614)

hanno-becker:
@dkostic Ok this is done now. You have developer rights, so should be able to push branches to the main repo and open PRs from there. Can you do that for this PR?

mlkem/native/x86_64/src/basemul.c
dkostic: done.

dkostic:
> @dkostic Ok this is done now. You have developer rights, so should be able to push branches to the main repo and open PRs from there (closing this one). Can you do that for this PR?

I'd prefer if we can finish this PR as is. I can start pushing to the main repo with the next one.

hanno-becker:
@dkostic i'm afraid we definitely need to move the PR because the full CI does not run otherwise.

dkostic:
> @dkostic I'm afraid we definitely need to move the PR because the full CI does not run otherwise.

oh, so it was you who started the CI for my last update?

nevertheless, is this the development process for the project:
1. clone the main repo
2. create a local branch
3. do changes
4. push the local branch to the main repo
5. open a pr

hanno-becker:
> oh, so it was you who started the CI for my last update?

No I think the last update was automatic; but because of the way we authenticate to EC2 (OICD), large parts of the CI can currently only run for PRs from the main repo.

> clone the main repo
> create a local branch
> do changes
> push the local branch to the main repo
> open a pr

Yes!

dkostic:
closing in favor of #624

PR #583, opened 2024-12-27
[TEST] Run nix setup in compatibility tests

hanno-becker:
This is a test-only PR to check if the Docker based compatibility tests pass when a nix shell is used instead of the built-in shell. In particular, this tests the nix setup in those containers.


hanno-becker:
nix installer does not support being run as root, which is the default user in the docker images used in the compatibility tests.

PR #586, opened 2024-12-28
Improved error messages for dev scripts

mkannwischer:
Currently the format and autogenerate_files.py scripts fail with unuseful error messages if run outside of a nix shell.
We are currently assuming that developers use nix.
This commit adds some error messages nudging the user to install nix and install the project's dependencies through nix.


hanno-becker (approved):
LGTM

PR #588, opened 2024-12-28
Force explcit CYCLES make variable for benchmarking

mkannwischer:
Currently, when CYCLES is not set explicitly when
building the benchmarking binaries, we default to
NO which results in all 0 cycle counts which is
confusing.
This commit changes the buidl so that it fails
with a error when attempting to build the benchmarking binaries without explicitly setting CYCLES.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
-.PHONY: mlkem kat nistkat clean quickcheck buildall checkall all
+.PHONY: mlkem kat nistkat clean quickcheck buildall checkall all check-defined-CYCLES
 .DEFAULT_GOAL := buildall
 all: quickcheck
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -43,7 +43,13 @@ mlkem: \
   $(MLKEM768_DIR)/bin/test_mlkem768 \
   $(MLKEM1024_DIR)/bin/test_mlkem1024
 
-bench: \
+# Enforce setting CYCLES make variable when 
+# building benchmarking binaries
+check_defined = $(if $(value $1),, $(error $2))
+check-defined-CYCLES:
+	@:$(call check_defined,CYCLES,CYCLES undefined. Benchmarking requires setting one of NO PMU PERF M1)
```
hanno-becker, line 50: Mark this as PHONY?

hanno-becker (changes requested):
Thanks @mkannwischer BUILDING.md needs adjusting as well.

Makefile
mkannwischer: fixed

mkannwischer:
> Thanks @mkannwischer BUILDING.md needs adjusting as well, and `tests bench`?

I adjusted BUILDING.md and also made it a required argument for `tests bench`.

hanno-becker (approved):
LGTM!

PR #593, opened 2024-12-30
AVX2: Autogenerate lookup table for rejection sampling

hanno-becker:
Analogous to the corresponding table used in the AArch64 rejection sampling assembly.

* Based on #591


mkannwischer (approved):
Thanks! Yet another table of magic constants covered!

PR #596, opened 2024-12-30
Tests/CI: Move QEMU choice logic from `tests` into workflow

hanno-becker:
Previously, `tests` would automatically detect which QEMU binary to use for emulation. With this commit, this is instead made explicit in the logic of the workflow. This is a step towards simplifying the `tests` script and eventually replacing it by calls to `make`.


mkannwischer (approved):
Looks good to me! 
I stapled on some clean-up of the Makefiles - there were some unused QEMU targets/variables in there.

PR #597, opened 2024-12-31
ACVP: Remove duplicated ACVP test logic between `acvp_client.py` and `tests` script

hanno-becker:


mkannwischer (approved):
Code looks fine and I tested it on my machine. 
I double checked that this actually tests both the opt and no-opt implementations. 

Something seems wrong with CI and it does not seem to be related to this PR as far as I can tell.

PR #602, opened 2024-12-31
Makefile: Don't include HAL into libmlkem.a when building benchmarks

### LINKED ISSUE #515: Makefile: Don't try-include goalname as Makefile
The current `Makefile` includes this line which interprets the goalname as a Makefile name and tries to import it.
```
-include mk/$(MAKECMDGOALS).mk
```

In my case, this led to a build failure because I still had an old `acvp.mk` lying around which then got triggered by `make avcp` -- a valid target!.

**Task:** Rewrite the `Makefile` so that we specifically include what we need per goal, rather than try-including the goalname as a Makefile.

### LINKED ISSUE #601: `make bench` adds HAL to libmlkem.a
When running `make bench` or `make bench_components`, the chosen cycle counting HAL is added to the libmlkemX.a libraries. This should not be so, and the HAL instead be compiled alongside the benchmarking source.

hanno-becker:
Previously, `make bench` or `make bench_components` would add hal.c to SOURCES, which would make it part of the libmlkemXXX.a libraries. That should not be so, and the HAL implementation instead be compiled alongside the benchmarking source. This commit changes the Makefiles accordingly.

* Fixes #601 
* Fixes #515

### DIFF
--- a/mk/schemes.mk
+++ b/mk/schemes.mk
@@ -39,6 +39,20 @@ $(BUILD_DIR)/$(1)/bin/$(2)$(shell echo $(1) | tr -d -c 0-9): LDLIBS += -L$(BUILD
 $(BUILD_DIR)/$(1)/bin/$(2)$(shell echo $(1) | tr -d -c 0-9): $(BUILD_DIR)/$(1)/test/$(2).c.o $(BUILD_DIR)/lib$(1).a
 endef
 
+$(MLKEM512_DIR)/bin/bench_mlkem512: CFLAGS += -Itest/hal
+$(MLKEM768_DIR)/bin/bench_mlkem768: CFLAGS += -Itest/hal
+$(MLKEM1024_DIR)/bin/bench_mlkem1024: CFLAGS += -Itest/hal
+$(MLKEM512_DIR)/bin/bench_components_mlkem512: CFLAGS += -Itest/hal
+$(MLKEM768_DIR)/bin/bench_components_mlkem768: CFLAGS += -Itest/hal
+$(MLKEM1024_DIR)/bin/bench_components_mlkem1024: CFLAGS += -Itest/hal
+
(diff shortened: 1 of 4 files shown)


mk/schemes.mk
```diff
@@ -39,6 +39,20 @@ $(BUILD_DIR)/$(1)/bin/$(2)$(shell echo $(1) | tr -d -c 0-9): LDLIBS += -L$(BUILD
...
+$(MLKEM768_DIR)/bin/bench_components_mlkem768: CFLAGS += -Itest/hal
+$(MLKEM1024_DIR)/bin/bench_components_mlkem1024: CFLAGS += -Itest/hal
+
+$(MLKEM512_DIR)/bin/bench_mlkem512: $(MLKEM512_DIR)/test/hal/hal.c.o
+$(MLKEM768_DIR)/bin/bench_mlkem768: $(MLKEM768_DIR)/test/hal/hal.c.o
+$(MLKEM1024_DIR)/bin/bench_mlkem1024: $(MLKEM1024_DIR)/test/hal/hal.c.o
+$(MLKEM512_DIR)/bin/bench_components_mlkem512: $(MLKEM512_DIR)/test/hal/hal.c.o
+$(MLKEM768_DIR)/bin/bench_components_mlkem768: $(MLKEM768_DIR)/test/hal/hal.c.o
+$(MLKEM1024_DIR)/bin/bench_components_mlkem1024: $(MLKEM1024_DIR)/test/hal/hal.c.o
+
```
rod-chapman, line 55: Built "tests bench" from scratch and all ran OK. Looks good.

rod-chapman (approved):
Looks good to me. Please merge.

rod-chapman (approved):
Looks good. Please merge.

potsrevennil (approved):
This looks good to me!

PR #599, opened 2024-12-31
Makefile: Minor cleanup and introduction of level-specific targets

hanno-becker:
Based on #602 

- Use `{func, kat, nistkat, acvp, ...}` to build test binaries.
- Use `run_{func, kat, nistkat, acvp, bench, bench_components}` to run and check test binaries.
- Introduce level-specific targets
- Other minor cleanups

PR #605, opened 2025-01-01
Make `tests` a shallow wrapper around `make`

hanno-becker:
* Based on #602 
* Based on #599 
* Based on #608 

This PR simplifies `tests` to merely become a wrapper around various `make` invocations, which are printed. The only exceptions are:
* Benchmarking: Here, `tests` additionally does some postprocessing
* GH summary generation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,11 +12,11 @@
 	run_bench_512 run_bench_768 run_bench_1024 run_bench \
 	bench_components_512 bench_components_768 bench_components_1024 bench_components \
 	run_bench_components_512 run_bench_components_768 run_bench_components_1024 run_bench_components \
-	buildall checkall all \
+	build test all \
 	clean quickcheck check-defined-CYCLES
 
-.DEFAULT_GOAL := buildall
-all: quickcheck
+.DEFAULT_GOAL := build
+all: build
 
 W := $(EXEC_WRAPPER)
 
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -12,24 +12,24 @@
 	run_bench_512 run_bench_768 run_bench_1024 run_bench \
 	bench_components_512 bench_components_768 bench_components_1024 bench_components \
 	run_bench_components_512 run_bench_components_768 run_bench_components_1024 run_bench_components \
-	buildall checkall all \
+	build run all \
```
mkannwischer, line 15: should this be test instead of run?

scripts/lib/mlkem_test.py
```diff
@@ -252,9 +204,52 @@ def __init__(self, args):
...
+
+        github_log("::endgroup::")
+
+        if p.returncode != 0:
+            sys.exit(1)
+
     def func(self):
         def _func(opt):
-            self._func.compile(opt)
+            self._compile_schemes(TEST_TYPES.MLKEM, opt)
```
mkannwischer, line 200: Please rename the test type to FUNC for consistency.

scripts/lib/mlkem_test.py
mkannwischer, line 1: Why is this needed? I thought the `.WAIT` causes those to not run in parallel.

mkannwischer (changes requested):
A few small comments. 
Please also address the ` ## TODO What is happening here?` comment.

scripts/lib/mlkem_test.py
hanno-becker: This is to keep the command line minimal. You are right that, functionally, it makes no difference.

Makefile
hanno-becker: Yes, good catch!

mkannwischer (approved):
Thanks! LGTM.

PR #606, opened 2025-01-01
README: Minor improvement to Ubuntu quickstart

hanno-becker:
Strictly speaking, `git` may need to be installed to on a fresh Ubuntu system, so reorder the git clone and apt install steps, and add git to the latter.


rod-chapman (approved):
Looks good. Please merge

PR #518, opened 2024-12-12
CBMC: Delete unneeded files and boilerplate

hanno-becker:
* Resolves #500

PR #609, opened 2025-01-02
Add fast, layer-merged forward and inverse NTT code.

rod-chapman:
**Fast NTT and Inverse NTT in C**

This PR introduces optimized C implementation of the forward and inverse NTT for MLKEM.

**Significant changes**

1. Complete re-write of ntt.c - see comments at the top of that file
for an overview of the optimization approach.
2. New per-layer Zeta tables are now auto-generated (by updated autogenerate_file.py script)
into mlkem/zetas.i  Note that zetas.i is NOT a standalone translation unit in C, but is
intended to be included by cpp in the body of ntt.c, so that the literal constant
values in the tables are available to the compiler and optimizer for auto-vectorization.
3. Small change to poly.c/poly_mulcache_compute() to use these new Zeta tables.
4. The function basemul_cached() has moved from ntt.c to poly.c.  It is only
ever called in the latter unit, so really has no place in ntt.c. This also means
that ntt.c implements the NTT and its inverse and nothing else, so greater
cohesion of this unit. basemul_cached() is now declared "static" in poly.c and is
available for inlining and auto-vectorization at the compiler's discretion.
With GCC 14.2.0 -O3 on Apple M1, this saves approx 5000 cycles in all 3 top-level operations.
5. All CBMC proofs updated accordingly.
6. Contracts in ntt.h have NOT changed, so the new implementation meets the
same proof obligations as before.

**Verification**

1. Lint OK
2. All functional tests OK
3. All CBMC proofs OK with MLKEM_K=[2,3,4]

**Performance**

Results on top-level operations will appear from CI runs.

Results for low-level poly_ntt() and poly_invntt_tomont() using the
"--components" switch of "tests bench" on various platforms:

**Graviton 3 (c7g instance), GCC 13.2.0 -O3**

Clock cycles, lower is better

| Operation | Before | After |
| --------- | -----: | ----: |
| NTT       | 1853   | 705   |
| Inv NTT   | 4340   | 908   |

**x86_64 (c7i instance), GCC 13.2.0 -O3**

Clock cycles, lower is better

| Operation | Before | After |
| --------- | -----: | ----: |
| NTT       | 2367   | 1172  |
| Inv NTT   | 5420   | 1901  |

**Apple M1 (MacBook Pro), GCC 14.2.0 -O3**

Clock cycles, lower is better

| Operation | Before | After |
| --------- | -----: | ----: |
| NTT       | 1368   | 366   |
| Inv NTT   | 1619   | 439   |

**To come**

More results to come:

1. CPUs that can vectorize, but not AArch64 or Intel (e.g. PowerPC)
2. CPUs that cannot vectorize (e.g. low-end embedded platforms, such as 32-bit RISC-V)

**Code Size**

On Apple M1 with GCC 14.2.0 -O3

Size of text segment of ntt.o:

Before: 2212 bytes
After:  6924 bytes


cbmc/invntt_layer321/invntt_layer321_harness.c
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * @file invntt_layer321_harness.c
+ * @brief Implements the proof harness for invntt_layer321 function.
+ */
+
+/*
+ * Insert project header files that
+ *   - include the declaration of the function
+ *   - include the types needed to declare function arguments
```
mkannwischer, line 17: please remove these boilerplate comments as done in https://github.com/pq-code-package/mlkem-native/pull/518

hanno-becker:
@rod-chapman Please open this PR from a branch on the main repo.

PR #610, opened 2025-01-02
Add fast, layer-merged forward and inverse NTT code.

rod-chapman:
**Fast NTT and Inverse NTT in C**

This PR introduces optimized C implementation of the forward and inverse NTT for MLKEM.

**Significant changes**

1. Complete re-write of ntt.c - see comments at the top of that file
for an overview of the optimization approach.
2. New per-layer Zeta tables are now auto-generated (by updated autogenerate_file.py script)
into mlkem/zetas.i  Note that zetas.i is NOT a standalone translation unit in C, but is
intended to be included by cpp in the body of ntt.c, so that the literal constant
values in the tables are available to the compiler and optimizer for auto-vectorization.
3. Small change to poly.c/poly_mulcache_compute() to use these new Zeta tables.
4. The function basemul_cached() has moved from ntt.c to poly.c.  It is only
ever called in the latter unit, so really has no place in ntt.c. This also means
that ntt.c implements the NTT and its inverse and nothing else, so greater
cohesion of this unit. basemul_cached() is now declared "static" in poly.c and is
available for inlining and auto-vectorization at the compiler's discretion.
With GCC 14.2.0 -O3 on Apple M1, this saves approx 5000 cycles in all 3 top-level operations.
5. All CBMC proofs updated accordingly.
6. Contracts in ntt.h have NOT changed, so the new implementation meets the
same proof obligations as before.

**Minor changes**
1. cbmc/Makefile.common has been updated so that the final 4 lines
of logs/result.txt are shown following `make result`

**Verification**

1. Lint OK
2. All functional tests OK, both with and without MLKEM_DEBUG
3. All CBMC proofs OK with MLKEM_K=[2,3,4]

**Performance**

Results on top-level operations will appear from CI runs.

Results for low-level poly_ntt() and poly_invntt_tomont() using the
"--components" switch of "tests bench" on various platforms:

**Graviton 3 (c7g instance), GCC 13.2.0 -O3**

Clock cycles, lower is better

| Operation | Before | After |
| --------- | -----: | ----: |
| NTT       | 1853   | 705   |
| Inv NTT   | 4340   | 908   |

**x86_64 (c7i instance), GCC 13.2.0 -O3**

Clock cycles, lower is better

| Operation | Before | After |
| --------- | -----: | ----: |
| NTT       | 2367   | 1172  |
| Inv NTT   | 5420   | 1901  |

**Apple M1 (MacBook Pro), GCC 14.2.0 -O3**

Clock cycles, lower is better

| Operation | Before | After |
| --------- | -----: | ----: |
| NTT       | 1368   | 366   |
| Inv NTT   | 1619   | 439   |

**To come**

More results to come:

1. CPUs that can vectorize, but not AArch64 or Intel (e.g. PowerPC)
2. CPUs that cannot vectorize (e.g. low-end embedded platforms, such as 32-bit RISC-V)

**Code Size**

On Apple M1 with GCC 14.2.0 -O3

Size of text segment of ntt.o:

Before: 2212 bytes
After:  6924 bytes


hanno-becker:
Thanks a lot @rod-chapman. This is a very impressive performance on the C-level!

My concerns, as discussed offline before, are:

1. This is a large deviation from the reference implementation, which we intend to inherit trust from. Are we risking to loose a potential customer because of this? Is the performance uplift worth the deviation? I'm not sure yet.
2. Which customer will actually benefit from this? mlkem-native targets server, mobile and application cores, which I'd say are almost exclusively AArch64 or x86_64+AVX2, hence covered by the existing native backends. I agree that your code is likely faster than the reference [inv]NTT on other architectures as well, but arguing this way feels like putting the cart in front of the horse.

@mkannwischer Please join in and share your thoughts.

mlkem/ntt.c
```diff
@@ -9,137 +9,402 @@
...
+
+#define NTT_BOUND1 (MLKEM_Q - 1)
+#define NTT_BOUND2 (2 * MLKEM_Q - 1)
+#define NTT_BOUND4 (4 * MLKEM_Q - 1)
+#define NTT_BOUND6 (6 * MLKEM_Q - 1)
+#define NTT_BOUND7 (7 * MLKEM_Q - 1)
+#define NTT_BOUND8 (8 * MLKEM_Q - 1)
+
+/* "pc" == "polynomial coefficients" */
+typedef int16_t pc[MLKEM_N];
```
hanno-becker, line 80: I find this type abbreviation rather confusing -- can we remove it?

cbmc/Makefile.common
```diff
@@ -949,6 +949,7 @@ result:
 	$(MAKE) -B _result
 	@ echo Running 'litani build'
 	$(LITANI) run-build
+	tail -4 $(LOGDIR)/result.txt
```
hanno-becker, line 952: What's this for?

hanno-becker:
> The function basemul_cached() has moved from ntt.c to poly.c. It is only
ever called in the latter unit, so really has no place in ntt.c. This also means
that ntt.c implements the NTT and its inverse and nothing else, so greater
cohesion of this unit. basemul_cached() is now declared "static" in poly.c and is
available for inlining and auto-vectorization at the compiler's discretion.
With GCC 14.2.0 -O3 on Apple M1, this saves approx 5000 cycles in all 3 top-level operations.

NB: If we go down the route of further optimizing the C code at the cost of deviating from the reference implementation, the scalar-product base multiplication should be further rewritten to match the structure of the native AArch64 implementation: Compute one scalar product of vectors of polynomials of degree 2 at time, which reduces the number of modular reductions by a factor of MLKEM_K.

mlkem/zetas.i
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) 2024 The mlkem-native project authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*
```
hanno-becker, line 6: This should be an `.h` file, not `.i`? I think `.i` is typically used for snippets of assembly.

scripts/autogenerate_files.py
```diff
@@ -88,24 +88,77 @@ def gen_c_zetas():
         zeta.append(signed_reduce(pow(root_of_unity, i, modulus) * montgomery_factor))
 
     # The source code stores the zeta table in bit reversed form
-    yield from (zeta[bitreverse(i, 7)] for i in range(128))
+    zeta_reversed = []
+    for i in range(128):
+        zeta_reversed.append(zeta[bitreverse(i, 7)])
+    # return a finite sequence here, not a generator
```
hanno-becker, line 94: Why was this changed?

hanno-becker: Nit: You can keep it was it was, but use `zetas = list(gen_c_zetas())` on the caller.

mlkem/sys.h
```diff
@@ -65,17 +65,21 @@
 #if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
 #define INLINE inline
 #define ALWAYS_INLINE __attribute__((always_inline))
+#define NO_INLINE __attribute__((noinline))
 #elif defined(_MSC_VER)
 #define INLINE __inline
 #define ALWAYS_INLINE __forceinline
+#define NO_INLINE
```
hanno-becker, line 72: It looks like `noinline` is also supported by MSVC: `__declspec(noinline)`

mlkem/ntt.c
```diff
@@ -9,137 +9,402 @@
...
+
+/* Zeta tables are auto-generated into the file zetas.i */
+/* That file is included here in the body of the this   */
+/* translation unit to make the literal values of the   */
+/* constants directly available the compiler in support */
+/* of auto-vectorization and optimization of the code   */
+#include "zetas.i"
+
+STATIC_NO_INLINE_TESTABLE void ntt_layer123(pc r)
```
hanno-becker, line 89: Why do we need the NO_INLINE annotation? What performance difference does it make?

mlkem/ntt.c
```diff
@@ -9,137 +9,402 @@
...
+#define NTT_BOUND8 (8 * MLKEM_Q - 1)
+
+/* "pc" == "polynomial coefficients" */
+typedef int16_t pc[MLKEM_N];
+
+/* Zeta tables are auto-generated into the file zetas.i */
+/* That file is included here in the body of the this   */
+/* translation unit to make the literal values of the   */
+/* constants directly available the compiler in support */
+/* of auto-vectorization and optimization of the code   */
```
hanno-becker, line 86: Hm, I don't follow this -- what changes if `zetas.i` remains a standalone compilation unit? The specific values should not affect auto-vectorization or optimization.

mlkem/ntt.c
```diff
@@ -9,137 +9,402 @@
...
+    invariant(array_abs_bound(r,     192, j + 191, NTT_BOUND4))
+    invariant(array_abs_bound(r, j + 192,     223, NTT_BOUND1))
+    invariant(array_abs_bound(r,     224, j + 223, NTT_BOUND4))
+    invariant(array_abs_bound(r, j + 224, (MLKEM_N - 1), NTT_BOUND1)))
   {
-    int16_t t;
-    t = fqmul(r[j + len], zeta);
-    r[j + len] = r[j] - t;
-    r[j] = r[j] + t;
+    const int ci1 = j + 0;
```
hanno-becker, line 123: Nit: I'd find this more readable if the indexing was 0-based.

mlkem/ntt.c
```diff
@@ -9,137 +9,402 @@
...
+    const int ci3 = j + 64;
+    const int ci4 = j + 96;
+    const int ci5 = j + 128;
+    const int ci6 = j + 160;
+    const int ci7 = j + 192;
+    const int ci8 = j + 224;
+    int16_t t1, t2;
+
+    /* Layer 1 */
+    t1 = fqmul(r[ci5], mlkem_l1zeta1);
```
hanno-becker, line 134: This may be more readable if you introduced either a macro or a static inline function for a single butterfly.

mlkem/ntt.c
```diff
@@ -9,137 +9,402 @@
...
   {
-    int16_t zeta = zetas[k++];
-    ntt_butterfly_block(r, zeta, start, len, layer * MLKEM_Q - 1);
+    const int ci1 = j + start;
+    const int ci2 = ci1 + 8;
+    const int ci3 = ci1 + 16;
+    const int ci4 = ci1 + 24;
+    int16_t t1, t2;
+
+    /* Layer 4 */
```
hanno-becker, line 234: Similar here, let's use a helper macro/function.

hanno-becker (changes requested):
@rod-chapman Many thanks again for all the work. Notwithstanding the broader discussion of whether/why we want to integrate this, some more comments:

1. One should try to make the code more readable by folding CT/GS butterflies into a macro (or perhaps helper function), and in particular avoid/hide all the `const int rcXXX` definitions. I realize you have done a lot of experimentation, but overall I find the code too verbose at the moment and would like to have it more compact; if we loose a % of [inv]NTT-performance on some compilers, this won't matter overall.
2. I am confused by the reasoning about what facilitates auto-vectorization. The code mentions that it's the amount of coefficients processed per inner loop, but I don't think it's true. For AVX2, it is actually unclear to me why the merging is ideal, because you begin to need intra-vector operations already at layer 5. It would be interesting to see how compilers handle this.
3. If possible, I'd like to keep the PR to just the changes to the [inv]NTT, and not doing any code motion, e.g. of the base multiplication.

mlkem/ntt.c
```diff
@@ -9,137 +9,402 @@
...
+ * and loop invariants of each layer and their inner functions.
+ *
+ * Auto-Vectorization
+ * ------------------
+ * The code is written in a style that is amenable to auto-vectorization
+ * on targets that can support it, and using comtemporary compilers - for
+ * example, Graviton/AArch64 or Intel/AVX2 using recent versions of GCC
+ * or CLANG.
+ *
+ * Inner loops of the butterfly functions process at most 8 coefficients
```
hanno-becker, line 63: Hm, I don't follow this reasoning. It's not the number of coefficients processed per loop that's important, but the uniformity and striding of the loop itself? For example, for 128-bit vectors the compiler must merge 8 iterations of the layer123 loop for vectorization; the loop bodies themselves don't vectorize. The amount of layer merging -- here 3, leading to 8 coefficients in flight -- is mostly relevant for instruction-level parallelism and reduction of redundant load/stores.

Things become difficult as soon as intra-vector operations would be required for vectorization. For AArch64, that starts to happen at layer 6, while for AVX2, it starts to happen at layer 5. So on an intuitive level, it makes sense to isolate layer 6,7 for AArch64 because they require intra vector shuffling, but for AVX2, already layer 5 is problematic.

mlkem/ntt.c
```diff
@@ -9,137 +9,402 @@
...
+ * The code defers reduction of coefficients in the core
+ * "butterfly" functions in each layer, so that the coefficients
+ * grow in absolute magnitiude by MLKEM_Q after each layer.
+ * This growth is reflected and verification in the contracts
+ * and loop invariants of each layer and their inner functions.
+ *
+ * Auto-Vectorization
+ * ------------------
+ * The code is written in a style that is amenable to auto-vectorization
+ * on targets that can support it, and using comtemporary compilers - for
```
hanno-becker, line 59: Nit: Typo

mkannwischer:
Thanks @rod-chapman - the results are truly impressive and this looks like a lot of work.
I agree with @hanno-becker that as it stands right now the code is a bit too verbose - we should try to simplify that as much as possible and I'd rather have something readable that is a few percent slower on some platforms.

On the discussion of whether or not this change aligns with the goals of mlkem-native: Currently the goal was to deviate from the reference implementation only to (1)  facilitate formal verification, or (2) to harden against side-channel attacks, or (3) to use the native backends. This change fits none of these categories and arguably we'd move quite far away from the reference implementation. 

I see two solutions:
(1) We adopt the new NTT as an alternative backend and keep the current C implementation as the default backend. 
(2) We change the goals of mlkem-native to also provide a high-speed C backend. 

(1) would make using mlkem-native more complicate and it would also be duplicating the effort of maintenance. Hence, I'm not a big fan.

(2) needs careful consideration and - in my view - only makes sense if we have a (potential) consumer for which this makes a real difference. So far my impression was that x86_64 and AArch64 are the only platforms that our anticipated consumers care about. Do we have a potential consumer for which the performance of the C backend is very important? Or what are other reasons for making this change?

scripts/autogenerate_files.py
rod-chapman: Oh... I thought it had to be a sequence not a genertor, so it could be indexed. Happy to defer to whatever is considered better style in Python.

mlkem/ntt.c
rod-chapman: I tried inling _all_ of these functions into the top-level (e.g. poly_ntt() function) and it was not good for performance. I think the size of the resulting single function and the number of loops in it confounded the optimizer. It's also a _lot_ easier to read and review the resulting assembly code if these functions are not inlined.

mlkem/ntt.c
rod-chapman: I think the vectorizer needs to know the literal values so it can turn a single table entry into a vector of 8 of them to be loaded in a single instruction. If it only has the declaration of the lookup table, then it won't be able to do that.

scripts/autogenerate_files.py
hanno-becker: > I thought it had to be a sequence not a genertor

That's true, but you can convert the generator into a list with `list(...)` at the caller.

mlkem/ntt.c
rod-chapman: That sounds like a good idea. I will try.

hanno-becker: Please see #611

mlkem/ntt.c
hanno-becker: Hm, I can't imagine. Let's try.

cbmc/Makefile.common
rod-chapman: On a successful proof run, it shows
```

** 0 of 394 failed (1 iterations)
VERIFICATION SUCCESSFUL

```
which is what I want to see. Without that line I find myself having to type
```
less logs/result.txt
```
after every single `make result` run, so I decided to make it do that every time.

mlkem/ntt.c
hanno-becker: Yes but one can just not have any (NO|ALWAYS)_INLINE annotation.

mlkem/ntt.c
rod-chapman: Removed.

scripts/autogenerate_files.py
rod-chapman: Would you strongly prefer to change this back, or leave it as is?

mlkem/sys.h
rod-chapman: Do we have any CI in Windows to verify that this won't break anything?

mlkem/ntt.c
rod-chapman: I tried. I put the zeta tables into a distinct compilation unit in their own right. zetas.h declares the constants as "extern" but with no actual values. zetas.c contains the full declaration of the tables with the initializing values or aggregates. Surprised to find this adds only approx 100 cycles to all 3 top-level operations for all parameter sets on Apple M1. Code-gen is completely different. Let's discuss.

mlkem/ntt.c
rod-chapman: All this code has been re-factored into inner "ct_butterfly()" and "gs_butterfly_reduce()" and "gs_butterfly_defer()" functions.

mlkem/ntt.c
rod-chapman: Fixed

rod-chapman:
Thanks for your thoughts on this PR, and apologies for latency
in replying.

I sympathize with your main point - why did I do this?!?
(And is it compatible with the main goals of mlkem-native anyway?)

I had a bit of think about the first question.  My own goal for
doing this are probably something like

1. To provide a demonstration of proof-driven optimization and that
   "formal is fast" is achievable using C and CBMC.
2. To force improvement of CBMC.
3. To provide a "faster, slightly bigger" C backend for mlkem-native
   (more of which below...).
4. To provide a fast C backend for targets other than AArch64/NEON and x86-64/AVX2,
   for example
      AArch64/SVE2 (if and when compilers can do it)
      x86_64/AVX512 (if and when compilers can do it)
      PowerPC/AltiVec
      RISC-V64 with the Vector extension
      Any machine which can't vectorize at all e.g. lots of 32-bit devices.

5. To provide a vehicle for teaching CBMC and type-safe programming in C.
6. To provide some interesting tests cases to see how well SLOTHY
   performs on compiler-generated NEON code.

While I realize that most of these are not goals of mlkem-native,
I don't think any of them are obviously contradictory or incompatible
with mlkem-native.

On point 3, I kinda like the idea of keeping 1 C backend, but offering
a build option for "small code size, slow" (i.e. current reference NTT code)
or "larger code size, faster" (i.e. my new code). Many other crypto libraries
(e.g. WolfCrypt) have such build-time options for code-size vs performance
trade-off.

Thinking about "Who would use or want this?" I can see a few use cases:

1. Embedded systems people who want reasonable performance on all targets,
   but aren't worried about power analysis attacks.
2. LibOQS - does LibOQS adoption of mlkem-native change things? Do they
   have more targets in mind other an AArch64 and Intel? Would a fast reference
   C backend be useful for them?
3. I will reach out to AWS Devices team to see what targets they have in play.

rod-chapman:
I realize that AArch64/SVE2 and x86_64/AVX512 are something of a red-herring, since anyone with a CPU implementing those things will also have NEON and AVX2, so they'd naturally use the existing back-ends for those.

mlkem/ntt.c
rod-chapman: Done

hanno-becker:
The performance impact has gone down since the switch to unsigned variables has miraculously improved performance by 10-30%

rod-chapman:
This PR will be closed now, and a new one #683 opened following to re-organization of the sources in #674.

PR #612, opened 2025-01-03
Add support for single-CU build

### LINKED ISSUE #553: Demonstrate single-compilation-unit build
Some build environments, e.g. AWS-LC's FIPS build, require all .c sources to be compiled in one giant `.c` file. 

- [x] Autogenerate an `.c` file which includes all mlkem-native .c source files, excluding native backend code.
- [x] Demonstrate single-CU build with single config in an example, without native backends.
- [x] Make monolithic .c file standalone by `#undef`'ing all `#define` defined within this giant `.c` file.
- [x] Demonstrate single-CU build with multiple configs in an example, without native backends.

Incorporating native code will be tracked separately.

hanno-becker:
* Fix #553 

This PR modifies mlkem-native to enable multi-level builds in single compilation units. The necessary changes are:
1. Provide an auto-generated compilation unit `examples/monolithic_build/mlkem_native_all.c` which includes all source files.
2. In `mlkem_native_all.c`,`#undef` all `#define`'s in all header and source files.
3. Namespace statics and structs.

Step 1 is enough for a single-unit build with a single configuration. Steps 2 and 3 are needed for a single-unit build with multiple configuration files.

Steps 1 and 2 are autogenerated, while step 3 is currently done manually; we may want to automate this in the future.

A full example `examples/monolithic_build` is added which shows how to use the monolithic unit with multiple configurations. The prime example here is to build mlkem-native multiple times with different security levels.

**Limitations:**
* The single-CU build can only be used with the C backend for now.
* We make no effort at present to avoid duplication of functionality when including mlkem-native multiple times. For example, if it's merely the security level that needs to be varied, a lot of code is only needed once, but is, at present, included once per security level. We may address this in the future.


hanno-becker:
@mkannwischer Eventually, I could see the monolithic build become the default (though not yet in this PR). What are your thoughts on that?

examples/monolithic_build/mlkem_native_all.c
```diff
@@ -74,11 +69,6 @@
 #undef _16XMONTSQLO
 #endif
 
-/* mlkem/poly.h:715 */
-#if defined(poly_add)
```
mkannwischer: How are the undefs ordered? There are a lot of changes to this file in this commit. Many more than necessary. Can we order the undefs by file?

mlkem/verify.h
```diff
@@ -11,6 +11,19 @@
 #include "cbmc.h"
 #include "params.h"
 
+/* Static namespacing */
```
mkannwischer, line 14: Can you add a section to the README regarding single-CU builds, and refer to the section in this comment?
I think this is rather confusing for most C developers.

examples/monolithic_build/config_a.h
```diff
@@ -43,15 +43,16 @@
  * Description: The macros to use to namespace global symbols
  *              from mlkem/.
  *****************************************************************************/
-#define MLKEM_NAMESPACE(sym) MLKEM_DEFAULT_NAMESPACE(sym)
+#define CONCAT(a, b) a##b
+#define MLKEM_NAMESPACE(sym) CONCAT(namespace_a_, sym)
```
mkannwischer, line 47: why not just use the default namespace? it includes the parameter set so there should not be any clashes?

mlkem/verify.h
mkannwischer: I see you added it in a later commit. Do you want to refer to it here?

examples/monolithic_build/main.c
```diff
@@ -0,0 +1,18 @@
...
+
+#define MLKEM_NATIVE_CONFIG_FILE "config_b.h"
+#include "mlkem_native_all.c"
+#undef MLKEM_NATIVE_CONFIG_FILE
+
+int main(void)
+{
+  /* Nothing to do here -- only test that it builds */
+  return 0;
+}
```
mkannwischer, line 18: Please add a basic test for both parameter sets. That should be very easy.

mkannwischer (changes requested):
Thanks @hanno-becker. This looks great - this isn't much cleaner than I expected. Thank you.  
The changes to the mlkem-native sources itself are minimal and all look good (vast majority of the line changes are adding the new example). 
I've tested that the build works without problems. 

Here are a few minor suggestions.

mkannwischer:
> @mkannwischer Eventually, I could see the monolithic build become the default (though not yet in this PR). What are your thoughts on that?

What would be the reason for changing the default? It feels rather unnatural for me. The requirements of a single CU seems to be very specific to me and likely won't apply to many consumers, or am I missing something? 
I think for the average C developer it will make it quite a bit more complicated to use.

hanno-becker:
> What would be the reason for changing the default? It feels rather unnatural for me. The requirements of a single CU seems to be very specific to me and likely won't apply to many consumers, or am I missing something?
I think for the average C developer it will make it quite a bit more complicated to use.

The reason to change would be that it's simpler to use: One would move all code from `mlkem` into (say) `mlkem/src`, and in `mlkem` otherwise only have `api.h` and `mlkem.c`; the developer then only needs to (a) compile `mlkem.c`, (b) use `api.h`.

examples/monolithic_build/config_a.h
hanno-becker: Actually in this case there would still be clashes for FIPS202, which does not include the level.

examples/monolithic_build/mlkem_native_all.c
hanno-becker: Yes, good point. Using file+line_no ordering now.

examples/monolithic_build/main.c
hanno-becker: Addressed.

hanno-becker:
Thanks @mkannwischer for the review! Please have another look.

mkannwischer (approved):
Thanks for the changes. Looks good to me now!

PR #615, opened 2025-01-06
Remove STATIC_TESTABLE and STATIC_INLINE_TESTABLE wrappers

hanno-becker:
STATIC_TESTABLE and STATIC_INLINE_TESTABLE were previously used as wrappers
around `static` and `static inline` that were defined as empty for a CBMC
build, thereby lifting visibility of static (inline) functions so that CBMC
would see them.

This commit achieves the same without STATIC_TESTABLE and
STATIC_INLINE_TESTABLE, both of which are removed, by defining `static`
and `INLINE` as the empty string in a CBMC build.

### DIFF
--- a/cbmc/Makefile.common
+++ b/cbmc/Makefile.common
@@ -535,6 +535,10 @@ ifndef MLKEM_K
 endif
 
 DEFINES += -DMLKEM_K=$(MLKEM_K)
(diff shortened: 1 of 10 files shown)


mkannwischer (approved):
Very nice - I didn't know this works.
LGTM.

PR #618, opened 2025-01-06
Remove backend name from default namespace

hanno-becker:
Previously, the backend type would enter the default namespace
used by mlkem-native.

This means that before the MLKEM_NAMESPACE/FIPS202_NAMESPACE macros
can be used, the backend header has to be included.

This is inconvenient for the introduction of a public API header which
solely depends on the config file, since that header would need to also
include the backend header, and hence be compiled with the same include
paths as the mlkem-native build itself. This is unintuitive and unnecessarily
complicated for the customer.

This PR removes the backend type from the default namespace. Customers
who need the backend type, or at least the target architecture, to be
reflected in the namespace, can overwrite the default namespace as desired.

Further smaller changes related to the config are made; see commits.


hanno-becker:
cc @bhess: With this, you need to explicitly set the namespace to reflect the backend/architecture. Please let us know if this is of any concern.

mkannwischer (approved):
LGTM.

bhess:
> cc @bhess: With this, you need to explicitly set the namespace to reflect the backend/architecture. Please let us know if this is of any concern.

Thanks for tagging me, @hanno-becker. Explicitly setting the namespace isn’t a problem, but it would be more convenient if the namespace prefix (e.g., PQCP_MLKEM_NATIVE_MLKEM1024_AARCH64_OPT) could be set via CFLAGS.

Unless I missed something, it appears that currently, one needs to overwrite the MLKEM_DEFAULT_NAMESPACE(x) macro by either patching or providing a custom config.h. Would it be possible to introduce an optional compile-time parameter -e.g., MLKEM_NAMESPACE_PREFIX - that defines MLKEM_DEFAULT_NAMESPACE(x) as MLKEM_NAMESPACE_PREFIX_##x?

hanno-becker:
@bhess That's no problem, I will open a PR.

hanno-becker:
@bhess Can you have a look at #635 and see if that's what you had in mind, and whether it makes your life in OQS easier?

bhess:
Great, thanks! Looks like what I had in mind 👍 .

PR #619, opened 2025-01-06
Introduce header for mlkem-native's public API

hanno-becker:
* Based on #618 

TLDR This PR aims to introduce a public header `mlkem_native.h` specifying the stable API of mlkem-native.

----

Application sources using an mlkem-native build should not have to
add the source tree of mlkem-native to the include paths. Instead,
there should be a single standaloe header sufficient to include
all necessary declarations.

Previously, this was not true: Applications would need to include
`kem.h` from the source tree, which in turn would pull in, for example,
`config.h` and `cbmc.h`.

On the other hand, the header of mlkem-native is not static, but
depends on information from the build, namely the security level
and the namespacing: The security level determines the sizes of
key material which callers need to know about. The namespacing
determines the name of the function symbols, which callers need
to know about.
It is therefore not possible to completely detach the mlkem-native
API header from the source tree used to build mlkem-native.

One way to address this is to merely require the build configuration
header to be included in the public API header. This works well for
single-level builds, but is inconvenient if the consumer wants to use
multiple instances of mlkem-native for different security levels; in
this case, including the public header multiple times would pull in
multiple configuration files, which would lead to clashes of #define's
clauses.
On the other hand, requiring the consumer to duplicate the level and
namespacing information in the API header is inconvenient for simple
tests or single-level builds.

This introduces a public API header mlkem-native.h, which addresses
the above as follows:

mlkem-native.h requires the build level and build namespace. By default,
it is obtained by including the underlying build configuration
(assumed to be present in the include paths). Alternatively, the user
can set BUILD_INFO_LVL and BUILD_INFO_NAMESPACE explicitly in mlkem-native.h,
and thereby avoid any dependency between mlkem-native.h and config.h.
In this setting, mlkem-native.h leaves no other traces than static size
declarations and build-specific API declarations. In particular, it can
be included multiple times.

This commit adjusts all tests and examples to no longer include `kem.h`,
but only `mlkem-native.h`. In all cases but the monolithic build example,
no further change is required, with the API header pulling the underlying
configuration automatically.

The monolithic build test `examples/monolithic_build` is adjusted as follows:
Where previously it included two instances of mlkem-native, it now includes
three, one per security level. The previous `mlkem_native_all.c` is renamed
to `mlkem_native_monobuild.c`, and is the source file needed for a monolithic
build of a _single_ configuration. This is then included 3 times in `mlkem_native_all.c`.
To provide a header for this file, we include `mlkem-native.h` 3 times, using
manually set `BUILD_INFO_LVL` and `BUILD_INFO_NAMESPACE` to avoid nameclashes
from pulling in the full configuration.


hanno-becker:
@bhess FYI. Feel free to chime in in terms of suitability for OQS.

mlkem/config.h
```diff
@@ -97,4 +97,60 @@
...
 
+/******************************************************************************
+ * Name:        MLKEM_NATIVE_API_STANDARD
+ *
+ * Description: Define this to extend api.h to also export key sizes and public
+ *              API in the CRYPTO_xxx and crypto_kem_xxx format as used e.g. by
+ *              SUPERCOP.
+ *
+ *****************************************************************************/
+#define MLKEM_NATIVE_API_STANDARD
```
mkannwischer, line 108: This is never used?

mlkem/config.h
```diff
@@ -97,4 +97,60 @@
 #define MLKEM_NATIVE_FIPS202_BACKEND "fips202/native/default.h"
 #endif /* MLKEM_NATIVE_FIPS202_BACKEND */
 
+/******************************************************************************
+ * Name:        MLKEM_NATIVE_API_STANDARD
+ *
+ * Description: Define this to extend api.h to also export key sizes and public
```
mkannwischer, line 103: We don't have a api.h anymore?

mlkem/params.h
```diff
@@ -50,12 +53,12 @@
...
 
-#define MLKEM_PUBLICKEYBYTES (MLKEM_INDCPA_PUBLICKEYBYTES)
+#define MLKEM_INDCCA_PUBLICKEYBYTES (MLKEM_INDCPA_PUBLICKEYBYTES)
 /* 32 bytes of additional space to save H(pk) */
-#define MLKEM_SECRETKEYBYTES                                   \
+#define MLKEM_INDCCA_SECRETKEYBYTES                            \
   (MLKEM_INDCPA_SECRETKEYBYTES + MLKEM_INDCPA_PUBLICKEYBYTES + \
    2 * MLKEM_SYMBYTES)
-#define MLKEM_CIPHERTEXTBYTES (MLKEM_INDCPA_BYTES)
+#define MLKEM_INDCCA_CIPHERTEXTBYTES (MLKEM_INDCPA_BYTES)
```
mkannwischer, line 61: I think this change is undesirable as it requires understanding the difference between CCA and CPA security. 
I'd rather hide the CPA names.

mlkem/mlkem_native.h
```diff
@@ -0,0 +1,233 @@
...
+#if !defined(BUILD_INFO_LVL)
+
+/* Option 1: Extract from config */
+#if defined(MLKEM_NATIVE_CONFIG_FILE)
+#include MLKEM_NATIVE_CONFIG_FILE
+#else
+#include "config.h"
+#endif
+
+#if MLKEM_K == 2
```
mkannwischer, line 48: What if MLKEM_K is not defined?

mlkem/mlkem_native.h
```diff
@@ -0,0 +1,233 @@
...
+ * Remove this if you don't need it, or if you need multiple instances
+ * of this header. */
+
+#if !defined(BUILD_INFO_NO_STANDARD_API)
+#define CRYPTO_SECRETKEYBYTES MLKEM_SECRETKEYBYTES(BUILD_INFO_LVL)
+#define CRYPTO_PUBLICKEYBYTES MLKEM_PUBLICKEYBYTES(BUILD_INFO_LVL)
+#define CRYPTO_CIPHERTEXTBYTES MLKEM_CIPHERTEXTBYTES(BUILD_INFO_LVL)
+
+#define CRYPTO_SYMBYTES MLKEM_SYMBYTES
+#define CRYPTO_BYTES MLKEM_BYTES
```
mkannwischer, line 217: should these be namespaced?
That's what PQClean does https://github.com/PQClean/PQClean/blob/master/crypto_kem/ml-kem-1024/clean/api.h

It allows to duplicate the source tree for multiple different parameter sets and then include multiple `api.h`s into one file.

mkannwischer (changes requested):
Thanks @hanno-becker - I think this a step in the right direction!
A few suggestions.

mlkem/params.h
hanno-becker: Hm, but this is internal?

mlkem/mlkem_native.h
hanno-becker: You can't namespace `#define`'s -- if you need this multiple times, you use the main API which is already namespaced. 

Do you have a better idea?

mlkem/mlkem_native.h
hanno-becker: A valid config must define `MLKEM_K`

mlkem/config.h
hanno-becker: This needs to go

mlkem/config.h
hanno-becker: This needs to go

mlkem/mlkem_native.h
mkannwischer: maybe error here with an appropriate error then, instead of just continuing and running into some error somewhere else?

mlkem/params.h
mkannwischer: okay

mlkem/mlkem_native.h
hanno-becker: Ok

mlkem/config.h
hanno-becker: Done

mkannwischer (approved):

PR #626, opened 2025-01-09
Use Keccak-x4 in poly_getnoise_eta1122_4x

mkannwischer:


hanno-becker:
Looks worthwhile on native backends, but slightly detrimental for the C backend. Not sure how to best handle this. @mkannwischer WDTY.

mkannwischer:
> Looks worthwhile on native backends, but slightly detrimental for the C backend. Not sure how to best handle this. @mkannwischer WDTY.

It is also detrimental on platforms with poor vector performance (e.g., Cortex-A55). More specifically if 2xKeccakx4 take longer than 6xKeccakx1. 

I'm also undecided. 10% on high-end x86_64 looks important enough though. Apple Silicon should also benefit from this significantly. 

An alternative approach would be to do one Keccakx4 permutation, extract the two states that need another permutation and do that using Keccakx1. This should be better on all platforms. But it's a bit hacky.

mlkem/poly.c
```diff
@@ -424,27 +424,35 @@ void poly_getnoise_eta1122_4x(poly *r0, poly *r1, poly *r2, poly *r3,
...
+  memcpy(extkey0, seed, MLKEM_SYMBYTES);
+  memcpy(extkey1, seed, MLKEM_SYMBYTES);
+  memcpy(extkey2, seed, MLKEM_SYMBYTES);
+  memcpy(extkey3, seed, MLKEM_SYMBYTES);
+  extkey0[MLKEM_SYMBYTES] = nonce0;
+  extkey1[MLKEM_SYMBYTES] = nonce1;
+  extkey2[MLKEM_SYMBYTES] = nonce2;
+  extkey3[MLKEM_SYMBYTES] = nonce3;
+
+  prf_eta1_x4(buf0, buf1, buf2, buf3, extkey0, extkey1, extkey2, extkey3);
```
hanno-becker, line 450: In principle, one can guard this by `defined(MLKEM_USE_FIPS202_NATIVE_X2) || defined(MLKEM_USE_FIPS202_NATIVE_X4)`, which probably does the right thing. Can you change the PR accordingly so we see the benchmarks?

It's still not clear though whether we want to introduce this distinction, because it makes to achieve full CBMC coverage.

hanno-becker (approved):

PR #627, opened 2025-01-09
Test: Try 4xKeccak AVX2 for 1xKeccak

mkannwischer:


mkannwischer:
Okay, this is much slower. As expected: You should not use AVX2 for a single Keccak.

PR #629, opened 2025-01-10
CBMC: Big-endian proofs for KeccakF1600_State{Extract,XOR}Bytes

hanno-becker:
Those functions have an explicit distinction between little and big endian systems and therefore need a proof for either case.

Care has to be taken to allow overwriting the directives in the host system's preprocessor; we disable the auto-detection of endianness if SYS_LITTLE_ENDIAN or SYS_BIG_ENDIAN have been set manually.


mkannwischer (approved):
Thanks - LGTM. 
I double checked that this code path is actually what's evaluated. 
I also double checked that there are no other code locations where we depend on endianness. This should complete the CBMC proofs - great!

PR #630, opened 2025-01-10
Monobuild: Share FIPS202 sources

hanno-becker:
The monobuild example `examples/monolithic_build` demonstrates how to build three instances of mlkem-native for MLKEM-{512,768,1024} inside a single compilation unit.

Previously, however, the monobuild would include the FIPS202 sources 3x as well, even though the functionality is the same in all cases and should be shared across the security levels.

This commit enables the sharing of FIPS202 sources in multi-level monobuilds, as follows:
- The autogenerated `monobuild.c` recognizes the new define `MLKEM_NATIVE_MONOBUILD_KEEP_FIPS202_HEADERS`. The effect of this define is that the `#define`'s from FIPS202 header are kept at the end of `monobuild.c`. (Remember that, by default, _all_ `#define`'s from the entire source tree are `#undef`'d at the end of `monobuild.c`)
- The autogenerated `monobuild.c` recognizes the new define `MLKEM_NATIVE_MONOBUILD_NO_FIPS202_SOURCES`. The effect of this define is that the FIPS202 source files are not included.

Those defines are sufficient to support a 3-level monobuild with shared FIPS202; this commit adjusts the monobuild example accordingly.

More sharing is desirable between levels since many functions are independent of the security level (e.g. the NTT), but this is left for future commits.


mkannwischer (approved):
Thanks! Oh wow - quite amazing that you did not have to touch the mlkem sources at all.

PR #631, opened 2025-01-10
CBMC: Use exclusive upper bounds for `forall` primitive

hanno-becker:
This PR modifies the CBMC predicates `forall(...)`,
`array_bound(...)` and `array_abs_bound` to use exclusive
upper bounds for the indices.

This simplifies all call sites which previously manually issued
a `-1` to get from an array length to a maximal index.

It will also enable switching to unsigned loop indices everywhere;
the previous versions of `forall(...)`, `array_bound(...)` and
`array_abs_bound(...)` prevented this because they did not work
with the exclusive unsigned upper bound `0`.

Note that the _coefficient_ bounds are not changed and remain inclusive.


mkannwischer (approved):
Thanks! This was bothering me for a while.

PR #635, opened 2025-01-10
Configure namespace prefix instead of namespace macro

hanno-becker:
Previously, the user could customize namespacing by providing macros `MLKEM_NAMESPACE(sym)` and `FIPS202_NAMESPACE(sym)`.

This mode of configuration is slightly inconvenient in that the namespace cannot be controlled from CFLAGS.

This commit addresses this, as follows: Instead of making the namespace macro configurable, make the namespace _prefix_ configurable, and derive the namespace macro: The config provides MLKEM_NAMESPACE_PREFIX and FIPS202_NAMESPACE_PREIFX, and then the derived MLKEM_NAMESPACE(sym) and FIPS202_NAMESPACE(sym) are the concatenation of that, an underscore `_`, and the symbol name `sym`.

All configs and examples are adjusted accordingly.

### DIFF
--- a/examples/custom_backend/mlkem_native/custom_config.h
+++ b/examples/custom_backend/mlkem_native/custom_config.h
@@ -34,25 +34,20 @@
 /* #define MLKEM_NATIVE_CONFIG_FILE "config.h" */
 
 /******************************************************************************
- * Name:        MLKEM_NAMESPACE
- *              _MLKEM_NAMESPACE
+ * Name:        MLKEM_NAMESPACE_PREFIX
  *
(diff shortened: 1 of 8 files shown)


bhess (approved):
Thank you @hanno-becker, looks good to me!

This makes it convenient - for example for liboqs - to define separate namespaces for the architectures and backends, via a CFLAG definition of MLKEM_NAMESPACE_PREFIX.

mkannwischer (approved):
LGTM. We should have done that in the first place.

PR #636, opened 2025-01-11
CBMC: Update proof guide

### LINKED ISSUE #382: Update Proof Guide to cover use of OBJECT_UPTO rather than OBJECT_WHOLE
The Proof Guide document (cbmc/proofs/proof_guide.md) talks of using the OBJECT_WHOLE macro within ASSIGNS contracts. We now know this is unwise, and that the OBJECT_UPTO macro is almost always a better choice.

The guide should be updated to explain why this is so, and update the existing examples to use OBJECT_UPTO where appropriate.

I also suggest adding a sub-section under "proof patterns" called "Updating composite objects" that gives guidance on how to write contracts for functions that update a composite object (e.g. an array or a struct) via a mutable by-reference parameter.

hanno-becker:
* Fix #382 

- Provide some more details on annotations
- Shorten in various places
- Adjust to changed syntax and convention

### DIFF
--- a/cbmc/proof_guide.md
+++ b/cbmc/proof_guide.md
@@ -2,57 +2,151 @@
 
 # CBMC Proof Guide and Cookbook for mlkem-native
 
-This doc acts as a guide to developing proofs of C code using [CBMC](https://model-checking.github.io/cbmc-training/). It concentrates
-on the use of _contracts_ to achieve _unbounded_ and _modular_ proofs of type-safety
-and correctness properties.
+This document acts as a guide to how we develop proofs of mlkem-native's C code using
+[CBMC](https://model-checking.github.io/cbmc-training/). It concentrates on the use of _contracts_ to achieve
+_unbounded_ and _modular_ proofs of type-safety and correctness properties.
 
-This document uses the abbreviated forms of the [CBMC contracts](https://diffblue.github.io/cbmc/contracts-user.html) defined by macros in the
-[cbmc.h](../mlkem/cbmc.h) header file in the mlkem-native sources.
+## Scope
 
-## Common Proof Patterns
+Our CBMC proofs confirm the absence of certain classes of undefined behaviour, such as integer overflow or out of bounds
+memory accesses -- for the precise list of conditions checked, see the CBMC configuration in
+[Makefile.common](Makefile.common). For many arithmetic functions, we additionally specify how they affect coefficient
+bounds: For example, we show that the result of `poly_invntt_tomont()` has coefficients bound by
+`INVNTT_BOUND`. Finally, some simple functions have their full functional behaviour specified: For example, the
+specification of the constant-time `ct_memcmp()` shows that, functionally, it is just an ordinary `memcmp()`.
 
-### Loops (general advice)
+## CBMC annotations
 
-1. A function should contain at most one outermost loop statement. If the function you're trying to prove has more than one outermost loop, then re-factor it into two or more functions.
+CBMC proofs are largely automatic; there are no proof scripts as in interactive theorem proving. Instead, CBMC consumes
+_annotated_ C code and checks that it does not exhibit the configured classes of undefined behaviour within the
+context described by the annotations; if the annotations add further constraints, those are checked, too. For example, a
+function contract annotation provides contextual assumptions about a function as _preconditions_ to CBMC, and adds
+further _constraints_ for the program state at function return.
 
-2. The one outermost loop statement should be the _final_ statement before the function returns. Don't have complicated code _after_ the loop body.
+In mlkem-native, we use abbreviated forms of the CBMC annotations defined by macros in the [cbmc.h](../mlkem/cbmc.h). We
+now list the most prominent.
+
+### Function contracts
+
+A _function contract_ can be added where a function is declared. For `static` functions where the sites of
+declaration and definition coincide, it is part of the definition. mlkem-native's syntax for function contracts is
+
+```c
+int foo(args...)
+__contract(
+  requires(...)
+  ...
+  requires(...)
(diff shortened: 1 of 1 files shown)


cbmc/proof_guide.md
```diff
@@ -2,57 +2,151 @@
...
+function contract annotation provides contextual assumptions about a function as _preconditions_ to CBMC, and adds
+further _constraints_ for the program state at function return.
 
-2. The one outermost loop statement should be the _final_ statement before the function returns. Don't have complicated code _after_ the loop body.
+In mlkem-native, we use abbreviated forms of the CBMC annotations defined by macros in the [cbmc.h](../mlkem/cbmc.h). We
+now list the most prominent.
+
+### Function contracts
+
+A _function contract_ can be added when a function is declared or defined; typically, non-`static` functions should have
```
mkannwischer, line 31: Are you sure you can define a contract in the function definition? 
Sounds like a foodgun because if in a proof using a certain function contract, you do not include the definition, it would see an empty contract from the declaration assuming no pre-conditions, while the proof of the contract would see the pre-conditions.

Should this be rephrased or am I misunderstanding something?

cbmc/proof_guide.md
```diff
@@ -2,57 +2,151 @@
...
+```c
+int foo(args...)
+__contract(
+  requires(...)
+  ...
+  requires(...)
+  ensures(...)
+  ...
+  ensures(...)
+  assigns(...)
```
mkannwischer, line 43: I think the order does not matter, but we commonly have requires, assigns, ensures.

cbmc/proof_guide.md
```diff
@@ -2,57 +2,151 @@
...
+with care: If a function precondition specified `requires(memory_no_alias(ptr, 42))` and `assigns(object_whole(ptr))`
+and is called in a context where `ptr` is, say, a slice of some larger structure, then the entire structure will be
+marked as tained by the function call. This is often not desired, hence the more fine-grained `memory_slice(...)` is
+desirable.
+
+### Quantifiers and bounds
+
+If you need to specify a quantified condition for use in `ensures(...)` or `requires(...)`, you can use the
+`forall(...)` wrapper. It has the shape `forall(k, low_bound, high_bound, condition)`, where `k` is a name for the
+quantified variable `[low_bound, ..., high_bound-1]` the quantification range, and `conditon` is the quantified
```
mkannwischer, line 85:
```suggestion
quantified variable `[low_bound, ..., high_bound-1]` the quantification range, and `condition` is the quantified
```

cbmc/proof_guide.md
```diff
@@ -2,57 +2,151 @@
...
+`memory_no_alias(...)` for both, because `&foo[0]`, `&foo[1]` point to the same object. In mlkem-native, we sometimes
+work around this by manually splitting statically-sized arrays into multiple separate objects.
+
+### Memory footprint
+
+The most common way to specify memory footprint in `assigns(...)` clauses is via `memory_slice(ptr, len)`. This asserts
+that the code in question may change the first `len` bytes starting from `ptr`.
+
+There is also `object_whole(ptr)`, which more coarsely asserts that the entire object can change. This has to be used
+with care: If a function precondition specified `requires(memory_no_alias(ptr, 42))` and `assigns(object_whole(ptr))`
```
mkannwischer, line 76:
```suggestion
with care: If a function precondition specifies `requires(memory_no_alias(ptr, 42))` and `assigns(object_whole(ptr))`
```

cbmc/proof_guide.md
```diff
@@ -2,57 +2,151 @@
...
+  assigns(...)
+  invariant(...)
+  ...
+  invariant(...))
+{
+   ...
+}
+```
+
+Here, one of more `invariant(...)` clauses describe the invariant maintained by the loop body. As for function
```
mkannwischer, line 109:
```suggestion
Here, one or more `invariant(...)` clauses describe the invariant maintained by the loop body. As for function
```

cbmc/proof_guide.md
```diff
@@ -269,41 +347,47 @@ Using contracts, this harness function should not need to contain any CBMC `assu
...
-  ensures());
+  ensures())
+{
+   ...
+}
 ```
-with a final semi-colon on the end of the last one. That final semi-colon is needed to terminate the function declaration as per normal C syntax.
 
-### Interior contracts and loop invariants
+Note that when added to a declarationn, the contract has to come before the final semicolon concluding the declaration.
```
mkannwischer, line 374:
```suggestion
Note that when added to a declaration, the contract has to come before the final semicolon concluding the declaration.
```

mkannwischer (changes requested):
Thanks for updating the proof guide. 
This looks up to date now and some things are much clearer.

A few comments.

cbmc/proof_guide.md
hanno-becker: This is mainly for `static` functions which often have definition and declaration at the same time. But you're right, it's a bit misleading, I'll rephrase.

cbmc/proof_guide.md
hanno-becker: okay

cbmc/proof_guide.md
hanno-becker: Fixed

cbmc/proof_guide.md
hanno-becker: fixed

cbmc/proof_guide.md
hanno-becker: Fixed

mkannwischer (approved):
Thanks for the changes. This looks good to me now.

PR #637, opened 2025-01-11
CBMC: Switch to exclusive coefficient upper bounds

### LINKED ISSUE #324: Pick uniform convention for inclusive vs.exclusive bounds in runtime and CBMC bounds checking macros
We have debug-build runtime checks (e.g. `POLY_BOUND(...)`) as well as CBMC assertions `ARRAY_IN_BOUND(...)` both expressing bounds on an int16_t array. However, the runtime assertions use exclusive bounds, while the CBMC assertions use inclusive bounds. This is confusing and was close to leading to a poor spec in #https://github.com/pq-code-package/mlkem-c-aarch64/pull/313.

- [ ] Make a pick and use inclusive or exclusive bounds in both runtime and CBMC bounds checking macros.

hanno-becker:
* Resolves #324 

Previously, the coefficient bounds in `array_bound(...)` and `array_abs_bound(...)` were inclusive. This was somewhat inconvenient since (a) it required a lot of `-1` in places where the exclusive upper bound was more natural to work woth, and (b) it was a deviation from the runtime debug assertions `POLY_BOUND` which would use exclusive bounds.

This commit switches `array_bound` to use an exclusive upper bound. The lower bound remains inclusive. For `array_abs_bound`, we switch to an exclusive absolute bound.

The macro `UINT12_MAX` (mapping to 4095) is replaced by UINT12_LIMIT, mapping to 4096. Uses of `array_abs_bound(..., UINT12_MAX) are replaced by `array_bound(..., 0, UINT12_LIMIT)`, since we never deal with negative numbers bound by UINT12_LIMIT.

### DIFF
--- a/examples/monolithic_build/mlkem_native_monobuild.c
+++ b/examples/monolithic_build/mlkem_native_monobuild.c
@@ -1559,8 +1559,8 @@
 #endif
 
 /* mlkem/params.h */
-#if defined(UINT12_MAX)
-#undef UINT12_MAX
+#if defined(UINT12_LIMIT)
+#undef UINT12_LIMIT
 #endif
 
 /* mlkem/params.h */
--- a/mlkem/cbd.c
+++ b/mlkem/cbd.c
@@ -73,7 +73,7 @@ static void cbd2(poly *r, const uint8_t buf[2 * MLKEM_N / 4])
   for (i = 0; i < MLKEM_N / 8; i++)
   __loop__(
     invariant(i >= 0 && i <= MLKEM_N / 8)
-    invariant(array_abs_bound(r->coeffs, 0, 8 * i, 2)))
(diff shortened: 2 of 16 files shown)


mkannwischer (approved):
Thanks! The changes look good to me. 
I have double checked if you have replaced all the instances. You have left out a few in the intermediate functions in ntt.c, but since the CBMC proofs are still going through, this seems to be fine. Was that intentional?

mkannwischer:
> Thanks! The changes look good to me. I have double checked if you have replaced all the instances. You have left out a few in the intermediate functions in ntt.c, but since the CBMC proofs are still going through, this seems to be fine. Was that intentional?

Nevermind. I had another look - those instances do not have to be changed as the bound is actually passed as a parameter and it's already changed in the calll site.

PR #641, opened 2025-01-12
Simplify include paths for mlkem-native builds

hanno-becker:
Previously, a host of include paths was needed when building mlkem-native, which is inconvenient for the user.

With this commit, all #include's from mlkem-native source files are relative to their own base directory, and therefore do not need any include path.

A slight inconvenience is that this forces a custom FIPS202 backend to live in a directory named `fips202` for now. This limitation will be lifted in a subsequent commit allowing to configure the name of a custom FIPS202 header.

The change also sheds light on the lack of a separate include directory for mlkem-native. This will be addressed separately.


examples/mlkem_native_as_code_package/Makefile
```diff
@@ -22,13 +22,7 @@ MLKEM_NATIVE_SOURCE=$(wildcard          \
...
 	mlkem_native/**/**/**/*.c	\
 	mlkem_native/**/**/**/**/*.c)
-
-INC=
-INC+=-Imlkem_native/mlkem
-INC+=-Imlkem_native/mlkem
-INC+=-Imlkem_native/mlkem/native
-INC+=-Imlkem_native/mlkem/fips202
-INC+=-Imlkem_native/mlkem/fips202/native
+#INC=-Imlkem_native/mlkem
```
mkannwischer, line 25: remove?

mlkem/native/x86_64/src/arith_native_x86_64.h
```diff
@@ -5,17 +5,16 @@
...
 #include "consts.h"
-#include "fips202.h"
-#include "polyvec.h"
 
-#define REJ_UNIFORM_AVX_NBLOCKS \
-  ((12 * MLKEM_N / 8 * (1 << 12) / MLKEM_Q + SHAKE128_RATE) / SHAKE128_RATE)
-#define REJ_UNIFORM_AVX_BUFLEN (REJ_UNIFORM_AVX_NBLOCKS * SHAKE128_RATE)
+#define REJ_UNIFORM_AVX_NBLOCKS 3 /* See MLKEM_GEN_MATRIX_NBLOCKS */
+#define REJ_UNIFORM_AVX_BUFLEN \
+  (3 * 168) /* REJ_UNIFORM_AVX_BUFLEN * SHAKE128_RATE */
```
mkannwischer, line 17: I don't see how this change is related to this commit. Please do it in a separate commit.

mkannwischer (changes requested):
Looks most good to me. Two small nits.

examples/mlkem_native_as_code_package/Makefile
hanno-becker: Yes

mlkem/native/x86_64/src/arith_native_x86_64.h
hanno-becker: Fair point. This is to avoid reliance on the FIPS202 header.

mkannwischer: Ahh okay - the `SHAKE128_RATE` - I see.

mkannwischer (approved):

PR #645, opened 2025-01-12
Remove NTT_BOUND_NATIVE and INVNTT_BOUND_NATIVE

hanno-becker:
The numeric macros NTT_BOUND_NATIVE and INVNTT_BOUND_NATIVE could be used by backends to specify an output bound for the forward and inverse NTT. We would then use STATIC_ASSERT's to check that those backend-specific bounds would imply the bounds (NTT_BOUND and INVNTT_BOUND) that the CBMC proofs work against.

The added value of NTT_BOUND_NATIVE and INVNTT_BOUND_NATIVE is questionable. In the end, what we care about is that the native NTT and invNTT adhere to the bounds that CBMC relies on. Any strengthening of the bounds may be good to know, but is otherwise an unnecessary complication of the code.

This commit removes NTT_BOUND_NATIVE and INVNTT_BOUND_NATIVE and uses the contractual bounds NTT_BOUND and INVNT_BOUND in their place.

------

Also, remove the need for `STATIC_ASSERT` by rewriting its use using a plain `#if`.

### DIFF
--- a/mlkem/indcpa.c
+++ b/mlkem/indcpa.c
@@ -405,9 +405,10 @@ __contract__(
   }
 }
 
-
-
-STATIC_ASSERT(NTT_BOUND + MLKEM_Q < INT16_MAX, indcpa_enc_bound_0)
+#if NTT_BOUND > INT16_MAX - MLKEM_Q
+#error \
+    "The bound for the forward NTT is too large to ensure non-overflow in indcpa_keypair_derand"
+#endif
 
 MLKEM_NATIVE_INTERNAL_API
 void indcpa_keypair_derand(uint8_t pk[MLKEM_INDCPA_PUBLICKEYBYTES],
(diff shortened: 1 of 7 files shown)


mlkem/indcpa.c
```diff
@@ -405,9 +405,10 @@ __contract__(
   }
 }
 
-
-
-STATIC_ASSERT(NTT_BOUND + MLKEM_Q < INT16_MAX, indcpa_enc_bound_0)
+#if NTT_BOUND > INT16_MAX - MLKEM_Q
+#error \
+    "The bound for the forward NTT is too large to ensure non-overflow in indcpa_keypair_derand"
+#endif
```
mkannwischer, line 411: Isn't this something that the CBMC proofs should garuantee now? 
I'm tempted to remove all such bound assertations.

mkannwischer (commented):
I'm fine with removing the native bounds as there is no real value in them. 
Should we also move the static bound assertions?

mlkem/indcpa.c
hanno-becker: Yes, CBMC checks the same. Ok, let's remove it.

hanno-becker: Done

mkannwischer: thx

PR #651, opened 2025-01-13
Fix includes in benchmarks

mkannwischer:
#641 forgot to adjust the benchmarks. This PR fixes that. 

I also extended the quickcheck to additionally run the benchmarks without counting cycles to avoid such errors in the future.


hanno-becker (approved):
Looks good -- let's wait for CI

PR #654, opened 2025-01-13
Add AWS-LC integration test

hanno-becker:
This commit extends the CI with an integration test for AWS-LC.

The CI checks that mlkem-native successfully integrates into
AWS-LC by building and running basic tests of AWS-LC in FIPS
and non-FIPS mode.

We test both against a fixed commit and main. Testing against a
fixed known-good commit in AWS-LC means we note if we break the
integration. Testing against main means we note if a change in
AWS-LC breaks the integration. The latter is out of our control,
and we may need to temporarily ignore failures, but we should
be aware.

### DIFF
--- a/.github/actions/setup-aws-lc/action.yml
+++ b/.github/actions/setup-aws-lc/action.yml
@@ -1,0 +1,34 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Setup AWS-LC
+description: Setup AWS-LC
+
+inputs:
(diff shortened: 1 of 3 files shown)


mkannwischer (approved):
Awesome! Thanks @hanno-becker for all the work onn the AWS-LC integration!!

This PR adding the integration tests looks good. It may be a bit excessive, but we can trim it down later in case CI is running too slowly.

Issue #652, opened 2025-01-13
`tests bench --opt ALL -c NO` not executing no_opt benchmarks


mkannwischer:
This is not what I'm expectin from `--opt ALL`

```
$ tests bench --opt ALL -c NO
INFO  > Benchmark          Compile     (native opt):     make bench OPT=1 AUTO=1 CYCLES=NO -j16
INFO  > Benchmark          ML-KEM-512  (native opt):     make run_bench_512
INFO  > Benchmark          ML-KEM-512  (native opt):     test/build/mlkem512/bin/bench_mlkem512
   keypair cycles = 0
    encaps cycles = 0
    decaps cycles = 0

           percentile      1     10     20     30     40     50     60     70     80     90     99
   keypair percentiles:      0      0      0      0      0      0      0      0      0      0      0
    encaps percentiles:      0      0      0      0      0      0      0      0      0      0      0
    decaps percentiles:      0      0      0      0      0      0      0      0      0      0      0

INFO  > Benchmark          ML-KEM-768  (native opt):     make run_bench_768
INFO  > Benchmark          ML-KEM-768  (native opt):     test/build/mlkem768/bin/bench_mlkem768
   keypair cycles = 0
    encaps cycles = 0
    decaps cycles = 0

           percentile      1     10     20     30     40     50     60     70     80     90     99
   keypair percentiles:      0      0      0      0      0      0      0      0      0      0      0
    encaps percentiles:      0      0      0      0      0      0      0      0      0      0      0
    decaps percentiles:      0      0      0      0      0      0      0      0      0      0      0

INFO  > Benchmark          ML-KEM-1024 (native opt):     make run_bench_1024
INFO  > Benchmark          ML-KEM-1024 (native opt):     test/build/mlkem1024/bin/bench_mlkem1024
   keypair cycles = 0
    encaps cycles = 0
    decaps cycles = 0

           percentile      1     10     20     30     40     50     60     70     80     90     99
   keypair percentiles:      0      0      0      0      0      0      0      0      0      0      0
    encaps percentiles:      0      0      0      0      0      0      0      0      0      0      0
    decaps percentiles:      0      0      0      0      0      0      0      0      0      0      0
    ````

PR #657, opened 2025-01-13
Fix `tests bench [--opt=ALL]` so it runs opt and non-opt benchmarks

hanno-becker:
* Fixes https://github.com/pq-code-package/mlkem-native/issues/652

Previously, `tests bench` and `tests bench --opt=ALL`  would only run the opt benchmarks. This was because of a bug in the helper `do_opt_all()` in the `tests` script. This commit fixes this bug, thereby restoring the desired behaviour.

### DIFF
--- a/scripts/tests
+++ b/scripts/tests
@@ -323,7 +323,7 @@ class Tests:
         return [f"-j{self.args.j}"]
 
     def do_opt_all(self):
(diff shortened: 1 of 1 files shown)


mkannwischer (approved):
LGTM - it does what it should now on my machine. 
Re-running benchmarking CI just to be sure everything is fine.

PR #659, opened 2025-01-14
autogen: Add checker for dangling register aliases in assembly files

### LINKED ISSUE #658: Always `.unreq` register aliases at the end of assembly routines
The `.req` primitive is heavily used in mlkem-native's AArch64 assembly for register aliases.

Since compilers usually just warn but not error out if a `.req` is re-written with a different alias, it is good hygiene to `.unreq` all register aliases once an assembly function has been defined. Otherwise, there is risk of things breaking in a hard-to-debug way when merging assembly files or having more than one assembly function per file.

- [x] Adjust all assembly files to have matching `.req`and `.unreq` directives
- [x] Add a test script to CI which flags dangling `.req`s in assembly files

hanno-becker:
* Fixes #658 

The .req primitive is heavily used in mlkem-native's AArch64 assembly for register aliases.

Since compilers usually just warn but not error out if a .req is re-written with a different alias, it is good hygiene to .unreq all register aliases once an assembly function has been defined. Otherwise, there is risk of things breaking in a hard-to-debug way when merging assembly files or having more than one assembly function per file.

This PR extends the `autogen` script to check that register aliases in assembly files are always undefined by the end of the file. If they are not, it suggests a block of code to be added. The code is not added automatically because the right placement isn't obvious to infer.


hanno-becker:
> I do not follow this one. Numberic labels do not hinder merging multiple assembly files into one. 1b and 1f uniquely reference the previous/next label called 1.

I seem to remember that there were compilation errors upon multiple numeric labels with the same name, but I may be wrong.

mkannwischer:
> > I do not follow this one. Numberic labels do not hinder merging multiple assembly files into one. 1b and 1f uniquely reference the previous/next label called 1.
> 
> I seem to remember that there were compilation errors upon multiple numeric labels with the same name, but I may be wrong.

I've many times used multiple `1` labels in a single file without issues. 
That's the only way to define a label within a macro that you want to use multiple times. 
Can you double check this please? 
It's not that I want numeric labels here, but I think checking for that is unnecessary.

hanno-becker:
> I've many times used multiple 1 labels in a single file without issues.
> That's the only way to define a label within a macro that you want to use multiple times.
> Can you double check this please?
> It's not that I want numeric labels here, but I think checking for that is unnecessary.

You are right. Duplicated numeric labels are OK, and don't even need namespacing. However, they are harder to read, and e.g. s2n-bignum/aws appear to convert to namespaced non-numeric labels (https://github.com/awslabs/s2n-bignum/pull/177). I'd suggest we stick with that for the time being as well?

mkannwischer (approved):
Thanks - this now looks good!

PR #660, opened 2025-01-14
Add monobuild with native backend

hanno-becker:
* Based on #659 

This PR adds support for a 'monobuild' with native backend, using a single *.c and a single *.S file.

In addition to #659, automatic namespace generation was needed for assembly labels and macros. At this point, the name-spacing is permanently added to the assembly files themselves. We may instead want to emit it in the monobuild file only -- this is up for discussion (cc @mkannwischer).


hanno-becker:
@mkannwischer This is not yet ready for merge, but I'd appreciate your input on the approach taken. In particular, I'd like to know a) how you feel about the label and macro namespacing and where it should go, b) if you have an idea to cleanly resolve the unused function warning which is just silenced in the last commit.

examples/monolithic_build_native/mlkem_native_monobuild.S
```diff
@@ -0,0 +1,4452 @@
...
+
+/* mlkem/native/aarch64/src/intt_clean.S */
+#if defined(push_stack)
+#undef push_stack
+#endif
+
+/* mlkem/native/aarch64/src/intt_clean.S */
+#if defined(restore_vregs)
+#undef restore_vregs
+#endif
```
mkannwischer, line 109: didn't you namespace those?

hanno-becker: Yes, but what you undef here is the name of the macro.

hanno-becker:
@mkannwischer Note the last commit where I couldn't help but hoist the poly* structs into a new header `structs.h` to avoid cyclic dependencies between `poly.h` and `arith_backend.h`. Maybe you have a better idea?

hanno-becker:
I misunderstood how .S files are integrated into AWS-LC, which was the primary driver for this change. A monolithic assembly file may not be needed. Closing this for now.

PR #662, opened 2025-01-15
Package hol_light and s2n-bignum into nix derivations

potsrevennil:
**Summary**:

**Steps**:

**Performed local tests**:
 - [ ] `lint` passing
 - [ ] `tests all` passing
 - [ ] `tests bench` passing
 - [ ] `tests cbmc` passing


proofs/hol_light/arm/Makefile
```diff
@@ -0,0 +1,503 @@
...
+OBJ = $(POINT_OBJ) $(BIGNUM_OBJ)
+
+# According to
+# https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms,
+# x18 should not be used for Apple platforms. Check this using grep.
+
+%.o : $(SRC_ARM)/%.S
+%.o :
+	$(Q)[ -d $(@D) ] || mkdir -p $(@D)
+	cat $(SRC_ARM)/$@ | $(PREPROCESS) | $(SPLIT) | grep -v -E '^\s+.quad\s+0x[0-9a-f]+$$' | $(ASSEMBLE) -o $@ -
```
hanno-becker, line 404: This does not seem to work -- do you mean `cat $(SRC_ARM)/$<`?

hanno-becker (changes requested):
Something is still fishy in the Makefile target for object files -- see comments.

Could you also add a brief `README` that explains what one should do to fully compile + run a proof? Is it feasible to run at least one proof in CI to confirm that the setup is functional? (e.g. a "hol-light sanity check" job or similar)

proofs/hol_light/arm/Makefile
potsrevennil: Hi, Thanks for the check, seems like I mistakenly change `$<` to `$@`
But I'm now very confused, as either one works on my machine...

potsrevennil: According to what I know now, there're some built-in implicit makefile rules involved 
Cancelling them is possible, but doesn't seem to be applicable to our case
(ref: https://www.gnu.org/software/make/manual/html_node/Canceling-Rules.html)

However we can make the rule explicit by using static pattern rules (ref: https://www.gnu.org/software/make/manual/make.html#Static-Pattern):
```
$(OBJ): %.o: $(SRC_ARM)/%.S:
   ...
```

hanno-becker (changes requested):
Thank you @potsrevennil for fixing the Makefile issue.

Something is still not right, though:
```
make -j8 p256/bignum_montmul_p256.correct
```
continues to fail with
```
Fatal error: exception Failure("dest_cons4: 4-byte inst code 2839547939 != first 4 bytes of []")
make: *** [Makefile:440: p256/bignum_montmul_p256.correct] Error 2
```

proofs/hol_light/arm/Makefile
hanno-becker: Thank you for figuring that out!

rod-chapman:
I have tried to reproduce proof results on macOS. In short:
1. pulled and checked out this branch.
2. Re-started NIX
3. cd proofs/hol_light/arm
4. make
terminates with
```
Fatal error: exception Failure("not an ELF file")
mlkem/mlkem_ntt.native had errors!
make: *** [Makefile:112: mlkem/mlkem_ntt.correct] Error 1
```

rod-chapman:
I note that
```
$ file mlkem/mlkem_ntt.o
mlkem/mlkem_ntt.o: Mach-O 64-bit arm64 object
```

which is obviously "not an ELF file". Is this the problem?

hanno-becker:
Yes, s2n-bignum/HOL-Light's object file parser only supports ELF

rod-chapman:
To confirm this PR is working OK, we need a testcase which is much smaller and faster than either NTT or INTT. What's the smallest, simplest, fastest proof that we can add?

hanno-becker:
We've tested small examples before. Now the PR is targeting the actual ML-KEM proofs.

rod-chapman:
I have reproduced the proofs successfully on an EC2 Graviton 3 (c7g) machine, running Ubuntu 24.04.

hanno-becker (approved):
Thanks @potsrevennil for the work!

PR #663, opened 2025-01-15
Debug: Reduce number of macros, assimilate to CBMC assertions

hanno-becker:
Previously, debug.h provided a host of bounds checking assertions, such as POLY_BOUND, POLY_UBOUND, POLY_BOUND_MSG, POLYVEC_BOUND, ...

This commit simplifies debug.h by reducing to three assertions
- debug_assert: For any boolean valued assertion
- debug_assert_bound: For a lower+upper bounds assertion on an array, akin to the CBMC assertion array_bound
- debug_assert_abs_bound: For an absolute bounds assertion on an array, akin to the CBMC assertion array_abs_bound.

Moreover, assertion strings are removed: The filename and linenumber already provide enough information to locate the issue when an assertion fails -- the additional text clutters the code and doesn't add much.

While adjusting the code, two smaller issues were noted and fixed:
- The debug assertions had gone out of sync with the CBMC specs for the basemul functions: The CBMC spec would assert that the a-input is coefficient-wise in [0..4095], while the debug assertions would assert an absolute bound. This is unified by using bound by [0..4095] as proved correct by CBMC.
- In `poly_basemul_montgomery_cached`, a debug assertion was placed on the mulcache what should have been placed on the a-input. This is fixed.
- The documentation of `poly_basemul_montgomery_cached` had gone out of sync with the CBMC assertions: The documentation still stated the older output bound by 3/2q, while the CBMC assertions showed only 2q. This change stemmed from the recent relaxation of input bounds from MLKEM_Q -> 4096. Also, there was no debug-assertion for the post-condition, which is added.

### DIFF
--- a/mlkem/polyvec.c
+++ b/mlkem/polyvec.c
@@ -15,7 +15,7 @@ void polyvec_compress_du(uint8_t r[MLKEM_POLYVECCOMPRESSEDBYTES_DU],
                          const polyvec *a)
 {
   unsigned i;
-  POLYVEC_UBOUND(a, MLKEM_Q);
+  debug_assert_bound(a, MLKEM_K * MLKEM_N, 0, MLKEM_Q);
 
   for (i = 0; i < MLKEM_K; i++)
   {
@@ -33,7 +33,7 @@ void polyvec_decompress_du(polyvec *r,
     poly_decompress_du(&r->vec[i], a + i * MLKEM_POLYCOMPRESSEDBYTES_DU);
   }
 
-  POLYVEC_UBOUND(r, MLKEM_Q);
+  debug_assert_bound(r, MLKEM_K * MLKEM_N, 0, MLKEM_Q);
 }
 
 MLKEM_NATIVE_INTERNAL_API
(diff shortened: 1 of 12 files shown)


mlkem/polyvec.c
```diff
@@ -85,9 +85,9 @@ void polyvec_basemul_acc_montgomery_cached(poly *r, const polyvec *a,
   unsigned i;
   poly t;
 
-  POLYVEC_BOUND(a, 4096);
-  POLYVEC_BOUND(b, NTT_BOUND);
-  POLYVEC_BOUND(b_cache, MLKEM_Q);
```
mkannwischer, line 90: This should have never worked .... thanks for fixing that.

mkannwischer (approved):
Thanks - looks much clearner.

mlkem/polyvec.c
hanno-becker: It does work, but it is a mismatch with the precondition which requires nothing about b and b_cache

PR #664, opened 2025-01-15
CBMC: Remove redundant `0 <= ...` assertions in loop invariants

hanno-becker:
Previously, we were using `int` as the type of loop indices, which requires loop invariants of the form `0 <= i`.

With the recent switch to `unsigned` indices, those invariants are automatic and can be removed for improved readability. This commit does that.

Along the way, it is noticed that some of the NTT proofs still used integral loop indices. Those are changed along the way.


mkannwischer (changes requested):
Please also adjust the proof guide. 
There are still some `0 <=` in there

mkannwischer (approved):

PR #665, opened 2025-01-15
autogen: Don't generate multiple #undef's for the same macro

hanno-becker:
Previously, the generation of the monobuild compilation unit using `autogen` would generate one `#undef` clause for every `#define` found in a file. If a single macro was defined multiple times in different branches, multiple #undef's would be generated.

This commit fixes `autogen` to only emit one `#undef ...` clause per macro per file. This is done by omitting the no longer used line number information from the function yielding the defines. Since the function is gathered into a set, this automatically removes duplicates.


mkannwischer (approved):
Great - thank you.

PR #667, opened 2025-01-17
AArch64: Remove literal pools from native code

hanno-becker:
This commit removes all literal pools from the native AArch64 assembly. Those literal pools are slightly easier to read, but impede verification using HOL-Light. Instead, constant vectors are prepared by loading immediates into GPRs and copying/broadcasting them into the target vector.


mkannwischer (approved):
Changes look good. I'd like to see one complete benchmarking run after your last force push just to be sure.

PR #672, opened 2025-01-20
Remove FIPS202_NAMESPACE_PREFIX configuration option

hanno-becker:
Previously, the namespace for FIPS202 symbols could be separately defined from the namespace for MLKEM-specific symbols. This was primarily to allow for a multi-level build where FIPS202 code is shared.

Now that multi-level builds are possible even with sharing of all level-independent code (not merely FIPS202), there is no longer a use of a separate FIPS202 namespace. Note also that when a custom FIPS202 implementation is used, the user is in full control of namespacing since they are expected to replace the FIPS202 headers.

In light of the above, this commit merges the FIPS202_NAMESPACE_PREFIX configuration option into MLKEM_NAMESPACE_PREFIX. The latter now controls prefixing of _all_ global symbols in an mlkem-native build.

If users want to share common code in a multilevel build, this should be done by setting MLKEM_NAMESPACE_PREFIX_ADD_LEVEL, and dropping the level from the MLKEM_NAMESPACE_PREFIX; all this is already demonstrated in the example `example/multilevel_build`. It is therefore no longer necessary to demonstrate a shared-FIPS202 build in the `Makefile`, which was done so far. Instead, the default build can be simplified to three independent builds of libmlkem{512,768,1024}, each with their own FIPS202. Again, usage examples are now provided in `examples/*`, while the user should not need to inspect the test Makefiles.


mkannwischer (approved):
LGTM.

PR #673, opened 2025-01-20
Move mlkem/debug/* into mlkem/*

hanno-becker:
The subdirectory mlkem/debug/* provides debugging functionality via debug.h and debug.c. However, contrary to other subdirectories which can be omitted from the build with appropriate configuration -- e.g. the in-built FIPS202 being replaced by a custom one, or the native backends removes because only the C backend is used -- the `debug` directory can not just be removed: At least `debug.h` is needed.

Since first integration attempts of mlkem-native have suggested that the deeply nested organization of mlkem-native is sometimes cumbersome to work with, this commit inlines mlkem/debug/* into mlkem/*.


mkannwischer (changes requested):
Please remove `$(wildcard mlkem/debug/*.c)` from components.mk

mkannwischer (approved):

PR #674, opened 2025-01-21
Introduce separate source files for compression, sampling, and arithmetic

hanno-becker:
* Based on #673 

This PR merges `cbd.*` and `rej_uniform.*` into `sampling.*`, now containing all generic sampling code. It also moves `ntt.*` and `reduce.h` into `poly.*`, now containing all arithmetic. Finally, compression-related routines are moved from `poly.*` into newly created `compress.*`.


hanno-becker:
@mkannwischer Request-for-comment. If we do this, we should also find a suitable new name for `polyvec.*` which is a misnomer by now.

mkannwischer (changes requested):
Let's do all restructuring in one go: 
I'm okay with your separation by compression, sampling, and arithmetic. 
For the k-dependent wrappers in `polyvec.c`, I suggest something like `poly_k.c`? 
The `polyvec` type can be moved to poly.h?

hanno-becker:
> For the k-dependent wrappers in polyvec.c, I suggest something like poly_k.c?

Ok

> The polyvec type can be moved to poly.h?

That doesn't work, poly.h is independent of k

hanno-becker:
@mkannwischer Done

mkannwischer (approved):
There is a very tiny performance regression for the optimized backend on c7i. Seems to be reproducible (+4%), but very close to usual noise (+-3%).
I'm not sure how moving around code like this can result in this much performance difference, but this is very likely very dependent on the compiler (version). 
I think we can merge this restructuring regardless.

PR #675, opened 2025-01-21
Backend API: Remove dependence on poly_* structs

hanno-becker:
* Based on #674 

This PR modifies the backend API to work directly with int16_t arrays rather than the poly_xxx structures.


mkannwischer (approved):
I reviewed the last commit - that is the only one that is not already in #674. LGTM.
Thanks for cleaning this up - this gave me a headache before as you can easily run into circular dependencies.

PR #677, opened 2025-01-21
NTT: Simplify computation of zeta index

### LINKED ISSUE #828: Remove scan-build false positive
A plain `scan-build make OPT=0` fails, complaining about a non-issue in `poly.c` about the loading of the zeta table. Yet, we need to find a way to make `scan-build` happy.

hanno-becker:
The n-th layer of the NTT needs the zeta table entries at indices `2^(n-1)..2^n-1`. Previously, the initial index was computed using a division by the `len` parameter, upholding the status of the `layer` parameter as a pure ghost variable only needed for the specification, but not the code itself.

This commit gives up the status of `layer` as a ghost variable and instead uses it to compute the initial zeta index using a shift instead of a division.

While `len` is not secret and there is, thus, no risk of information leakage when a `div` instruction is used, it still seems cleaner and faster to work with a shift.

* Resolves #828 

Suggested-by: @rod-chapman

### DIFF
--- a/mlkem/poly.c
+++ b/mlkem/poly.c
@@ -349,29 +349,22 @@ __contract__(
  * Compute one layer of forward NTT
  * Parameters:
  * - r: Pointer to base of polynomial
- * - len: Stride of butterflies in this layer.
- * - layer: Ghost variable indicating which layer is being applied.
- *          Must match `len` via `len == MLKEM_N >> layer`.
- * Note: `len` could be dropped and computed in the function, but
- *   we are following the structure of the reference NTT from the
- *   official Kyber implementation here, merely adding `layer` as
- *   a ghost variable for the specifications.
+ * - layer: Variable indicating which layer is being applied.
  */
 
 /* Reference: Embedded in `ntt()` in the reference implementation. */
-static void mlk_ntt_layer(int16_t r[MLKEM_N], unsigned len, unsigned layer)
+static void mlk_ntt_layer(int16_t r[MLKEM_N], unsigned layer)
 __contract__(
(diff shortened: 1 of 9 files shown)


rod-chapman:
CI looks good. Happy to merge.

mlkem/ntt.c
```diff
@@ -172,9 +170,9 @@ __contract__(
   ensures(array_abs_bound(r, 0, MLKEM_N, MLKEM_Q)))
 {
   unsigned start, k;
-  /* `layer` is a ghost variable used only in the specification */
-  ((void)layer);
-  k = MLKEM_N / len - 1;
+  /* Twiddle factors for layer n are at indices 2^(n-1)..2^n-1.
+   * The inverse NTT uses them in backwards order. */
+  k = (1u << layer) - 1;
```
rod-chapman, line 175: Looks good. CI is clean.

rod-chapman (approved):
A welcome simplification, and also removes the odd "ghost" status the layer parameter.

mkannwischer (changes requested):
Since layer is now no longer a ghost variable, please adjust the comments above the function accordingly.

mlkem/poly.c
```diff
@@ -397,17 +396,17 @@ __contract__(
 MLK_INTERNAL_API
```
hanno-becker, line 396: @mkannwischer Can you add a comment here? e.g.
```
/* Reference: `ntt()` in the reference implementation.
 * - Iterate over `layer` instead of `len` in the outer loop */
```

mkannwischer: done.

mkannwischer (approved):
LGTM

PR #679, opened 2025-01-22
Integration: Add liboqs META files

mkannwischer:
liboqs requires META files providing meta data on the available implementations and required flags for building them.
To ease integration, we include the files in mlkem-native.

This commit also adds script that checks that the source list is kept up to date and runs it in CI.

### DIFF
--- a/nix/util.nix
+++ b/nix/util.nix
@@ -43,6 +43,7 @@ rec {
       native-gcc
     ]
     ++ builtins.attrValues {
+      inherit (pkgs.python3Packages) pyyaml;
       inherit (pkgs)
         python3
         qemu; # 8.2.4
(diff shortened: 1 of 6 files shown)


mkannwischer:
@bhess, does this help?

nix/util.nix
```diff
@@ -43,6 +43,7 @@ rec {
       native-gcc
     ]
     ++ builtins.attrValues {
+      inherit (pkgs.python312Packages) pyyaml;
```
hanno-becker, line 46: `python3Packages`?

bhess (approved):
Thank you @mkannwischer, this helps a lot with the liboqs integration!

hanno-becker (approved):

PR #683, opened 2025-01-22
Add optimized NTT and invNTT in C

rod-chapman:
Re-introduces fast NTT C code, following source re-organization in #674 

See comments on older #610 . This commit addresses comments in that PR.

All tests, proofs, and lint OK.


rod-chapman:
Code size metrics using gcc 14.2.0 on Apple M1
gcc -c -O3 poly.c
size poly.o

yields

main: 4368 bytes
fastntt3: 8320 bytes

so increase of 3952 bytes

rod-chapman:
The benchmarks seem to be a bit better again following adoption of GCC 14...

rod-chapman:
What compiler is being used to run the "bench" tests on the Banana PI F3? Is it GCC 14.something?

hanno-becker:
@rod-chapman In doubt, you can always look up the CI actions. Yes, it's GCC-14 as installed into the default nix shell.

hanno-becker:
Closing this for now. It is unlikely we will merge this version of the NTT. We may consider a more general approach such as https://github.com/pq-code-package/mlkem-native/pull/784, but in general, improvements to the NTT are now a priority at the moment.

PR #684, opened 2025-01-22
Make arithmetic and FIPS202 backends independently configurable

hanno-becker:
Previously, the configuration option MLKEM_USE_NATIVE would guard whether native backends would be used or not. Specifically, it would control both the arithmetic backend and the FIPS202 backend.

The specific choice of backend would then be provided through MLKEM_NATIVE_{ARITH,FIPS202}_BACKEND.

This setup does allow the user to use a native arithmetic backend but no native FIPS202 backend -- however, it requires a custom config, since the default config would auto-define MLKEM_NATIVE_FIPS202_BACKEND when unset -- this cannot be altered through CFLAGS.

To make everything configurable through CFLAGS as well as a custom config file, this commit explicitly separates MLKEM_USE_NATIVE in two configurations options

- MLKEM_USE_NATIVE_BACKEND_ARITH
- MLKEM_USE_NATIVE_BACKEND_FIPS202,

which control whether an arithmetic resp. FIPS202 native backend should be used.

When those options are set, MLKEM_NATIVE_{ARITH,FIPS202}_BACKEND_FILE must either be unset -- in which case the default backends will be used -- or point to a backend metadata file.


hanno-becker:
@bhess Does this work for you?

bhess (approved):
I gave it a try, and it works, thank you for adding this feature!

mkannwischer (approved):

PR #687, opened 2025-01-23
Add Valgrind-based constant-time tests

### LINKED ISSUE #462: Run valgrind-based constant-time tests in CI
#460 hardend the functions in `verify.c` that have to be constant time to avoid timing attacks. 
We should test that this code (and all other code in mlkem-native operating on secrets) is actually constant time. 

I propose we use the [standard valgrind-based constant-time tests](https://github.com/agl/ctgrind)  with the patch from https://kyberslash.cr.yp.to/papers.html for detecting secret dependent divisons. 
This should be run in CI for as many compilers as possible with as many flag combinations as possible.

mkannwischer:
Resolves #462 

This extends the mlkem-native tests to support Valgrind-based constant-time tests following the ct-grind approach proposed by Adam Langley in https://github.com/agl/ctgrind.

The basic idea is to mark all output of randombytes as undefined (=secret). Valgrind will then track all values derived from that through the program and warn about any conditional branches and memory addressed depening on it.

In Kyber this leads to multiple false positives (i.e., branches depending on values that are actually public), that we resolve by explicitly declassifying that data (i.e., marking it defined). The cases where this happens are as follows:
1) In key generation: The expansion of the public matrix A uses rejection
   sampling which contains conditional branches. The seed used for that
   is output of randombytes, and, hence, Valgrind complains.
   However, this seed is public and part of the public ke.y
   We declassify the seed before calling gen_matrix.
2) In encapsulation: The same problem regarding gen_matrix occurs.
   We resolve this by declassifying the entire public key.
3) In encapsulation: Public key modulus check requires to check
   that each coefficient of the public key is correctly encoded.
   Currently this is implemented using a memcmp which has input-dependent
   timing. Declassifying the entire public key (as already done for
   2), resolves this already.
4) In decapsulation: The public key hash checks requires recomputing
   the hash of the public key and comparing it with the pre-computed hash.
   As both the (INDCPA) public key and the pre-computed hash are part
   of the secret key (and, hence, marked secret), we have to declassify
   both the public key and the hash of the public key.
5) In decapsulation: In re-encryption the matrix A needs to be expanded
   from the public seed using rejection sampling. We declassify the seed.
   (Technically, this is already covered by 4, but as we eventually
   may make the public key hash check optional, I opted for declassifying again)

We are testing
{x86 Ubuntu, Arm64 Ubuntu}
x
{gcc4,gcc7,gcc11,gcc12,gcc13,gcc14,clang14,clang15,clang16,clang17,clang18,clang19}
x
{-Oz,-Os,-O3,-O2,-O1,-O0,-Ofast,-O3 -ffast-math}
x
{C backend, native backend}

What is still left to do is adding a patched Valgrind covering variable-latency instructions from https://kyberslash.cr.yp.to/papers.html. I'll do that as a follow up.

### DIFF
--- a/.github/actions/ct-test/action.yml
+++ b/.github/actions/ct-test/action.yml
@@ -1,0 +1,17 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: CT Test
+description: Builds the library with given flags and runs Valgrind-based constant-time tests
+
+inputs:
+  cflags:
+    description: CFLAGS to pass to compilation
+    default: ""
+
+runs:
+  using: composite
+  steps:
+      - shell: ${{ env.SHELL }}
+        run: |
+          make clean
+          tests func --exec-wrapper="valgrind --error-exitcode=1 --track-origins=yes" --cflags="-DENABLE_CT_TESTING -DNTESTS=50 ${{ inputs.cflags }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -298,6 +298,9 @@ jobs:
          - name: clang-18
            shell: ci_clang18
            darwin: True
+         - name: clang-19
+           shell: ci_clang19
+           darwin: True
     runs-on: ${{ matrix.target.runner }}
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -312,7 +312,6 @@ jobs:
           kat: false
           acvp: false
           nix-shell: ${{ matrix.compiler.shell }}
-          cflags: "-O1" # _FORTIFY_SOURCE requires compiling with optimization
       - name: native build+functest (C90)
         if: ${{ matrix.compiler.darwin || matrix.target.runner != 'macos-latest' }}
         uses: ./.github/actions/multi-functest
@@ -324,7 +323,7 @@ jobs:
           kat: false
           acvp: false
           nix-shell: ${{ matrix.compiler.shell }}
-          cflags: "-O1 -std=c90" # _FORTIFY_SOURCE requires compiling with optimization
+          cflags: "-std=c90"
       - name: native build+functest (C99)
         if: ${{ matrix.compiler.darwin || matrix.target.runner != 'macos-latest' }}
         uses: ./.github/actions/multi-functest
@@ -336,7 +335,7 @@ jobs:
           kat: false
           acvp: false
           nix-shell: ${{ matrix.compiler.shell }}
-          cflags: "-O1 -std=c99" # _FORTIFY_SOURCE requires compiling with optimization
+          cflags: "-std=c99"
       - name: native build+functest (C11)
         if: ${{ matrix.compiler.darwin || matrix.target.runner != 'macos-latest' }}
         uses: ./.github/actions/multi-functest
@@ -348,7 +347,7 @@ jobs:
           kat: false
           acvp: false
           nix-shell: ${{ matrix.compiler.shell }}
-          cflags: "-O1 -std=c11" # _FORTIFY_SOURCE requires compiling with optimization
+          cflags: "-std=c11"
       - name: native build+functest (C17)
         if: ${{ (matrix.compiler.darwin || matrix.target.runner != 'macos-latest') &&
                 matrix.compiler.c17 }}
@@ -361,7 +360,7 @@ jobs:
           kat: false
           acvp: false
           nix-shell: ${{ matrix.compiler.shell }}
-          cflags: "-O1 -std=c17" # _FORTIFY_SOURCE requires compiling with optimization
+          cflags: "-std=c17"
   # The purpose of this job is to test non-default yet valid configurations
   config_variations:
     name: Non-standard configurations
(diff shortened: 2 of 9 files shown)


mkannwischer:
This is ready for review now. 
I temporarily added the [clangover](https://github.com/antoonpurnal/clangover) vulnerability here:  https://github.com/pq-code-package/mlkem-native/commit/f7e04bee5b0ce6406f1b1a9a57c5f5c296eb4a23. 
You can see the run here: https://github.com/pq-code-package/mlkem-native/actions/runs/12926095230/job/36048394763?pr=687

Exactly as one would expect the tests catch clang 15 and newer producing vulnerable code for x86.

.github/actions/ct-test/action.yml
```diff
@@ -0,0 +1,27 @@
...
+    default: ""
+
+runs:
+  using: composite
+  steps:
+      - shell: ${{ env.SHELL }}
+        run: |
+          make clean
+          OPT=1 CFLAGS="-DENABLE_CT_TESTING -DNTESTS=50 ${{ inputs.cflags }}" make func
+          valgrind --error-exitcode=1 --track-origins=yes ./test/build/mlkem512/bin/test_mlkem512
```
hanno-becker, line 18: Nit: You could simplify this by using `tests func --exec-wrapper ...` I think.

flake.nix
```diff
@@ -44,11 +45,27 @@
...
 
           devShells.ci_clang18 = util.wrapShell (util.mkShellWithCC pkgs.clang_18) { packages = [ pkgs.python3 ]; };
+          devShells.ci_clang19 = util.wrapShell (util.mkShellWithCC inputs.nixpkgs2411.legacyPackages.${system}.clang_19) { packages = [ pkgs.python3 ]; };
           devShells.ci_gcc48 = util.wrapShell (util.mkShellWithCC pkgs.gcc48) { packages = [ pkgs.python3 ]; };
           devShells.ci_gcc49 = util.wrapShell (util.mkShellWithCC pkgs.gcc49) { packages = [ pkgs.python3 ]; };
           devShells.ci_gcc7 = util.wrapShell (util.mkShellWithCC pkgs.gcc7) { packages = [ pkgs.python3 ]; };
           devShells.ci_gcc11 = util.wrapShell (util.mkShellWithCC pkgs.gcc11) { packages = [ pkgs.python3 ]; };
           devShells.ci_gcc14 = util.wrapShell (util.mkShellWithCC pkgs.gcc14) { packages = [ pkgs.python3 ]; };
+
+          devShells.ci_valgrind_clang14 = util.wrapShell (util.mkShellWithCC pkgs.clang_14) { packages = [ pkgs.valgrind ]; hardeningDisable = [ "fortify" ]; };
```
hanno-becker, line 55: I'm not sure this needs a separate shell; it's a bug in nix that causes FORTIFY_SOURCE to be set unconditionally, right? Normally it wouldn't be, so I'd suggest to unset it in the default shells and add a pointer to the respective nix issue.

hanno-becker (changes requested):
Thanks @mkannwischer, this is a valuable addition!

Two smaller points on the nix and CI setup.

flake.nix
mkannwischer: The main reason for the extra shell is the additional valgrind dependency. The fortify stuff indeed needs to be added to the other shells as well - I will do that shortly in a separate PR.

We could install valgrind always if you insist - then we need half the shells.

hanno-becker: Yes, always having valgrind makes sense to me

mkannwischer: Okay, I've added it everywhere and removed the extra shells.

.github/actions/ct-test/action.yml
mkannwischer: Trying that now.

.github/workflows/ci.yml
```diff
@@ -348,7 +348,6 @@ jobs:
           kat: false
           acvp: false
           nix-shell: ${{ matrix.compiler.shell }}
-          cflags: "-O1 -std=c11" # _FORTIFY_SOURCE requires compiling with optimization
```
hanno-becker, line 351: The `-std=...` lines have gone missing here

.github/workflows/ct-tests.yml
```diff
@@ -0,0 +1,61 @@
...
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Setup nix
+        uses: ./.github/actions/setup-shell
+        with:
+          gh_token: ${{ secrets.GITHUB_TOKEN }}
+          nix-shell: ${{ matrix.nix-shell }}
+      - name:  Build and run test (-Os)
+        uses: ./.github/actions/ct-test
+        with:
+          cflags: -Os
```
hanno-becker, line 45: Is it worth having `-Oz` as well? From https://clang.llvm.org/docs/CommandGuide/clang.html:
> [-Oz](https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-O0) Like [-Os](https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-O0) (and thus [-O2](https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-O0)), but reduces code size further.

.github/workflows/ci.yml
mkannwischer: uff. Good catch. Thanks - readded.

.github/workflows/ct-tests.yml
mkannwischer: added.

hanno-becker (approved):
Thanks @mkannwischer, great work!

Issue #689, opened 2025-01-23
Normalize on "unsigned" not "unsigned int" in sources


rod-chapman:
The main sources in mlkem/*.[hc] use plain "unsigned" for types of parameters and variables.
The FIPS202 sources appear to use "unsigned int".

Should we settle on "unsigned" and normalize the FIPS202 sources to that standard?

PR #691, opened 2025-01-23
Use unsigned rather than unsigned int everywhere

mkannwischer:
Resolves https://github.com/pq-code-package/mlkem-native/issues/689


hanno-becker (approved):
I checked that there are no `unsigned int` instances remaining in `.[ch]` files

hanno-becker (changes requested):
Some bad replacements and bad alignments

hanno-becker (approved):

PR #692, opened 2025-01-24
CI: Limit max # of parallel constant-time-check tasks

hanno-becker:


mkannwischer (approved):
LGTM. I just restarted the workflow to see if it works reliably.

PR #699, opened 2025-01-25
Fix exec wrapper for `tests bench --components`

hanno-becker:


mkannwischer (approved):
Thanks! Seems to work now also on A55 and BPi.

PR #701, opened 2025-01-26
Set default to -march=native

mkannwischer:
In our benchmarks we always set -march=explicitly to get the best performance on the current platform.
While we know that this is needed, users of mlkem-native may not and simply run the bench script and be disappointed about performance.

This commit sets the default -march to native. It can be overwritten by passing a different march flag through the CFLAGS env variable.

### DIFF
--- a/test/mk/config.mk
+++ b/test/mk/config.mk
@@ -42,6 +42,7 @@ CFLAGS := \
 	-Wno-unknown-pragmas \
 	-Wno-unused-command-line-argument \
 	-O3 \
+	-march=native \
 	-fomit-frame-pointer \
 	-std=c99 \
 	-pedantic \


mkannwischer:
Okay, no benchmarks changed - that's good as we are setting `-march=` explicitly for each of the benchmarks. 

Unfortunately, you cannot use `-march=native` on all platforms. If I understand correctly, we can never use it when cross-compiling and some platform / compiler version combinations don't work. 

We can probably guard it depending on if it is supported or not. `cmake` makes that easy - in make we'll have to hack it I suppose. Converting this to draft for now.

mkannwischer:
I have no good idea right now how to work around some compilers not supporting this flag. closing for now.

PR #703, opened 2025-01-26
Add balign to all AArch64 asm

mkannwischer:
Resolves https://github.com/pq-code-package/mlkem-native/issues/702

This adds .balign 16 to all global assembly functions to resolve some warnings that show up with gcc 14.2.0.
I also switched all .align's to .balign for consistency. .balign 4 seems to be prefered over .align 4 for aligning to a 4-byte boundary as it is more consistent accross systems (balign always takes bytes, while align takes bytes one some systems and the two-power on most systems).

### DIFF
--- a/mlkem/fips202/native/aarch64/src/keccak_f1600_x4_v8a_v84a_scalar_hybrid_asm_opt.S
+++ b/mlkem/fips202/native/aarch64/src/keccak_f1600_x4_v8a_v84a_scalar_hybrid_asm_opt.S
@@ -877,8 +877,8 @@
 
 .global MLKEM_ASM_NAMESPACE(keccak_f1600_x4_scalar_v8a_v84a_hybrid_asm_opt)
 .text
-.align 4
 
+.balign 4
 MLKEM_ASM_NAMESPACE(keccak_f1600_x4_scalar_v8a_v84a_hybrid_asm_opt):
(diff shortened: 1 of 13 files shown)


hanno-becker:
Why 16 byte alignment, not 4?

mkannwischer:
> Why 16 byte alignment, not 4?
You tell me :-) We had both `.align 4` and `.balign 16` in various places before. 
`.align 4` is the same as `.balign 16`.

mlkem/fips202/native/aarch64/src/keccak_f1600_x4_v8a_v84a_scalar_hybrid_asm_opt.S
```diff
@@ -875,9 +875,9 @@
     ror sAsu, sAsu,#(64-55)
 .endm
 
+.balign 16
```
hanno-becker, line 878: Isn't this misdisplaced? Don't you need it within the target section?

hanno-becker: Changed

hanno-becker:
Switched to 4 byte alignment

hanno-becker (approved):

PR #706, opened 2025-01-27
Don't emit dummy definitions in empty CUs

hanno-becker:

PR #709, opened 2025-01-27
Add AVX2 native polynomial compress/decompression

hanno-becker:
This commit extends the arithmetic natice interface by the
polynomial compression routines
- poly_compress_d4
- poly_compress_d5
- poly_compress_d10
- poly_compress_d11
- poly_decompress_d4
- poly_decompress_d5
- poly_decompress_d10
- poly_decompress_d11

It also imports the native implementation in AVX2 intrinsics
provided by the official Kyber implementation.

Care has to be taken to not overflow the input buffer during
the vectorization. The official Kyber implementation _does_
overflow them, but ensures at the call-sites that sufficient
space is available, and that no relevant data is overwritten.

While this seems to work, it is difficult to reason about this
code, and it moreover prevents a unified interfaces for the native
(de)compression routines.

To remedy, this commit adds tail-loop handling for the (de)compression
routines that overflow their input buffers in the offical Kyber implementation,
reading/writing only as much as is actually needed.

A test is added which invokes the individual (de)compression routines
with minimal input buffers. When used with address sanitization, this
would catch a potential overflow.

### DIFF
--- a/mlkem/native/x86_64/src/compress_avx2.c
+++ b/mlkem/native/x86_64/src/compress_avx2.c
@@ -1,0 +1,367 @@
+/*
+ * Copyright (c) 2024 The mlkem-native project authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/* Implementation from Kyber reference repository
+ * https://github.com/pq-crystals/kyber/blob/main/avx2 */
+
+#include "../../../common.h"
+
+#if defined(MLKEM_NATIVE_ARITH_BACKEND_X86_64_DEFAULT)
+
+#include <immintrin.h>
+#include <stdint.h>
+#include <string.h>
+#include "arith_native_x86_64.h"
+#include "consts.h"
+
+#if defined(MLKEM_NATIVE_MULTILEVEL_BUILD_WITH_SHARED) || \
+    (MLKEM_K == 2 || MLKEM_K == 3)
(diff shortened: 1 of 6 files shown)


mlkem/native/x86_64/src/compress_avx2.c
```diff
@@ -0,0 +1,370 @@
...
+}
+
+#endif /* MLKEM_NATIVE_MULTILEVEL_BUILD || MLKEM_K == 4 */
+
+#else /* MLKEM_USE_NATIVE_X86_64 && SYS_X86_64_AVX2 */
+
+/* Dummy declaration for compilers disliking empty compilation units */
+#define empty_cu_avx2_polyvec_compress \
+  MLKEM_NAMESPACE(empty_cu_avx2_polyvec_compress)
+int empty_cu_avx2_polyvec_compress;
```
mkannwischer, line 368: Should this be changed in the same way as https://github.com/pq-code-package/mlkem-native/pull/706?

mkannwischer (approved):
Looks good to me. Thanks for cleaning this up! 
One nit - but I am not sure if I understand this correctly.

mlkem/native/x86_64/src/compress_avx2.c
hanno-becker: Yes, fixed

Issue #696, opened 2025-01-24
Consider Z3 4.13.4 and Bitwuzla 0.7


rod-chapman:
Recently, both Z3 4.13.4 and Bitwuzla 0.7 have been released.

We should consider bringing these into the NIX environment. I have re-generated all proofs here locally (on macOS) with success.

hanno-becker:
z3 update here https://github.com/pq-code-package/mlkem-native/pull/711.

Not sure about the bitwuzla update. @potsrevennil can you chime in?

PR #711, opened 2025-01-28
Update z3 to 4.13.4 and bitwuzla to 0.7.0

hanno-becker:
Resolves https://github.com/pq-code-package/mlkem-native/issues/696


hanno-becker:
@potsrevennil When you have a moment, could have a look at what to fix here?

hanno-becker (commented):
I confirm that this installs the expected versions of `z3` and `bitwuzla`

Issue #712, opened 2025-01-28
CI: Ensure HOL-Light proofs and source files are in sync


hanno-becker:
The source files that the HOL-Light proofs in `proofs/hol_light/arm/proofs` reason about are currently stored in `proofs/hol_light/arm/mlkem/`. It is at present not checked automatically that the code that mlkem-native ships with (`mlkem/native/aarch64/src/*.S`) matches those files.

- [ ] HOL-Light proofs scripts and CI are adjusted so that HOL-Light proofs reason about byte code directly derived from the shipped assembly files in `mlkem/native/aarch64/*.S`.

PR #713, opened 2025-01-28
Clarify that HOL-Light proofs are not yet synced with shipped ASM

hanno-becker:
This will be addressed through #712, but until then we should be clear that we do not yet claim that mlkem-native's AArch64 [inv]NTT are formally verified.


rod-chapman (approved):
Approve

PR #715, opened 2025-01-29
Proposal: New API

mkannwischer:
This PR rebases an old proposal for extending the API to allow operating on validated+expanded keys. 
Benefit of this is that in case you can keep the keys expanded, you get much better performance. This is particularly useful for an ephemeral use-case (e.g., TLS), where the secret key never has to leave memory.

This is far from being ready, but I'd like to kick off a discussion on how the API should look like.
One point I do not like about this anymore is that it _changes_ the signature of the previous API - It would be cleaner to just add new functions and implement the old 3-function API on top of the new API. Downside of that is that we will have a lot of functions.

Once we have consensus on _how_ to do this, there is still some work to do here: I have not addressed C90 complaince, constant-time tests, and CBMC proofs. Also currently the user (through `mlkem_native.h`) requires to know the details of the key structs - this should be changed to be a generic `uint8_t` array of the right size.

### DIFF
--- a/mlkem/indcpa.c
+++ b/mlkem/indcpa.c
@@ -28,52 +28,29 @@
  * This is to facilitate building multiple instances
  * of mlkem-native (e.g. with varying security levels)
  * within a single compilation unit. */
-#define pack_pk MLKEM_NAMESPACE_K(pack_pk)
-#define unpack_pk MLKEM_NAMESPACE_K(unpack_pk)
-#define pack_sk MLKEM_NAMESPACE_K(pack_sk)
-#define unpack_sk MLKEM_NAMESPACE_K(unpack_sk)
 #define pack_ciphertext MLKEM_NAMESPACE_K(pack_ciphertext)
 #define unpack_ciphertext MLKEM_NAMESPACE_K(unpack_ciphertext)
 #define matvec_mul MLKEM_NAMESPACE_K(matvec_mul)
 /* End of static namespacing */
 
-/*************************************************
- * Name:        pack_pk
- *
- * Description: Serialize the public key as concatenation of the
- *              serialized vector of polynomials pk
- *              and the public seed used to generate the matrix A.
- *
- * Arguments:   uint8_t *r: pointer to the output serialized public key
- *              polyvec *pk: pointer to the input public-key polyvec.
- *                Must have coefficients within [0,..,q-1].
- *              const uint8_t *seed: pointer to the input public seed
- **************************************************/
-static void pack_pk(uint8_t r[MLKEM_INDCPA_PUBLICKEYBYTES], polyvec *pk,
-                    const uint8_t seed[MLKEM_SYMBYTES])
+
+MLKEM_NATIVE_INTERNAL_API
+void indcpa_serialize_pk(uint8_t pks[MLKEM_INDCPA_PUBLICKEYBYTES],
+                         const mlkem_indcpa_public_key *pk)
 {
-  debug_assert_bound_2d(pk, MLKEM_K, MLKEM_N, 0, MLKEM_Q);
-  polyvec_tobytes(r, pk);
-  memcpy(r + MLKEM_POLYVECBYTES, seed, MLKEM_SYMBYTES);
+  debug_assert_bound_2d(pk->pkpv, MLKEM_K, MLKEM_N, 0, MLKEM_Q);
+  polyvec_tobytes(pks, &pk->pkpv);
+  memcpy(pks + MLKEM_POLYVECBYTES, pk->seed, MLKEM_SYMBYTES);
 }
 
-/*************************************************
- * Name:        unpack_pk
- *
- * Description: De-serialize public key from a byte array;
- *              approximate inverse of pack_pk
- *
- * Arguments:   - polyvec *pk: pointer to output public-key polynomial vector
- *                  Coefficients will be normalized to [0,..,q-1].
- *              - uint8_t *seed: pointer to output seed to generate matrix A
- *              - const uint8_t *packedpk: pointer to input serialized public
- *                  key.
- **************************************************/
-static void unpack_pk(polyvec *pk, uint8_t seed[MLKEM_SYMBYTES],
-                      const uint8_t packedpk[MLKEM_INDCPA_PUBLICKEYBYTES])
+MLKEM_NATIVE_INTERNAL_API
+void indcpa_deserialize_pk(mlkem_indcpa_public_key *pk,
+                           const uint8_t pks[MLKEM_INDCPA_PUBLICKEYBYTES])
 {
-  polyvec_frombytes(pk, packedpk);
-  memcpy(seed, packedpk + MLKEM_POLYVECBYTES, MLKEM_SYMBYTES);
+  polyvec_frombytes(&pk->pkpv, pks);
+  memcpy(pk->seed, pks + MLKEM_POLYVECBYTES, MLKEM_SYMBYTES);
+  gen_matrix(pk->at, pk->seed, 1);
+
 
   /* NOTE: If a modulus check was conducted on the PK, we know at this
    * point that the coefficients of `pk` are unsigned canonical. The
@@ -81,35 +58,19 @@ static void unpack_pk(polyvec *pk, uint8_t seed[MLKEM_SYMBYTES],
    * work with the easily provable bound by UINT12_LIMIT. */
 }
 
-/*************************************************
- * Name:        pack_sk
- *
- * Description: Serialize the secret key
- *
- * Arguments:   - uint8_t *r: pointer to output serialized secret key
- *              - polyvec *sk: pointer to input vector of polynomials (secret
- *key)
- **************************************************/
-static void pack_sk(uint8_t r[MLKEM_INDCPA_SECRETKEYBYTES], polyvec *sk)
+MLKEM_NATIVE_INTERNAL_API
+void indcpa_serialize_sk(uint8_t sks[MLKEM_INDCPA_SECRETKEYBYTES],
+                         const mlkem_indcpa_secret_key *sk)
 {
-  debug_assert_bound_2d(sk, MLKEM_K, MLKEM_N, 0, MLKEM_Q);
-  polyvec_tobytes(r, sk);
+  debug_assert_bound_2d(&sk->skpv, MLKEM_K, MLKEM_N, 0, MLKEM_Q);
+  polyvec_tobytes(sks, &sk->skpv);
 }
 
-/*************************************************
- * Name:        unpack_sk
- *
- * Description: De-serialize the secret key; inverse of pack_sk
- *
- * Arguments:   - polyvec *sk: pointer to output vector of polynomials (secret
- *                key)
- *              - const uint8_t *packedsk: pointer to input serialized secret
- *                key
- **************************************************/
-static void unpack_sk(polyvec *sk,
-                      const uint8_t packedsk[MLKEM_INDCPA_SECRETKEYBYTES])
+MLKEM_NATIVE_INTERNAL_API
+void indcpa_deserialize_sk(mlkem_indcpa_secret_key *sk,
+                           const uint8_t sks[MLKEM_INDCPA_SECRETKEYBYTES])
 {
-  polyvec_frombytes(sk, packedsk);
+  polyvec_frombytes(&sk->skpv, sks);
 }
 
 /*************************************************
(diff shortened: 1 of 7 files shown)


mkannwischer:
@hanno-becker - any thoughts on this?

hanno-becker:
@mkannwischer What stands in the way of keeping the old API and providing the new one as an (optional) addition?

mlkem/kem.h
```diff
@@ -35,6 +36,77 @@
...
+} mlk_secret_key;
+
+typedef struct
+{
+  mlk_indcpa_public_key indcpa_pk;
+  uint8_t hpk[MLKEM_SYMBYTES];
+} mlk_public_key;
+
+
+MLK_EXTERNAL_API
```
hanno-becker, line 55: MLK_MUST_CHECK_RETURN_VALUE is missing in a few places here?

mlkem/kem.h
```diff
@@ -35,6 +36,77 @@
...
+  uint8_t seed[MLKEM_SYMBYTES];
+  uint8_t z[MLKEM_SYMBYTES];
+  uint8_t hpk[MLKEM_SYMBYTES];
+} mlk_secret_key;
+
+typedef struct
+{
+  mlk_indcpa_public_key indcpa_pk;
+  uint8_t hpk[MLKEM_SYMBYTES];
+} mlk_public_key;
```
hanno-becker, line 52: Add alignment?

mlkem/kem.h
```diff
@@ -35,6 +36,77 @@
...
 #endif
 
+typedef struct
+{
+  mlk_indcpa_secret_key indcpa_sk;
+  mlk_indcpa_public_key indcpa_pk;
+  uint8_t seed[MLKEM_SYMBYTES];
+  uint8_t z[MLKEM_SYMBYTES];
+  uint8_t hpk[MLKEM_SYMBYTES];
+} mlk_secret_key;
```
hanno-becker, line 46: Add alignment?

mlkem/kem.c
```diff
@@ -120,32 +120,30 @@ static int mlk_check_sk(const uint8_t sk[MLKEM_INDCCA_SECRETKEYBYTES])
...
 
 #if defined(MLK_KEYGEN_PCT)
 /* Specification:
  * Partially implements 'Pairwise Consistency Test' [FIPS 140-3 IG] and
  * [FIPS 203, Section 7.1, Pairwise Consistency]. */
 
 /* Reference: Not implemented in the reference implementation. */
-static int mlk_check_pct(uint8_t const pk[MLKEM_INDCCA_PUBLICKEYBYTES],
-                         uint8_t const sk[MLKEM_INDCCA_SECRETKEYBYTES])
+static int mlk_check_pct(const mlk_public_key *pk, const mlk_secret_key *sk)
```
hanno-becker, line 134: It is unclear if this will remain FIPS compliant -- we should keep the previous form of the PCT for now?

mlkem/kem.c
```diff
@@ -183,34 +180,149 @@ static int mlk_check_pct(uint8_t const pk[MLKEM_INDCCA_PUBLICKEYBYTES],
...
+}
+
+
+MLK_EXTERNAL_API
+int crypto_kem_keypair_derand_struct(mlk_public_key *pk, mlk_secret_key *sk,
+                                     const uint8_t coins[2 * MLKEM_SYMBYTES])
+{
+  mlk_indcpa_keypair_derand(&sk->indcpa_pk, &sk->indcpa_sk, coins);
+
+  // pre-compute H(pk)
```
hanno-becker, line 244: Not C90

mlkem/kem.c
```diff
@@ -183,34 +180,149 @@ static int mlk_check_pct(uint8_t const pk[MLKEM_INDCCA_PUBLICKEYBYTES],
...
+  uint8_t pks[MLKEM_INDCCA_PUBLICKEYBYTES];
```
hanno-becker, line 245: Not C90

mlkem/kem.c
```diff
@@ -183,34 +180,149 @@ static int mlk_check_pct(uint8_t const pk[MLKEM_INDCCA_PUBLICKEYBYTES],
...
+  mlk_public_key pks;
+  mlk_secret_key sks;
+
+
+  res = crypto_kem_keypair_derand_struct(&pks, &sks, coins);
+
+
+  crypto_kem_marshal_pk(pk, &pks);
+  crypto_kem_marshal_sk(sk, &sks);
+
```
hanno-becker, line 317: Referring to the PCT comment, I think we need to retain the PCT here for the time being.

hanno-becker (commented):
This is going in a good direction I think. You may defer updating `mlkem_native.h` for the time being, keeping that to the core API, and using `kem.h` directly from the test exercising the new API.

mlkem/kem.c
mkannwischer: No - If PCT is enabled, _all_ public APIs need to do PCT - so we cannot move it here, otherwise `crypto_kem_keypair_struct` doesn't peform PCT. We can perform serialization in the PCT if needed, but that is _very_ expensive.

hanno-becker: The current code does not perform the PCT on the key material that is returned to the caller -- it is thus not clear if it is compliant. Please re-introduce the previous form of the PCT.

I don't know how to best introduce a PCT in the internal-format API -- but this is secondary to me at the moment, because the FIPS-compliance of that API is still questionable anyway.

mlkem/kem.h
mkannwischer: yes.

mlkem/kem.h
mkannwischer: yes.

mlkem/kem.h
mkannwischer: yes.

mlkem/kem.c
mkannwischer: yes

mlkem/kem.c
mkannwischer: yes

mlkem/kem.c
mkannwischer: Okay - I have implemented what you said - we should be explicit somewhere that PCT is ONLY implemented when using `crypto_kem_keypair`, but not when using any of the other functions.
AWS-LC may want to consider changing this in the medium-term - it's much faster if you can work on the internal formats.

mlkem/kem.h
```diff
@@ -33,8 +34,105 @@
...
+
+#define crypto_kem_marshal_pk MLK_NAMESPACE_K(marshal_pk)
+#define crypto_kem_parse_pk MLK_NAMESPACE_K(parse_pk)
+#define crypto_kem_marshal_sk MLK_NAMESPACE_K(marshal_sk)
+#define crypto_kem_parse_sk MLK_NAMESPACE_K(parse_sk)
+#define crypto_kem_keypair_derand_struct MLK_NAMESPACE_K(keypair_derand_struct)
+#define crypto_kem_keypair_struct MLK_NAMESPACE_K(keypair_struct)
+#define crypto_kem_enc_derand_struct MLK_NAMESPACE_K(enc_derand_struct)
+#define crypto_kem_enc_struct MLK_NAMESPACE_K(enc_struct)
+#define crypto_kem_dec_struct MLK_NAMESPACE_K(dec_struct)
```
hanno-becker, line 46: This should be `MLK_ADD_LEVEL` for the structs

mkannwischer: Not sure I can follow. For the struct I am using using `MLK_ADD_LEVEL`,

mlkem/kem.c
hanno-becker: > AWS-LC may want to consider changing this in the medium-term - it's much faster if you can work on the internal formats.

Of course it is -- but it's about FIPS compliance. When it's clear that the internal PCT is still enough, I'm confident it would be changed.

mlkem/kem.h
hanno-becker: True

mlkem/indcpa.h
```diff
@@ -20,6 +20,70 @@
 #include "common.h"
 #include "poly_k.h"
 
+#define mlk_indcpa_secret_key MLK_NAMESPACE_K(mlk_indcpa_secret_key)
+typedef struct
+{
+  mlk_polyvec skpv;
```
hanno-becker, line 26: Have you considered caching the mulcache? This would be useful in decapsulation.

mlkem/indcpa.c
```diff
@@ -331,15 +363,15 @@ __contract__(
  *            - We include buffer zeroization.
```
hanno-becker, line 363: All `Reference: ..` comments need extending

mlkem/indcpa.c
```diff
@@ -357,49 +389,48 @@ void mlk_indcpa_keypair_derand(uint8_t pk[MLKEM_INDCPA_PUBLICKEYBYTES],
    */
   MLK_CT_TESTING_DECLASSIFY(publicseed, MLKEM_SYMBYTES);
 
-  mlk_gen_matrix(a, publicseed, 0 /* no transpose */);
+  mlk_gen_matrix(pk->at, publicseed, 0 /* no transpose */);
```
hanno-becker, line 392: Can you leave a comment that the matrix will be transposed later? Otherwise, it's confusing why the variable is named `at`.

mlkem/indcpa.c
```diff
@@ -357,49 +389,48 @@ void mlk_indcpa_keypair_derand(uint8_t pk[MLKEM_INDCPA_PUBLICKEYBYTES],
...
+  mlk_polyvec_tomont(pk->pkpv);
+
+  mlk_polyvec_add(pk->pkpv, e);
+  mlk_polyvec_reduce(pk->pkpv);
+  mlk_polyvec_reduce(sk->skpv);
 
-  mlk_polyvec_add(pkpv, e);
-  mlk_polyvec_reduce(pkpv);
-  mlk_polyvec_reduce(skpv);
+  mlk_transpose_matrix(pk->at);
```
hanno-becker, line 425: Can you leave a comment what the matrix is transposed for at this point?

mlkem/indcpa.h
```diff
@@ -45,7 +109,7 @@ __contract__(
   requires(memory_no_alias(a, sizeof(mlk_polymat)))
   requires(memory_no_alias(seed, MLKEM_SYMBYTES))
   requires(transposed == 0 || transposed == 1)
-  assigns(object_whole(a))
+  assigns(memory_slice(a, sizeof(mlk_poly) * MLKEM_K * MLKEM_K))
```
hanno-becker, line 112: Nit: `sizeof(mlk_polymat)`?

mlkem/indcpa.h
```diff
@@ -102,12 +172,16 @@ __contract__(
...
                     const uint8_t coins[MLKEM_SYMBYTES])
 __contract__(
   requires(memory_no_alias(c, MLKEM_INDCPA_BYTES))
   requires(memory_no_alias(m, MLKEM_INDCPA_MSGBYTES))
-  requires(memory_no_alias(pk, MLKEM_INDCPA_PUBLICKEYBYTES))
+  requires(memory_no_alias(pk, sizeof(mlk_indcpa_public_key)))
+  requires(forall(k0, 0, MLKEM_K,
+    array_bound(pk->pkpv[k0].coeffs, 0, MLKEM_N, 0, MLKEM_UINT12_LIMIT)))
+  requires(forall(x, 0, MLKEM_K * MLKEM_K,
+      array_bound(pk->at[x].coeffs, 0, MLKEM_N, 0, MLKEM_Q)))
```
hanno-becker, line 184:
```suggestion
    array_bound(pk->at[x].coeffs, 0, MLKEM_N, 0, MLKEM_Q)))
```

mlkem/indcpa.c
```diff
@@ -193,6 +199,32 @@ __contract__(
   ensures(array_bound(data, 0, MLKEM_N, 0, MLKEM_Q))) { ((void)data); }
 #endif /* !MLK_USE_NATIVE_NTT_CUSTOM_ORDER */
 
+static void mlk_transpose_matrix(mlk_polymat a)
```
hanno-becker, line 202: Can you add this to the component benchmarks so we get a sense of the performance cost?

mlkem/kem.c
```diff
@@ -203,6 +203,119 @@ static int mlk_check_pct(uint8_t const pk[MLKEM_INDCCA_PUBLICKEYBYTES],
...
+  /* pre-compute H(pk) */
+  mlk_indcpa_marshal_pk(pks, &sk->indcpa_pk);
+  mlk_hash_h(sk->hpk, pks, MLKEM_INDCCA_PUBLICKEYBYTES);
+
+  /*
+   * copy over indcpa pk and H(pk) to public key
+   * pk is NULL during parsing before decaps as the pk is not needed
+   **/
+  if (pk != NULL)
```
hanno-becker, line 281: The CBMC contract says `pk` must be valid. Where do we actually need this?

PR #717, opened 2025-01-29
Add script for assembly simplification

hanno-becker:
This PR adds `scripts/simpasm`, which simplifies assembly through
assembling + disassembling.

The primary simplification achieved this way is the removal of assembly
macros (`.macro` directive) and register aliases (`.req` directives),
which some custom assembly processing tools may struggle with. In such
situation, the user can call `simpasm` to remove those directives,
while otherwise keeping the assembly usable as before.


mkannwischer (approved):
The changes to the asm look good to me. I tested that the script works fine. 
Whether or not the simplifications is what we need for integration is something that I dont know how to check, but I guess we will find out in the follow up PR. 

Maybe it is a good idea to also test this on MacOS in CI?

PR #718, opened 2025-01-30
Run HOL-Light proofs on mlkem-native assembly

hanno-becker:
* Fixes https://github.com/pq-code-package/mlkem-native/issues/712 

This commit uses the `simpasm` script to auto-generate the
input assembly for the HOL-Light proofs from the assembly that's
shipped with mlkem-native.

A minor adjustment is made to `simpasm` along the way, changing
the meaning of `-p` from preserving the header/footer to preserving
the preprocessor directives in the header. That is, the header will
_always_ be preserved, and if `-p` is not set, then preprocessor
directives will be dropped, making the resulting assembly file
standalone. We always preserve to header to ensure that licensing
notices are kept.

The `autogen` script is adjusted to ensure that the assembly sources
in `proofs/hol_light/arm/mlkem` are in sync with `mlkem/*`. Note that
`simpasm` in particular ensures that simplified assembly has the same
byte code as the original one, so the HOL-Light proofs apply to the
assembly shipped in mlkem-native. Previous warning notices in READMEs
are therefore removed.


mkannwischer (approved):
Thanks @hanno-becker for completing this big milestone. This is awesome!
I did not run the proofs locally, but I looked at the proofs in CI and it looks good!

One problem I encountered is that `autogen` created different assembly file for me likely due to a differen `objdump` being used. Linting fails accordingly. Here is some of the diff I am getting:
```
<         ldr	q26, [x0, #64]
<         ldr	q29, [x0, #128]
<         ldr	q20, [x0, #192]
<         ldr	q23, [x0, #256]
<         ldr	q11, [x0, #448]
<         mul	v2.8h, v23.8h, v0.h[0]
<         ldr	q17, [x0, #320]
<         mul	v15.8h, v11.8h, v0.h[0]
<         ldr	q13, [x0, #384]
---
>         ldr	q26, [x0, #0x40]
>         ldr	q29, [x0, #0x80]
>         ldr	q20, [x0, #0xc0]
>         ldr	q23, [x0, #0x100]
>         ldr	q11, [x0, #0x1c0]
>         mul.8h	v2, v23, v0[0]
>         ldr	q17, [x0, #0x140]
>         mul.8h	v15, v11, v0[0]
>         ldr	q13, [x0, #0x180]
```

We should fix that in a followup PR.

PR #719, opened 2025-01-30
Split AArch64 clean vs. opt into separate subdirectories

hanno-becker:
Based on #718


mkannwischer (approved):
Thanks for splitting this up - mixing opt and clean code in one folder bothered me for a while. 

Currenlty the clean backend seems to be default - please switch that back to the opt backend.

My impression is that the clean code is only used as a source for SLOTHY optimizations, and we are expecting our consumers to only use the opt code. 
So, thus far, the clean code is unused and remains unused except for testing, right? 
Could we then move that code out of the main mlkem source tree into our test folder?
I'll leave it up to you if you do that as a part of this PR, or if we do it in a follow up PR that also adds tests for the clean code.

scripts/autogen
```diff
@@ -362,7 +361,11 @@ def gen_aarch64_fwd_ntt_zeta_file(dry_run=False):
         yield ""
 
     update_file(
-        "mlkem/native/aarch64/src/aarch64_zetas.c", "\n".join(gen()), dry_run=dry_run
+        "mlkem/native/aarch64/src/aarch64_zetas.c", "\n".join(gen("opt")), dry_run=dry_run
+    )
+
+    update_file(
+        "mlkem/native/aarch64_clean/src/aarch64_zetas.c", "\n".join(gen("clean")), dry_run=dry_run
```
mkannwischer, line 368: same here.

scripts/autogen
```diff
@@ -413,7 +415,13 @@ def gen_aarch64_rej_uniform_table(dry_run=False):
...
     update_file(
         "mlkem/native/aarch64/src/rej_uniform_table.c",
-        "\n".join(gen()),
+        "\n".join(gen("opt")),
+        dry_run=dry_run,
+    )
+
+    update_file(
+        "mlkem/native/aarch64_clean/src/rej_uniform_table.c",
+        "\n".join(gen("clean")),
```
mkannwischer, line 424: this looks wrong?

hanno-becker: Fixed

scripts/autogen
hanno-becker: Fixed

Issue #238, opened 2024-10-15
Bump GCC version in nix to gcc-14


hanno-becker:
We're currently using gcc-13, but as @mkannwischer noticed in #224, there is a meaningful performance impact from moving to gcc-14. Moreover, there even appears to be differences even between gcc-14 in nix, and gcc-14 in arch.

- [ ] Adjust nix flake to use gcc-14
- [ ] Make sure nix flake build time and memory usage is within acceptable bounds for our CI
- [ ] Validate that performance matches that of gcc-14 from arch

hanno-becker:
@mkannwischer Please add some colour here based on your experiments.

mkannwischer:
Is there any reason to still hold this back? @potsrevennil tells me this is going be very easy to change.

PR #723, opened 2025-01-31
Nix: switch default compiler to gcc14

mkannwischer:
Resolves https://github.com/pq-code-package/mlkem-native/issues/238.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -315,7 +315,7 @@ jobs:
            shell: ci_gcc11
            darwin: True
          - name: gcc-13
-           shell: ci
+           shell: ci_gcc13
            darwin: True
          - name: gcc-14
            shell: ci_gcc14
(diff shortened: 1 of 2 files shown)


hanno-becker (changes requested):
The compiler tests need changing. They currently use
```
         - name: gcc-13
           shell: ci
           darwin: True
         - name: gcc-14
           shell: ci_gcc14
           darwin: True
```
relying on `ci` to use `gcc13`. One should just add an explicit `ci_gcc13` and use that here, not relying on the `ci` shell.

mkannwischer:
> The compiler tests need changing. They currently use
> 
> ```
>          - name: gcc-13
>            shell: ci
>            darwin: True
>          - name: gcc-14
>            shell: ci_gcc14
>            darwin: True
> ```
> 
> relying on `ci` to use `gcc13`. One should just add an explicit `ci_gcc13` and use that here, not relying on the `ci` shell.

Fixed.

hanno-becker (approved):

PR #727, opened 2025-02-01
Simplify base multiplication

hanno-becker:
Previously, the vector-vector base multiplication would be split in three stages:
- deg-2 polynomial base multiplication
- deg-256 polynomial base multiplication
- polyvec base multiplication

Conceptually, this separation makes sense, but is otherwise unnecessarily complicated: The entire polyvec base multiplication can be written in one compact and readable function. This saves code, proof burden, and also has the added benefit of not requiring a stack-allocated accumulator which may need to be zeroized for FIPS compliance. Finally, the monolithic version is faster because it requires 2-4x fewer scalar reductions, and is more amenable to vectorization.


mkannwischer (approved):
LGTM. Small performance improvement of up to 5%.

PR #733, opened 2025-02-03
Add example for multilevel monobuild with native code

hanno-becker:


examples/monolithic_build_multilevel_native/mlkem_native_all.c
```diff
@@ -0,0 +1,37 @@
...
+ * Copyright (c) 2024 The mlkem-native project authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#define MLKEM_NATIVE_MULTILEVEL_BUILD
+
+/* Three instances of mlkem-native for all security levels */
+
+/* Include level-independent code */
+#define MLKEM_NATIVE_MULTILEVEL_BUILD_WITH_SHARED 1
```
mkannwischer, line 11: Why is this set as 1?

mkannwischer (approved):
LGTM. I looked at the resulting binary and it does contain the assembly that I would expect. 

This PR doesn't touch the mlkem files, so there is no risk this breaks anything.

examples/monolithic_build_multilevel_native/mlkem_native_all.c
hanno-becker: So it doesn't give a warning with how `-D...` in the CFLAGS is done

PR #735, opened 2025-02-03
autogen: Auto-generate header guards

hanno-becker:
This commit modifies `autogen` to automatically insert/correct standardized header guards

```bash
   #ifndef MLKEM_NATIVE_{FILE_NAME}_H
   #define MLKEM_NATIVE_{FILE_NAME}_H

   ...

   #endif MLKEM_NATIVE_{FILE_NAME}_H
```

Here, `FILE_NAME` is the file path relative to `mlkem/`, with `/` replaced by `_`.

Such namespaced header guards reduce the risk of inadvertent
header clashes in single-compilation-unit builds.


scripts/autogen
```diff
@@ -799,6 +799,62 @@ def gen_hol_light_asm(dry_run=False):
...
+    with open(header_file, "r") as f:
+        content = f.read().split('\n')
+
+    # Use full filename as the header guard, with '/' and '.' replaced by '_'
+    guard_name = header_file.removeprefix("mlkem/").replace("/","_").replace(".","_").upper()
+    guard_name = "MLKEM_NATIVE_" + guard_name
+
+    def gen_copyright():
+        yield "/*"
+        yield " * Copyright (c) 2024 The mlkem-native project authors"
```
mkannwischer, line 814: Should we change this to 2024-2025? or something like that?

mkannwischer (changes requested):
linting fails for me.

scripts/autogen
hanno-becker: Good point, will add this

hanno-becker: Changed this everywhere... hopefully

PR #740, opened 2025-02-04
Use 'raw' assembly in main source tree

hanno-becker:
This PR modifies the way backends are developed and consumed.

Originally, we had AArch64 clean and optimized backends in `mlkem/native/aarch64`. Recently, we moved the clean backend to `test/aarch64_clean`, but kept the optimized backend.

Now some potential consumers, e.g. AWS-LC, need 'simplified' assembly without register macros or aliases. Previous PRs have introduced the `simpasm` script to provide this simplification. Yet, the simplified files are not part of the source tree, so when importing the library one has to call `simpasm`, which -- to cover the whole assembly tree -- needs to use both a native and a cross-compiler. This is inconvenient.

Moreover, we did not have a mechanism to install the `aarch64_clean` backend since its move to `test/aarch64_clean`, other than copy-paste -- not the end of the world, but not convenient either.

This commit moves both the clean and optimized aarch64 to a new directory `dev/`. What is shipped in the main source tree `mlkem/*`, in turn, it merely the _simplified_ assembly. The `autogen` script ensures that the simplified assembly files and the SLOTHY-optimized ones are in sync. 

What's more, `autogen` can now be called with `--aarch64-clean`, which similarly established `mlkem/native/aarch64/*` based on the _clean_ assembly -- this provides a convenient mechanism to test changes made to the clean backend. Also, `autogen` offers `--no-simplify`, which skips assembly simplification.


proofs/hol_light/arm/mlkem/mlkem_intt.S
mkannwischer, line 1: Why do we need to keep a copy of this file for the HOL Light Proof? Can't we use the simplified assembly directly? If that is the case I would prefer not duplicating the files.

README.md
```diff
@@ -80,7 +80,7 @@ offers three backends for C, AArch64 and x86_64 - if you'd like contribute new b
 PR.
 
 Our AArch64 assembly is developed using [SLOTHY](https://github.com/slothy-optimizer/slothy): We write
-'clean' assembly by hand and automate micro-optimizations (e.g. see the [clean](test/aarch64_clean/src/ntt_clean.S)
+'clean' assembly by hand and automate micro-optimizations (e.g. see the [clean](dev/aarch64_clean/src/ntt_clean.S)
```
mkannwischer, line 83: Please add a link to `dev/README.md` for those interest in how the final assembly was produced.

scripts/autogen
mkannwischer, line 1: Not related to your PR (sorry!): 
Currently linting errors with 
```
Autogenerated file proofs/hol_light/arm/mlkem/mlkem_ntt.S needs updating. Have you called scripts/autogenerated.py?
Writing new version to proofs/hol_light/arm/mlkem/mlkem_ntt.S.new
```

1) Could you adjust that error message `scripts/autogenerated.py` is now `scripts/autogen` (two places where this needs to change). 

2) Why is linting writing `proofs/hol_light/arm/mlkem/mlkem_ntt.S.new`? Linting should just error in case files are outdated.

scripts/autogen
```diff
@@ -772,20 +771,66 @@ def check_asm_register_aliases():
...
+    # Check native architecture
+    if platform.machine().lower() in ["arm64", "aarch64"]:
+        native_arch = "aarch64"
+    else:
+        native_arch = "x86_64"
+
+    if native_arch != source_arch:
+        if platform.system() == "Darwin":
+            print(
+                f"WARNING: Skipping simplification of file {infile} -- "
```
mkannwischer, line 799: Discussions with @potsrevennil yesterday suggest that this limitation can be easily overcome. Please open an issue referencing this line so we do not forget to address this.

mkannwischer (changes requested):
Thanks @hanno-becker!

1) Linting fails with 
```
  Autogenerated file mlkem/native/aarch64/src/arith_native_aarch64.h needs updating. Have you called scripts/autogenerated.py?
  Writing new version to mlkem/native/aarch64/src/arith_native_aarch64.h.new
  5,6c5,6
  < #ifndef MLKEM_NATIVE_NATIVE_AARCH64_SRC_ARITH_NATIVE_AARCH64_H
  < #define MLKEM_NATIVE_NATIVE_AARCH64_SRC_ARITH_NATIVE_AARCH64_H
  ---
  > #ifndef MLKEM_NATIVE_DEV_AARCH64_OPT_SRC_ARITH_NATIVE_AARCH64_H
  > #define MLKEM_NATIVE_DEV_AARCH64_OPT_SRC_ARITH_NATIVE_AARCH64_H
  77c77
  < #endif /* MLKEM_NATIVE_NATIVE_AARCH64_SRC_ARITH_NATIVE_AARCH64_H */
  ---
  > #endif /* MLKEM_NATIVE_DEV_AARCH64_OPT_SRC_ARITH_NATIVE_AARCH64_H */
```
Running `autogen` locally (on MacOS) does not resolve this problem - I don't understand why. 

But running autogen locally updates mlkem/fips202/native/x86_64/meta.h - which is not flagged by CI linting. Here is the patch: 
```
diff --git a/mlkem/fips202/native/x86_64/meta.h b/mlkem/fips202/native/x86_64/meta.h
index 88f42739..f077bc75 100644
--- a/mlkem/fips202/native/x86_64/meta.h
+++ b/mlkem/fips202/native/x86_64/meta.h
@@ -3,8 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#ifndef MLKEM_NATIVE_FIPS202_NATIVE_X86_64_XKCP_H
-#define MLKEM_NATIVE_FIPS202_NATIVE_X86_64_XKCP_H
+#ifndef MLKEM_NATIVE_FIPS202_NATIVE_X86_64_META_H
+#define MLKEM_NATIVE_FIPS202_NATIVE_X86_64_META_H
 /* Default FIPS202 assembly profile for AArch64 systems */
 
 #ifdef MLKEM_NATIVE_FIPS202_PROFILE_H
@@ -26,4 +26,4 @@
 
 #endif /* MLKEM_NATIVE_FIPS202_PROFILE_H */
 
-#endif /* MLKEM_NATIVE_FIPS202_NATIVE_X86_64_XKCP_H */
+#endif /* MLKEM_NATIVE_FIPS202_NATIVE_X86_64_META_H */
```

proofs/hol_light/arm/mlkem/mlkem_intt.S
hanno-becker: The HOL Light ASM file is completely standalone and does not even have the header anymore. We could change that, but I would not want to do it in this PR.

README.md
hanno-becker: Ok

scripts/autogen
hanno-becker: 1. Ok
2. To ease debugging

proofs/hol_light/arm/mlkem/mlkem_intt.S
mkannwischer: Okay - would be nice to have only one file.

scripts/autogen
hanno-becker: Changed the behaviour here to print the warning whenever a cross-chain is unavailable, not specifically when it's Darwin.

README.md
hanno-becker: Done

scripts/autogen
```diff
@@ -794,22 +848,130 @@ def gen_hol_light_asm_file(infile, outfile, dry_run=False):
...
+    if transform is not None:
+        content = transform(content)
+
+    update_file(outfile_full, content, dry_run=dry_run)
+
+
+def update_via_remove(filename, dry_run=False):
+    if dry_run is True:
+        print(
+            f"Autogenerated file {filename} needs removing. Have you called scripts/autogenerated.py?",
```
mkannwischer, line 882: change to autogen, please

mkannwischer (approved):
Thanks @hanno-becker - I tested that the autogen script works with the different flag variations (`autogen --aarch64-clean`, `autogen --aarch64-clean --no-simplify`). 

One nit and then I'm happy with this PR.

scripts/autogen
hanno-becker: done

PR #744, opened 2025-02-05
CI: Only cache one valgrind per arch

mkannwischer:
We are using a patched valgrind for the ct tests
which is built in CI and  then cached in the Github Cache. Currently we are using many different compilers for these tests. Currently, for each of these compilers valgrind is recompiled and cached taking up 600-800 MB per
each of the 26 compiler/arch combinations.
We only have 10 GB of cache, so we are constantly triggering cache evictions.
This leads to painfully slow tests for other tests that due to that have to recompile gcc which takes around 45 minutes.

This commit changes the cache key for valgrind to only contain the OS and arch, but not the compiler version.

### DIFF
--- a/.github/actions/ct-test/action.yml
+++ b/.github/actions/ct-test/action.yml
@@ -14,6 +14,28 @@ inputs:
 runs:
   using: composite
   steps:
+      - name: System info
+        shell: ${{ env.SHELL }}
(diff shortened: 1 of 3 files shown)


hanno-becker (approved):
Thanks @mkannwischer !

PR #746, opened 2025-02-05
Minor changes to sys.h

hanno-becker:
**Summary**:

**Steps**:

**Performed local tests**:
 - [ ] `lint` passing
 - [ ] `tests all` passing
 - [ ] `tests bench` passing
 - [ ] `tests cbmc` passing


mlkem/verify.h
mkannwischer, line 1: The commit message states  a `defined(__asm__)` guard is added.
I cannot see that guard. 

Also I am not sure if I can follow what this commit is doing.
sys.h so far was setting asm to `__asm` for Windows and `__asm__` otherwise.
Replacing asm with `__asm__` will probably break it on Windows or what am I missing here?

mlkem/sys.h
mkannwischer, line 1: What's the rationale for this change? It's never used, is it?

examples/monolithic_build_multilevel_native/mlkem_native_all.c
```diff
@@ -10,8 +10,10 @@
 /* Include level-independent code */
 #define MLKEM_NATIVE_MULTILEVEL_BUILD_WITH_SHARED 1
 /* Include C files accompanying native code */
-#define MLKEM_NATIVE_MONOBUILD_WITH_NATIVE_ARITH
-#define MLKEM_NATIVE_MONOBUILD_WITH_NATIVE_FIPS202
+#define MLKEM_NATIVE_MONOBUILD_WITH_NATIVE_ARITH_AARCH64
+#define MLKEM_NATIVE_MONOBUILD_WITH_NATIVE_FIPS202_AARCH64
+#define MLKEM_NATIVE_MONOBUILD_WITH_NATIVE_ARITH_X86_64
+#define MLKEM_NATIVE_MONOBUILD_WITH_NATIVE_FIPS202_X86_64
```
mkannwischer, line 16: What's the reason for the refinement? Can't you achieve the same by guarding the 
```
#define MLKEM_NATIVE_MONOBUILD_WITH_NATIVE_ARITH
#define MLKEM_NATIVE_MONOBUILD_WITH_NATIVE_FIPS202
``` 
depending on the architecture you are building for?

mkannwischer (changes requested):
The PR title looks outdated and I for some commits it's a bit unclear why the change is made. Please add the rationale to the commit message.

mlkem/sys.h
hanno-becker: Simplifies writing custom backends

mlkem/verify.h
hanno-becker: __asm was never used on Windows because we don't use the asm-based value barrier there.

examples/monolithic_build_multilevel_native/mlkem_native_all.c
hanno-becker: If you want a build of mlkem-native with, say, aarch64 backend but without x86_64 backend, the mechanism you propose does not work, and one would need to manually adjust the monobuild file.

mlkem/verify.h
mkannwischer: okay.

hanno-becker: The commit message is stale, I'll fix it.

examples/monolithic_build_multilevel_native/mlkem_native_all.c
hanno-becker: Mentioned that case in the commit message.

mkannwischer: Why can I not do
```
#if defined(__AARCH64EL__)
#define MLKEM_NATIVE_MONOBUILD_WITH_NATIVE_ARITH
#define MLKEM_NATIVE_MONOBUILD_WITH_NATIVE_FIPS202
#endif 
```
?

hanno-becker: Yes that works, thank you. Adjusted

mkannwischer (dismissed):

mlkem/params.h
```diff
@@ -18,7 +18,7 @@
 #define MLKEM_N 256
 #define MLKEM_Q 3329
 #define MLKEM_Q_HALF ((MLKEM_Q + 1) / 2) /* 1665 */
-#define UINT12_LIMIT 4096
+#define MLK_UINT12_LIMIT 4096
```
mkannwischer, line 21: Do you want to change this to MLKEM_UINT12_LIMIT? I know this is hardly a scheme "parameter" (and, hence, should not be named MLKEM_ or be placed in params.h), but I don't know where else it should go and having a single `MLK_` define in this file looks weird.

mkannwischer (changes requested):
I went through all the renaming and found a couple of small mistakes (some are not related to your renaming, but it would be great if you could resolve them as well): 

1) `proofs/cbmc/Makefile.common` contains
```
# Give visibility to all static functions
DEFINES += -Dstatic=
DEFINES += -DINLINE=
DEFINES += -DALWAYS_INLINE=
```

which needs to update? Why does this even work?

2) `FORCE_AARCH64` + `FORCE_AARCH64_EB` + `FORCE_X86_64) is not properly replaced in all places. In Makefiles and .ymls it's still without the `MLK_` namespace. 

3) (unrelated, sorry) Double header guards in align.h

4) MLKEM_NATIE_CONFIG_FILE is not properly replaced in some comments in examples (as it contains a typo)

5) (unrelated, sorry) FIPS202_NAMESPACE can be removed from `proofs/cbmc/Makefile_params.common`

mlkem/fips202/native/aarch64/src/cortex_a55_impl.h
```diff
@@ -3,8 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#ifndef MLKEM_NATIVE_FIPS202_NATIVE_AARCH64_SRC_CORTEX_A55_IMPL_H
-#define MLKEM_NATIVE_FIPS202_NATIVE_AARCH64_SRC_CORTEX_A55_IMPL_H
+#ifndef MLK_FIPS202_NATIVE_AARCH64_SRC_CORTEX_A55_IMPL_H
+#define MLK_FIPS202_NATIVE_AARCH64_SRC_CORTEX_A55_IMPL_H
 /* FIPS202 assembly profile targeting Cortex-A55 */
 
 #ifdef FIPS202_NATIVE_PROFILE_IMPL_H
```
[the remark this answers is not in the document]
hanno-becker: This is still deliberate and guards against double-inclusion of backends

mlkem/params.h
hanno-becker: Ok

hanno-becker:
Super, thank you @mkannwischer for the careful review!

> proofs/cbmc/Makefile.common contains
>  Give visibility to all static functions 
>  DEFINES += -Dstatic=
>  DEFINES += -DINLINE=
>  DEFINES += -DALWAYS_INLINE=
>  which needs to update? Why does this even work?

Fixed

> FORCE_AARCH64 + FORCE_AARCH64_EB + FORCE_X86_64) is not properly replaced in all places. In Makefiles and .ymls it's still without the MLK_` namespace.

Fixed

> (unrelated, sorry) Double header guards in align.h

Not sure which file you mean -- if it's the FIPS native backend header, the double-check is deliberate because it applies to backends in general.

> MLKEM_NATIE_CONFIG_FILE is not properly replaced in some comments in examples (as it contains a typo)

Wow, good catch. Fixed too.

> (unrelated, sorry) FIPS202_NAMESPACE can be removed from proofs/cbmc/Makefile_params.common

Fixed, too.

mlkem/fips202/native/aarch64/src/cortex_a55_impl.h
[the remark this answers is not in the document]
hanno-becker: Ah that's what you mean. Renaming as well.

mkannwischer (approved):
Thanks for the changes. LGTM now.

Issue #749, opened 2025-02-05
`autogen --aarch64-clean` causes `tests all` to fail on macOS


rod-chapman:
on macOS... running within the NIX shell, on the HEAD of main:

```
make clean
autogen --aarch64-clean
tests all
```
fails with
```
INFO  > Example (multilevel build, native)       Compile     (native):         make -C examples/multilevel_build_native -j10
duplicate symbol '_mlkem_rej_uniform_asm_clean' in:
    /Users/rodchap/Desktop/rod/projects/crypto/pqcp/pqcp/examples/multilevel_build_native/build/mlkem512/mlkem_native/mlkem/native/aarch64/src/rej_uniform_asm_clean.S.o
    /Users/rodchap/Desktop/rod/projects/crypto/pqcp/pqcp/examples/multilevel_build_native/build/mlkem768/mlkem_native/mlkem/native/aarch64/src/rej_uniform_asm_clean.S.o
    /Users/rodchap/Desktop/rod/projects/crypto/pqcp/pqcp/examples/multilevel_build_native/build/mlkem1024/mlkem_native/mlkem/native/aarch64/src/rej_uniform_asm_clean.S.o
ld: 1 duplicate symbols
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [Makefile:140: build/test_binary] Error 1
ERROR > Example (multilevel build, native)       Compile     (native):         make failed: 2
INFO  > Example (multilevel build, native)       All         (native):         make run -C examples/multilevel_build_native
ERROR > Example (multilevel build, native)       All         (native):         'make run -C examples/multilevel_build_native' failed with with 2
ERROR > Example (multilevel build, native)       All         (native):         duplicate symbol '_mlkem_rej_uniform_asm_clean' in:
    /Users/rodchap/Desktop/rod/projects/crypto/pqcp/pqcp/examples/multilevel_build_native/build/mlkem512/mlkem_native/mlkem/native/aarch64/src/rej_uniform_asm_clean.S.o
    /Users/rodchap/Desktop/rod/projects/crypto/pqcp/pqcp/examples/multilevel_build_native/build/mlkem768/mlkem_native/mlkem/native/aarch64/src/rej_uniform_asm_clean.S.o
[...]

PR #751, opened 2025-02-05
Fix include guard for clean AArch64 native implementation

hanno-becker:
* Fix https://github.com/pq-code-package/mlkem-native/issues/749


rod-chapman (approved):
Looks good to me. Please merge.

PR #752, opened 2025-02-06
Various namespacing changes

hanno-becker:
This PR makes a series of smaller namespacing changes, largely adding MLK_ to macros.

Also, it changes the monobuild file to #undef _all_ macros, including those defined in K-generic files. Those were previously omitted in the belief they would not harm, but in fact they leave some macros dangling, e.g. `poly`, which was kept as `MLKEM_NAMESPACE(poly)` while `MLKEM_NAMESPACE` was #undef'ed.


dev/x86_64/src/arith_native_x86_64.h
```diff
@@ -11,9 +11,9 @@
 #include <stdint.h>
 #include "consts.h"
 
-#define REJ_UNIFORM_AVX_NBLOCKS 3 /* See MLKEM_GEN_MATRIX_NBLOCKS */
-#define REJ_UNIFORM_AVX_BUFLEN \
-  (3 * 168) /* REJ_UNIFORM_AVX_BUFLEN * SHAKE128_RATE */
+#define MLK_AVX2_REJ_UNIFORM_NBLOCKS 3 /* See MLKEM_GEN_MATRIX_NBLOCKS */
```
mkannwischer, line 14: This isn't actually used.

dev/x86_64/src/arith_native_x86_64.h
```diff
@@ -11,9 +11,9 @@
...
+#define MLK_AVX2_REJ_UNIFORM_BUFLEN \
+  (3 * 168) /* REJ_UNIFORM_BUFLEN * SHAKE128_RATE */
```
mkannwischer, line 16: I don't quite understand this comment. should be nblocks I suppose?

mlkem/native/x86_64/src/default_impl.h
```diff
@@ -19,7 +19,7 @@
 
 #define MLK_USE_NATIVE_NTT_CUSTOM_ORDER
 
-#define MLK_USE_NATIVE_REJ_UNIFORM
+#define MLK_USE_NATIVE_MLK_AVX2_REJ_UNIFORM
```
mkannwischer, line 22: looks wrong.

mkannwischer (approved):
Overall looks good. Some minor points.

dev/x86_64/src/arith_native_x86_64.h
hanno-becker: Removed

dev/x86_64/src/arith_native_x86_64.h
hanno-becker: Done

mlkem/native/x86_64/src/default_impl.h
hanno-becker: Fixed

PR #755, opened 2025-02-06
Make custom FIPS-202 headers configurable

### LINKED ISSUE #743: CI: Test liboqs integration
mlkem-native has been integrated into liboqs (see #653).
We should test in our CI that future changes in mlkem-native do not break that integration. 
The approach should be similar to what is done for AWS-LC (#654). 
Here are the steps needed for liboqs: 
 - clone the main branch of liboqs (if this is too flaky, we can fix to the last release version)
 - Patch the [copy_from_upstream.yml](https://github.com/open-quantum-safe/liboqs/blob/a554b36dd321e94c276e85c025f350c70740f328/scripts/copy_from_upstream/copy_from_upstream.yml#L37) with the new commit. 
 - Run the [copy_from_upstream.py](scripts/copy_from_upstream/copy_from_upstream.py) (that has some python dependencies)
 - Run the liboqs build as per README
    - Once with `cmake -DOQS_DIST_BUILD=ON -DOQS_OPT_TARGET=auto -DCMAKE_BUILD_TYPE=Release` (C backend)
    - Once with `cmake -DOQS_DIST_BUILD=OFF -DOQS_OPT_TARGET=auto -DCMAKE_BUILD_TYPE=Release` (native backend - should do this on both x86 and AArch64)
 - Run tests `pytest -k ML-KEM`

@bhess, does this make sense?

hanno-becker:
While consumers of mlkem-native can bring their own FIPS-202 implementation (as demonstrated by libOQS and the example/bring_your_own_fips202) it is slightly rigid in that the FIPS-202 headers have to be reachable through a path "fips202/fips202.h" and "fips202/fips202x4.h". In the case of libOQS, this is the reason why a patch is needed.

This commit makes the installation of a custom FIPS-202 implementation more convenient, by adding the configuration options
- MLK_FIPS202_CUSTOM_HEADER
- MLK_FIPS202X4_CUSTOM_HEADER If set, those must be the names of files replacing fips202/fips202.h and fips202/fips202x4.h If they are not set (default), mlkem-native's own FIPS-202 implementation will be used.

This is demonstrated in the case of example/bring_your_own_fips202, where we can now freely rename the directory holding the tiny_sha3 implementation.

Also, add basic integration test for libOQS. 

* Resolves #743


mkannwischer (commented):
Nice! I'm happy with those changes. 

Would be good to get confirmation that this works for oqs before merging. 
Could you also adjust the flags in the liboqs meta files, please?

hanno-becker:
@mkannwischer I tested that this works in libOQS with the following:
```diff
--- a/scripts/copy_from_upstream/copy_from_upstream.yml
+++ b/scripts/copy_from_upstream/copy_from_upstream.yml
@@ -33,13 +33,12 @@ upstreams:
   -
     name: mlkem-native
     git_url: https://github.com/pq-code-package/mlkem-native.git
-    git_branch: v1.0.0-alpha.2
-    git_commit: 3f1b9fc214a3c3f18e88b144f68814ea7ae88625
+    git_branch: fips202_inc
+    git_commit: 759d8f52c466627bf7b94d29618b2c0f47a6328e
     kem_meta_path: 'integration/liboqs/{pretty_name_full}_META.yml'
     kem_scheme_path: '.'
-    patches: [mlkem-native-ml_kem.patch]
- ```

Good to merge?

hanno-becker:
@mkannwischer @bhess I added a basic integration test with libOQS.

mkannwischer (approved):
Thanks @hanno-becker this is great! I tested that the oqs tests break if I change the testvector hash in one of the oqs yml files, so it seems to be correclty importing the new version.

bhess:
That's great, thanks for adding that and checking the liboqs integration!

PR #761, opened 2025-02-08
x86_64: Add CET instructions and check in CI

### LINKED ISSUE #762: Support Intel's Control-Flow Protections (CET)
**Context** See https://lpc.events/event/7/contributions/729/attachments/496/903/CET-LPC-2020.pdf
Intel added hardware support to stop Return-Oriented Programming (ROP) attacks. IT consists of a Shadow Stack (SHSTK) and Indirect Branch Tracking (IBT). 

Our current x86 assembly does not implement CET. That means that when compiling with `-fcf-protection` (only available on x86), the resulting binary won't actually have cf protections enabled.
To fix our assembly two things need to happen:

1) Each object file generated needs to contain the appropriate properties signalling that it does support CET. That basically means adding
```
	.section	.note.gnu.property,"a"
.align 8
	.long	 1f - 0f
	.long	 4f - 1f
	.long	 5
0:
	.string	 "GNU"
1:
	.align 8
	.long	 0xc0000002
	.long	 3f - 2f
2:
	.long	 0x3
 3:
   .align 8
4:
```

or including <cet.h> in the assembly files.

2) Each potential target of an indirect branch needs to be an `ENDBR64` instruction. `<cet.h>` provides a convienent `_CET_ENDBR` macro for that which is empty if CET is disabled, `ENDBR64` when it's enabled on 64-bit platforms, and `ENDBR32` when it's enabled on 32-bit platforms.

mkannwischer:
Modern x86_64 support hardware features through the CET instructions to protect
against return-oriented programming (ROP) attacks. Those protections can be
enabled through -fcf-protection=.
For this to work, all compilation units (including assembly) need to support
CET.

This commit adds support and contains two main changes:

1) Each assembly file needs to set the the appropriate note.gnu.property
 section to signal that it does support CET. We achieve this by including cet.h
 in all assembly files which sets the required properties.
 Note that this applies also to empty compilation units (e.g., the aarch64
 assembly files when compiling for x86_64).

2) Each label that can potentially be a target of an indirect branch needs
 to start with en endbr64 instruction, otherwise the branch faults.
 Our assembly does not use indirect branches and, hence, any internal branching
 is unaffected by this.
 However, the global symbols may potentially be targets of indirect branches
 (they don't seem to be though). To address this, we add endb64 to every global
 symbol by using the _CET_ENDBR macro provided by cet.h. Note that this is
 only adding the instruction in case CET is enabled.
 We introduce a new macro MLK_ASM_FN_SYMBOL that adds this automatically for
 X86 systems. This way our assembly simplification scripts are unaffected as
 the endbr64 instructions are out of scope.

Resolves #762

### DIFF
--- a/mlkem/sys.h
+++ b/mlkem/sys.h
@@ -109,4 +109,30 @@
 #define MLK_ALIGN /* No known support for alignment constraints */
 #endif
 
+
+/* New X86_64 CPUs support Conflow-flow protection using the CET instructions.
+ * When enabled (through -fcf-protection=), all compilation units (including
+ * empty ones) need to support CET for this to work.
+ * For assembly, this means that source files need to signal support for
+ * CET by setting the appropriate note.gnu.property section.
+ * This can be achieved by including the <cet.h> header in all assembly file.
+ * This file also provides the _CET_ENDBR macro which needs to be placed at
+ * every potential target of an indirect branch.
+ * If CET is enabled _CET_ENDBR maps to the endbr64 instruction, otherwise
+ * it is empty.
+ * In case the compiler does not support CET (e.g., <gcc8, <clang11),
+ * the __CET__ macro is not set and we default to nothing.
+ * Note that we only issue _CET_ENDBR instructions through the MLK_ASM_FN_SYMBOL
+ * macro as the global symbols are the only possible targets of indirect
+ * branches in our code.
+ */
+#if defined(MLK_SYS_X86_64)
+#if defined(__CET__)
+#include <cet.h>
+#define MLK_CET_ENDBR _CET_ENDBR
+#else
+#define MLK_CET_ENDBR
+#endif
+#endif
+
 #endif /* MLK_SYS_H */
--- a/scripts/autogen
+++ b/scripts/autogen
@@ -1056,7 +1056,14 @@ def synchronize_backends(
         delete=delete,
         force_cross=force_cross,
         no_simplify=no_simplify,
-        cflags="-Imlkem/native/x86_64/src/ -mavx2",
+        # Turn off control-flow protection (CET) explicitly. Newer versions of
+        # clang turn it on by default and insert endbr64 instructions at every
+        # global symbol.
+        # We insert endbr64 instruction manually via the MLK_ASM_FN_SYMBOL
+        # macro.
+        # This leads to duplicate endbr64 instructions causing a failure when
+        # comparing the object code before and after simplification.
+        cflags="-Imlkem/native/x86_64/src/ -mavx2 -fcf-protection=none",
     )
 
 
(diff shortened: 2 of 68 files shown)


scripts/simpasm
```diff
@@ -278,7 +278,7 @@ def simplify(logger, args, asm_input, asm_output=None):
         else:
             autogen_header += [
                 f".global {sym}",
-                f"{sym}:",
+                f"{sym.replace('MLK_ASM_NAMESPACE', 'MLK_ASM_FN_SYMBOL')}",
```
hanno-becker, line 281: Why don't we use the macro in the dev files too?

mlkem/sys.h
```diff
@@ -109,4 +109,29 @@
...
+ * the __CET__ macro is not set and we default to nothing.
+ * Note that we only issue _CET_ENDBR instructions through the MLK_ASM_FN_SYMBOL
+ * macro as the global symbols are the only possible targets of indirect
+ * branches in our code.
+ */
+#if defined(MLK_SYS_X86_64)
+#if defined(__CET__)
+#include <cet.h>
+#else
+#define _CET_ENDBR
```
hanno-becker, line 133: Wrap this macro in MLK_XXX? otherwise there is risk it unintentionally gets undefined at the end of the monobuild

scripts/simpasm
hanno-becker: Or wait, we seem to... so why is this needed?

hanno-becker (changes requested):
I would prefer not to dummy-define _CET_xxx but use a MLK_ wrapper. Otherwise we risk breaking this with the undef section in the monobuild

scripts/simpasm
mkannwischer: The symbol name is grepped out of the .global directive. 
Before we were using MLK_ASM_NAMESPACE() for both the .global and the label. 
Now we use `MLK_ASM_NAMESPACE` for the .global, but `MLK_ASM_FN_SYMBOL` for the label. 
Would be nice to clean this up in the future and combine everything into one macro, but I did not manage to get that to work.

mlkem/sys.h
mkannwischer: yes, good point. Changed.

scripts/simpasm
hanno-becker: Ack

scripts/autogen
```diff
@@ -1056,7 +1056,14 @@ def synchronize_backends(
         delete=delete,
         force_cross=force_cross,
         no_simplify=no_simplify,
-        cflags="-Imlkem/native/x86_64/src/ -mavx2",
+        # Turn off control-flow protection (CET) explicitly. Newer versions of
+        # clang turn it on by default and insert endbr64 instructions at every
+        # global symbol.
+        # We insert endbr64 instruction manually via the MLK_ASM_FN_SYMBOL
+        # macro.
+        # This leads to duplicate endbr64 instructions causing a failure when
```
hanno-becker, line 1064: This comment seems a bit misplaced -- it should be moved up?

hanno-becker (approved):
Thank you @mkannwischer !

PR #763, opened 2025-02-08
Add option for destruction of intermediate values

hanno-becker:
FIPS-203, Section 3.3 demands the destruction of intermediate values before returning to the caller. This commit adds a compile-time option MLK_DESTRUCT_INTERMEDIATES which implements that.

When set, MLK_DESTRUCT_INTERMEDIATES causes all intermediate stack buffers to be wiped prior to function return, via a new function ct_zeroize() in verify.h. (NB: The constant-time'ness is not relevant here, but conceptually the function is still best-placed in verify.h alongside other functions which, when naively implemented, would be prone to harmful compiler optimization -- ct_zeroize fits this category).

By default, ct_zeroize() is implemented via SecureZeroMemory on Windows, and a plain memset + compiler barrier otherwise. Using memset_s would be preferred, but there is no portable way of detecting whether it is available. If users need to register a custom ct_zeroize, they can do so by defining MLK_CT_ZEROIZE_NATIVE and defining ct_zeroize_native, similar to how the arithmetic and FIPS-202 backends work.

Even if MLK_DESTRUCT_INTERMEDIATES is not set, there are still dummy function calls to ct_zeroize() in the code. Those are no-ops, but are kept nonetheless to force the CBMC proofs to consider them, based on a contract for ct_zeroize().


hanno-becker:
@mkannwischer I don't expect this to be perfect yet, but would appreciate a first review on the approach taken.

hanno-becker:
In my view, the performance drop is small enough that this could be enabled by default.

mlkem/verify.h
```diff
@@ -314,4 +314,53 @@ __contract__(
...
+#else
+#error MLK_DESTRUCT_INTERMEDIATES set but no default implementation available.
+#endif
+#else  /* MLK_DESTRUCT_INTERMEDIATES */
+static MLK_INLINE void ct_zeroize(void *ptr, size_t len)
+{
+  ((void)ptr);
+  ((void)len);
+}
+#endif /* MLK_ZEROIZE_SECRET_STACK */
```
mkannwischer, line 364: rename  to `MLK_DESTRUCT_INTERMEDIATES`

mlkem/config.h
```diff
@@ -235,6 +235,41 @@
...
+ *              and a memset + compiler barrier otherwise. If you want to
+ *replace this default implementation, set MLK_USE_CT_ZEROIZE_NATIVE and provide
+ *a definition for `ct_zeroize_native(void *ptr, size_t len);`.
+ *
+ *****************************************************************************/
+#define MLK_DESTRUCT_INTERMEDIATES
+/* #define MLK_USE_CT_ZEROIZE_NATIVE
+   #if !defined(__ASSEMBLER__)
+   #include "sys.h"
+   static MLK_INLINE void ct_zeroize_native(void *ptr, size_t len)
```
mkannwischer, line 266: If I uncomment this and add my own implementation - I get a "
```
mlkem/config.h:266:56: error: unknown type name 'size_t'
  266 |    static MLK_INLINE void ct_zeroize_native(void *ptr, size_t len)
 ```
 
 If I add a `#include <stddef.h>` it works fine

**Review by mkannwischer — CHANGES_REQUESTED** [review]
Overall looks good to me. I tested that I can provide my own native implementation. 

Why did you not zeroize the following functions?
poly_rej_uniform - buf 
check_pk - p_reencoded
check_sk - check
indcpa_dec - b + b_cache

File: mlkem/config.h [hunk 3]
```diff
@@ -235,6 +235,41 @@
...
+ *              intermediate stack buffers prior to returning.
+ *
+ *              WARNING:
+ *              This feature reduces the likelihood of secrets leaking
+ *              on the stack, but does not eliminate it!
+ *              The C standard makes no guarantee about where a compiler
+ *              allocates structures and whether/where it makes copies of
+ *              them. Also, in addition to entire structures, there may also
+ *              be potentially exploitable leakage of individual values on the
+ *stack.
```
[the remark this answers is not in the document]
hanno-becker: oof. clang-format can't handle this, sadly.

mlkem/verify.h
hanno-becker: done

mlkem/config.h
hanno-becker: done

mkannwischer (changes requested):
Apologies - I found another missed buffer in `polyvec_basemul_acc_montgomery_cached_avx2`.

Another question is how we handle Keccak states - should those be zeroed out as well?

hanno-becker:
@mkannwischer I checked what AWS-LC does, which should be our initial orientation. They zeroize intermediate Keccak states as well, so I added it.

mkannwischer (approved):
Thanks for the changes!

PR #765, opened 2025-02-10
Upgrade nix to use 24.11 channel

potsrevennil:
**Summary**:
- Upgrade nix channel to 24.11
- Keep 24.05 for gcc48 and gcc49, as both are deprecated in 24.11.
- Clean up redundant Nix wrapper functions for easier maintenance.
- Set Clang 15 as the default compiler for macOS to prevent namespace checking issues in CI (Clang 16 seems to cause failures). A follow-up PR is needed to properly resolve this.
- As noted in `nix/util.nix`: 
  - Tools in /Library/Developer/CommandLineTools/usr/bin on macOS are inaccessible in the Nix shell. This issue is addressed in https://github.com/NixOS/nixpkgs/pull/353893 but hasn’t been merged into the 24.11 channel yet. As a workaround, we include the `git` dependency for macOS temporary. 
- Refactored Nix installation and cache setup in setup-nix action:  
  - Nix cache is now built per os-arch instead of combinations of OS, architecture, shell, and root/non-root users.
  - Only the `build_nix_cache` job in `nix` workflow can build, save, and replace the cache.
  - For other workflows that depend on the Nix cache:
    - If started before the Nix cache is built:
      - They fetch an old cache that matches the prefix `$os-$arch`.
      - They build missing dependencies from scratch but do not save derivations as a cache.
    - If started after the Nix cache is built:
      - They successfully fetch the Nix cache using the primary key (`$os-$arch-$hash`).
      - Even if they modify the Nix store, they do not save the updated cache.
  - With the current approach, we can avoid race conditions when using the Nix cache, thereby reducing the risk of hitting the GitHub cache limit frequently.

**Performed local tests**:
 - [ ] `lint` passing
 - [ ] `tests all` passing
 - [ ] `tests bench` passing
 - [ ] `tests cbmc` passing

### DIFF
--- a/.github/workflows/ci_ec2_reusable.yml
+++ b/.github/workflows/ci_ec2_reusable.yml
@@ -119,15 +119,10 @@ jobs:
           nix-shell: ci-linter
           gh_token: ${{ secrets.AWS_GITHUB_TOKEN }}
           nix-verbose: ${{ inputs.verbose }}
-      - name: Preprocess
-        id: preprocess
-        shell: bash
-        run: |
-          echo "nix-shell=${{ inputs.cbmc && 'ci-cbmc' || 'ci' }}${{ (inputs.compile_mode == 'cross' || inputs.compile_mode == 'all') && '-cross' || '' }}" >> $GITHUB_OUTPUT
       - name: Functional Tests
         uses: ./.github/actions/multi-functest
         with:
-          nix-shell: ${{ steps.preprocess.outputs.nix-shell }}
+          nix-shell: ${{ (inputs.compile_mode == 'cross' || inputs.compile_mode == 'all') && 'ci-cross' || 'ci' }}
           nix-cache: ${{ inputs.cbmc || inputs.compile_mode == 'cross' || inputs.compile_mode == 'all' }}
           nix-verbose: ${{ inputs.verbose }}
           gh_token: ${{ secrets.AWS_GITHUB_TOKEN }}
@@ -142,7 +137,7 @@ jobs:
         if: ${{ inputs.cbmc && (success() || failure()) }}
         uses: ./.github/actions/cbmc
         with:
-          nix-shell: ${{ steps.preprocess.outputs.nix-shell }}
+          nix-shell: ci-cbmc
           nix-verbose: ${{ inputs.verbose }}
           mlkem_k: ${{ inputs.cbmc_mlkem_k }}
           gh_token: ${{ secrets.AWS_GITHUB_TOKEN }}
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -103,4 +103,4 @@ jobs:
           echo "$(dirname $(which nix))" >> $GITHUB_PATH
       - name: nix develop
         run: |
-          ${{ matrix.target.install == 'apt' && 'sudo' || '' }} nix develop --experimental-features "nix-command flakes"
+          ${{ matrix.target.install == 'apt' && 'sudo' || '' }} nix develop --experimental-features "nix-command flakes" --access-tokens "github.com=${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -24,6 +24,28 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  build_nix_cache:
+    permissions:
+      actions: 'write'
+      contents: 'read'
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: [ ubuntu-24.04, ubuntu-24.04-arm, macos-latest ]
+    name: build nix cache (${{ matrix.runner }})
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: ./.github/actions/setup-nix
+        with:
+          cache: true
+          verbose: true
+          save_cache: true
+          devShell: ci
+          gh_token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            nix develop --profile tmp
+            nix-collect-garbage
   develop_environment:
     strategy:
       fail-fast: false
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -18,6 +18,7 @@ on:
       - 'flake.lock'
       - 'flake.nix'
       - 'nix/**'
+  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -44,9 +45,20 @@ jobs:
           devShell: ci
           gh_token: ${{ secrets.GITHUB_TOKEN }}
           script: |
+            # NOTE: we're not running cross compilation tests on macOS currently
+            #       and building cross compilation toolchain on macOS runner took
+            #       extremely long time
+            if [[ ${{ runner.os }} != 'macOS' ]]; then
+              nix develop .#ci-cross --profile tmp-cross
+              # GH ubuntu-24.04 image tend to run outof space
+              if [[ ${{ matrix.runner }} == 'ubuntu-24.04' ]]; then
+                nix-collect-garbage
+              fi
+            fi
             nix develop --profile tmp
             nix-collect-garbage
   develop_environment:
+    needs: [ build_nix_cache ]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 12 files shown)


nix/cbmc/default.nix
```diff
@@ -1,36 +1,22 @@
...
 , cadical
-, libgcc
 , z3
 }:
 
 buildEnv {
   name = "pqcp-cbmc";
   paths =
     builtins.attrValues {
-      cbmc = cbmc.overrideAttrs (old: rec {
```
mkannwischer, line 17: We are updating cbmc quite often - can you please keep the fixed version so that it is easy to update the version number?

potsrevennil:
Running into a GitHub Actions cache issue ([discussion](https://github.com/orgs/community/discussions/151267)) that prevents the workflows from replacing wrongly cached entries when Nix files change, leading to very slow workflows.

Fortunately, this can be worked around by making slight changes to the Nix files, forcing workflows to create new cache entries.

Additionally, there’s a memory limitation due to the increased dependencies in the default shell after upgrading to 24.11. This can be mitigated by running nix-collect-garbage to free unused Nix derivations.

test/mk/rules.mk
```diff
@@ -27,13 +27,15 @@ $(BUILD_DIR)/%.a: $(CONFIG)
...
+ifneq ($(findstring Darwin,$(HOST_PLATFORM)),) # if is on native macOS
+ifeq ($(CROSS_PREFIX),)
 	$(Q)echo "int main(void) {return 0;}" \
 		| $(CC) -x c - -L$(BUILD_DIR) $(_CFLAGS) \
 		 -all_load -Wl,-undefined,dynamic_lookup -l$(_LIB) \
 		 -Imlkem $(wildcard test/notrandombytes/*.c) -o $(@:%.a=%_tmp.a.out)
 	$(Q)rm -f $(@:%.a=%_tmp.a.out)
-else                                           # if not on macOS
+endif
+else                                           # if not on macOS or cross compiling on macOS
```
mkannwischer, line 38: I'm not sure what is going on here. Before we have two cases, one for macos and one not macos. 
Now we have 3 cases: macos+native, not-macos, macos+cross. The last one is currently empty. Is that intentional? The commit message does not seem to reflect that.

.github/workflows/ci_ec2_reusable.yml
```diff
@@ -123,7 +123,7 @@ jobs:
         id: preprocess
         shell: bash
         run: |
-          echo "nix-shell=${{ inputs.cbmc && 'ci-cbmc' || 'ci' }}${{ (inputs.compile_mode == 'cross' || inputs.compile_mode == 'all') && '-cross' || '' }}" >> $GITHUB_OUTPUT
+          echo "nix-shell=${{ (inputs.cbmc || inputs.compile_mode == 'cross' || inputs.compile_mode == 'all') && 'ci-cross' || 'ci' }}" >> $GITHUB_OUTPUT
```
mkannwischer: This switches the shell to ci-cross for cbmc? The commit message does not reflect that.

flake.nix
```diff
@@ -39,65 +39,74 @@
...
+          };
+          devShells.ci-bench = util.mkShell {
+            packages = builtins.attrValues { inherit (config.packages) toolchains_native; };
+          };
+          devShells.ci-cbmc = util.mkShell {
+            packages = builtins.attrValues { inherit (config.packages) cbmc toolchains_native; };
+          };
+          devShells.ci-cross = util.mkShell {
+            packages = builtins.attrValues { inherit (config.packages) linters cbmc toolchains; };
+          };
```
mkannwischer: Why add cbmc here?

flake.nix
```diff
@@ -74,7 +74,7 @@
             packages = builtins.attrValues { inherit (config.packages) cbmc toolchains_native; };
           };
           devShells.ci-cross = util.mkShell {
-            packages = builtins.attrValues { inherit (config.packages) linters cbmc toolchains; };
+            packages = builtins.attrValues { inherit (config.packages) linters toolchains; };
```
mkannwischer: here cbmc is removed again - please clean this up.

.github/workflows/nix.yml
```diff
@@ -18,6 +18,7 @@ on:
       - 'flake.lock'
       - 'flake.nix'
       - 'nix/**'
+  workflow_dispatch:
```
mkannwischer, line 21: can this be removed?

.github/workflows/ci.yml
```diff
@@ -10,6 +10,16 @@ on:
...
+    paths:
+      - '**'
+      - '!flake.lock'
+      - '!flake.nix'
+      - '!nix/**'
+  workflow_run:
+    workflows:
+      - Nix
+    types:
+      - completed
```
mkannwischer, line 22: This deserves a comment. 
Do I understand correctly that: 
This runs on push if nix files did not change. If nix files change, it runs once the nix workflow completed?

mkannwischer (changes requested):
Thanks @potsrevennil for this update and clean-up. I have some questions and suggestions.

test/mk/rules.mk
potsrevennil: It's still 2 cases actually
`-all_load` seemed to only work with clang on macOS (which means without cross compilation)

`--whole-archive` works for gcc, including the cross compilation toolchains on macOS

not-macOS and macOS+cross should all fall into the `else` case I suppose ?

.github/workflows/nix.yml
potsrevennil: Yes, this can be removed.
However, I keep this just in case that if the cache got evicted, we can manually build the cache, and avoid slowing down the other workflows
In this case, do you still prefer to remove this ?

.github/workflows/ci.yml
potsrevennil: actually, I forgot to remove this
you understanding of my intention is correct, however the action doesn't seem to work as I expected, therefore it's better to remove it to avoid confusion

test/mk/rules.mk
mkannwischer: If this is supposed to be just two cases, then the code is wrong.
There is an `endif else` there. The macos + cross case does nothing.

.github/workflows/ci.yml
mkannwischer: Okay. 
How did you implement it then?

potsrevennil: So the current workflow works like this:
- only `build_nix_cache` in `nix` workflow can build, save and replace cache.
- for all the other workflows that depends on nix cache:
  - if they started before nix cache is built
    - they will first fetch an old cache that first matches the prefix `$os-$arch`
    - they then build the missing dependencies from scratch, but won't save the derivations as a cache
  - if they started after nix cache is built
    - they can successfully fetch the nix cache with the primary key (`$os-$arch-$hash`)
    - similarly, the they won't save nix cache even if they change anything in the nix store
    
I believe that only `build_nix_cache` can create/replace nix cache can avoid the race condition of using nix cache, and alos avoid running into GH cache limit frequently

mkannwischer: okay. I understand.

test/mk/rules.mk
potsrevennil: The first `endif` is for ending the `ifeq ($(CROSS_PREFIX),)` clause
then there's `else` and the second `endif` which are the same level as `ifneq ($(findstring Darwin,$(HOST_PLATFORM)),)` clause

or do I misunderstand something ?

mkannwischer: but what if `ifeq ($(CROSS_PREFIX),)` is false?

mkannwischer:
Why is qemu listed twice? Once as `qemu; # 8.2.4` and once as `qemu = pkgs-unstable.qemu; # 9.2.0`.

test/mk/rules.mk
mkannwischer: This is not yet resolved, right?

mkannwischer (changes requested):
`nix develop` still takes ages on my MacOS machine. 
It is installing things like `libgcc-aarch64-unknown-linux-gnu-13.3.0` that take ages to compile. This sounds related to the CI workarounds for slow builds on MacOS.

test/mk/rules.mk
mkannwischer: sorry - it is fixed now.

potsrevennil:
```
overlays = [
  (_:_: {
    gcc48 = pkgs-2405.gcc48;
    gcc49 = pkgs-2405.gcc49;
    qemu = pkgs-unstable.qemu; # 9.2.0
  })
];
```

in here, I specified that changing qemu version in the `pkgs` set (which is the nixpkgs 24.11 channel) to use the qemu version (9.2.0) from pkgs-unstable (which is the nixpkgs unstable channel)
and from then on, in the flake.nix or `nix/**.nix`, when we specified qemu, it will be the version of 9.2.0 as if we're using the version from the unstable channel

The `8.2.4` was a mistaken comment, I've removed it now.

mkannwischer (approved):
Thanks @potsrevennil! This looks good now!

PR #767, opened 2025-02-10
Rename ENABLE_CT_TESTING -> MLK_ENABLE_CT_TESTING

hanno-becker:
In keeping with namespacing of other macros.


mlkem/indcpa.c
```diff
@@ -302,16 +298,13 @@ void indcpa_keypair_derand(uint8_t pk[MLKEM_INDCPA_PUBLICKEYBYTES],
...
-#ifdef ENABLE_CT_TESTING
   /*
    * Declassify the public seed.
    * Required to use it in conditional-branches in rejection sampling.
    * This is needed because all output of randombytes is marked as secret
    * (=undefined)
    */
-  VALGRIND_MAKE_MEM_DEFINED(publicseed, MLKEM_SYMBYTES);
-#endif
+  MLK_CT_DECLASSIFY(publicseed, MLKEM_SYMBYTES);
```
mkannwischer, line 307: I'm not sure if which this change it is clear that this is _only_ needed for the tests. 
Could we keep those within `#ifdef MLK_CT_TESTING_ENABLED` guards?
Or alternatively rename the macro to `MLK_CT_TESTING_DECLASSIFY` and `MLK_CT_TESTING_SECRET`?

mkannwischer (approved):
Thanks for cleaning this up @hanno-becker!
One suggestion - but not essential.

mlkem/indcpa.c
hanno-becker: Done

mkannwischer: Thanks!

PR #769, opened 2025-02-11
Add Pairwise Consistency Test (PCT) for FIPS-203 IG compliance

hanno-becker:
This PR adds the Pairwise Consistency Check (PCT) to key generation.

The declassification of the public key is moved to key generation, past the PCT.

The toplevel API is modified to force the checking of return values, and tests are adjusted accordingly.

PCT is disabled by default, but most tests in CI are configured to enable PCT (Generally speaking, we want the CI to run on "full" configurations, even if those aren't the default).


mkannwischer (approved):
Thanks @hanno-becker for adding support for PCT!

I am unsure if it's allowed to do PCT with a fixed seed. But we can change that later if required.

PR #771, opened 2025-02-14
CT: Mark random coins as secret outside of randombytes()

hanno-becker:
Previously, valgrind CT tests would assume that the underlying randombytes() function would tag its output bytes as secret.

This commit changes this to explicitly tag the coins as secret in the ML-KEM source code.

This simplifies integration with other libraries using a custom randombytes() implementation, and makes the secret/declassify annotations with mlkem-native self-contained.

### DIFF
--- a/mlkem/kem.c
+++ b/mlkem/kem.c
@@ -188,7 +188,11 @@ int crypto_kem_keypair(uint8_t pk[MLKEM_INDCCA_PUBLICKEYBYTES],
 {
   int res;
   MLK_ALIGN uint8_t coins[2 * MLKEM_SYMBYTES];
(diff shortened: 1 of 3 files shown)


mkannwischer (changes requested):
Yes, this makes sense to me, but you missed one call to randombytes: https://github.com/pq-code-package/mlkem-native/blob/a1afd46eab1ee60d2cce3abd8c10e8b19b345c33/mlkem/kem.c#L239

hanno-becker:
@mkannwischer Thanks, fixed.

mkannwischer (approved):

PR #772, opened 2025-02-14
Explicitly mark external API with MLK_EXTERNAL_API

hanno-becker:
This commit adds MLK_EXTERNAL_API annotations to the toplevel API of mlkem-native. By default, MLK_EXTERNAL_API is defined as the empty string, hence giving external linkage to the public API.

However, in a monobuild where both mlkem-native and the consuming application are in the same compilation unit, MLK_EXTERNAL_API can be defined as `static`, hence giving internal linkage to mlkem-native.

This is exemplified through modification of the example `monolithic_build_multilevel_native`.


examples/monolithic_build_multilevel_native/Makefile
```diff
@@ -125,16 +120,11 @@ $(BUILD_DIR)/%.S.o: %.S
...
-	$(Q)[ -d $(@D) ] || mkdir -p $(@D)
-	$(Q)$(AR) rcs $@ $^
-	$(Q)strip -S $@
-
-$(BINARY_NAME_FULL): $(APP_SOURCE) $(LIB_NAME_FULL)
+$(BINARY_NAME_FULL): $(APP_SOURCE) $(MLK_OBJ_ASM)
 	$(Q)echo "CC  $@"
 	$(Q)[ -d $(@D) ] || mkdir -p $(@D)
 	$(Q)$(CC) $(CFLAGS) $(INC) $^ -o $@
+	$(Q)strip -S $@
```
mkannwischer, line 127: Should this include the cross prefix?

mkannwischer (approved):
Looks good to me. I have one question concerning cross-compliation, but that's unrelated to your PR.

examples/monolithic_build_multilevel_native/Makefile
hanno-becker: Good question. If it is not necessary -- which appears to be the case -- I'd prefer to leave it as is.

PR #773, opened 2025-02-14
Faster INTT on AArch64 by more efficient reduction

rod-chapman:
This PR introduces a small change in AArch64 intt_clean.S that reduces the number of intermediate
reduction steps from 4 to 3, through more thorough tracking of coefficient bounds.

TODO
1. Applicatioin of SLOTHY to this code to also update intt_opt.S
2. Re-proof of both Clean and Opt versions with HOL-Light
3. Benchmarks


hanno-becker (commented):
@rod-chapman Good catch, this may indeed be a little improvement, but there's more work to do in this PR?

rod-chapman:
Yes... is "slothy-cli" installed by the NIX environment? If not, then what platform should I be running SLOTHY on?

hanno-becker (approved):
LGTM -- thanks @rod-chapman !

PR #781, opened 2025-02-17
Change unsigned to __CPROVER_size_t inside quantified expressions

rod-chapman:
CBMC team have suggested that "__CPROVER_size_t" would be more appropriate for the type of the quantified variable inside a quantified expression.

This PR makes that change inside the macros in cbmc.h only to determine potential impact of this change.

### DIFF
--- a/mlkem/cbmc.h
+++ b/mlkem/cbmc.h
@@ -81,14 +81,14 @@
 #define forall(qvar, qvar_lb, qvar_ub, predicate)                 \
   __CPROVER_forall                                                \
   {                                                               \
-    unsigned qvar;                                                \
+    __CPROVER_size_t qvar;                                        \
     ((qvar_lb) <= (qvar) && (qvar) < (qvar_ub)) ==> (predicate)   \
   }
 
-#define EXISTS(qvar, qvar_lb, qvar_ub, predicate)         \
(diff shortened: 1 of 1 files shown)


mlkem/cbmc.h
```diff
@@ -118,7 +118,7 @@
                          value_lb, value_ub)                           \
   __CPROVER_forall                                                     \
   {                                                                    \
-    unsigned qvar;                                                     \
+    __CPROVER_size_t qvar;                                             \
```
tautschnig, line 121: I wonder what the performance would be with `__CPROVER_ssize_t` instead, which would avoid a type cast (at present, CBMC uses signed integers for index expressions (the C standard is not clear as whether index expressions have signed or unsigned integer type)).

rod-chapman: I tried CPROVER_ssize_t and it makes no difference - still much slower. It seems the difference is something to do with "unsigned" being "32 bit" but ssize_t and size_t are "64 bit". Look at the generated GOTO files.

hanno-becker:
Unless @tautschnig has ideas why the move to size_t is detrimental here, I'd suggest closing and sticking to the current approach of using `unsigned` indices.

rod-chapman:
Agreed - no change to mlkem-native code for now, but and Issue for CBMC team to look into.

PR #783, opened 2025-02-18
Add magic constant checker

hanno-becker:
This commit adds a script scripts/check-magic which looks for magic constants and checks that correct symbolic explanations for them have been given.

Some constants in the AVX2 code are not yet explained, and a TODO is left. The magic checker is temporarily disabled for those through special annotations in the code.

A magic constant should now be prefixed or suffixed with an annotation of the form `/* check-magic: MAGIC_VAL == EXPR */`, where `EXPR` is a symbolic expression which may involve

- unsigned_mod
- signed_mod
- pow


mkannwischer (approved):
Thanks @hanno-becker - I think this is useful.
Some work left, but we can do that in a follow-up PR.

PR #784, opened 2025-02-18
Add generalized 2- and 3-layer merging of the forward NTT

rod-chapman:
This PR adds functions in poly.c that perform generalized 2- and 3-layer merges for the forward NTT.
This adds to the existing single-layer processing function, so that many different "merge strategies"
can be implemented and measures for performance on different platforms.

This commit initially implements a "3211" merging that has been found experimentally to be reasonably efficient
on AArch64 platforms using GCC 13 and 14.

CBMC proofs of these new functions are TBD.

Further benchmarks and experiments will determine the best layer merge for additional targets and compilers.


hanno-becker:
I am closing this for now. We have native backends for x86_64, AArch64 and soon RV64, and not enough data to support this change on other platforms. We may revisit this in the future, of course.

PR #786, opened 2025-02-19
Move zetas.c -> zetas.inc

### LINKED ISSUE #732: Drop zeta table for C [inv]NTT if native [inv]NTT are used

hanno-becker:
* Resolves #732 

This commit changes zetas.c from a standalone compilation unit
to an .inc file which is included into poly.c if the zeta table
is needed.

This resolves two issues:
- The zeta table should have static linkage in a monobuild,
  which it previously didn't have. Now, it always has static
  linkage.
- The zeta table should be omitted if NTT, invNTT and mulcache
  computation have been replaced by native implementations.

Also, a minor CBMC change:

Previously, `array_bound(x, 0, N, lb, ub)` required `N > 0`, hence
the prefix `N > 0 ==> ...` would be needed if only `N == 0` was a
possibility.

This has since been corrected, and `array_bound(x, 0, N, lb, ub)`
works correctly even for `N == 0`.

Accordingly, this commit drops unnecessary `N > 0 ==> ...` prefixes
from some loop invariant in `poly_rej_uniform_x4`. The corresponding
invariant in `poly_rej_uniform` was already simplified.


mlkem/poly.c
```diff
@@ -228,6 +228,14 @@ void poly_sub(poly *r, const poly *b)
   }
 }
 
+/* Include zeta table unless NTT, invNTT and mulcache computation
+ * have been replaced by native implementations. */
+#if !defined(MLK_USE_NATIVE_POLY_MULCACHE_COMPUTE) && \
+    !defined(MLK_USE_NATIVE_NTT) && !defined(MLK_USE_NATIVE_INTT)
+#include "zetas.inc"
```
mkannwischer, line 235: This is wrong isn't it? This will include it only if _all_ three are not defined, but you want to include it if _any_ of the three is not defined.

mkannwischer (changes requested):
Minor mistake (unless I am misunderstanding). 
Rest looks good!

mlkem/poly.c
hanno-becker: Good spot, fixed

mkannwischer (approved):
Thanks for the fix.

PR #788, opened 2025-02-19
Add runtime option to break PCT

hanno-becker:
This commit introduces a config option MLK_KEYGEN_PCT_BREAKAGE_TEST. When set, the user must provide a runtime (!) function mlk_break_pct() to indicate if the PCT should be made fail. If set, the shared secret in the PCT will be deliberately corrupted to make the PCT fail. A test is added to exercise the runtime breakage logic.


mkannwischer (approved):
I think neither PCT itself, nor testing it using it a runtime configuration option are a good idea.... but this looks okay.

PR #794, opened 2025-02-21
Delete cbmc-proof.txt marker files

mkannwischer:
Before, we placed an empty cbmc-proofs.txt file into a directory in proofs/cbmc to mark that this is a CBMC proof.
However, all directories except for lib are proofs. This commit changes the script to consider all directories in proofs/cbmc/ that have a Makefile as CBMC proofs.

### DIFF
--- a/.github/actions/cbmc/action.yml
+++ b/.github/actions/cbmc/action.yml
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 name: CBMC
-description: Run CBMC proofs for MLKEM-C_AArch64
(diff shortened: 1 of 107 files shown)


hanno-becker (commented):
I realise this is a draft, but I may as well mention: The proof guide and some CBMC documentation mention `cbmc-proof.txt`

mkannwischer:
> I realise this is a draft, but I may as well mention: The proof guide and some CBMC documentation mention `cbmc-proof.txt`

Yes, good catch! I've adjusted the proof guide as well.

hanno-becker (approved):

PR #796, opened 2025-02-21
Remove opt clean suffixes from AArch64 backend

### LINKED ISSUE #793: Remove `xxx_clean` vs. `xxx_opt` suffix for clean/opt assembly
Initially, the suffixes were needed to tell apart clean vs. opt implementation in the same build. Now, we never build them together, so the name distinction is unnecessary.

- [ ] Remove all `xxx_opt` and `xxx_clean` suffixes from the (a) dev/aarch64_clean/..., (b) dev/aarch64_opt/... (c) mlkem/native/.... aarch64 assembly, covering both file names and global symbols. In particular, the backend headers need adjusting.

rod-chapman:
Fixes #793 

This PR removes the "_opt" and "_clean" suffix from filename, function names, and symbols in the AArch64 back-end.

The "autogen" script is also updated to correctly copy, move and generate these files as appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ undefined behaviour in C, including out of bounds memory accesses and integer ov
 all C code in [mlkem/*](mlkem) and [mlkem/fips202/*](mlkem/fips202) involved in running mlkem-native with its C backend.
 See [proofs/cbmc](proofs/cbmc) for details.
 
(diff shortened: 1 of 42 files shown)


hanno-becker (changes requested):
Thank you @rod-chapman for working through this -- I think it will be less confusing to not have the clean vs. opt distinction in the names anymore.

Could you please squash the format / autogen commits into the commits they apply to? Ideally, every commit should have happy format+autogen.

rod-chapman:
Still waiting for 2 failed CI jobs to re-run... they appear to be stuck waiting for a runner.
Yes... will squash and re-push soon.

hanno-becker (approved):

PR #797, opened 2025-02-22
TEST: clang20

mkannwischer:
clang 20.1.0 is expected to be released next month.
This PR adds a nix shell with clang 20.1.0rc2 and runs the compiler tests and constant-time tests with it.

This is mostly intended to identify any issues early. We can merge it once clang 20 is released.


mkannwischer:
All looks fine on Linux. clang20 currently doesn't build on MacOS. I'll check this again in a few weeks.

PR #807, opened 2025-02-25
CI: Split ci.yml in three workflows

hanno-becker:
base.yml runs 'base tests'. If base.yml completes successfully, it runs cbmc.yml and ci.yml.

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -1,0 +1,236 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: Base tests
+permissions:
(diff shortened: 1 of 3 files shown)


potsrevennil:
Just a reminder, I wanted to try to do similar things, run a workflow after another is completed. However run into this github bug: https://github.com/actions/runner/issues/2324

hanno-becker:
@potsrevennil We have no file-dependency here, only workflow dependencies. But I still didn't figure out how to do it so far.

mkannwischer (approved):

PR #815, opened 2025-02-27
Call extended tests and CBMC after completion of base tests only

hanno-becker:


mkannwischer (approved):
Thanks! this is a great improvement.

PR #821, opened 2025-03-01
HOL-Light: Add proof for Keccak-F1600-x1 aarch64 asm

hanno-becker:
This commit integrates John Harrison's HOL-Light proof
for functional correctness of the scalar aarch64 implementation
of Keccak-F1600.


mkannwischer (approved):
Awesome! Thanks for the work!

LGTM! I checked that the proof runs successful in CI and that the autogen script automatically updates asm (and, hence, linting will complain if there are changes in the sources).

PR #822, opened 2025-03-01
FIPS202/AArch64: Interleave first iteration in triple hybrid Keccak

hanno-becker:
The scalar-v8a-v84a triple-hybrid Keccak-f1600-x4 implementation interleaves scalar, Neon and Neon+SHA3 implementations of Keccak-f1600.

However, in the previous code the interleaving of the initial Keccak-f1600 round was missing.

This commit interleaves the initial round as well. No algorithmic changes are made.


hanno-becker:
@mkannwischer It makes virtuall no difference -- on Gv3, one sees a 1% bump, but that's it. Still, I think it makes more sense to apply optimization measures uniformly or not at all.

mkannwischer (approved):
LGTM. Not much performance gain, but makes sense to do this consistently.

PR #831, opened 2025-03-03
CI/HOL-Light: Only run proof if respective files have changed

hanno-becker:
This PR modifies the HOL-Light CI workflow to run a proof only if the source or .ML file for that proof have changed.

It also removes all`_clean` and `_opt` suffixes from the AArch64 FIPS202 files, and align src and proof file names in `proofs/hol_light` with those in the main source tree.


mkannwischer (approved):
LGTM. Thanks! Looks good and CI confirms that only the changed proofs are executed. 
Thanks also for the renaming.

PR #834, opened 2025-03-03
CBMC: Pull and test weakening is_fresh in contract applications

### LINKED ISSUE #764: CBMC: Avoid splitting multi-dimensional arrays into multiple allocations
To work with the semantics of `__CPROVER_is_fresh / memory_no_alias`, we are currently forced to split multi-dimensional arrays into separate declarations, so that the subarrays are seen as separate objects in the CBMC sense. This is quite inconvenient and introduces a fair amount of visual clutter. 

Study if there is a way to modify the relevant specs so they don't require the arguments to be in separate objects.

hanno-becker:
* Resolves #764 

This PR updates CBMC to include the change from https://github.com/diffblue/cbmc/pull/8603 which weakens the interpretation of `memory_no_alias` (or `__CPROVER_is_fresh`) in function contract applications: Rather than asserting that every pointer featuring in a `memory_no_alis/__CPROVER_is_fresh` clause be in a separate object, pointers can now point to the same object, so long as the slices they represent are disjoint.

This allows us to undo some artificial splitting of 2D arrays into separate objects, which this PR also does.

### DIFF
--- a/examples/monolithic_build/mlkem_native_monobuild.c
+++ b/examples/monolithic_build/mlkem_native_monobuild.c
@@ -196,6 +196,7 @@
 #undef mlk_polyvec_tomont
 /* mlkem/sys.h */
 #undef MLK_ALIGN
+#undef MLK_ALIGN_UP
 #undef MLK_ALWAYS_INLINE
 #undef MLK_CET_ENDBR
 #undef MLK_CT_TESTING_DECLASSIFY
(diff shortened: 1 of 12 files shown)


mkannwischer (approved):
LGTM - Thanks @hanno-becker. 
Very welcome simplification of the contracts/proofs.

PR #805, opened 2025-02-24
Add a Makefile and documentation for automatically running SLOTHY

rod-chapman:
This PR adds a suitable Makefile and README.md that explains how to automatically run
SLOTHY on the "clean" AArch64 back-end to produce the "opt" back-end.

PR #850, opened 2025-03-05
Apply SLOTHY to AArch64 poly_tobytes_asm.S

rod-chapman:
This PR applies SLOTHY to poly_tobytes_asm.S to determine impact on performance.


hanno-becker:
It is surprising that this does not even have a meaningful impact on Cortex-A55.

mkannwischer:
main on A55:
```
mlk_poly_tobytes cycles=244
```

this PR: 
```
mlk_poly_tobytes cycles=193
```

hanno-becker:
Converting this to a draft so long as clean and optimized are out of sync (at the moment, opt is unrolled). @rod-chapman Was this merely to do a benchmarking test or were you suggesting to merge this before #805?

rod-chapman:
The improvement on A55 looks reasonable. The version with explicit unroll=4 seems to make little difference on other platforms, but happy to keep it that way if it's better on A55.

rod-chapman:
The Makefile proposed under #805 has the `-c sw_pipelining.unroll=4` in there, so we should merge that first, and rebase this branch to continue any further experiments.

rod-chapman:
On Apple M1, I see poly_tobytes_asm() drop from 49 on main to 38 cycles on this branch.

hanno-becker (changes requested):
Actually, @rod-chapman, while you are at it, would you mind emulating the LD2 by LDR+UZP just as its done for the basemul? See the `ld2_wrap` macro. 

This makes things more uniform and facilitates verification.

hanno-becker (changes requested):
Now that we are optimizing the code, we need to put a stack store/reload logic in place, otherwise the ABI is not respected.

dev/aarch64_opt/src/Makefile
```diff
@@ -12,12 +12,19 @@ TARGET_MICROARCH=Arm_Cortex_A55
...
              -c inputs_are_outputs \
-             -c reserved_regs="[x18--x30,sp]" \
              -c sw_pipelining.minimize_overlapping=False \
              -c sw_pipelining.allow_post \
              -c variable_size \
              -c constraints.stalls_first_attempt=64
 
+# For some units, we allow SLOTHY to use ALL the V/Q registers. These units MUST
+# save and restore v8-v15 as part of their entry/exit. Registers x18-x30 and sp
+# must always be avoided to respect the ABI.
```
hanno-becker, line 22: Not necessarily, X19-X30 could be stashed, too, but the SIMD-heavy routines we're optimizing here don't need them. It's different for Keccak, for example, where you really every register.

rod-chapman: I think I know what you mean... let me try to draft something.

dev/aarch64_opt/src/Makefile
```diff
@@ -12,12 +12,24 @@ TARGET_MICROARCH=Arm_Cortex_A55
...
+#
+# These units do not need all of the scalar X registers, so we tell SLOTHY
+# to avoid using x18-x30 and sp to meet the ABI. Therefore, these units do
+# NOT need to explictly save and restore these registers.
+RESERVE_X_ONLY_FLAG=-c reserved_regs="[x18--x30,sp]"
+
+# For some units, we additionally forbid SLOTHY from using v8-v15 to repect the ABI.
+# These units therefore do NOT need to save/restore these registers,
+# giving a small performance gain.
+RESERVE_ALL_FLAG=-c reserved_regs="[x18--x30,sp,v8--v15]"
```
hanno-becker, line 31: Let's keep it simple:
```suggestion
# For kernels which stash callee-saved v8-v15 but don't stash callee-saved GPRs x19-x30. 
# Allow SLOTHY to use all V-registers, but only caller-saved GPRs.
RESERVE_X_ONLY_FLAG=-c reserved_regs="[x18--x30,sp]"

# Used for kernels which don't stash callee-saved registers.
# Restrict SLOTHY to caller-saved registers.
RESERVE_ALL_FLAG=-c reserved_regs="[x18--x30,sp,v8--v15]"
```

hanno-becker:
@mkannwischer Pleas have a look. I missed a criticial bug before.

dev/aarch64_clean/src/poly_tobytes_asm.S
```diff
@@ -30,16 +40,17 @@ MLK_ASM_FN_SYMBOL(poly_tobytes_asm)
 
         mov count, #16
 poly_tobytes_asm_asm_loop_start:
-        ld2 {data0.8h, data1.8h}, [src], #32
+        ld2_wrap data, src
+//        ld2 {data0.8h, data1.8h}, [src], #32
```
mkannwischer, line 44:
```suggestion
```

mkannwischer (changes requested):
The changes look good to me.
Can you please clean-up the commit history? No need to keep all these versions of poly_tobytes_asm.S.

mkannwischer (approved):
LGTM. Thanks for the work @rod-chapman!

dev/aarch64_clean/src/poly_tobytes_asm.S
rod-chapman: Thanks

PR #852, opened 2025-03-06
Scripts: Make simpasm, autogen, tests callable from any subdirectory

### LINKED ISSUE #847: Support `autogen` in any subdirectory of repository
`autogen` and `format` should work in any subdirectory of the mlkem-native repository without change of behaviour (that is, it should apply to all files in the repository).

hanno-becker:
Previously, those scripts would need to be called from the root directory of the repository. With this change, they can be called from anywhere within the repository.

* Resolves #847

### DIFF
--- a/scripts/autogen
+++ b/scripts/autogen
@@ -1242,6 +1242,8 @@ def _main():
 
     args = parser.parse_args()
 
+    os.chdir(os.path.join(os.path.dirname(__file__), ".."))
+
     check_asm_register_aliases()
 
     gen_c_zeta_file(args.dry_run)
--- a/scripts/simpasm
+++ b/scripts/simpasm
@@ -345,6 +345,8 @@ def _main():
 
     args = parser.parse_args()
 
+    os.chdir(os.path.join(os.path.dirname(__file__), ".."))
+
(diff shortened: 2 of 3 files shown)


scripts/autogen
```diff
@@ -26,6 +26,21 @@ montgomery_factor = pow(2, 16, modulus)
...
+    while current_path != "/":
+        if os.path.exists(os.path.join(current_path, ".git")):
+            git_root = current_path
+            break
+        current_path = os.path.dirname(current_path)
+
+    if git_root is None:
+        raise Exception("Could not find root of repository")
+
+    os.chdir(git_root)
```
mkannwischer, line 41: Why not just 
```
os.chdir(os.path.join(os.path.dirname(__file__), ".."))
```
? 
feels much cleaner to me.

mkannwischer (changes requested):

scripts/autogen
hanno-becker: Fair point!

hanno-becker: Done

mkannwischer (approved):

PR #853, opened 2025-03-07
Remove metadata vs. implementation distinction in backends

### LINKED ISSUE #140: Test all implementation variants in CI
We have multiple implementations variants for Keccak now.
For NTTs, we currently have C and asm as well now, and we'll see mulitple asm variants in the future. 

Currently, for each platform exactly one (ideally the fastest) is tested.
CI should be extended to test all (functional) variants on all platforms -- not that not all platforms can run all implementations - for example, SHA-3 instructions are not available on platforms < v8.4. 
Benchmarks should continue to be limited to only the fastest implementation.

hanno-becker:
* Resolves #140 

Previously, every backend would be described via a combination
of a (a) metadata file, (b) implementation file. The metadata file would
give a name to the backend and describe where the implementation file
could be found. The metadata file would always be included by common.h,
while the implementation file would only be included when xxx_backend.h
was requested.

This commit aims to simplify the organization of material by merging
metadata and implementation files. There is, then, only one file which
specifies the USE_XXX clauses defining the scope of the backend, and
implementation the respective backend C wrappers. This single file
is unconditionally included from common.h; there is no xxx_backend.h
anymore.

Care has to be taken to make the backend file compatible with assembly.
Previously, the metadata file would include preprocessor directives only,
while the implementation file would contain C code. Now, the unified
backend file includes directives as well as C code, but guards the latter
via !__ASSEMBLER__ to make it safe to include from assembly files as well.

Every backend file also specifies custom directives indicating which
backend source files should be included. For example, the x4_v84a_scalar.h
backend sets MLK_FIPS202_AARCH64_NEED_X4_V8A_SCALAR_HYBRID which guards
the assembly in keccak_f1600_x4_v8a_scalar_hybrid_asm.S.


mkannwischer (approved):
LGTM. Thanks @hanno-becker for cleaning that up.

PR #860, opened 2025-03-10
Add ML-KEM related specs and common proof tools

### LINKED ISSUE #861: HOL-Light: Include functional specs in source tree
Currently we import ML-KEM related functional specs as part of the s2n-bignum import in nix. In particular, the functional specs are _not_ part of mlkem-native's source tree, which makes it impossible to validate the assembly specs by inspection of the source tree alone.

Copy the relevant functional specifications into the mlkem-native source tree.

hanno-becker:
* Resolves #861 

The specs were previously imported through s2n-bignum using nix,
but were not visible in the repository, which made it impossible
to validate the specs through inspection.

This commit imports the relevant `arm/proofs/utils/mlkem.ml`
directly into the source tree of mlkem-native.

It also hoists common proof tactics shared between the NTT, invNTT
and poly_tomont proof into a new `mlkem_utils.ml`.

Unfortunately, the way proofs are currently built prevents
the specification of arbitrary include paths: While, in general,
the `load_path` reference controls where HOL-Light will look for
files, the initial 'flattening' of the proof script using
`$HOLLIGHT_DIR/inline_load.ml` only considers the PWD and the
HOL-Light dir -- this is enough for s2n-bignum so long as one
`cd`s into s2n-bignum's base directory, but it's not enough for
mlkem-native, where do need all three of (a) hol-light, (b) s2n-bignum
and (c) mlkem-native proof directories.

As a remedy, this commit includes a copy of `build-proof.sh`
and adapts it to use a modified version of `inline_load.ml` which
adds a 3rd load path before running the flattening. This import
of `build-proof.sh` makes the prior nix path obsolete.


mkannwischer (approved):
Thanks @hanno-becker - it's much easier to understand the proofs now. LGTM. 

Same should be done for the Keccak specs.
In the long term, I think it would be good to establiash some connection between the code in s2n-bignum and the one in mlkem-native and ensure that they are kept in sync.

PR #864, opened 2025-03-11
HOL-Light: Add functional Keccak spec

### LINKED ISSUE #863: HOL-Light: Add functional spec for Keccak
Like #861, we should have the functional spec for Keccak in the mlkem-native source tree

hanno-becker:
This commit imports the functional spec for Keccak in HOL into the mlkem-native source tree. Previously, those were indirectly imported from s2n-bignum as part of the runtime nix setup, preventing direct inspection of the assembly specs from the source tree.

The HOL-Light Github workflow is refined slightly to allow for the specification of per-proof dependencies.

* Resolves #863

### DIFF
--- a/.github/workflows/hol_light.yml
+++ b/.github/workflows/hol_light.yml
@@ -25,15 +25,24 @@ jobs:
       fail-fast: false
       matrix:
         proof:
-          - mlkem_ntt
-          - mlkem_intt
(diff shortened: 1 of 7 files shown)


mkannwischer (approved):
Thanks! LGTM.

PR #865, opened 2025-03-11
README: Document multi-level and C-only build, + various minor fixes

### LINKED ISSUE #804: Document multi-level build in README
The README should explain that mlkem-native is at its core a single-level library, and explain how multi-level builds can be derived from it, including a pointer to the examples which do just that.

hanno-becker:
* Resolves #804


mkannwischer (approved):
LGTM.

PR #871, opened 2025-03-12
CI: Fix c17 tests and add c23 tests

mkannwischer:
Our tests were never actually running c17 tests as the condition in the Github workflow was always false. This commit fixes that.

In addition, this commit adds C23 tests (supported starting from clang18 and gcc14).


hanno-becker (approved):
Good spot, thanks @mkannwischer !

Issue #640, opened 2025-01-12
Make name of `randombytes` configurable


hanno-becker:
At the moment, the name of the PRNG is hardcoded to `randombytes`, which is inconvenient.

Make the function name of `randombytes` configurable in `config.h`.

PR #873, opened 2025-03-13
Allow customization of randombytes function name 

mkannwischer:
Resolves https://github.com/pq-code-package/mlkem-native/issues/640

This commit makes the name of the randombytes function configurable through
the config.h. Thus far, we have been assuming the function to be called
randombytes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -329,6 +329,17 @@ jobs:
           kat: true
           acvp: true
           examples: false # Some examples use a custom config themselves
+      - name: "Custom randombytes"
+        uses: ./.github/actions/multi-functest
+        with:
+          gh_token: ${{ secrets.GITHUB_TOKEN }}
+          compile_mode: native
+          cflags: "-std=c11 -D_GNU_SOURCE -DMLK_CONFIG_FILE=\\\\\\\"../test/custom_randombytes_config.h\\\\\\\" -fsanitize=address -fsanitize=undefined -fno-sanitize-recover=all"
+          func: true
+          nistkat: true
+          kat: true
+          acvp: true
+          examples: false # Some examples use a custom config themselves
       - name: "MLKEM_GEN_MATRIX_NBLOCKS=1"
         uses: ./.github/actions/multi-functest
         with:
--- a/integration/liboqs/config_c.h
+++ b/integration/liboqs/config_c.h
@@ -116,6 +116,32 @@
 */
 
 /******************************************************************************
+ * Name:        MLK_CUSTOM_RANDOMBYTES
+ *
+ * Description: mlkem-native does not provide a secure randombytes
+ *              implementation. Such an implementation has to provided by the
+ *              consumer.
+ *
+ *              If this option is not set, mlkem-native expects a function
+ *              void randombytes(uint8_t *out, size_t outlen).
+ *
+ *              Set this option and define `mlk_randombytes` if you want to
+ *              use a custom method to sample randombytes with a different name
+ *              or signature.
+ *
+ *****************************************************************************/
+#define MLK_CUSTOM_RANDOMBYTES
+#if !defined(__ASSEMBLER__)
+#include <oqs/rand.h>
+#include <stdint.h>
+#include "../../mlkem/sys.h"
+static MLK_INLINE void mlk_randombytes(uint8_t *ptr, size_t len)
+{
+  OQS_randombytes(ptr, len);
+}
+#endif
+
+/******************************************************************************
  * Name:        MLK_NO_ASM
  *
  * Description: If this option is set, mlkem-native will be built without
--- a/integration/liboqs/config_c.h
+++ b/integration/liboqs/config_c.h
@@ -76,18 +76,18 @@
 #define MLK_FIPS202X4_CUSTOM_HEADER "../integration/liboqs/fips202x4_glue.h"
 
 /******************************************************************************
- * Name:        MLK_USE_ZEROIZE_NATIVE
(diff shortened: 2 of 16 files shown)


hanno-becker (commented):
This looks good @mkannwischer -- before merge, though, I want to double-check that this will work for AWS-LC integration, which I will get to either today or tomorrow.

hanno-becker (changes requested):
The current implementation requires the custom `MLK_RANDOMBYTES_FUNC` to be of external linkage, which forces an indirection if the native implementation does not have exactly the same signature -- for example, if it has a return value.

I would suggest we change the approach to a boolean flag `MLK_CUSTOM_RANDOMBYTES`: If that's set, it is the user's responsiblity to provide in a header a static inline definition of `mlk_randombytes` that is used instead of the fallback to `randombytes()`.

The same approach should also be applied to custom zeroization,  which is conceptually similar.

mkannwischer:
> The current implementation requires the custom `MLK_RANDOMBYTES_FUNC` to be of external linkage, which forces an indirection if the native implementation does not have exactly the same signature -- for example, if it has a return value.
> 
> I would suggest we change the approach to a boolean flag `MLK_CUSTOM_RANDOMBYTES`: If that's set, it is the user's responsiblity to provide in a header a static inline definition of `mlk_randombytes` that is used instead of the fallback to `randombytes()`.
> 
> The same approach should also be applied to custom zeroization, which is conceptually similar.

Okay - I will adjust accordingly later today.

mlkem/config.h
```diff
@@ -247,15 +247,39 @@
...
+
+/******************************************************************************
+ * Name:        MLK_CUSTOM_RANDOMBYTES
+ *
+ * Description: mlkem-native does not provide a secure randombytes
+ *              implementation. Such an implementation has to provided by the
+ *              consumer.
+ *
+ *              If this option is not set, mlkem-native expects a function
+ *              randombytes(uint8_t *out, size_t outlen).
```
hanno-becker, line 272: Add return type

mlkem/randombytes.h
```diff
@@ -9,11 +9,21 @@
...
+#if defined(MLK_CUSTOM_RANDOMBYTES)
+static MLK_INLINE void mlk_randombytes(uint8_t *out, size_t outlen)
 __contract__(
   requires(memory_no_alias(out, outlen))
-  assigns(memory_slice(out, outlen))
-);
+  assigns(memory_slice(out, outlen)));
+#else
+
+void randombytes(uint8_t *out, size_t outlen);
```
hanno-becker, line 22: Don't you need a CBMC spec in this case, too?

integration/liboqs/config_c.h
```diff
@@ -100,21 +100,45 @@
...
+ *              or signature.
+ *
+ *****************************************************************************/
+#define MLK_CUSTOM_RANDOMBYTES
+#if !defined(__ASSEMBLER__)
+#include <oqs/rand/rand.h>
+#include <stdint.h>
+static MLK_INLINE void mlk_randombytes(uint8_t *ptr, size_t len)
+{
+  OQS_randombytes(prt, len)
```
hanno-becker, line 138: missing `;`

mlkem/config.h
mkannwischer: done.

mlkem/randombytes.h
mkannwischer: yes - stupid me. fixed.

integration/liboqs/config_c.h
mkannwischer: fixed.

.github/workflows/ci.yml
```diff
@@ -312,7 +312,7 @@ jobs:
         with:
           gh_token: ${{ secrets.GITHUB_TOKEN }}
           compile_mode: native
-          cflags: "-std=c11 -D_GNU_SOURCE -DMLK_CONFIG_FILE=\\\\\\\"../test/custom_zeroize_config.h\\\\\\\" -fsanitize=address -fsanitize=undefined -fno-sanitize-recover=all"
+          cflags: "-std=c11 -D_GNU_SOURCE -DMLK_CONFIG_FILE=\\\\\\\"../test/custom_zeroize_config.h\\\\\\\" -fsanitize=address -fsanitize=undefined -fno-sanitize-recover=all -Wno-redundant-decls"
```
hanno-becker, line 315: Why is this necessary?

mlkem/randombytes.h
```diff
@@ -9,11 +9,21 @@
 #include <stdint.h>
 
 #include "cbmc.h"
+#include "common.h"
 
-void randombytes(uint8_t *out, size_t outlen)
+
+#if defined(MLK_CUSTOM_RANDOMBYTES)
+static MLK_INLINE void mlk_randombytes(uint8_t *out, size_t outlen)
```
hanno-becker, line 16: This declaration is not needed, is it? We don't use `MLK_CUSTOM_RANDOMBYTES` with CBMC, and it's the user's responsibility to provide a declaration.

hanno-becker (changes requested):
I'd like this to work without redundant declarations. Also, can we have a custom config test, similar to `custom_zeroize_config.h`? It may just use `randombytes` again, but we should exercise the new option.

.github/workflows/ci.yml
mkannwischer: removed the declaration. now it's not necessary anymore.

mlkem/randombytes.h
mkannwischer: removed.

hanno-becker (approved):
LGTM, thanks @mkannwischer !

PR #878, opened 2025-03-14
Remove outdated CC0 notices

hanno-becker:
mlkem-native is Apache-2.0 licensed as indicated in the toplevel
LICENSE file. This commit removes multiple outdated CC0 license
notes from the initial import of the reference implementation.

The note on the origin of the FIPS-202 C implementation is still
present in mlkem/fips202/keccakf1600.c.


mkannwischer (approved):
Thanks @hanno-becker  - these are indeed outdated.

PR #883, opened 2025-03-18
HOL-Light: Add low-level spec and proof for AArch64 basemul

hanno-becker:
This commit adds low-level specifications and functional correctness proofs for the AArch64 polynomial base multiplication routines

- poly_basemul_acc_montgomery_cached_k2
- poly_basemul_acc_montgomery_cached_k3
- poly_basemul_acc_montgomery_cached_k4

The specs are 'low-level' in the sense that they do only provide a description in HOL of what is being computed on raw words; they do, however, not yet give a high-level description of what the computation does mod-3329: A scalar product in `Prod_i F_q[X]/(X^2-zeta_i)`.

The addition of a high-level spec and the connection of low-level and high level spec will be in a future commit.

s2n-bignum is updated to incorporate the recent addition of various long multiplication instructions.

The other ML-KEM arithmetic proofs needed updating since a core tactic SIMD_SIMPLIFY_TAC for the simplification of SIMD expressions was generalized to allow for a parametrized list of definitions to fold. Previously, this was hardcoded to the definitions used by the NTT and invNTT.

### DIFF
--- a/proofs/hol_light/arm/proofs/mlkem_poly_basemul_acc_montgomery_cached_k2.ml
+++ b/proofs/hol_light/arm/proofs/mlkem_poly_basemul_acc_montgomery_cached_k2.ml
@@ -1,0 +1,369 @@
+(*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT-0
+ *)
+
+needs "arm/proofs/base.ml";;
+
+needs "proofs/mlkem_specs.ml";;
+needs "proofs/mlkem_utils.ml";;
+
+(**** print_literal_from_elf "mlkem/mlkem_poly_basemul_acc_montgomery_cached_k2.o";;
+ ****)
+
+
+let poly_basemul_acc_montgomery_cached_k2_mc = define_assert_from_elf
+    "poly_basemul_acc_montgomery_cached_k2_mc" "mlkem/mlkem_poly_basemul_acc_montgomery_cached_k2.o"
+    [
+      0xd10103ff;       (* arm_SUB SP SP (rvalue (word 64)) *)
+      0x6d0027e8;       (* arm_STP D8 D9 SP (Immediate_Offset (iword (&0))) *)
+      0x6d012fea;       (* arm_STP D10 D11 SP (Immediate_Offset (iword (&16))) *)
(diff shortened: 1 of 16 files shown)


proofs/hol_light/arm/proofs/mlkem_poly_basemul_acc_montgomery_cached_k2.ml
```diff
@@ -0,0 +1,369 @@
...
+             )
+           )
+           (word 3329)
+         ) x
+      ) (16, 16)`;;
+
+let pmul_acc2 = define
+    `pmul_acc2 (x00: 16 word) (x01 : 16 word) (y00 : 16 word) (y01 : 16 word)
+          (x10: 16 word) (x11 : 16 word) (y10 : 16 word) (y11 : 16 word) =
+      montred (pmull_acc2 x00 x01 y00 y01 x10 x11 y10 y11)`;;
```
mkannwischer, line 214: It is somewhat confusing that here y00 and y01 are actually b0, and b1*twiddle in the even case, and b1 and b0 in the even case. 
In case you have to touch this again, maybe it makes sense to rename and reorder the arguments so that two coeffients next to each other are multiplied?

mkannwischer (approved):
Thanks @hanno-becker - this looks good to me.
I reviewed the specifications  and they make sense. I cannot review the proof itself. 

Some common code could be moved into the common files. Some of the inner functions may be more readable with some refactoring. 
But both these changes can be done in the next step.

k=2 and k=3 proofs look good in CI. k=4 is still running.

proofs/hol_light/arm/proofs/mlkem_poly_basemul_acc_montgomery_cached_k2.ml
hanno-becker: Ok!

proofs/hol_light/arm/proofs/mlkem_poly_basemul_acc_montgomery_cached_k2.ml
```diff
@@ -0,0 +1,369 @@
...
+           word_mul (
+             (word_sx : 16 word -> 32 word) (
+               word_mul (
+                 word_subword x (0,16)
+               ) (word 3327)
+             )
+           )
+           (word 3329)
+         ) x
+      ) (16, 16)`;;
```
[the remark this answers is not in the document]
hanno-becker: Yes, we can move this to a common file 👍

PR #894, opened 2025-03-19
Install slothy-cli in nix shell and run optimization of all code in CI

### LINKED ISSUE #536: Run SLOTHY in CI
We are currently not running `optimize.sh` in CI, which means we won't catch if changes are made to the code and/or SLOTHY that render both incompatible. Apparently, this has already happened: `optimize.sh` does not work on the current code.

- [ ] CI contains composable action to setup SLOTHY.
- [ ] `optimize.sh` has a 'dry run' option, similar to `example.py` in the SLOTHY repository.
- [ ] CI runs `optimize.sh` in dry-run, and builds + tests the mlkem-native afterwards.

mkannwischer:
See individual commit messages. 

Resolves #536

### DIFF
--- a/nix/slothy/default.nix
+++ b/nix/slothy/default.nix
@@ -1,0 +1,123 @@
+# SPDX-License-Identifier: Apache-2.0
+
+{ stdenvNoCC
+, fetchFromGitHub
+, python311
+, pkgs
+, callPackage
+, python311Packages
+, fetchPypi
+, stdenv
+, cmake
+, pkg-config
+, llvm
+, gcc
+}:
+
+
+let
+  ortools97 = python311Packages.buildPythonPackage rec {
+    pname = "ortools";
+    version = "9.7.2996";
+
+    format = "wheel";
+
+    src = fetchPypi {
+      inherit pname version;
+      format = "wheel";
+      dist = "cp311";
+      python = "cp311";
+      abi = "cp311";
+      platform =
+        if stdenv.isDarwin then
+          if stdenv.isAarch64 then "macosx_11_0_arm64"
+          else "macosx_10_15_x86_64"
+        else if stdenv.isLinux then
+          if stdenv.isAarch64 then "manylinux_2_17_aarch64.manylinux2014_aarch64"
+          else "manylinux_2_17_x86_64.manylinux2014_x86_64"
+        else throw "Unsupported platform";
+
+      hash =
+        if stdenv.isDarwin then
+          if stdenv.isAarch64 then "sha256-Oc8xz9iRuOiDyZBc85qlS57u0efP/f4cDpi2k9ZlCQI="
+          else "sha256-0ax/I7604BumV+VRV7Y5fNDs0XrxkK+ocr9yU1oFMDQ="
+        else if stdenv.isLinux then
+          if stdenv.isAarch64 then "sha256-B8QfDoYT2OxrpoUzSQTNduRbxG2j1S8XPpMBZo/yI90="
+          else "sha256-pUerlb2cykE9cYraz68MnuBKDM4V1Du0Ta3yve16K/o="
+        else throw "Unsupported platform";
+    };
+
+    propagatedBuildInputs = with python311Packages; [
+      numpy
+      pandas
+      protobuf
+    ];
+
+  };
+
+  unicorn_2_1_3 = python311Packages.buildPythonPackage rec {
+    pname = "unicorn";
+    version = "2.1.3";
+
+    propagatedBuildInputs = [
+      python311Packages.setuptools
+    ];
+
+    build-system = with python311Packages; [
+      setuptools
+    ];
+
+    dontConfigure = true;
+    nativeBuildInputs =
+      [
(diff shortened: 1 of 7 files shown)


nix/slothy/default.nix
```diff
@@ -0,0 +1,50 @@
...
+    rev = version;
+    sha256 = "sha256-bD+ryt9JO3OwApTIz5I+8eI/q6Bp/rqfaWS4ZUAZRTc=";
+  };
+
+  nativeBuildInputs = [ pkgs.makeWrapper ];
+  dontConfigure = true;
+
+  installPhase = ''
+    mkdir -p $out/bin
+    cp slothy-cli $out/bin/
```
hanno-becker, line 34: Is that necessary? Can we not just amend the `PATH`?

mkannwischer: Not sure what you mean. `$out` is the only thing that lives past the build, so we need to copy it at some point.

mkannwischer: See for example, our hol_light or the litani build - we always copy to some path with `$out`.

nix/slothy/default.nix
```diff
@@ -0,0 +1,50 @@
...
+
+{ stdenvNoCC
+, fetchFromGitHub
+, python311
+, pkgs
+}:
+
+let
+  pythonEnv = python311.withPackages (ps: with ps; [
+    ortools
```
hanno-becker, line 11: (How) are the versions fixed here?

mkannwischer: currently it just takes the default version in the current version. For ortools that happens to be 9.7. 
I don't quite know how to change that - we will likely need to take it from a different channel, but that can quickly get impossible due to Python versions. In that case we would have to build it ourselves by overwriting the package definitions just like we do for cbmc.

mkannwischer: it's not pip, so you cannot expect to just be able to throw in a any version that is cached in pypi.

mkannwischer: Actually, there seems to be a `fetchPypi`, so we can probably make it work for any version in pypi somehow.

hanno-becker (changes requested):
Thank you @mkannwischer, this is a great start!

I noticed that CTRL-C during optimization does not abort the optimization, which is inconvenient for interactive use; it should not matter for CI, though.

I'd like to have (more) confidence that this will actually work in CI. So, could you add a CI job which a) regenerates aarch64_opt, b) runs autogen, c) runs the tests?

For the CI, it is not important that we get good results -- we just need to make sure the overall flow works. So I would suggest to allow an `SLOTHY_EXTRA_FLAGS` or so in the Makefile, and use that to set a small timeout when SLOTHY is run for CI. Alternatively, I think there's an option in SLOTHY which stops search after the first solution is found -- that may work as well.

mkannwischer:
> Thank you @mkannwischer, this is a great start!
> 
> I noticed that CTRL-C during optimization does not abort the optimization, which is inconvenient for interactive use; it should not matter for CI, though.

This is fixed now. The cause was the `makeWrapper` which creates a wrapper that swallowed the SIGINT.

> I'd like to have (more) confidence that this will actually work in CI. So, could you add a CI job which a) regenerates aarch64_opt, b) runs autogen, c) runs the tests?
> 
> For the CI, it is not important that we get good results -- we just need to make sure the overall flow works. So I would suggest to allow an `SLOTHY_EXTRA_FLAGS` or so in the Makefile, and use that to set a small timeout when SLOTHY is run for CI. Alternatively, I think there's an option in SLOTHY which stops search after the first solution is found -- that may work as well.

Done. I had a CI pass that optimizes and tests everything on a Github MacOS runner in 53 minutes with a timeout set to 300 seconds. The same take around 5 minutes on my Apple M4 (without any timeouts, obviously). 
I'm trying it again with a 180 second timeout. Let's see if that still manages to find solutions and terminates faster. 

> 
> It would also be good to confirm the exact version of OR-Tools that's being run here -- the logs say v9.7.9999. Depending on how close it is to v9.8, it may or may not already have whatever change causes the massive performance drop since v9.8.

The version was exactly v9.7, but it was a custom build in the nixpkgs (which explains the 9999). I am not quite sure how this differs from the pypi packages. I switches to installing ortools from pypi directly (using the same version we have pinned in SLOTHY - v9.7.2996) and it is indeed much faster. 
I also briefly tried v9.11 but it's vastly shower even on my Apple M4. 

cc @dop-amin

hanno-becker:
@mkannwischer Instead of a more aggressive timeout, you could also abort optimization with the first solution found?

hanno-becker (approved):
Thank you _very_ much @mkannwischer, this is a great improvement, and I appreciate all the time you spent fiddling with nix!

nebeid:
I think the readme needs to be updated https://github.com/pq-code-package/mlkem-native/blob/83d85fe224bd6cf1b75f096a2b2fa01033b3dfda/dev/aarch64_opt/src/README.md?plain=1#L14

PR #895, opened 2025-03-20
Define `mlk_polyvec[_mulcache]` as typedef for array

### LINKED ISSUE #893: Define `mlkem_polyvec` as array rather than struct wrapper
`mlkem_polyvec` is currently defined as a struct wrapper around `mlkem_polyvec[MLKEM_K]`. This renders a [cast](https://github.com/pq-code-package/mlkem-native/blob/cc6c39877a0a73532018456c0b27f1fa1ae23754/mlkem/indcpa.c#L252) problematic, since _in theory_, a compiler could introduce padding `mlkem_polyvec` -- even though that's very unlikely given that `mlkem_poly` has size 512 bytes.

Consider making `mlkem_polyvec` a `typedef` for `mlkem_poly[MLKEM_K]`, rather than a struct wrapper.

hanno-becker:
* Resolves #893

This commit changes the definition of `mlk_polyvec` from a struct
wrapper `struct { mlk_poly vec[MLKEM_K] }` to a typedef for
`mlk_poly[MLKEM_K]`. Similarly, it introduces `mlk_polymat`
as a typedef for `mlk_poly[MLKEM_K * MLKEM_K]`.

This ensures that when working with polynomial matrices, 
the polynomial entries are consecutive in memory. 
This needs to be assumed for the correctness of `gen_matrix()`.

From all we know, is a theoretical concern that without this
change, compilers would be free to introduce padding into
`mlk_polyvec`, breaking `gen_matrix()`. Yet, the behaviour of
`gen_matrix` is strictly speaking UB, so even if a compiler
does not use padding in `mlk_polyvec`, it is better to change it.

Defining `mlk_polymat` as a typedef for `mlk_poly[MLKEM_K * MLKEM_K]`
and not `mlk_polyvec[MLKEM_K]` is deliberate: In the latter case, one
would not be able to pass a `mlk_polymat` to a function expecting
`const polymat` because of how qualifiers are (not) propagated, 
forcing cumbersome manual casts.

### DIFF
--- a/examples/monolithic_build/mlkem_native_monobuild.c
+++ b/examples/monolithic_build/mlkem_native_monobuild.c
@@ -181,6 +181,7 @@
 #undef mlk_poly_getnoise_eta1_4x
 #undef mlk_poly_getnoise_eta2
 #undef mlk_poly_getnoise_eta2_4x
+#undef mlk_polymat
 #undef mlk_polyvec
 #undef mlk_polyvec_add
 #undef mlk_polyvec_basemul_acc_montgomery_cached
--- a/mlkem/indcpa.c
+++ b/mlkem/indcpa.c
@@ -36,7 +36,7 @@
  *              and the public seed used to generate the matrix A.
  *
  * Arguments:   uint8_t *r: pointer to the output serialized public key
- *              mlk_polyvec *pk: pointer to the input public-key mlk_polyvec.
+ *              mlk_polyvec pk: pointer to the input public-key mlk_polyvec.
  *                Must have coefficients within [0,..,q-1].
  *              const uint8_t *seed: pointer to the input public seed
  *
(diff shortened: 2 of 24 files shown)


hanno-becker:
I suggest _not_ to do this. There is a major pain point here in that `T (*)[N]` does not cast to `const T (*)[N]`. Concretely, if you have a function `void foo(const T[N][M])`, and a local `T x[N][M]`, then calling `foo(x)` is not valid C. You would need to explicitly cast `foo((const T(*)[M]) x)` (@tautschnig @remi-delmas-3000 Would such cast actually be UB?) For `mlk_polyvec` alone, this only surfaces twice -- when calling `matvec_mul`. But it's a fair question why one would not also make `mlk_poly` a typedef, and then the above issue would surface countless times and make the entirely unreadable and also less safe.

mkannwischer (approved):
LGTM. Thanks @hanno-becker for resolving this!
I have no concerns about this change.

PR #896, opened 2025-03-21
Switch mlkem_poly_basemul_acc_montgomery_cached_k2 proof to integer spec

jargh:
This changes the specification style used in this proof from raw word operations to integer congruences, under some additional input bound assumptions |x0_i|, |x1_i| <= 2^12.

This relies on some improvements to the underlying automation for congruences and bounds, which are updated in line with some similar changes in s2n-bignum.

  https://github.com/jargh/s2n-bignum-dev/commit/50e617bce53a737ef11a605b7660dfe2bb97f01d

So far mlkem_poly_basemul_acc_montgomery_cached_k[34] are left as they are, except that the definition of "montred" is further type-constrained to match the definition used in the updated automation, which is in any case the type instance used in the specifications.

**Summary**:

**Steps**:

**Performed local tests**:
 - [ ] `lint` passing
 - [ ] `tests all` passing
 - [ ] `tests bench` passing
 - [ ] `tests cbmc` passing

### DIFF
--- a/proofs/hol_light/arm/proofs/mlkem_poly_basemul_acc_montgomery_cached_k2.ml
+++ b/proofs/hol_light/arm/proofs/mlkem_poly_basemul_acc_montgomery_cached_k2.ml
@@ -184,45 +184,29 @@ let poly_basemul_acc_montgomery_cached_k2_mc = define_assert_from_elf
     ];;
 
 let pmull = define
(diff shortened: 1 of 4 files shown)


hanno-becker:
I reopened this as #897 so we can run the full CI.

PR #902, opened 2025-03-22
SLOTHY: Update ortools to 9.12 + LLVM nix clean-up

mkannwischer:
See individual commits


hanno-becker (approved):
Thanks @mkannwischer for the detailed commit messages. LGTM

PR #905, opened 2025-03-24
CBMC tests: Allow wildcard pattern in list of proofs to run

hanno-becker:
This commit adds support for the use of wildcard pattern `*` in invocations of `tests cbmc`.

Example:
```
tests cbmc -p "*poly*"
```

This will run all tests for functions containing "poly".

As before, multiple patterns can be provided, e.g.

```
tests cbmc -p "*poly*" "*native"
```

### DIFF
--- a/scripts/tests
+++ b/scripts/tests
@@ -10,6 +10,7 @@ See the command line interface for more information."""
 import platform
 import argparse
 import os
+import re
 import sys
 import time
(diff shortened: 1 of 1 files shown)


mkannwischer (approved):
Nice. Thanks @hanno-becker. This works like a charm.

PR #906, opened 2025-03-24
CBMC: Refine bounds for input and output of base multiplication

hanno-becker:
* Builds on: #1115 

Previously, the base multiplication would assume that one of its inputs is bound by 4096 in absolute value, but make no assumptions about the other input ("b-input" henceforth) and its mulcache.

This commit refines the bounds slightly, as follows:
- The b-input is assumed to be bound by MLK_NTT_BOUND in absolute value. This comes for free since all values for b _are_ results of the NTT.
- The b-cache-input is assumed to be bound by MLKEM_Q in absolute value.

With those additional bounds in place, it can be showed that the result of the base multiplication is below INT16_MAX/2 in absolute value. Accordingly, this can be added as a precondition for the inverse NTT.

For the native AVX2 backend, the new output bound for the mulcache forces an explicit zeroization of the mulcache. This is not ideal since the cache is in fact entirely unused, but the performance penalty should be marginal (if the compiler can't eliminate the zeroization in the first place).

### DIFF
--- a/dev/aarch64_clean/src/arith_native_aarch64.h
+++ b/dev/aarch64_clean/src/arith_native_aarch64.h
@@ -113,6 +113,19 @@ __contract__(
   assigns(memory_slice(r, MLKEM_POLYBYTES))
 );
 
+/* Output bounds for the basemul-accumulate routines.
+ *
+ * The AArch64 backend leaves the schoolbook accumulator unreduced, so the
+ * output bound grows with k.  Each value is an *exclusive* bound matching
+ * the array_abs_bound() convention (abs(r) < BOUND) and the same-named
+ * constant in the HOL Light specification, where the proof establishes
+ * abs(r) <= BOUND-1. */
+/* check-magic: off */
+#define MLK_BASEMUL_ACC_K2_BOUND 8324
+#define MLK_BASEMUL_ACC_K3_BOUND 11653
+#define MLK_BASEMUL_ACC_K4_BOUND 14982
+/* check-magic: on */
+
 #define mlk_polyvec_basemul_acc_montgomery_cached_k2_aarch64_asm \
   MLK_NAMESPACE(polyvec_basemul_acc_montgomery_cached_k2_aarch64_asm)
 void mlk_polyvec_basemul_acc_montgomery_cached_k2_aarch64_asm(
@@ -127,7 +140,10 @@ __contract__(
     requires(memory_no_alias(b, sizeof(int16_t) * 2 * MLKEM_N))
     requires(memory_no_alias(b_cache, sizeof(int16_t) * 2 * (MLKEM_N / 2)))
     requires(array_abs_bound(a, 0, 2 * MLKEM_N, MLKEM_UINT12_LIMIT + 1))
+    requires(array_abs_bound(b, 0, 2 * MLKEM_N, 8 * MLKEM_Q))
+    requires(array_abs_bound(b_cache, 0, 2 * (MLKEM_N / 2), MLKEM_Q))
     assigns(memory_slice(r, sizeof(int16_t) * MLKEM_N))
+    ensures(array_abs_bound(r, 0, MLKEM_N, MLK_BASEMUL_ACC_K2_BOUND))
 );
 
 #define mlk_polyvec_basemul_acc_montgomery_cached_k3_aarch64_asm \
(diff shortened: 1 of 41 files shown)


hanno-becker:
CI is quite a bit slower. ~30min for MLKEM-1024.

mkannwischer (approved):
LGTM.

mkannwischer (dismissed):
Something must have broke when rebasing - CBMC is not happy and it looks to me like it's correct.

rod-chapman:
Looks good to me.

rod-chapman (dismissed):
Looks good to me. All proofs OK.

hanno-becker:
The combination of this and #895 seems to have a bad effect on CBMC performance. It's too slow to be merged at present.

hanno-becker:
CBMC for ML-KEM-512,768 only fails because the native contracts need adjusting and the HOL-Light proofs reworking accordingly. Checking what goes wrong with ML-KEM-1024.

hanno-becker:
This still seems to be hanging on `polyvec_basemul_acc_montgomery_cached_native` for ML-KEM-1024.

hanno-becker:
BLOCKER: The AVX2 basemul does **not** obey the new basemul bounds because it does not use lazy reduction. Unless/Until this is reworked -- which I'm not sure we should do as it was barely finished (#1097) -- we cannot proceed with this PR.

rod-chapman:
Still to do on this PR:
1. Resolve proof scaling issues with mlk_matvec_mul() when MLKEM_K=3 or 4. May be improved by CBMC 6.8.0 and/or Z3 4.15.3 releases to come.
2. Re-check post-conditions of AArch64 HOL-Light specification of polyvec_basemul_acc_cached_native_asm_k[234] and re-prove against C contracts to prove that AArch64 asm code yields coefficients bounded by INT16_MAX/2
3. Re-implement and re-prove x86_64 native implementations of the same.

hanno-becker:
@dkostic Now that you have done all the proofs including the basemul. How much work would you expect it to be to adjust the basemul to use a (slightly) lazy reduction strategy? That is, only reduce after computing each polynomial product, but before accumulation.

hanno-becker:
@dkostic Would you (and Claude) have time to do this, and at the same time resolve https://github.com/pq-code-package/mlkem-native/issues/1421?

rod-chapman:
I hope that CBMC 6.9.0 will unblock progress on this one. I will rebase the branch first and see if the proofs complete OK.

hanno-becker:
Thanks @rod-chapman! But this also needs significant work on the x86_64 backend.

rod-chapman:
Oh yes... the x86 back-end has to brought into line with the AArch64 and C back-ends to produce tighter bounds on the output of basemul, right?

I have re-based and re-running proofs now on the C stuff.

rod-chapman:
Update. I am having trouble getting one proof to work - it's polvec_basemul_acc_montgomery_cached_native().
The problem is the need to carry the stronger pre- and post-conditions across the call to the native back-end in a way that they "survive" so that, if the first call to the native backend responds with "FALLBACK", they are still true for the second call to the C backend.  The problem is that the native backend changes representation of the types (e.g. from `polyvec *` to `int16_t b[MLKEM_K * MLKEM_N` which makes the reasoning sufficiently complex to defeat Z3.

rod-chapman:
C Proofs are OK now. I found a Z3 tactic that was effective. @dkostic The next move is to prove that the bounds on the outputs of the AArch64 native implementation of `basemul` are as expected, and as given by the specification of those 3 functions in native/aarch64/src/arith_native_aarch64.h on the `refine_bounds` branch. The CBMC spec of that for K=2 is here:

  https://github.com/pq-code-package/mlkem-native/blob/8d3997766159c9c8c406cde2c406c0c480ed8253/mlkem/src/native/aarch64/src/arith_native_aarch64.h#L134

The critical contract is `ensures(array_abs_bound(r, 0, MLKEM_N, INT16_MAX/2))`

We need to know that's true for the assembly language versions (3 times for K=2,3,4), so that means amending the HOL-Light proof to explicitly state that post-condition. For K=2, that's here or thereabouts:

https://github.com/pq-code-package/mlkem-native/blob/8d3997766159c9c8c406cde2c406c0c480ed8253/proofs/hol_light/aarch64/proofs/polyvec_basemul_acc_montgomery_cached_k2_aarch64_asm.ml#L328

Does that make sense?

dkostic:
Before and after performance measured on m8i EC2 instance with Intel(R) Xeon(R) 6975P-C:
```
  ┌──────────────┬──────────────────┬────────┬────────┬──────────────────┬────────┬────────┬───────────────────┬─────────┬─────────┐
  │              │ mlkem512 keypair │ encaps │ decaps │ mlkem768 keypair │ encaps │ decaps │ mlkem1024 keypair │ encaps  │ decaps  │
  ├──────────────┼──────────────────┼────────┼────────┼──────────────────┼────────┼────────┼───────────────────┼─────────┼─────────┤
  │ no Barrett   │ 37 359           │ 43 892 │ 55 961 │ 60 371           │ 68 916 │ 84 209 │ 90 615            │ 104 190 │ 122 639 │
  ├──────────────┼──────────────────┼────────┼────────┼──────────────────┼────────┼────────┼───────────────────┼─────────┼─────────┤
  │ with Barrett │ 37 345           │ 43 784 │ 55 847 │ 60 867           │ 69 213 │ 84 345 │ 90 648            │ 104 279 │ 122 820 │
  ├──────────────┼──────────────────┼────────┼────────┼──────────────────┼────────┼────────┼───────────────────┼─────────┼─────────┤
  │ Δ cycles     │ −14              │ −108   │ −114   │ +496             │ +297   │ +136   │ +33               │ +89     │ +181    │
  ├──────────────┼──────────────────┼────────┼────────┼──────────────────┼────────┼────────┼───────────────────┼─────────┼─────────┤
  │ Δ %          │ −0.04%           │ −0.25% │ −0.20% │ +0.82%           │ +0.43% │ +0.16% │ +0.04%            │ +0.09%  │ +0.15%  │
  └──────────────┴──────────────────┴────────┴────────┴──────────────────┴────────┴────────┴───────────────────┴─────────┴─────────┘
```

PR #907, opened 2025-03-24
CBMC: Remove convoluted implementation of 2D bounds check

hanno-becker:
A previous issue in CBMC

https://github.com/diffblue/cbmc/issues/8570

had forced to break bounds assertions over a 2D array into nested bounds assertions.

This issue has since been fixed, allowing for the simplification of the respective assertions, which is what this commit does.


hanno-becker:
Causing problems with CBMC, and not a priority. Closing for now.

Issue #604, opened 2025-01-01
Wait + Retry EC2 runner spinup when hitting request limit


hanno-becker:
Currently, CI fails when too many EC2 instances are attempted to be created simultaneously. Instead, our fork of `ec2-github-runner` should just wait and retry a few times.

hanno-becker:
@mkannwischer Can you do this?

hanno-becker:
@mkannwischer There is no urgency, but I'd like to fix this at some point. Can you mark yourself down for this?

PR #909, opened 2025-03-24
CI: Implement simple retry-once logic for starting EC2 runners

hanno-becker:
* Resolves https://github.com/pq-code-package/mlkem-native/issues/604

We often need to manually restart the CI because too many EC2 instances are attempted to be started at the same time.

This commit implements a basic retry-once logic in ci_ec2_reusable.yml: Upon failure to start an EC2 instance, it waits ~30s and retries once before failing. A slight random delay is inserted to reduce the likelihood of clashes if multiple jobs fail at the same time.

### DIFF
--- a/.github/workflows/ci_ec2_container.yml
+++ b/.github/workflows/ci_ec2_container.yml
@@ -78,8 +78,8 @@ jobs:
                         # avoiding race conditions where an instance is started,
                         # but isn't yet done registering as a runner and reporting back.
     outputs:
-      label: ${{ steps.start-ec2-runner.outputs.label }}
-      ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
+      label: ${{ steps.remember-runner.outputs.label }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_ec2_container.yml
```diff
@@ -109,6 +110,35 @@ jobs:
           ec2-instance-type: ${{ inputs.ec2_instance_type }}
           subnet-id: subnet-07b2729e5e065962f
           security-group-id: sg-0ab2e297196c8c381
+      - name: Start EC2c runner (wait before retry)
```
mkannwischer, line 113:
```suggestion
      - name: Start EC2 runner (wait before retry)
```

mkannwischer (changes requested):
Small nit - rest looks good. I'll test a bit more.

.github/workflows/ci_ec2_container.yml
hanno-becker: Fixed

mkannwischer (approved):
Thank you @hanno-becker. This seems to work without problems and will save us a lot of time and pain.

PR #921, opened 2025-03-31
Autogen: Autogenerate comments for #else and #endif 

### LINKED ISSUE #917: Consider autogeneration of #else ... #endif comments
We often annotate preprocessor directives with the relevant macro being tested, but don't distinguish positive/negative clauses. For example, the following is common:

```c
#if defined(foo)
...
#else /* foo */
...
#endif /* foo */
```

It would be nicer if the code would also correctly indicate whether the flag was positively/negatively guarded, e.g.

```c
#if defined(foo)
...
#else /* foo */
...
#endif /* !foo */
```

Doing this manually is rather tedious and I'm not sure it's worth it.

However, we could consider extending `autogen` to autogenerate those comments.

hanno-becker:
* Resolves #917 

In parts of the mlkem-native source, we annotate `#else`
and `#endif` directives with the condition to which they apply.
This can be useful, but is also tedious to do by hand, and difficult
to do consistently.

This commit extends `autogen` with the ability to autogenerate
the `#else` and `#endif` annotations.

It was found that appyling comments everywhere impedes readability
when `#if`, `#else` and `#endif` are very close to each other and
there is no need to remind the reader of what condition is being
closed. We therefore only apply comments if the distance between
`#if` and `#else` and `#endif` is larger than a certain threshold
(for now, it's 5), or if there are nested `#if`s inside.

A very basic parser and simplifier for conditionals is implemented
based on `pyparsing`, which takes care of removing unnecessary brackets,
normalizing spacing, and removing double-negations.


dev/x86_64/meta.h
```diff
@@ -143,8 +143,7 @@ static MLK_INLINE void mlk_poly_decompress_d10_native(
 {
   mlk_poly_decompress_d10_avx2((__m256i *)r, a);
 }
-#endif /* defined(MLK_MULTILEVEL_BUILD_WITH_SHARED) || (MLKEM_K == 2 \
-          || MLKEM_K == 3) */
+#endif /* MLK_MULTILEVEL_BUILD_WITH_SHARED || MLKEM_K == 2 || MLKEM_K == 3 */
```
mkannwischer, line 146: what happened to the defined()?

dev/x86_64/src/basemul.c
```diff
@@ -66,10 +66,10 @@ void mlk_polyvec_basemul_acc_montgomery_cached_avx2(unsigned k,
...
-          !defined(MLK_MULTILEVEL_BUILD_NO_SHARED) */
+#else /* MLK_ARITH_BACKEND_X86_64_DEFAULT && !MLK_MULTILEVEL_BUILD_NO_SHARED \
+       */
 
 MLK_EMPTY_CU(avx2_basemul)
 
-#endif /* defined(MLK_ARITH_BACKEND_X86_64_DEFAULT) && \
-          !defined(MLK_MULTILEVEL_BUILD_NO_SHARED) */
+#endif /* !(MLK_ARITH_BACKEND_X86_64_DEFAULT && \
+          !MLK_MULTILEVEL_BUILD_NO_SHARED) */
```
mkannwischer, line 75: Same here.

mkannwischer (changes requested):

dev/x86_64/meta.h
hanno-becker: It's removed for readability.

mkannwischer: Okay makes sense.

mkannwischer (approved):
Thanks @hanno-becker. This is great! You cleaned up a lof of very inconsistent comments. 
A potential follow-up would be to also auto-generate the comments in the asm files.

PR #922, opened 2025-03-31
autogen: Autogenerate #else and #endif comments for .S files

hanno-becker:
A previous commit extended `autogen` to autogenerate comments for `#else` and `#endif` directives, indicating the conditions they guard.

This commit extends this feature to .S files which were previously excluded because we cannot run `clang-format` on them. This is circumvented here by running `clang-format` on individual lines with preprocessor directives.


mkannwischer (approved):
Awesome - thanks! Amazing how simple this turned out.

PR #923, opened 2025-03-31
[TEST]: Add CT-LLVM

### LINKED ISSUE #915: Use CT-LLVM in CI
CT-LLVM is a compiler plugin that can prove certain functions constant-time on an LLVM-IR level. 
See https://eprint.iacr.org/2025/338 https://github.com/Neo-Outis/CT-LLVM-Artifact

I have just tried it out locally on the monobuild and it is looking promising. It can prove a good number of functions with the right configuration options. 
After discussing with Zhiyuan, it is looking like some more functions can be proved by patching the tool to work with our empty inline-assembly. Another problem we see is that some functions perform branches on public inputs (mostly length arguments that are fixed) - those can be proved by marking those inputs as public.
We should add the tool to our CI, so we can try it out on a few more versions of clang.
It should work from clang 14 as that added `-fpass-plugin` API that is used by CT-LLVM.

mkannwischer:
Work in progress

Eventually resolves #915.

### DIFF
--- a/.github/actions/ct-test/action.yml
+++ b/.github/actions/ct-test/action.yml
@@ -10,6 +10,9 @@ inputs:
   valgrind_flags:
     description: Extra flags to pass to valgrind
     default: ""
+  ctllvm:
+    description: whether to run CT-LLVM or not (requires LLVM)
+    default: "false"
 
 runs:
   using: composite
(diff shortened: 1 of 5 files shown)


hanno-becker:
@mkannwischer Can you summarize the status of this PR please?

mkannwischer:
> @mkannwischer Ping

I don't have bandwidth right now to continue this. Status is that I managed to build and use CT-LLVM locally, but I was thus far not able to build it within nix.
We can close this PR for now if you want.

PR #930, opened 2025-04-01
Use MLK_CONFIG_XXX pattern for all configuration options

hanno-becker:
* Based on #933 

This commit renames all configuration options to follow the pattern
`MLK_CONFIG_XXX`.

One exception is MLKEM_K, which stays as-is in the bulk of the code.
However, instead of exposing MLKEM_K directly in the config, it is
now determined by the value of a new configuration option
MLK_CONFIG_SECURITY_LEVEL, which takes the value 512/768/1024.


mkannwischer (changes requested):
Thanks @hanno-becker. Overall looks good - one small nit.

mkannwischer:
Let's merge this in one go with #934.

PR #931, opened 2025-04-01
tests: Fix naming of `mlkem_native_as_code_package` test

hanno-becker:
Some previous systematic renaming of mlkem_native_ -> mlk_ must have erroneously renamed the test name in the `tests` script, rendering it impossible to run the respective example via `tests examples -l ...`. It would only be run as part of `tests all`.


mkannwischer (approved):
Good catch! I confirmed that it breaks prior to this PR and is fixed with the PR. 
Thanks!

PR #934, opened 2025-04-01
Monobuild: Don't `#undef` configuration options

### LINKED ISSUE #928: Don't `#undef` configuration options in monolithic compilation unit
Currently, the monolithic compilation unit for mlkem-native `#undef`s everything that is `#define`'ed somewhere within `mlkem/*`.

For configuration options, whether an option is `#undef`ed or not thus depends on whether it has a default in `mlkem/config.h`. This is confusing.

Moreover, the `#undef`s include the `MLK_CONFIG_H` guard for `mlkem/config.h` itself, which means that the config file will be re-imported in a multilevel build. This was intentional so far, but it does not work when using a custom configuration file which does not use the same `MLK_CONFIG_H` guard. Concretely, if a custom config file did not use the same header guard, it would not be re-imported in a multi-level build, yet parts of the default config would have been `#undef`'ed. Clearly, this is messy and error-prone.

The situation will be much simplified if the monolithic compilation unit never touches any configuration options, nor the header guard for the configuration file. This means that in a monolithic multi-level build, the configuration file will not be imported multiple times, and the user will need to tweak it in between inclusions of the monofile. Realistically, however, this will only mean `#undef MLKEM_K` followed by `#define MLKEM_K XXX`.

hanno-becker:
* Resolves #928 

-------------

This commit renames all configuration options to follow the pattern
MLK_CONFIG_XXX.

One exception is MLKEM_K, which stays as-is in the bulk of the code.
However, instead of exposing MLKEM_K directly in the config, it is
now determined by the value of a new configuration option
MLK_CONFIG_SECURITY_LEVEL, which takes the value 512/768/1024.

-------------

Previously, the `autogen`erated monobuild file `mlkem_native_monobuild.c`
would `#undef` all directives set in `config.h`.

This approach was error prone for at least two reasons:
- Only options with defaults in `config.h` would be `#undef`ed.
- A consuming application using a custom config does not use `config.h`.
  In particular, that custom config could use a different header guard
  than config.h, in which case it would inadvertently be included multiple
  times.

This commit changes `autogen` to not `#undef` configuration options
at the end of `mlkem_native_monobuild.c`. In a multilevel build, it
is the user's responsibility to use manual undefine+define to make
changes to the configuration.

The affected examples

- monolithic_build_multilevel
- monolithic_build_multilevel_native

are adjusted accordingly.


mkannwischer (approved):
Thanks @hanno-becker - this is looking good to me. 
Benchmarks in #930 suggest everything is in order.
Please update the PR title to reflect that #930 got included here before merging.

PR #939, opened 2025-04-02
CI: Update dependencies 

mkannwischer:
Combination of #935 #936 #937 + ec2-github-runner update.


mkannwischer:
I removed the liboqs update as that breaks our CI. We need to address that separately.

hanno-becker (approved):

PR #940, opened 2025-04-02
Parallelize scripts/autogen

### LINKED ISSUE #932: Parallelize `autogen`
Even on a powerful machine like an Apple M1, `autogen` takes ~20s, mainly because it's sequential. 

Parallelize the per-file tasks done by `autogen`, such as `simpasm` or header/comment autogeneration.

hanno-becker:
* Resolves #932 

The `autogen` script performs a few tasks which are parametrized over the source files in the mlkem-native repository.

This commit parallelizes those tasks for better performance.

On my Apple M1, time drops from ~44s to ~12s.

### DIFF
--- a/scripts/autogen
+++ b/scripts/autogen
@@ -13,6 +13,9 @@ import sys
 import pyparsing as pp
 import os
 
+from concurrent.futures import ThreadPoolExecutor
+from functools import partial
+
 modulus = 3329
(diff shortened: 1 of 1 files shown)


mkannwischer (approved):
I confirmed that this works. 
On my M4 excution time drops from 19 seconds to 5 seconds.

PR #942, opened 2025-04-02
Simplify mlkem_native.h

hanno-becker:
The goal of mlkem/mlkem_native.h is to capture the public API of mlkem-native in a way that is independent of the main source tree.

Previously, mlkem_native.h could be used in two ways:
- Standalone: Here, prior to inclusion of mlkem_native.h, the user would set various MLK_BUILD_INFO_XXX flags indicating parameter set and namespacing.
- Config-based: Here, the user would provide the configuration used to build mlkem-native, and have MLK_BUILD_INFO_XXX derived automatically.

The second case is mainly for convenience during development: When the mlkem-native source code is available, one may as well include kem.h and use it as the API. We include mlkem_native.h in this case primarily to test consistency of mlkem_native.h and kem.h.

This commit aims to simplify the use of mlkem_native.h and to align it with other configuration of mlkem-native, as follows:

First, MLK_BUILD_INFO_XXX are renamed to follow the naming pattern MLK_CONFIG_API_XXX. In particular, this is a sub-pattern of the pattern MLK_CONFIG_XXX used for options that the user may tweak -- which is the case here.

The options controlling the use of mlkem_native.h, are then:

- MLK_CONFIG_API_PARAMETER_SET [required]

  The parameter set: 512, 768, 1024

  Corresponds to MLK_CONFIG_PARAMETER_SET used in the config.h.

- MLK_CONFIG_API_NAMESPACE_PREFIX [required]

  The namespace prefix, including level.

  This almost corresponds to MLK_CONFIG_NAMESPACE_PREFIX used during the build, but includes the level.

- MLK_CONFIG_API_NO_SUPERCOP [optional]

  This used to be MLK_BUILD_INFO_NO_STANDARD_API, but the new name seems more apt.

- MLK_CONFIG_API_CONSTANTS_ONLY [optional]

  This can be set if the header is included only to be the sizes of key material. This is needed in monolithic multilevel builds.

Second, the config-based usage of mlkem_native.h is removed. The only way to use mlkem_native.h is to set a subset of the above MLK_CONFIG_API_XXX options and then include mlkem_native.h.

To not loose the convenience of importing mlkem_native.h during a) an mlkem-native build for consistency checking, or b) a test application, the standard header mlkem/common.h is adjusted to set MLK_CONFIG_API_XXX consistent with MLK_CONFIG_XXX, so that one can just import mlkem_native.h whenever needed in any context that has included mlkem/common.h.


mlkem/kem.h
```diff
@@ -13,7 +13,12 @@
 #if defined(MLK_CHECK_APIS)
 /* Include to ensure consistency between internal kem.h
  * and external mlkem_native.h. */
+#if defined(MLK_CONFIG_API_NO_SUPERCOP)
+#error Unexpected configuration of MLK_CONFIG_API_XXX during build
```
mkannwischer, line 17:
```suggestion
#error Unexpected configuration of MLK_CONFIG_API_NO_SUPERCOP during build
```

mlkem/mlkem_native.h
```diff
@@ -5,76 +5,50 @@
...
- */
-
-#include <stdint.h>
-
-/*************************** Build information ********************************/
-
-/*
- * Provide security level (MLK_BUILD_INFO_LVL) and namespacing
- * (MLK_BUILD_INFO_NAMESPACE)
+ * make sure to rename or #undefine the header guard.
```
mkannwischer, line 49:
```suggestion
 * make sure to rename or #undef the header guard.
```

mlkem/mlkem_native.h
```diff
@@ -5,76 +5,50 @@
...
+ *   SUPERCOP naming convention crypto_kem_xxx. If you don't want/need this,
+ *   set MLK_CONFIG_API_NO_SUPERCOP. You must set this for a multilevel build.
  *
- * - The config.h used for the build is available in the include paths.
- * - The values of MLK_BUILD_INFO_LVL and MLK_BUILD_INFO_NAMESPACE are set,
- * reflecting the security level (512/768/1024) and namespace of the build.
+ * - MLK_CONFIG_API_CONSTANTS_ONLY [optional]
+ *
+ *   If you don't want this header to expose any function declarations,
+ *   but only constants, set MLK_CONFIG_API_CONSTANTS_ONLY. In this case,
```
mkannwischer, line 40: should we elborate here that the constants are refering to public key/ciphertext/etc sizes? constants is a bit vague

mlkem/mlkem_native.h
```diff
@@ -140,10 +132,11 @@
...
  **************************************************/
-MLK_MUST_CHECK_RETURN_VALUE
-int MLK_BUILD_INFO_NAMESPACE(keypair_derand)(
-    uint8_t pk[MLKEM_PUBLICKEYBYTES(MLK_BUILD_INFO_LVL)],
-    uint8_t sk[MLKEM_SECRETKEYBYTES(MLK_BUILD_INFO_LVL)], const uint8_t *coins);
+MLK_API_MUST_CHECK_RETURN_VALUE
+int MLK_API_NAMESPACE(keypair_derand)(
+    uint8_t pk[MLKEM_PUBLICKEYBYTES(MLK_CONFIG_API_PARAMETER_SET)],
+    uint8_t sk[MLKEM_SECRETKEYBYTES(MLK_CONFIG_API_PARAMETER_SET)],
+    const uint8_t *coins);
```
mkannwischer, line 139:
```suggestion
    const uint8_t coins[2 * MLKEM_SYMBYTES]);
```

examples/monolithic_build/Makefile
```diff
@@ -104,17 +104,17 @@ $(LIB1024_FULL): $(MLK_SOURCE)
 $(BIN512_FULL): $(APP_SOURCE) $(LIB512_FULL)
 	$(Q)echo "$@"
 	$(Q)[ -d $(@) ] || mkdir -p $(@D)
-	$(Q)$(CC) $(CFLAGS) -DMLK_CONFIG_FILE="\"config_512.h\"" $(INC) $^ -o $@
+	$(Q)$(CC) $(CFLAGS) -DMLK_CONFIG_API_PARAMETER_SET=512 $(INC) $^ -o $@
```
mkannwischer, line 107: Does this mean the `config_{512,768,1024}.h` headers can be deleted?

mkannwischer (changes requested):
Thanks @hanno-becker for this simplification - looks mostly good to me.
Here are a few nits.

mlkem/kem.h
hanno-becker: Removing. This is checked in `common.h` already.

mlkem/mlkem_native.h
hanno-becker: Ok

mlkem/mlkem_native.h
hanno-becker: Ok

mlkem/mlkem_native.h
hanno-becker: While at it, ok..

examples/monolithic_build/Makefile
hanno-becker: No, they are needed for the library build.

mkannwischer: ahh right - makes sense.

mkannwischer (approved):
Thanks for the changes! LGTM
I'll do a follow-up PR changing the `coins` argument.

PR #947, opened 2025-04-03
Add and apply "InsertBraces" option for clang-format

### LINKED ISSUE #946: Consider "InsertBraces" option for clang-format
For all C code, we might consider adding "InsertBraces: true" to our .clang-format file.
This forces the use of braces to enclose a single statement nested inside an enclosing compound statement such as a for-loor or an if statement. We have been doing this, but it would be good auto-enforce.

So we not allow
```
if (E)
  s;
```
but insist on
```
if (E)
{
  s;
}
```
which is normally considered "safe" programming style in C, and avoids the potential for a future mistakes like Apple's notorious "goto fail" defect.

On current main branch, this change only affects mlkem/debug.c and a few files in the "examples" subdirectory.

rod-chapman:
Fixes #946 

1. Adds "InsertBraces: true" to .clang-format file
2. autogen
3. format
4. lint

All tests OK.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -16,6 +16,7 @@ IncludeBlocks: Preserve
 # Designate CBMC contracts/macros that appear in .h files
 # as "attributes" so they don't get increasingly indented line after line
 BreakBeforeBraces: Allman
+InsertBraces: true
 WhitespaceSensitiveMacros: ['__contract__', '__loop__' ]
 Macros:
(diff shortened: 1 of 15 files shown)


mkannwischer:
LGTM. Can you make the same change to mldsa-native, please?

rod-chapman:
Yes... next stop MLDSA...
Please merge.

hanno-becker (approved):

PR #949, opened 2025-04-05
tests: Integrate namespace check into `tests func`

### LINKED ISSUE #372: Namespacing should be checked for both opt and no-opt implementation
#234 refactored the build scripts and now we no longer check the namespacing for the no-opt implementations. 

The namespace check should just be integrated into the test scripts, so that it gets checked whenever a binary is built. Then we can remove the explicit namespacing steps in the CI.

hanno-becker:
* Resolves #372 

This commit adds the parameter --check-namespace to `tests func`. It is false by default. If true, `scripts/check-namespace` is called after both the C and the native build to check namespacing.

Previously, a call to `check-namespace` after `tests func` would only apply to the final binary produced by `tests func`, which was the native one.

### DIFF
--- a/scripts/tests
+++ b/scripts/tests
@@ -540,6 +540,14 @@ class Tests:
     def func(self):
         def _func(opt):
             self._compile_schemes(TEST_TYPES.FUNC, opt)
+            if self.args.check_namespace is True:
+                p = subprocess.run(
+                    ["python3", "check-namespace"],
+                    stdout=subprocess.DEVNULL if not self.args.verbose else None,
+                    cwd="scripts",
+                )
+                if p.returncode != 0:
+                    self.fail(f"Namespacing failed for opt={opt}")
             if self.args.run:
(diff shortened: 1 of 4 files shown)


scripts/tests
```diff
@@ -934,6 +942,13 @@ def cli():
         "all", help="Run all tests (except benchmark for now)", parents=[common_parser]
     )
 
+    all_parser.add_argument(
+        "--check-namespace",
+        help="Check namespacing of binaries",
+        action="store_true",
+        default=False,
+    )
+
```
mkannwischer, line 951: Why is this argument added to both the `all_parser` and the `func_parser`? 
The commit message suggests it should only be present for the `func` tests?

hanno-becker: Func can be run through `tests all`

mkannwischer (approved):
LGTM! Thanks!

PR #978, opened 2025-04-20
autogen: Include all source files in most operations

hanno-becker:
autogen conducts numerous changes, such as the automatic addition of header guards, #undefs at the bottom of a source file, or commenting of `#if/#else/#endif` blocks.

Previously, those operations would only be applied to 'core' source files in `mlkem/*` or `dev/*`. However, many of them make sense on a broader set of source files. For example, we were currently not applying the autogeneration of `#if/#else/#endif` comments on non-'core' source files, which led to wrong/inconsistent comments in those files.

This commit extends the scope of autogen by considering all source files by default, and restricting it explicitly to 'core' files in the case of 1. autogeneration of header guards, 2. autogeneration of #undefs: Enabling autogeneration of header guards everywhere leads to unscrutably long guard names for example files, and also breaks because of expectations on the copyright header. Autogeneration of `#undef`s is not useful outside of the core source files.


mkannwischer (approved):
LGTM. Thank you @hanno-becker!

PR #979, opened 2025-04-20
README: Simplify and shorten

hanno-becker:
This commit shortens the README, in particular removing the full list of AArch64 assembly routines which are covered by the HOL-Light proofs.


mkannwischer (approved):
LGTM

PR #986, opened 2025-04-29
Add statement of origin and terms of use for NIST KAT rng

hanno-becker:


mkannwischer (approved):
Thanks @hanno-becker!

PR #989, opened 2025-04-30
nix, ci: Add cross toolchain and tests for ppc64le

hanno-becker:
This commit adds a gcc ppc64le cross toolchain to the `ci-cross` nix shell. The existing `qemu` installation already provides `qemu-ppc64le`, so no changes are needed.

It also extends the CI to build and test mlkem-native for ppc64le.

Finally, `sys.h` is extended to detect RV64 and PPC64LE, and the CI modified to use `MLK_FORCE_{ARCH}` to double-check that system detection works as expected.


mkannwischer (approved):
Thanks @hanno-becker - these changes look good to me!

PR #990, opened 2025-04-30
CI: Run extended tests only after optional nix cache rebuild

hanno-becker:
This commit moves the nix workflow nix.yml into all.yml
and triggers the extended tests ci.yml only once the nix
workflow completes. This prevents concurrent rebuilding
of the expensive ci-cross shell. With the current patch,
the ci-cross shell is built and cached once in nix.yml,
and can then be reused by ci.yml.

Also, the dependency of the nix development tests on the
nix cache build for ci-cross is unnecessary and removed.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -4,20 +4,7 @@ name: Nix
 permissions:
   contents: read
 on:
-  push:
-    branches: ["main"]
-    # Only run upon changes to nix files
-    paths:
-      - 'flake.lock'
-      - 'flake.nix'
-      - 'nix/**'
-  pull_request:
-    branches: ["main"]
-    # Only run upon changes to nix files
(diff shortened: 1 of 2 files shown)


mkannwischer (approved):

.github/workflows/nix.yml
```diff
@@ -83,7 +106,6 @@ jobs:
             install: 'installer'
     name: nix setup test (${{ matrix.target.container != '' && matrix.target.container || matrix.target.runner }}, nix via ${{ matrix.target.install }})
     runs-on: ${{ matrix.target.runner }}
-    if: github.repository_owner == 'pq-code-package' && !github.event.pull_request.head.repo.fork
```
hanno-becker, line 86: (@mkannwischer This was never needed)

PR #996, opened 2025-05-02
AVX2 basemul with mulcache

dkostic:
**Summary**:
AVX2 implementation of mulcache_compute and basemul
with mulcache. The performance stays unchanged, but the
aarhc64 and x86_64 backends are now unified in the approach
to basemul.


hanno-becker (approved):
Thanks a lot @dkostic for your work!

PR #1000, opened 2025-05-10
AArch64: Zeroize rejection sampling stack buffer

hanno-becker:
The AArch64 rejection sampling routine samples data into a temporary stack buffer. This stack buffer is copied in full to the output buffer at the end of the routine, regardless of how many bytes were actually sampled. This leaks parts of the original stack values to the output buffer. While the call-site eventually fills the output buffer fully, thereby overwriting the temporary values copied from the stack, this temporary leakage is still undesirable and should be avoided.

This commit avoids unnecessarily leaking uninitialized stack values by explicitly zeroing the temporary stack buffer prior to starting the sampling.

### DIFF
--- a/dev/aarch64_clean/src/rej_uniform_asm.S
+++ b/dev/aarch64_clean/src/rej_uniform_asm.S
@@ -56,6 +56,7 @@
 
     /* Temporary registers */
     tmp                         .req w11
+    initial_zero_count          .req w11
     final_copy_count            .req w11
 
     rec_idx_0                   .req w12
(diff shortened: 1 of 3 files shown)


mkannwischer (approved):
Good catch @hanno-becker! Thank you!

A55 results: 
before:
```
mlk_poly_rej_uniform_x4 cycles=22768
```

after: 
```
mlk_poly_rej_uniform_x4 cycles=22918
``` 

Looks acceptable.

PR #1014, opened 2025-05-14
HOL-Light: Autogenerate rejection sampling constant table

### LINKED ISSUE #1013: HOL-Light: Autogenerate constant table for rejection sampling in proof script
The large constant table used for rejection sampling on AArch64 is embedded in the proof script. It should be autogenerated by the same core function that also autogenerates the same table in the source -- making it obvious that they are the same.

This is similar to #868

hanno-becker:
* Fixes #1013 

The HOL-Light proof script for AArch64 rejection sampling has embedded in it the large precomputed constant table used by the assembly. The same table is also embedded in the C source of mlkem-native, and autogenerated from `scripts/autogen`.

To make it as obvious as possible that the constant table assumed by the proof script matches the constant table used by the source code, this commit extends `scripts/autogen` to autogenerate the constant table embedded in the proof script from the same Python code that also generates the constant table in the C source.


mkannwischer (approved):
Thanks! Nice improvement @hanno-becker!

PR #1026, opened 2025-05-18
Fix API consistency check for backends

hanno-becker:
The backend API `native/api.h` only declares backend functions for which the relevant `MLK_USE_XXX` macro has been set. Those macros are set by the backend header, which hence must be included _before_ inclusion of `native/api.h`. The code in `common.h` which includes both `native/api.h` and the backend header says as much, but then plainly does the opposite: It includes `native/api.h` first, then the backend header. This renders the API consistency checks useless.

This commit fixes this by moving the inclusion of the backend headers ahead of the backend API headers.

### DIFF
--- a/mlkem/common.h
+++ b/mlkem/common.h
@@ -100,6 +100,7 @@
 #endif
 
 #if defined(MLK_CONFIG_USE_NATIVE_BACKEND_ARITH)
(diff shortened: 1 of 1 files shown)


mkannwischer (approved):
Uff - good catch, @hanno-becker! Thank you!

Issue #1025, opened 2025-05-18
Create CBMC specifications for AArch64 assembly and prove native backend spec against them


hanno-becker:
The frontend CBMC proofs work against specifications for the backend. In the case of the AArch64 backend, those are shim inline wrappers around the AArch64 backend assembly routines. Those routines are specified and proved correct in HOL-Light.

- [ ] Create CBMC specifications of the AArch64 assembly routines, mirroring the relevant aspects of the HOL-Light specifications. Since we don't express functional correctness in CBMC, those specifications will necessarily be weaker than the HOL-Light ones.
- [ ] Prove the AArch64 backend API against CBMC specs of assembly routines.

mkannwischer:
- [X] mlk_ntt_asm (#1029)
- [X] mlk_intt_asm (#1029)
- [ ] mlk_poly_reduce_asm
- [X] mlk_poly_tomont_asm (#1029)
- [ ] mlk_poly_mulcache_compute_asm
- [X] mlk_poly_tobytes_asm (#1029)
- [X] mlk_polyvec_basemul_acc_montgomery_cached_asm_k2 (#1028) 
- [X] mlk_polyvec_basemul_acc_montgomery_cached_asm_k3 (#1028)
- [X] mlk_polyvec_basemul_acc_montgomery_cached_asm_k4 (#1028)
- [X] mlk_rej_uniform_asm (#1027)

mkannwischer:
Should we do the same for the Keccak implementations?

PR #1028, opened 2025-05-18
CBMC: Prove AArch64 native API for basemul atop HOL-Light spec

hanno-becker:
* Part of #1025

This commit shows that the AArch64 implementations of `polyvec_basemul_acc_montgomery_cached_k{2,3,4}_native` satisfiy their CBMC spec atop the HOL-Light spec for the underlying assembly routines.

More precisely, we translate relevant aspects of the HOL-Light specs to CBMC specs, and then show that assuming those specs, the AArch64 implementations uphold their CBMC contracts defined in `native/api.h`.

### DIFF
--- a/dev/aarch64_clean/src/arith_native_aarch64.h
+++ b/dev/aarch64_clean/src/arith_native_aarch64.h
@@ -54,21 +54,54 @@ void mlk_poly_tobytes_asm(uint8_t *r, const int16_t *a);
 void mlk_polyvec_basemul_acc_montgomery_cached_asm_k2(int16_t *r,
                                                       const int16_t *a,
                                                       const int16_t *b,
-                                                      const int16_t *b_cache);
+                                                      const int16_t *b_cache)
+/* This must be kept in sync with the HOL-Light specification in
+ * proofs/hol_light/arm/proofs/mlkem_poly_basemul_acc_montgomery_cached_k2.ml.
+ */
+__contract__(
+    requires(memory_no_alias(r, sizeof(int16_t) * MLKEM_N))
+    requires(memory_no_alias(a, sizeof(int16_t) * 2 * MLKEM_N))
+    requires(memory_no_alias(b, sizeof(int16_t) * 2 * MLKEM_N))
+    requires(memory_no_alias(b_cache, sizeof(int16_t) * 2 * (MLKEM_N / 2)))
+    requires(array_bound(a, 0, 2 * MLKEM_N, 0, MLKEM_UINT12_LIMIT))
+    assigns(memory_slice(r, sizeof(int16_t) * MLKEM_N))
+);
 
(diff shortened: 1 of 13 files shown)


proofs/hol_light/arm/proofs/mlkem_poly_basemul_acc_montgomery_cached_k2.ml
```diff
@@ -378,6 +378,7 @@ let poly_basemul_acc_montgomery_cached_k2_SPEC = prove(poly_basemul_acc_montgome
     CONV_TAC INT_RING
 );;
 
+(* NOTE: This needs to be kept in sync with the CBMC spec in native/api.h *)
```
mkannwischer, line 381: Should be `mlkem/native/aarch64/src/arith_native_aarch64.h`?

dev/aarch64_clean/src/arith_native_aarch64.h
```diff
@@ -54,21 +54,54 @@ void mlk_poly_tobytes_asm(uint8_t *r, const int16_t *a);
...
+                                                      const int16_t *b_cache)
+/* This must be kept in sync with the HOL-Light specification in
+ * proofs/hol_light/arm/proofs/mlkem_poly_basemul_acc_montgomery_cached_k3.ml.
+ */
+__contract__(
+    requires(memory_no_alias(r, sizeof(int16_t) * MLKEM_N))
+    requires(memory_no_alias(a, sizeof(int16_t) * 3 * MLKEM_N))
+    requires(memory_no_alias(b, sizeof(int16_t) * 3 * MLKEM_N))
+    requires(memory_no_alias(b_cache, sizeof(int16_t) * 3 * (MLKEM_N / 2)))
+    requires(array_bound(a, 0, 3 * MLKEM_N, 0, MLKEM_UINT12_LIMIT))
```
mkannwischer, line 84: The HOL Light spec says `abs(ival(x0 i)) <= &2 pow 12`, right? So that does not quite match. 
Should be `array_abs_bound(... , MLKEM_UINT12_LIMIT + 2)`?

mkannwischer (changes requested):
Thanks @hanno-becker.
I'm confused by the bounds. If I understand this correctly, then the asm bounds are a bit less restrictive. Can we match what it is in the HOL Light spec?

dev/aarch64_clean/src/arith_native_aarch64.h
hanno-becker: You are right!

proofs/hol_light/arm/proofs/mlkem_poly_basemul_acc_montgomery_cached_k2.ml
hanno-becker: Yes

mkannwischer (approved):
Thanks for the changes @hanno-becker! Looks good to me now.

PR #1030, opened 2025-05-18
CBMC: Prove AArch64 native API for poly_reduce, poly_mulcache_compute atop HOL-Light spec

mkannwischer:
This commit shows that the AArch64 implementations of `poly_reduce_native` and `poly_mulcache_compute_native` satisfiy their CBMC spec atop the HOL-Light spec for the underlying assembly routines.

More precisely, we translate relevant aspects of the HOL-Light specs to CBMC specs, and then show that assuming those specs, the AArch64 implementations uphold their CBMC contracts defined in native/api.h.

### DIFF
--- a/dev/aarch64_opt/src/arith_native_aarch64.h
+++ b/dev/aarch64_opt/src/arith_native_aarch64.h
@@ -62,7 +62,14 @@ __contract__(
 );
 
 #define mlk_poly_reduce_asm MLK_NAMESPACE(poly_reduce_asm)
-void mlk_poly_reduce_asm(int16_t *);
+void mlk_poly_reduce_asm(int16_t *p)
+/* This must be kept in sync with the HOL-Light specification
+ * in proofs/hol_light/arm/proofs/mlkem_poly_reduce.ml */
+__contract__(
+  requires(memory_no_alias(p, sizeof(int16_t) * MLKEM_N))
+  assigns(memory_slice(p, sizeof(int16_t) * MLKEM_N))
+  ensures(array_bound(p, 0, MLKEM_N, 0, MLKEM_Q))
+);
 
 #define mlk_poly_tomont_asm MLK_NAMESPACE(poly_tomont_asm)
 void mlk_poly_tomont_asm(int16_t *p)
--- a/dev/aarch64_opt/src/arith_native_aarch64.h
+++ b/dev/aarch64_opt/src/arith_native_aarch64.h
@@ -93,7 +93,6 @@ __contract__(
   requires(zetas == mlk_aarch64_zetas_mulcache_native)
   requires(zetas_twisted == mlk_aarch64_zetas_mulcache_twisted_native)
   assigns(object_whole(cache))
-  ensures(array_abs_bound(cache, 0, MLKEM_N / 2, MLKEM_Q))
 );
 
 #define mlk_poly_tobytes_asm MLK_NAMESPACE(poly_tobytes_asm)
--- a/dev/aarch64_opt/src/arith_native_aarch64.h
+++ b/dev/aarch64_opt/src/arith_native_aarch64.h
@@ -82,8 +82,19 @@ __contract__(
 );
 
 #define mlk_poly_mulcache_compute_asm MLK_NAMESPACE(poly_mulcache_compute_asm)
-void mlk_poly_mulcache_compute_asm(int16_t *, const int16_t *, const int16_t *,
-                                   const int16_t *);
+void mlk_poly_mulcache_compute_asm(int16_t *cache, const int16_t *mlk_poly,
+                                   const int16_t *zetas,
+                                   const int16_t *zetas_twisted)
+/* This must be kept in sync with the HOL-Light specification
+ * in proofs/hol_light/arm/proofs/mlkem_poly_mulcache_compute.ml */
+__contract__(
+  requires(memory_no_alias(cache, sizeof(int16_t) * (MLKEM_N / 2)))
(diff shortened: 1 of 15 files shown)


mlkem/native/aarch64/src/arith_native_aarch64.h
```diff
@@ -75,8 +82,19 @@ __contract__(
...
+                                   const int16_t *zetas_twisted)
+/* This must be kept in sync with the HOL-Light specification
+ * in proofs/hol_light/arm/proofs/mlkem_poly_mulcache_compute.ml */
+__contract__(
+  requires(memory_no_alias(cache, sizeof(int16_t) * (MLKEM_N / 2)))
+  requires(memory_no_alias(mlk_poly, sizeof(int16_t) * MLKEM_N))
+  requires(zetas == mlk_aarch64_zetas_mulcache_native)
+  requires(zetas_twisted == mlk_aarch64_zetas_mulcache_twisted_native)
+  assigns(object_whole(cache))
+  ensures(array_abs_bound(cache, 0, MLKEM_N / 2, MLKEM_Q))
```
hanno-becker, line 96: Where does the HOL-Light spec say that?

hanno-becker: It doesn't, and judging from the comments in the proof script, deliberately so. Indeed, `poly_mulcache_compute()` has no bounds post-condition. Somehow it slipped into the spec for `poly_mulcache_compute_native()`, but there too it should be unnecessary.

hanno-becker (changes requested):
There's an incosistency in the presence/absence of a postcondition bound in `poly_mulcache_compute`: `poly_mulcache_compute()` has no output bound. `poly_mulcache_compute_native()` does. `poly_mulcache_compute_native_asm()` does. The HOL-Light spec does not.

For the time being, the output bound should be removed everywhere, as it is not needed in the higher level reasoning.

mlkem/native/aarch64/src/arith_native_aarch64.h
mkannwischer: Oh I just copied from the contract of native function and misunderstood `(mod &3329)`. I'll update. 
Good catch!

dev/aarch64_opt/src/arith_native_aarch64.h
```diff
@@ -75,8 +82,18 @@ __contract__(
...
+                                   const int16_t *zetas,
+                                   const int16_t *zetas_twisted)
+/* This must be kept in sync with the HOL-Light specification
+ * in proofs/hol_light/arm/proofs/mlkem_poly_mulcache_compute.ml */
+__contract__(
+  requires(memory_no_alias(cache, sizeof(int16_t) * (MLKEM_N / 2)))
+  requires(memory_no_alias(mlk_poly, sizeof(int16_t) * MLKEM_N))
+  requires(zetas == mlk_aarch64_zetas_mulcache_native)
+  requires(zetas_twisted == mlk_aarch64_zetas_mulcache_twisted_native)
+  assigns(object_whole(cache))
```
hanno-becker, line 95: Nit: It is somewhat inconsistent though not harmful to use `object_whole` here. Was that deliberate?

proofs/hol_light/arm/proofs/mlkem_poly_reduce.ml
```diff
@@ -118,6 +118,9 @@ let overall_lemma = prove
   REWRITE_TAC[MATCH_MP lemma_rem (CONGBOUND_RULE `barred x`)] THEN
   BITBLAST_TAC);;
 
+(* NOTE: This must be kept in sync with the CBMC specification
```
hanno-becker, line 121: Nit: While this is the spec for the actual code, the spec for the entire function (including potential stack spilling and the final `ret`) is `MLKEM_POLY_REDUCE_SUBROUTINE_CORRECT` at the bottom. The comment would thus be slightly better placed there.

mkannwischer: Does this also apply to the other functions, e.g., rej_uniform?

dev/aarch64_opt/src/arith_native_aarch64.h
mkannwischer: Inherited from the `mlk_poly_mulcache_compute_native` contract. 
The cache is not part of a larger structure, so there is no reason to not use `object_whole`?

proofs/hol_light/arm/proofs/mlkem_poly_reduce.ml
hanno-becker: Yes -- I may have made the same mistake there! Can you check?

dev/aarch64_opt/src/arith_native_aarch64.h
hanno-becker: Okay.

proofs/hol_light/arm/proofs/mlkem_poly_reduce.ml
hanno-becker: (We're not entirely consistent here -- some functions have only one contract, but most are split in the above way, I think)

mkannwischer: okay - I will adapt it for all

hanno-becker: Great, thank you!

mkannwischer: Done. Please double check if I got that right.

hanno-becker (approved):

PR #1037, opened 2025-05-19
Add RISCV64 RVV backend

mjosaarinen:
**Summary**:
rv64v support (risc-v vector extension 1.0, which is available on newer application-class silicon.)

yes (risc-v only)

Roughly 2.5x perf on silicon with vector hardware.

### DIFF
--- a/mlkem/src/native/riscv64/src/rv64v_poly.c
+++ b/mlkem/src/native/riscv64/src/rv64v_poly.c
@@ -740,7 +740,7 @@ void mlk_rv64v_poly_reduce(int16_t *r)
 unsigned int mlk_rv64v_rej_uniform(int16_t *r, unsigned int len,
                                    const uint8_t *buf, unsigned int buflen)
 {
-  size_t n, ctr, pos;
+  unsigned n, ctr, pos;
   vuint16m1_t x, y;
   vbool16_t lt;
 
@@ -749,8 +749,8 @@ unsigned int mlk_rv64v_rej_uniform(int16_t *r, unsigned int len,
 
   while (ctr < len && pos < buflen)
   {
-    const size_t vl = __riscv_vsetvl_e16m1((buflen - pos) * 8 / 12);
-    const size_t vl23 = (vl * 24) / 32;
+    const unsigned vl = (unsigned)__riscv_vsetvl_e16m1((buflen - pos) * 8 / 12);
+    const unsigned vl23 = (vl * 24) / 32;
 
     const vuint16m1_t vid = __riscv_vid_v_u16m1(vl);
     const vuint16m1_t srl12v = __riscv_vmul_vx_u16m1(vid, 12, vl);
@@ -772,7 +772,7 @@ unsigned int mlk_rv64v_rej_uniform(int16_t *r, unsigned int len,
 
       lt = __riscv_vmsltu_vx_u16m1_b16(x, MLKEM_Q, vl);
       y = __riscv_vcompress_vm_u16m1(x, lt, vl);
-      n = __riscv_vcpop_m_b16(lt, vl);
+      n = (unsigned)__riscv_vcpop_m_b16(lt, vl);
 
       if (ctr + n > len)
       {
--- a/mlkem/src/native/riscv64/meta.h
+++ b/mlkem/src/native/riscv64/meta.h
@@ -61,7 +61,10 @@ static MLK_INLINE int mlk_rej_uniform_native(int16_t *r, unsigned len,
                                              const uint8_t *buf,
                                              unsigned buflen)
 {
-  return mlk_rv64v_rej_uniform(r, len, buf, buflen);
+  /* The cast from unsigned to signed integer is safe
+   * because the return value is <= len, which we asssume
+   * to be bound by 4096 and hence <= INT_MAX. */
+  return (int)mlk_rv64v_rej_uniform(r, len, buf, buflen);
 }
 
 static MLK_INLINE int mlk_poly_reduce_native(int16_t data[MLKEM_N])


hanno-becker:
@mjosaarinen If you have `nix` setup, running `autogen` should hopefully resolve the linting issues.

rod-chapman:
Note that on the "fastntt3" branch, there are layer-merged implementations of the NTT and INTT that are highly amenable to auto-vectorization with compilers like GCC 14. Benchmarks of that code on an RV64v target were encouraging, so might provide some inspiration for a fully vectorized, hand-written back-end.

mjosaarinen:
> Note that on the "fastntt3" branch, there are layer-merged implementations of the NTT and INTT that are highly amenable to auto-vectorization with compilers like GCC 14. Benchmarks of that code on an RV64v target were encouraging, so might provide some inspiration for a fully vectorized, hand-written back-end.

Yeah you can easily double the speed with autovectorization alone, and some Google folks were of the opinion that they wanted to rely on that entirely in BoringSSL (RISC-V Android etc), rather than maintain a hand-optimized version. The resulting code is pretty wild; I looked at that when considering RISC-V ISA extensions ( see slides 17 for example in https://mjos.fi/doc/20240325-rwc-riscv.pdf ). It was almost "too good" -- I suspect that Google has used those NTTs as a microbenchmark when developing LLVM autovectorizers :)

mjosaarinen:
> @mjosaarinen If you have `nix` setup, running `autogen` should hopefully resolve the linting issues.

Yeah, sorry for abusing your CI like that (I wasn't expecting it to be that extensive), I could have just read the documentation. I'll set up this nix thing.

hanno-becker:
@mjosaarinen Sorry, we should have pointed that out earlier. With the `nix` environment, you should not have problems anymore to make linter happy. Just run `format && autogen` before pushing.

mlkem/native/riscv64/src/rv64v_settings.h
```diff
@@ -0,0 +1,40 @@
...
+#define MLK_MONT_R1 2285
+
+/*  r2  = lift(Mod(r, q)^2) */
+#define MLK_MONT_R2 1353
+
+/*  in  = lift(Mod(n / 2, q)^-1) */
+/*  nr  = (in * r^2) % q */
+#define MLK_MONT_NR 1441
+
+/* check-magic: on */
```
hanno-becker, line 38:
```suggestion
/* check-magic: 3327 == pow(-MLKEM_Q, -1, 2^16) */
#define MLKEM_QI 3327

/* check-magic: 2285 == unsigned_mod(2^16, MLKEM_Q) */
#define MLK_MONT_R1 2285

/* check-magic: 1353 == pow(MLK_MONT_R1, 2, MLKEM_Q) */
#define MLK_MONT_R2 1353

/* check-magic: 1441 == pow(2,32 - 7,MLKEM_Q) */
#define MLK_MONT_NR 1441

```

This auto-checks the magic number explanations in CI.

mlkem/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,820 @@
...
+  vp = __riscv_vrgatherei16_vv_i16m2(vp, v2p1, vl2);
+
+  return vp;
+}
+
+void mlk_rv64v_poly_ntt(int16_t *r)
+{
+  /* zetas can be compiled into vector constants; don't pass as a pointer */
+  /* check-magic: off */
+  const int16_t zeta[0x80] = {
```
hanno-becker, line 232: Those should ultimately be autogenerated via `autogen` similar to the other twiddles tables (e.g. `zetas.inc`). You will be able to copy-paste adjust most of it, I think -- from a cursory look, this is different from `zetas.inc` only in the order of the twiddles.

Do you have time to look into that, or shall me/Matthias do it as a follow-up?

mjosaarinen: I will be returning to this code in a couple of weeks, and I don't mind if someone scripts them in the meanwhile. All three tables were generated with throw-away gp-pari statements but are  obvious to "reverse engineer" as you note; just a some combo of various orderings of roots-of-unity powers and Montgomery constants. Anyway, I understand what you're after here.

hanno-becker:
@mkannwischer @mjosaarinen The RV functional tests in the ci-cross shell don't seem to pass.

```bash
% CROSS_PREFIX=riscv64-unknown-linux-gnu- make func OPT=1 AUTO=1 -j32
% EXEC_WRAPPER=qemu-riscv64 make run_func_512 -j32 OPT=1 AUTO=1
qemu-riscv64 test/build/mlkem512/bin/test_mlkem512
ERROR (test/test_mlkem.c,41)
ERROR (test/test_mlkem.c,225)
make: *** [Makefile:53: run_func_512] Error 1
```

Same for `OPT=0`.

Could you investigate?

hanno-becker (changes requested):
The RV64 cross tests in CI are failing. I don't know if this is an issue with the code or the CI setup, but it needs looking into.

mjosaarinen:
> The RV64 cross tests in CI are failing. I don't know if this is an issue with the code or the CI setup, but it needs looking into.

It looks like some things are running fine, and some things have build problems. I suspect that platform auto-detection code (in the C header files + makefies) is to blame (failing no_opt means "no optimization"?) It's a quite complicated CI -- can you tell me what would be the best way to debug this stage of the CI build process locally?

hanno-becker:
> It looks like some things are running fine, and some things have build problems

So far, it seems to me that the RV64 build works fine, but that all tests fail at runtime.

> It's a quite complicated CI -- can you tell me what would be the best way to debug this stage of the CI build process locally?

Sure. If you have nix installed, then it's exactly what I posted above: You open the nix shell with 

```bash
% nix develop --extra-experimental-features 'nix-command flakes' .#ci-cross
```

Building using cross compiler:

```
% CROSS_PREFIX=riscv64-unknown-linux-gnu- make func AUTO=1 -j32
```

And then running:

```
% qemu-riscv64 test/build/mlkem512/bin/test_mlkem512
% qemu-riscv64 test/build/mlkem768/bin/test_mlkem768
% qemu-riscv64 test/build/mlkem1024/bin/test_mlkem1024
```

The base command line used by `AUTO=1` above is:

```
riscv64-unknown-linux-gnu-gcc -Wall -Wextra -Werror=unused-result -Wpedantic -Werror -Wmissing-prototypes -Wshadow -Wpointer-arith -Wredundant-decls -Wno-long-long -Wno-unknown-pragmas -Wno-unused-command-line-argument -O3 -fomit-frame-pointer -std=c99 -pedantic -MMD  -march=rv64gcv_zvl256b -DMLK_FORCE_RISCV64 -DMLK_CONFIG_USE_NATIVE_BACKEND_ARITH -DMLK_CONFIG_USE_NATIVE_BACKEND_FIPS202 -DMLK_CONFIG_PARAMETER_SET=1024 
```

What is confusing to me is that this happens regardless of whether `OPT=0` or `OPT=1` (it defaults to `OPT=0`). If `OPT=0`, the new files aren't even compiled.

hanno-becker:
I adjusted the CI. Most tests pass, but there are some issues in the ACVP tests and the monobuild examples. The ACVP one seems to be an issue in the test script (not expecting parameters in the exec-wrapper), while the monobuild failure seems to be an architecture confusion. I'll take a look.

mlkem/native/riscv64/src/rv64v_settings.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef MLK_NATIVE_RISCV64_SRC_RV64V_SETTINGS_H
+#define MLK_NATIVE_RISCV64_SRC_RV64V_SETTINGS_H
+
+/*  === vector configuration */
+#ifndef MLK_RVV_VLEN
+#define MLK_RVV_VLEN 256
+#endif
+
+/*
+#define MLK_RVV_WIDENING_MUL
```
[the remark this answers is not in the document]
hanno-becker: @mjosaarinen What determines whether this option can/should be used?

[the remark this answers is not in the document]
hanno-becker: Thanks Markku! Let's remove it then -- will you do it?

mlkem/native/riscv64/riscv64-notes.md
```diff
@@ -0,0 +1,84 @@
+[//]: # (SPDX-License-Identifier: CC-BY-4.0)
+
+#	rv64v-notes.md
+
+Notes on RISC-V Vector support.
+
+**WARNING** This is highly experimental. Currently vlen=256 only.
```
hanno-becker: Most of this README can now be removed, right? We can refer to `nix develop .#ci-cross` for the required toolchain, and `spike` is not needed for now since we don't import the Keccak backend yet?

hanno-becker: @mkannwischer @mjosaarinen Are you planning to open a separate PR for leveraging the Keccak extension?

hanno-becker:
@mkannwischer @mjosaarinen (When) Are you planning to progress this? This is great work, and we've overcome the main challenges, so it would be a shame to see this bitrot.

mjosaarinen:
> @mkannwischer @mjosaarinen (When) Are you planning to progress this? This is great work, and we've overcome the main challenges, so it would be a shame to see this bitrot.

To answer your question -- yes, it is alive and we want to keep this PR up to date until it is appropriate for merge. The plan was to conduct some tests with real silicon, and especially try out VLEN=512, etc.

mkannwischer:
I'll continue working on this now. I've rebased on top of main and will resolve any remaining issues.

mlkem/native/riscv64/src/rv64v_settings.h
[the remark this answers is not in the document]
mkannwischer: I removed it.

mlkem/native/riscv64/riscv64-notes.md
mkannwischer: I have trimed down the README.

mkannwischer:
Okay, CI seems happy now and I have addressed the minor open points. 
I hope the rebase did not break anything, there were quite a few conflicts. 

I'll review this PR again tomorrow to see if there is anything else we need to change.

mlkem/src/native/meta.h
```diff
@@ -18,4 +18,8 @@
 #include "x86_64/meta.h"
 #endif
 
+#ifdef MLK_SYS_RISCV64
```
hanno-becker: How can we detect statically whether we are compiling for a RV64 platform with vector extension and VL=256? At the moment, we're pulling this in for any RV64 target, which is not correct.

mlkem/src/native/riscv64/riscv64-notes.md
```diff
@@ -0,0 +1,8 @@
+[//]: # (SPDX-License-Identifier: CC-BY-4.0)
```
hanno-becker: Minor: Rename file to `README.md`

mlkem/src/native/riscv64/riscv64-notes.md
```diff
@@ -0,0 +1,8 @@
...
+
+#	rv64v-notes.md
+
+Notes on RISC-V Vector support.
+
+**WARNING** This is highly experimental. Currently vlen=256 only.
```
hanno-becker: Let's try to be more concrete. In what way is this implementation inferior to the AArch64 or AVX2 backends?
- Verification: No verification
- Test: Same tests as for AVX2 and AArch64 as far as I can see?
- Audit: New piece of code, has not yet seen many 👀 nor use

mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,811 @@
...
+
+#if (MLK_RVV_VLEN == 256)
+
+/*************************************************
+ * Name:        poly_ntt
+ *
+ * Description: Computes negacyclic number-theoretic transform (NTT) of
+ *              a polynomial in place;
+ *              inputs assumed to be in normal order, output in
+ *bitreversed order
```
hanno-becker: Nit: formatting

mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,811 @@
...
+#undef MLK_RVV_BFLY_FX
+#undef MLK_RVV_BFLY_FV
+
+/*************************************************
+ * Name:        poly_invntt_tomont
+ *
+ * Description: Computes inverse of negacyclic number-theoretic transform (NTT)
+ *              of a polynomial in place;
+ *              inputs assumed to be in bitreversed order,
+ *output in normal order
```
hanno-becker: Nit: formatting

mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,811 @@
...
+
+/*  Kyber's middle field GF(3329)[X]/(X^2) multiplication   */
+
+static inline void mlk_rv64v_poly_basemul_mont_add_k(int16_t *r,
+                                                     const int16_t *a,
+                                                     const int16_t *b,
+                                                     unsigned kn)
+{
+  /* check-magic: off */
+  const int16_t roots[MLKEM_N] = {
```
hanno-becker: Those should be autogenerated by `autogen`

mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,811 @@
...
+  size_t n, ctr, pos;
+  vuint16m1_t x, y;
+  vbool16_t lt;
+
+  pos = 0;
+  ctr = 0;
+
+  while (ctr < len && pos + (vl23 * 2) <= buflen)
+  {
+    x = __riscv_vle16_v_u16m1((uint16_t *)&buf[pos], vl23);
```
hanno-becker: @mjosaarinen Can you briefly describe how this implementation works? It seems much simpler than AArch64 and AVX2.

mlkem/src/native/riscv64/src/rv64v_settings.h
```diff
@@ -0,0 +1,37 @@
...
+#ifndef MLK_RVV_VLEN
+#define MLK_RVV_VLEN 256
+#endif
+
+
+
+/*  vl value for a 16-bit wide type */
+#define MLK_RVV_E16M1_VL (MLK_RVV_VLEN / 16)
+
+/* check-magic: off */
```
hanno-becker: Remove this and convert the constant explanations below in a form that `check-magic` understands

hanno-becker (changes requested):
Some smaller change requests, see comments

mlkem/src/native/riscv64/riscv64-notes.md
mkannwischer: What do you think about removing this file entirely and adding the important info the our root README?

mlkem/src/native/riscv64/src/rv64v_poly.c
mkannwischer: I implemented autogeneration in 559a7fad7aebf105693f0673b56d8f945d9dd8c4

mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,750 @@
...
+#include "rv64v_settings.h"
+
+static inline vint16m1_t fq_redc(vint16m1_t rh, vint16m1_t rl, size_t vl)
+{
+  vint16m1_t t;
+  vbool16_t c;
+
+  t = __riscv_vmul_vx_i16m1(rl, MLK_RVV_QI, vl); /* t = l * -Q^-1  */
+  t = __riscv_vmulh_vx_i16m1(t, MLKEM_Q, vl);    /* t = (t*Q) / R  */
+  c = __riscv_vmsne_vx_i16m1_b16(rl, 0, vl);     /* c = l == 0     */
```
hanno-becker: The comment looks wrong here. Should it be `c = (l != 0)`?

Why are we not using a subtract on the high parts instead, avoiding the carry-in?

mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,750 @@
...
+  return rx;
+}
+
+/*  Conditionally subtract Q (if Q or above) */
+
+static inline vint16m1_t fq_csub(vint16m1_t rx, size_t vl)
+{
+  vbool16_t bn;
+
+  bn = __riscv_vmsge_vx_i16m1_b16(rx, MLKEM_Q, vl);       /*   if x >= 0:  */
```
hanno-becker: This should be `if x >= Q`

mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,750 @@
...
+ * Arguments:   - uint16_t *r: pointer to in/output polynomial
+ **************************************************/
+
+/*  reverse butterfly operation */
+
+#define MLK_RVV_BFLY_RX(u0, u1, ut, uc, vl) \
+  {                                         \
+    ut = __riscv_vsub_vv_i16m1(u0, u1, vl); \
+    u0 = __riscv_vadd_vv_i16m1(u0, u1, vl); \
+    u0 = fq_csub(u0, vl);                   \
```
hanno-becker: @mkannwischer @mjosaarinen Overall, the implementations seems to work with signed rather than unsigned values, hence I would expect every normalization operation to be symmetric -- but here we are only reducing values > Q, while keeping values < -Q as they are. That would suggest that with bad luck, one could get negative arithmetic overflow.

Can you clarify?

mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,750 @@
...
+    u1 = fq_mul_vx(ut, uc, vl);             \
+  }
+
+#define MLK_RVV_BFLY_RV(u0, u1, ut, uc, vl) \
+  {                                         \
+    ut = __riscv_vsub_vv_i16m1(u0, u1, vl); \
+    u0 = __riscv_vadd_vv_i16m1(u0, u1, vl); \
+    u0 = fq_csub(u0, vl);                   \
```
hanno-becker: Same question here

mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,750 @@
...
+  const vint16m1_t z8 = __riscv_vle16_v_i16m1(&izeta[0x40], vl);
+  const vint16m1_t za = __riscv_vle16_v_i16m1(&izeta[0x50], vl);
+  const vint16m1_t zc = __riscv_vle16_v_i16m1(&izeta[0x60], vl);
+  const vint16m1_t ze = __riscv_vle16_v_i16m1(&izeta[0x70], vl);
+
+  vint16m1_t vt;
+  vint16m1_t v0, v1, v2, v3, v4, v5, v6, v7, v8, v9, va, vb, vc, vd, ve, vf;
+  vint16m2_t vp;
+
+  vp = mlk_rv64v_intt2(__riscv_vle16_v_i16m2(&r[0x00], vl2), z0);
```
hanno-becker, line 486: Blocker: We currently make no bounds assumption on the inputs, which forces an implementation to either reduce the inputs upfront, or frontload the Montgomery normalization (which sufficiently reduces as a side-effect).

I am surprised that we don't get a functional failure here in some cases.

hanno-becker (changes requested):
I think the forward and inverse NTT still need to be aligned to the bounds pre/post conditions worked with in the higher-level code:
* Forward NTT: As far as I see, we should get the same coefficient growth as in the C implementation and other backends, so by the end of the forward NTT, we should be within absolute value < 8 * MLKEM_Q as required by the CBMC spec. The final reduction is thus not necessary, though not harmful either.
* Inverse NTT: The CBMC specs don't make any assumptions about the input bounds in the input NTT. We therefore must reduce or front-load the Montgomery normalization -- as it stands, I believe the initial butterflies could overflow, and am surprised why we didn't see this happen yet (I may be wrong, of course, and that's why)
* Inverse NTT: We seem to be doing a conditional subtraction to contain coefficient growth in the inverse NTT, but that would only positive coefficient, not negative ones -- what prevents exponential _negative_ coefficient growth?

hanno-becker:
I see some bounds assertion failing for `poly_reduce()`. It seems that the output is sometimes slightly non-canonical: values -1, -2, -3 are observed.

mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,768 @@
...
+static inline vint16m1_t fq_barrett(vint16m1_t a, size_t vl)
+{
+  vint16m1_t t;
+  const int16_t v = ((1 << 26) + MLKEM_Q / 2) / MLKEM_Q;
+
+  t = __riscv_vmulh_vx_i16m1(a, v, vl);
+  t = __riscv_vadd_vx_i16m1(t, 1 << (25 - 16), vl);
+  t = __riscv_vsra_vx_i16m1(t, 26 - 16, vl);
+  t = __riscv_vmul_vx_i16m1(t, MLKEM_Q, vl);
+  t = __riscv_vsub_vv_i16m1(a, t, vl);
```
hanno-becker, line 68: Blocker: This is failing bounds checks -- it sometimes produces slightly negative values (-1,-2,-3)? Needs to be understood.

hanno-becker: I think this is not exactly equivalent to the Kyber reduction because it does not resort to 32-bit in the interim. Compare Kyber:
```c
int16_t barrett_reduce(int16_t a) {
  int16_t t;
  const int16_t v = ((1<<26) + KYBER_Q/2)/KYBER_Q;

  t  = ((int32_t)v*a + (1<<25)) >> 26;
  t *= KYBER_Q;
  return a - t;
}
```

Compare `poly.c:`
```c
/* Reference: `barrett_reduce()` in the reference implementation @[REF]. */
static MLK_INLINE int16_t mlk_barrett_reduce(int16_t a)
__contract__(
  ensures(return_value > -MLKEM_Q_HALF && return_value < MLKEM_Q_HALF)
)
{
  /* Barrett reduction approximates
   * ```
   *     round(a/MLKEM_Q)
   *   = round(a*(2^N/MLKEM_Q))/2^N)
   *  ~= round(a*round(2^N/MLKEM_Q)/2^N)
   * ```
   * Here, we pick N=26.
   */
  const int32_t magic = 20159; /* check-magic: 20159 == round(2^26 / MLKEM_Q) */

  /*
   * PORTABILITY: Right-shift on a signed integer is
   * implementation-defined for negative left argument.
   * Here, we assume it's sign-preserving "arithmetic" shift right.
   * See (C99 6.5.7 (5))
   */
  const int32_t t = (magic * a + (1 << 25)) >> 26;

  /*
   * t is in -10 .. +10, so we need 32-bit math to
   * evaluate t * MLKEM_Q and the subsequent subtraction
   */
  int16_t res = (int16_t)(a - t * MLKEM_Q);

  mlk_assert_abs_bound(&res, 1, MLKEM_Q_HALF);
  return res;
}
```

The AArch64 code does not have this issue because thanks to the inherent doubling in `SQDMULH`, one can approximate with 2^27.

hanno-becker: I am struggling to get to the bottom of this and need to leave this for now. @mkannwischer @mjosaarinen Please have a look. I wrote unit tests for this piece of functionality, and it seems to be working as expected -- and yet the output-assertion in `mlk_poly_reduce()` fails, claiming that it finds values `-1,-2,-3`. It could be a compiler bug..

hanno-becker: Ignore this. I tested using the wrong qemu flag. The other comments regarding forward and inverse NTT still stand.

mlkem/src/native/riscv64/src/rv64v_poly.c
hanno-becker: Thanks!

mlkem/src/native/riscv64/src/rv64v_poly.c
hanno-becker: Fixed

mlkem/src/native/riscv64/src/rv64v_poly.c
hanno-becker: Fixed the comment, but the other question still stands.

hanno-becker:
@mjosaarinen Are you planning to continue working on this?

mjosaarinen:
> @mjosaarinen Are you planning to continue working on this?

Everyone can contribute, but yes, I have more time allocated for it in September. I hope it is not blocking development elsewhere -- if it is, feel free to change those things, and we can revisit these issues that you mention (which seem easy to address).

The overall goal is the RISC-V Summit NA in October, where we have an announcement talk in the main program ( https://riscvsummit2025.sched.com/event/28OTv/pqcp-support-for-risc-v-vector-future-keccak-extensions-markku-juhani-o-saarinen-tampere-university )  In August, the allocated time was somewhat wasted due to numerous technical issues with a prototype development board (RVV VLEN=512) that was intended to serve as a target for RISC-V implementations.

hanno-becker:
> Everyone can contribute, but yes, I have more time allocated for it in September. I hope it is not blocking development elsewhere -- if it is, feel free to change those things, and we can revisit these issues that you mention (which seem easy to address).

It's not blocking, no. I was just curious because it would be great to see this progress and get merged eventually.

mlkem/src/native/riscv64/src/rv64v_poly.c
mjosaarinen: The result `ut` of subtraction is only used by the following multiplication `fq_mul_vv`, which has a Montgomery reduction, keeping the result within its bounds. However, the first very coefficient is never subjected to Montgomery reduction and hence may require `csub` (well, probably not on every layer.)

mlkem/src/native/riscv64/src/rv64v_poly.c
mjosaarinen: Same here, the temporary result `ut` is only used by the Montgomery multiplication which is the next line on that macro.

mjosaarinen:
Hi, spent a couple of hours on this: updated the middle-layer API to match with current convention, but didn't touch the actual risc-v intrinsics code yet. At least the code passes the functional tests (kat, acvp) fine.

Note: RVA23 (with rvv1.0) is the expected to be the most common ISA in Application-class RISC-V CPUs. RVA23 is already required to run e.g. in RISC-V Ubuntu 25.10 ( https://www.phoronix.com/news/Ubuntu-25.10-To-Require-RVA23 ). But vectors still need to be explicitly enabled in these nix versions of QEMU and the cross compilers.

One needs to pass `-cpu rv64,v=true,vlen=256` to QEMU and `-march=rv64gcv` to clang or gcc.  The NTT part still requires vlen==256 exactly. Other lengths will be common, so we will need to address this shortly.

Anyway, compiling and running example:
```
$ EXEC_WRAPPER="qemu-riscv64 -cpu rv64,v=true,vlen=256" CROSS_PREFIX=riscv64-unknown-linux-gnu- make func AUTO=1 test
[...]
Running encapDecap test case 165 (encapsulationKeyCheck) ... done
Comparing results with test/.acvp-data/v1.1.0.40/files/ML-KEM-encapDecap-FIPS203/expectedResults.json
OK
ALL GOOD!
  Everything checks fine!
```

hanno-becker:
@mjosaarinen Thank you! I will have a look this week.

mlkem/src/native/riscv64/src/rv64v_poly.c
hanno-becker: @mjosaarinen Ok, but still `fq_csub` seems to only reduce if the value is `>= Q`, so for _negative_ coefficients of large absolute value, you could still get exponential growth (wrt the number of layers)?

hanno-becker:
@mjosaarinen I took the liberty to force-push a rebase atop `main`, so we don't have merge commits in the history.

hanno-becker:
Rebased atop #1188 which adds backend unit tests. Those should automatically be exercised for RV64 as well now.

mjosaarinen:
> Rebased atop #1188 which adds backend unit tests. Those should automatically be exercised for RV64 as well now.

The INTT test had seemingly unconstrained inputs, requiring an additional initial reduction ( https://github.com/pq-code-package/mlkem-native/pull/1037/commits/fae033031102b2bdb1cf0e742c07367b46d5917c ). One could of course derive more lax bound assumptions for INTT inputs via interval arithmetic examination of ML-KEM.
btw, for that "mystery" on why the INTT didn't overflow, did you notice the "half-way" layer reduction? the NTT and INTT operations are split into two halfs, with ntt2 and innt2 operating on three inner layers that fit into a single vector.

hanno-becker:
> for that "mystery" on why the INTT didn't overflow, did you notice the "half-way" layer reduction? 

I'm afraid I still don't understand the one-sided reduction discussed in https://github.com/pq-code-package/mlkem-native/pull/1037#discussion_r2367588956 (reducing >= Q, but not <= -Q). If you could shed more light on that, I'd be glad. 

> The INTT test had seemingly unconstrained inputs

This is a real requirement from the implementation, not just the unit test -- we don't reduce after the basemul, and the outputs do get very close to the 16-bit boundary (from a first look that also seems to hold true for the RV64 basemul). So the iNTT really needs to reduce pretty soon 

> One could of course derive more lax bound assumptions for INTT inputs via interval arithmetic examination of ML-KEM.

There's hope https://github.com/pq-code-package/mlkem-native/pull/906 that one can still do one layer of iNTT, and hence reduce only half of the coefficients, but even achieving that is difficult. There's certainly not much that can be done without just introducing the reductions elsewhere.

hanno-becker:
@mjosaarinen Whops, apologies for the accidental close.

mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,771 @@
...
+
+  vint16m1_t t0, t1, c0, vt;
+
+  /*    initial permute */
+  vp = __riscv_vrgatherei16_vv_i16m2(vp, v2p0, vl2);
+  t0 = __riscv_vget_v_i16m2_i16m1(vp, 0);
+  t1 = __riscv_vget_v_i16m2_i16m1(vp, 1);
+
+  /*	initial reduction due to lack of input assumptions on INTT */
+  t0 = fq_mul_vx(t0, MLK_RVV_MONT_R1, vl);
```
hanno-becker: We should use the final normalization via `MLK_RVV_MONT_NR` instead of a dedicated reduction.

mlkem/src/native/riscv64/src/rv64v_poly.c
hanno-becker: @mjosaarinen Ping. This is the only blocker I see at the moment to progressing the PR.

hanno-becker (changes requested):
We're close I think 👍 

* The initial reduction can be replaced by the final normalization, as in the AArch64 and x86_64 backends.
* I still don't understand the conditional subtraction in the invNTT -- that's the only functional blocker at the moment.

@mjosaarinen We should extend `src/sys.h` with the necessary system capability and guard the backend accordingly -- see e.g. how it's [done for AVX2](https://github.com/pq-code-package/mlkem-native/blob/main/mlkem/src/native/x86_64/meta.h#L71). Otherwise, one cannot target hosts with heterogeneous capabilities with a single build (e.g. RV64 CPU with/without vector extension, or with vector extension but different length). Probably we want a capability for "Vector Extension of VL=256". What's the correct name here? Once we have that, we should extend CI to test the library on a CPU a) without vector extension, b) with vector extension, but different VL -- in both cases, it should just fall back to the C code, of course.

hanno-becker:
@mjosaarinen Regarding https://github.com/pq-code-package/mlkem-native/pull/1037/commits/e15004e6c668405327a85be498069d08beb0f203: Is there anything fundamentally different about the instructions available in RV64 that prevents the 'normal' signed arithmetic to be used throughout the invNTT? I don't understand the need to work with unsigned coefficients yet.

mjosaarinen:
> @mjosaarinen Regarding [e15004e](https://github.com/pq-code-package/mlkem-native/commit/e15004e6c668405327a85be498069d08beb0f203): Is there anything fundamentally different about the instructions available in RV64 that prevents the 'normal' signed arithmetic to be used throughout the invNTT? I don't understand the need to work with unsigned coefficients yet.

Well, the invNTT takes in signed values but internally keeps them in the canonical range [0,q-1] throughout (I don't think FIPS 203 requires signed integers anywhere) This could be seen as a precaution, but it is not so bad for performance: Since there is now an initial reduction, we can remove the final normalization which is relatively expensive: https://github.com/pq-code-package/mlkem-native/pull/1037/commits/b31855db1d7c488cd78c7d6588eca7a98568f886
The RVV ISA itself is not restrictive -- in many ways, it has almost too much choice; the core modular multiplication can be done in a couple of different ways (e.g., widening 16->32 bit or non-widening), the best of which may depend on micro-architecture. Due to a lack of concrete hardware benchmarking targets, it is difficult to say which one will be best.

hanno-becker:
@mjosaarinen Thanks, yes you are right this is a valid strategy.

On the other hand, we don't require normalized outputs and always normalised only as part the serialisation at the end (nowhere else). This works based on a moderate output bound of 8q. So we could just get rid of any final normalisation in the invNTT and it would be fine, assuming a selected set of reductions at a few coefficients in the middle of the invNTT -- but I would strongly suspect those to be much less expensive than a cadd+csub in every butterfly.

Would you mind trying this? Again, not saying your approach is wrong, but we don't really gain anything from it because the rest of the code does not need stronger output bounds.

mjosaarinen:
> @mjosaarinen Thanks, yes you are right this is a valid strategy.
> 
> On the other hand, we don't require normalized outputs and always normalised only as part the serialisation at the end (nowhere else). This works based on a moderate output bound of 8q. So we could just get rid of any final normalisation in the invNTT and it would be fine, assuming a selected set of reductions at a few coefficients in the middle of the invNTT -- but I would strongly suspect those to be much less expensive than a cadd+csub in every butterfly.
> 
> Would you mind trying this? Again, not saying your approach is wrong, but we don't really gain anything from it because the rest of the code does not need stronger output bounds.

Well, I tried little bit of that (and the code can pass the tests with some of those modifications) but I realized that I was just guessing again what those "selected set of reductions" could be.  Let's leave that for later, ok?

You know, an appropriate thing would be to work out the actual bounds systematically -- for every step and (proven-correct) arithmetic "gadget" -- and always have the proven output properties as input assumptions further down the line.

Anyway, at present, this is not a particularly optimized implementation -- there are some important things to address in relation to performance, like Keccak.

rod-chapman:
@mjosaarinen Hanno and I have done detailed analysis of the various reduction strategies for the INTT. See the "fastntt3" branch, for example, which does INTT (in C) using a "1,1,2,3" layer merging strategy. It also turns out that much better reduction strategy can be achieved if we can prove that the inputs to the INTT have magnitude bounded by INT16_MAX/2. We have proved this for the C code (on the "refine_bounds") branch and AArch64 code, but need to extend that proof to cover x86_64, which will require some modification of the basemul code. Perhaps we will return to this another time?

hanno-becker:
@mjosaarinen Let's do as you say. I think one can follow the Arm backend here, but if that's so -- I will try out to be sure -- it can be a follow-up.

Similarly, you mentioned you aim for a VL agnostic version. Can this be a follow up too?

hanno-becker:
> You know, an appropriate thing would be to work out the actual bounds systematically -- for every step and (proven-correct) arithmetic "gadget" -- and always have the proven output properties as input assumptions further down the line.

Yes! For the record, this is exactly what the HOL-Light proofs for Arm are doing. You will see tight bounds in the HOL-Light specs that would be hard to verify on paper and impossible by eye. However, the frontend never relies on those bounds. The bounds required by the frontend can so far be audited by eye. This means, we go as far into lazy reduction as can easily be annotated and followed in the code, but not further. That may leave some unnecessary reduction in the code, but at least for the invNTT the returns of more and more lazy reduction diminish quickly.

mjosaarinen:
> @mjosaarinen Let's do as you say. I think one can follow the Arm backend here, but if that's so -- I will try out to be sure -- it can be a follow-up.
> 
> Similarly, you mentioned you aim for a VL agnostic version. Can this be a follow up too?

I think it can be a follow-up - there is always a fallback to the C back-end.

The thing about that is; while some parts can be easily made VLEN-agnostic using the "strip-mining" approach (discussed in the RISC-V manual), it seems to me that for efficient NTT/INTT, one would need to have separate versions for VLEN in { 128, 256, 512, 1024, perhaps others } due to the actual register allocation. 

btw.. In RVV, the vector length VL is different fom VLEN -- VL holds the dynamic vector length (in elements, like bytes or floats) while VLEN is the hardware size (in bits) of each of the 32 vector registers. Vectors (VL) do not have to be a power-of-two in size and can span up to 8 vector registers (LMUL). The max hardware register size allowed by the spec is VLEN=65536 bits. Due to LMUL one can technically do a *lot* of stuff in a single instruction (operate on 8*256 = 2048 bits even with VLEN=256) so that's an another thing to explore. This code uses only LMUL=1 and LMUL=2.

hanno-becker:
@mjosaarinen Very interesting, thank you. I should say "VLEN" then, not "VL".

What else needs doing then for this PR, from your perspective? I will aim to go through everything again, probably later or tomorrow.

Do you have input on my note above on runtime detection of vector extension support?
> We should extend src/sys.h with the necessary system capability and guard the backend accordingly -- see e.g. how it's [done for AVX2](https://github.com/pq-code-package/mlkem-native/blob/main/mlkem/src/native/x86_64/meta.h#L71). Otherwise, one cannot target hosts with heterogeneous capabilities with a single build (e.g. RV64 CPU with/without vector extension, or with vector extension but different length). Probably we want a capability for "Vector Extension of VL=256". What's the correct name here? Once we have that, we should extend CI to test the library on a CPU a) without vector extension, b) with vector extension, but different VL -- in both cases, it should just fall back to the C code, of course. I can do that part, though.

hanno-becker:
0-diff rebase to remove merge commits.

hanno-becker:
Did some history cleanup, and re-added bounds assertions aiming to check that coefficients stay unsigned-canonical.

mjosaarinen:
> Did some history cleanup, and re-added bounds assertions aiming to check that coefficients stay unsigned-canonical.

It no longer even compiles (this file is needed.)

```
  CC      test/build/mlkem1024/unit/mlkem/src/native/riscv64/src/rv64v_poly.c.o
mlkem/src/native/riscv64/src/rv64v_poly.c:15:10: fatal error: rv64v_settings.h: No such file or directory
   15 | #include "rv64v_settings.h"
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
```

mjosaarinen:
> @mjosaarinen Very interesting, thank you. I should say "VLEN" then, not "VL".
> 
> What else needs doing then for this PR, from your perspective? I will aim to go through everything again, probably later or tomorrow.

Well, I was thinking of running some benchmarks on real RVV hardware (SpecemiT X60) on Monday-Tuesday and perhaps tweaking some things a bit based on results.

> Do you have input on my note above on runtime detection of vector extension support?
> 
> > We should extend src/sys.h with the necessary system capability and guard the backend accordingly -- see e.g. how it's [done for AVX2](https://github.com/pq-code-package/mlkem-native/blob/main/mlkem/src/native/x86_64/meta.h#L71). Otherwise, one cannot target hosts with heterogeneous capabilities with a single build (e.g. RV64 CPU with/without vector extension, or with vector extension but different length). Probably we want a capability for "Vector Extension of VL=256". What's the correct name here? Once we have that, we should extend CI to test the library on a CPU a) without vector extension, b) with vector extension, but different VL -- in both cases, it should just fall back to the C code, of course. I can do that part, though.

From a practical viewpoint, it should be ok to have the existence of RVV (yes/no) as a build-time flag. After all, I understand that the Android and Ubuntu kernel & userland will be built for an RVA23-like processor profile and wouldn't work without Vector. Now, the physical vector length (VLEN) is another matter; it is intended to be dynamic and is supposed to be handled by the programs directly. RVV code can do that with 1 instruction (it's designed to be fast) and without syscalls etc.

hanno-becker:
> From a practical viewpoint, it should be ok to have the existence of RVV (yes/no) as a build-time flag. After all, I understand that the Android and Ubuntu kernel & userland will be built for an RVA23-like processor profile and wouldn't work without Vector. Now, the physical vector length (VLEN) is another matter; it is intended to be dynamic and is supposed to be handled by the programs directly. RVV code can do that with 1 instruction (it's designed to be fast) and without syscalls etc.

If that's the case, the PR needs further work to make the dispatch to the backend dynamic -- at the moment, it is static. If I understand you correctly, at compile-time we should not have `MLK_SYS_RISCV64_V256` (currently in `sys.h`), but merely `MLK_SYS_RISCV64_RVV` indicating compile-time support for RVV. And the backend, one would use the "VLEN-instruction" you mentioned to dynamically dispatch to C or the backend code. Once that's done, the CI should be extended to exercise multiple VLENs (cf. `multi-functest/action.yml`). 

Can you please make the corresponding adjustments?

Is there a similarly convenient way to infer without syscalls whether RVV itself is present?

hanno-becker:
Current SpacemiT K1 8 (Banana Pi F3) benchmarks 👍 

| Benchmark suite | Current: 3fa9c6036d2339f012523f762316a2738c6354fc | Previous: 5aaca94bd516c0a9f4d212948cd30e873d4a2956 | Ratio |
|-|-|-|-|
| `ML-KEM-512 keypair` | `159073` cycles | `227126` cycles | `0.70` |
| `ML-KEM-512 encaps` | `169825` cycles | `271033` cycles | `0.63` |
| `ML-KEM-512 decaps` | `214185` cycles | `345293` cycles | `0.62` |
| `ML-KEM-768 keypair` | `266886` cycles | `376324` cycles | `0.71` |
| `ML-KEM-768 encaps` | `285341` cycles | `433349` cycles | `0.66` |
| `ML-KEM-768 decaps` | `348863` cycles | `529898` cycles | `0.66` |
| `ML-KEM-1024 keypair` | `403231` cycles | `560159` cycles | `0.72` |
| `ML-KEM-1024 encaps` | `433854` cycles | `636793` cycles | `0.68` |
| `ML-KEM-1024 decaps` | `516348` cycles | `758317` cycles | `0.68` |

mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,770 @@
...
+  const vuint16m1_t sll12v = __riscv_vsll_vx_u16m1(vid, 2, vl);
+
+  size_t n, ctr, pos;
+  vuint16m1_t x, y;
+  vbool16_t lt;
+
+  pos = 0;
+  ctr = 0;
+
+  while (ctr < len && pos + (vl23 * 2) <= buflen)
```
hanno-becker: If `buflen` is not a multiple of `vl23 * 2`, then this function may leave bytes unprocessed in the input buffer, which would be incorrect. We should reject the input upfront if that's the case (thereby falling back to C), or add tail-handling in support of this case. As far as I understand the code, it should be simple to basically handle the tail in the same way as the bulk of the loop, adapting `vl23`.

This starts to be an issue with `VLEN>=512`, where `vl23 = 48 = 3 * 16` which is not a multiple of the concrete buflens `MLK_XOF_RATE` and `MLK_XOF_RATE * MLKEM_GEN_MATRIX_NBLOCKS` this function is called with.

hanno-becker:
@mjosaarinen At the moment, we compile with "CFLAGS += -march=rv64gcv_zvl256b". Is this still the right option if we want to compile for RVV 1.0 but without fixing the vector length? Or does this merely signify a minimal vector length?

EDIT: Looks like we want just `-march=rv64gcv`. See https://github.com/pq-code-package/mlkem-native/commit/1f2fce3ee64d558d0e46ae6268c5d2820930002d

Btw, leaving aside https://github.com/pq-code-package/mlkem-native/pull/1037#discussion_r2425198727, it's pretty cool to see the length-agnostic implementation of `rej_uniform` :+1:

mlkem/src/native/riscv64/src/rv64v_poly.c
hanno-becker: The issue surfaces in the KATs as soon as one tests with VLEN=512.

hanno-becker:
```c
  while (ctr < len && pos < buflen)
  {
    const size_t vl = __riscv_vsetvl_e16m1((buflen - pos) * 8 / 12);
    const size_t vl23 = (vl * 24) / 32;

    const vuint16m1_t vid = __riscv_vid_v_u16m1(vl);
    const vuint16m1_t srl12v = __riscv_vmul_vx_u16m1(vid, 12, vl);
    const vuint16m1_t sel12v = __riscv_vsrl_vx_u16m1(srl12v, 4, vl);
    const vuint16m1_t sll12v = __riscv_vsll_vx_u16m1(vid, 2, vl);

   ...
```

seems to work. Experimented a bit with the other functions, too, and it looks like https://github.com/pq-code-package/mlkem-native/commit/1f2fce3ee64d558d0e46ae6268c5d2820930002d passes QEMU tests for all supported VLENs (128,256,512,1024), skipping [inv]NTT unless VLEN=256 for now.

~It's not yet clear to me why the [inv]NTT doesn't work for VLEN > 256 even if in the code one forces `vl = 16`. Should that behave as if VLEN=256?~ understood

mjosaarinen:
So, I did some further benchmarks. It's well-known that LLVM has a really good autovectorizer for RISC-V (with large contributions from Google, I think. BoringSSL's MLKEM is not too slow on RISC-V because of that.)

**gcc**: riscv64-unknown-linux-gnu-gcc (GCC) 14.2.1 20250322
**clang**: clang version 22.0.0git (I built this from main on Sep 8, 2025)
**Intrin**: ml-kem native intrinsics enabled (-O3 -march=rv64gcv)
**Autovec**: plain C, just autovectorization  (-O3 -march=rv64gcv)
**No vec**: plain c, no vector (-O3 -march=rv64gc)

The gcc intrin vs autovec numbers are within 1 or 2% of those previous benchmarks, probably down to measurement method (my runs were probably shorter).

|               |           |Intrin |Intrin |Autovec|Autovec|No vec |No vec |
|---------------|-----------|-------|-------|-------|-------|-------|-------|
|Algorithm      |Function   |clang  |gcc    |clang  |gcc    |clang  |gcc    |
|ML-KEM-512     |Keypair    |141480 |160647 |186322 |226757 |313511 |311440 |
|ML-KEM-512     |Encaps     |152256 |170002 |203173 |273024 |414068 |422153 |
|ML-KEM-512     |Decaps     |189825 |215589 |240965 |347203 |549109 |566565 |
|ML-KEM-768     |Keypair    |240108 |271747 |313054 |376235 |511832 |510837 |
|ML-KEM-768     |Encaps     |257651 |287527 |335101 |436653 |639801 |649390 |
|ML-KEM-768     |Decaps     |311529 |353817 |384829 |535102 |810516 |833938 |
|ML-KEM-1024    |Keypair    |363202 |406382 |471771 |557870 |744641 |736261 |
|ML-KEM-1024    |Encaps     |393483 |434324 |505658 |636611 |904048 |908036 |
|ML-KEM-1024    |Decaps     |464661 |518650 |567039 |761398 |1109203|1127421|

Some observations:
- LLVM 22git autovectorization is 29% faster than GCC 14.2.1 autovectorization.
- Intrinsics give a 28% (clang) or 48% (gcc) speedup over autovectorization.
- Using intrinsics + autovectorization means 2.42x (clang) or 2.18x (gcc) speedup over not using vector instructions.
- With plain rv64gc ISA, the two compilers are roughly the same -- clang often slower!

hanno-becker:
Thanks for the analysis @mjosaarinen! Looks like the move to intrinsics still brings a notably benefit over autovectorization.

Note that we omit `poly_add/sub` from the backend API because of the assumption that autovectorizers should have an easy time with them. `mlk_rv64v_poly_add` and `mlk_rv64v_poly_sub` can thus be removed from this PR.

rod-chapman:
@mjosaarinen Thanks for those data. Could you re-run the "Autovec" set with clang 22 on the "fastntt3" branch please? I'm intrigued to see how clang does on that code, which was specifically written to be amenable to auto-vectorization. Results using GCC 14 on AArch64 were surprisingly good.

mjosaarinen:
> @mjosaarinen Thanks for those data. Could you re-run the "Autovec" set with clang 22 on the "fastntt3" branch please? I'm intrigued to see how clang does on that code, which was specifically written to be amenable to auto-vectorization. Results using GCC 14 on AArch64 were surprisingly good.

Unfortunately I already packed away my own SpacemiT X60 board as I am traveling tomorrow. (But I think there is one plugged into the CI somehow, perhaps one could try that one.)

hanno-becker:
@mjosaarinen I took the liberty to make a few changes:
* Adjusted the backend so it's functional for all VLENs. For NTT and invNTT, VLENs != 256 fall back to C. For all other backend functions, I made the code VLEN agnostic, but please check that I did it correctly.
* Removed an unnecessary reduction at the end of the forward NTT; the frontend does not require the output to be reduced.
* Removed unused `poly_add/sub` implementations.
* Removed a carry-in in the Montgomery multiplication, switching to high-sub instead of high-add
* Extended the CI to test RV64 a) without RVV, b) VLEN={128, 256, 512, 1024} (QEMU does not support higher VLEN)
* I was too curious not to try out switching to lazy reduction in the invNTT. I followed the same approach as in the other backends: Implement only as much lazy reduction that it's still easy to follow in your head what's going on. Please check what I've done and whether you find the comments comprehensible. The benefit of this approach is a significantly faster invNTT, which on the high-level manifests itself in a little performance bump of 1-3% (right now, the not-yet-optimized Keccak obviously outweighs the remaining cost).
* Adjusted the README to no longer warn about the experimental state of the implementation. Are you OK with that? Please also consider whether you want to name yourself as the author of the backend -- you're not giving yourself any credit at the moment, as far as I can see.

Could you also re-benchmark the code on your board, when you have time?

I also experimented using Barret multiplication rather than Montgomery multiplication, which saves some multiplications and uses low-MLA rather than high-MLA; however, I found it to be detrimental at least on the BPi, so I dropped it.

Awaiting CI now to see if I did something dumb somewhere, but otherwise I think we're good to go. Of course, it would be nice to have an [inv]NTT for other VLENs as well, but that can be a follow-up.

hanno-becker:
@mkannwischer Could you please take a look as well?

hanno-becker:
@mjosaarinen Is there a portable way to determine at _runtime_ whether RVV is implemented?

hanno-becker (approved):
From my perspective this is good to go.

mjosaarinen:
> @mjosaarinen I took the liberty to make a few changes:
(..)



Thanks Hanno, great work. Apologies for being so preoccupied with other stuff..

One thing that would greatly enhance performance is the use of vector bitmanip (Zvbb extension -- not part of RVV 1.0, but already on some silicon) as it provides vectorized rotate and ANDN, making it very important overall. Zvbb is part of the standard vector crypto Zvk, which is expected to be common (likely required for Android).

While detection of VLEN is straightforward (and something binaries are expected to do), I don't have a definite answer on the best way to do autodetection of vector features, but I can find out from authoritative sources.

ps. Slides for the talk at RISC-V Summit next week ( https://sched.co/28OTv  my slides  https://mjos.fi/doc/20251023-rvsna-pqcp-keccak.pdf ). It's only a 15-minute slot and they require an early slide submission + use of the conference PowerPoint template. I hope we can find additional good RISC-V contributors through the meeting.

hanno-becker:
Thank you @mjosaarinen for taking a look. Are we good to go from your perspective, then, or do you need more time to look at the changes?

> One thing that would greatly enhance performance is the use of vector bitmanip (Zvbb extension -- not part of RVV 1.0, but already on some silicon) as it provides vectorized rotate and ANDN, making it very important overall. Zvbb is part of the standard vector crypto Zvk, which is expected to be common (likely required for Android).

How far does one get trying to implement Keccak based on RVV 1.0 alone?

> While detection of VLEN is straightforward (and something binaries are expected to do), I don't have a definite answer on the best way to do autodetection of vector features, but I can find out from authoritative sources.

Thank you! At the moment, the default build assumes presence of RVV, similar to AVX2 being assumed by default on x86_64 builds. A build which is meant to target RVV and non-RVV systems would require a capability check (see `mlkem/src/sys.h`), probably platform dependent and nothing we would want to maintain here _unless_ it's as simple as the VLEN-detection instruction.

hanno-becker:
There are a few low-hanging fruits remaining here:
* Put zetas in the order needed by the [inv]NTT, thereby avoiding zeta shuffling at runtime
* Use a mulcache. This should be completely straightforward but simplify and accelerate the basemul.

The fact that the C implementation uses the mulcache (which is likely auto-vectorized) but the intrinsics implementations doesn't, might eat away from the performance delta between intrinsics and autovectorization. It would be good to get this implemented.

hanno-becker:
> Use a mulcache. This should be completely straightforward but simplify and accelerate the basemul.

I was wrong. I tried out what seemed like the straightforward mulcache implementation to me:
```c
void mlk_rv64v_poly_mulcache_compute(int16_t x[MLKEM_N / 2],
                                     const int16_t y[MLKEM_N])
{
#include "rv64v_zetas_basemul.inc"

  size_t vl = __riscv_vsetvl_e16m1(MLKEM_N) / 2;
  size_t i;

  for (i = 0; i < MLKEM_N; i += 2 * vl)
  {
    vint16m1_t b1 = 
        __riscv_vget_v_i16m1x2_i16m1(__riscv_vlseg2e16_v_i16m1x2(&y[i], vl), 1);
    vint16m1_t z = __riscv_vle16_v_i16m1(&roots[i / 2], vl);
    vint16m1_t bt = fq_mul_vv(b1, z, vl);
    __riscv_vse16_v_i16m1(&x[i / 2], bt, vl);
  }
}

static inline void mlk_rv64v_poly_basemul_mont_add_k(int16_t *r,
                                                     const int16_t *a,
                                                     const int16_t *b,
                                                     const int16_t *bc,
                                                     unsigned kn)
{
  size_t vl = __riscv_vsetvl_e16m1(MLKEM_N) / 2;
  size_t i, j;

  for (i = 0; i < MLKEM_N; i += 2 * vl)
  {
    vint32m2_t acc0 = __riscv_vmv_v_x_i32m2(0, vl);
    vint32m2_t acc1 = __riscv_vmv_v_x_i32m2(0, vl);

    for (j = 0; j < kn; j += MLKEM_N)
    {
      vint16m1x2_t x = __riscv_vlseg2e16_v_i16m1x2(&a[i + j], vl);
      vint16m1x2_t y = __riscv_vlseg2e16_v_i16m1x2(&b[i + j], vl);

      vint16m1_t x0 = __riscv_vget_v_i16m1x2_i16m1(x, 0);
      vint16m1_t x1 = __riscv_vget_v_i16m1x2_i16m1(x, 1);
      vint16m1_t y0 = __riscv_vget_v_i16m1x2_i16m1(y, 0);
      vint16m1_t y1 = __riscv_vget_v_i16m1x2_i16m1(y, 1);
      vint16m1_t yt = __riscv_vle16_v_i16m1(&bc[(i + j) / 2], vl);

      vint32m2_t x0y0 = __riscv_vwmul_vv_i32m2(x0, y0, vl);
      vint32m2_t x0y1 = __riscv_vwmul_vv_i32m2(x0, y1, vl);
      vint32m2_t x1y0 = __riscv_vwmul_vv_i32m2(x1, y0, vl);
      vint32m2_t x1yt = __riscv_vwmul_vv_i32m2(x1, yt, vl);

      acc0 = __riscv_vadd_vv_i32m2(acc0, x0y0, vl);
      acc0 = __riscv_vadd_vv_i32m2(acc0, x1yt, vl);
      acc1 = __riscv_vadd_vv_i32m2(acc1, x0y1, vl);
      acc1 = __riscv_vadd_vv_i32m2(acc1, x1y0, vl);
    }

    __riscv_vsseg2e16_v_i16m1x2(
        &r[i],
        __riscv_vcreate_v_i16m1x2(fq_redc2(acc0, vl), fq_redc2(acc1, vl)), vl);
  }
}
```
But this is notably slower. Interestingly, even ignoring the mulcache computation itself (which is almost twice as slow as `tomont`, oddly), the basemul is slower than @mjosaarinen's, despite the fewer multiplications and additions.

@mjosaarinen What can I learn from this? My guess is that `vlseg2` and `vsseg2` are very slow operations, but that would be a bit surprising because it's ultimately a combination of gather and load/store, and in your code there's heavy shuffling as well. Anyway, any intuition you can share about what works, and what doesn't, is appreciated.

mlkem/mlkem_native.S
```diff
@@ -85,6 +85,8 @@
 #include "mlkem/src/native/x86_64/src/rej_uniform_asm.S"
 #include "mlkem/src/native/x86_64/src/tomont.S"
 #endif /* MLK_SYS_X86_64 */
+#if defined(MLK_SYS_RISCV64)
+#endif
```
mkannwischer, line 89: This seems unnecessary.

mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -0,0 +1,799 @@
...
+                                        const int16_t *b)
+{
+  mlk_rv64v_poly_basemul_mont_add_k(r, a, b, 3 * MLKEM_N);
+}
+
+void mlk_rv64v_poly_basemul_mont_add_k4(int16_t *r, const int16_t *a,
+                                        const int16_t *b)
+{
+  mlk_rv64v_poly_basemul_mont_add_k(r, a, b, 4 * MLKEM_N);
+}
```
mkannwischer, line 692: Should this be guarded?

mlkem/src/native/riscv64/src/rv64v_debug.c
mkannwischer, line 1: Do we really need to keep this debug code?

mkannwischer (approved):
Three minor comments/questions, but none of them is blocking.
Happy to go ahead and merge this PR and revisit it in a follow-up if needed. 

Thank you @mjosaarinen for contributing and thank you @hanno-becker for all the work you have put into this!

mlkem/mlkem_native.S
hanno-becker: This is just autogenerated; we can adjust `autogen` at some point, but not here.

mlkem/src/native/riscv64/src/rv64v_debug.c
hanno-becker: It does no harm does it? I did my best to make it visually as appealing as possible and I do think it's useful to double check bounds in debug builds.

mlkem/src/native/riscv64/src/rv64v_poly.c
hanno-becker: Yes, good catch.

hanno-becker:
Many thanks @mjosaarinen, this is a great addition.

When do you think you'll get to writing a Keccak backend, and adding [inv]NTTs for other VLENs?

PR #1039, opened 2025-05-21
CBMC: Don't declare structures in harnesses

hanno-becker:
When CBMC checks a function contract, it seems that the preconditions set out in the contract are overlayed by the context set out in the harness. For example, if a function takes a pointer to `foo`, and the harness passes `&f` for `foo f`, then the function would never be evaluated for the possibility of an invalid pointer being passed. For that reason, harnesses must be as minimal as possible, merely declaring uninitialized variables of the expected type and calling the function under proof.

This commit fixes a few instances in mlkem-native where this was not yet the case.


mkannwischer (approved):
Thanks @hanno-becker for fixing this.
Note that this was uncovered during proof development in mldsa-native: https://github.com/pq-code-package/mldsa-native/pull/254

PR #1041, opened 2025-05-23
Update to nixpkgs 25.05

### LINKED ISSUE #1022: Consider updating z3 to 4.15.0
https://github.com/Z3Prover/z3/releases/tag/z3-4.15.0

mkannwischer:
- Resolves #1022

[NixOS 25.05 ](https://nixos.org/blog/announcements/2025/nixos-2505/) has been released yesterday bringing various version updates also to nixpkgs.

Notable changes for us are: 
 - Default clang/llvm is now 19; we switch to the default in the core shells
 - Default gcc is now 14; we switch to it in the core shells
 - gcc 7, zig_0_10, zig_0_11 are not longer supported in 25.05; we instead
   take it from nixpkgs 24.05
 - zig_0_14 has been added; we test in CI now as well
 - Unicorn and protobuf versions are now compatible with SLOTHY simpliying the
   build
 - z3 has been updated to 4.15.0; we take it from 25.05
 - clang_20, bitwuzla, qemu now have suitable versions in 25.05; we no longer
   need to take them from unstable.
 - The upstream CBMC nixpkgs build (6.4.1) uses 3 patches instead of 2 used     
   before. Since we are using 6.6.0, we do not require the third patch.         
   We, hence, have to overwrite the patches here locally and use on the the     
   first two. This slighty complicates our flake. 
 - A few dependency updates for which we do not require specific versions

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -5,7 +5,7 @@
 
   inputs = {
     nixpkgs-2405.url = "github:NixOS/nixpkgs/nixos-24.05";
-    nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.11";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-25.05";
     nixpkgs-unstable.url = "github:NixOS/nixpkgs/nixos-unstable";
 
     flake-parts = {
@@ -23,9 +23,10 @@
           pkgs-unstable = inputs.nixpkgs-unstable.legacyPackages.${system};
           pkgs-2405 = inputs.nixpkgs-2405.legacyPackages.${system};
           util = pkgs.callPackage ./nix/util.nix {
+            # Keep those around in case we want to switch to unstable versions
             cbmc = pkgs.cbmc;
-            bitwuzla = pkgs-unstable.bitwuzla;
-            z3 = pkgs-unstable.z3_4_14;
+            bitwuzla = pkgs.bitwuzla;
+            z3 = pkgs.z3;
           };
           zigWrapCC = zig: pkgs.symlinkJoin {
             name = "zig-wrappers";
@@ -48,8 +49,9 @@
               (_:_: {
                 gcc48 = pkgs-2405.gcc48;
                 gcc49 = pkgs-2405.gcc49;
-                qemu = pkgs-unstable.qemu; # 9.2.2
-                clang_20 = pkgs-unstable.clang_20;
+                gcc7 = pkgs-2405.gcc7;
+                zig_0_10 = pkgs-2405.zig_0_10;
+                zig_0_11 = pkgs-2405.zig_0_11;
               })
             ];
           };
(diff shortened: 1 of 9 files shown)


mkannwischer:
This is ready for review. @potsrevennil, could you take a look please?
I'd welcome other suggestions regarding the git workaround. I expected this to be fixed with this release, but it isn't. 

Before we merge this (or run benchmarks), we need to wait for the Raspberry Pi's to recompile the cross compilers. I just started that, but it may take 24 hours or more.

flake.nix
```diff
@@ -109,6 +109,7 @@
           devShells.ci_zig0_11 = util.mkShellWithCC' (zigWrapCC pkgs.zig_0_11);
           devShells.ci_zig0_12 = util.mkShellWithCC' (zigWrapCC pkgs.zig_0_12);
           devShells.ci_zig0_13 = util.mkShellWithCC' (zigWrapCC pkgs.zig_0_13);
+          devShells.ci_zig0_14 = util.mkShellWithCC' (zigWrapCC pkgs.zig);
```
hanno-becker, line 112: Can one not make the version explicit here?

mkannwischer: No, there is no zig_0_14: https://search.nixos.org/packages?channel=25.05&from=0&size=50&sort=relevance&type=packages&query=zig

potsrevennil:
```
            overlays = [
              (_:_: {
                ...
                zig_0_14 = pkgs.zig;
              })
```
we could try to do something like this to create an alias, to be able to make the version expicit afterwards, if we want

mkannwischer: I don't think that solves the problem?
The only problem of using `zig` is that in a future channel update the version may change without us noticing it. 
Unless I am misunderstanding, your suggestion is even worse as it is making it appear as this is zig 0.14 even though it may not be in the future.

potsrevennil: I think what you worried can be resolved by specifying the version with `override` (similar to what we've done with CBMC)
`zig_0_14_1 = pkgs.zig.override {version = "0.14.1"; };`
nix should be smart enough to find out this evaluate to the same canonical form, and won't try to recompile zig

However I found out that the name "zig_0_14" seemed to be special in nix, which will result in infinite recursion when do this in overlays. But trying with other names like "foo", "bar", "zig_0_14_0"... etc, are fine.

flake.nix
```diff
@@ -142,8 +143,8 @@
             util = pkgs.callPackage ./nix/util.nix {
               inherit pkgs;
               cbmc = pkgs.cbmc;
-              bitwuzla = pkgs-unstable.bitwuzla;
-              z3 = pkgs-unstable.z3_4_14;
+              bitwuzla = pkgs.bitwuzla;
```
potsrevennil, line 146: I think this line can also be removed as in the `util` specified inside `flake-parts`, right ?

mkannwischer: keeping them here in case we need to switch to unstable versions later

flake.nix
```diff
@@ -24,8 +24,7 @@
           pkgs-2405 = inputs.nixpkgs-2405.legacyPackages.${system};
           util = pkgs.callPackage ./nix/util.nix {
             cbmc = pkgs.cbmc;
-            bitwuzla = pkgs-unstable.bitwuzla;
```
potsrevennil, line 27: Then I guess this should also stay, for consistency

potsrevennil (approved):

hanno-becker (approved):
Thanks a lot @mkannwischer @potsrevennil !

PR #1043, opened 2025-05-24
Nix: Simplify SLOTHY flake (use ortools from nixpkgs)

mkannwischer:
I experimented more with using the ortools packaged in nixpkgs rather than taking it from pypi. I do not see any performance difference after running more experiments.
I think we can safely switch, greatly simplifying the flake.


hanno-becker (approved):
That's much simpler indeed, thanks @mkannwischer !

PR #1059, opened 2025-05-30
Rename `CBMC` macro to `MLK_CBMC`

hanno-becker:


hanno-becker:
It seems that `-DCBMC=1` is also

PR #1061, opened 2025-05-30
Restructure monobuild

mkannwischer:
Moves `mlkem/*` to `mlkem/src/*`, moves SCU files to `mlkem/*`, moves `mlkem_native.h` to `mlkem/*`.


hanno-becker (commented):
I'm in two minds about this. On the one hand, I think it makes sense to (a) have the mlkem-native header separate from the source, (b) have the SCU file -- which is used in multiple examples -- not live hidden in one of the examples files.

On the other hand, just exposing an SCU .c and .h file is a bit confusing, too, because it only allows to SCU-build mlkem-native without native code. If you want a native backend, you still need to manually dig out the *.S files from `mlkem/src` Even if it's not used in AWS-LC, it would seem more consequential to also offer an SCU .S file -- but as the latest issues and PRs show, there's a bit more work to do for that.

hanno-becker (approved):
Thanks for kicking this off @mkannwischer !

PR #1065, opened 2025-05-30
Remove unnecessary MLK_CONFIG_MONOBUILD_XXX configuration options

hanno-becker:
Previously, we had a various configurations options

MLK_CONFIG_MONOBUILD_XXX

which would control what the SCU file mlkem_native_monobuild.c does.

However, most of those options are actually already covered by the mlkem-native configuration, which the SCU file must have access to anyway. This is different from the configuration options for mlkem_native.h, which may be used in a context where the original sources, including the full configuration, are not present -- in this case, separate configuration options are required.

For the SCU file, however, the following monobuild configuration options are subsumed by configuration options, and can just be suitably renamed:

- MLK_CONFIG_MONOBUILD_WITH_NATIVE_ARITH -> MLK_CONFIG_USE_NATIVE_BACKEND_ARITH

- MLK_CONFIG_MONOBUILD_WITH_NATIVE_FIPS202 -> MLK_CONFIG_USE_NATIVE_BACKEND_FIPS202

- MLK_CONFIG_MONOBUILD_CUSTOM_FIPS202 -> defined(MLK_CONFIG_FIPS202_CUSTOM_HEADER)

### DIFF
--- a/examples/monolithic_build/mlkem_native_monobuild.c
+++ b/examples/monolithic_build/mlkem_native_monobuild.c
@@ -33,14 +33,16 @@
  *
  * # Configuration
  *
- * - MLK_CONFIG_MONOBUILD_CUSTOM_FIPS202
+ * The following options from the mlkem-native configuration are relevant:
+ *
+ * - MLK_CONFIG_FIPS202_CUSTOM_HEADER
  *   Set this option if you use a custom FIPS202 implementation.
  *
- * - MLK_CONFIG_MONOBUILD_WITH_NATIVE_ARITH
+ * - MLK_CONFIG_USE_NATIVE_BACKEND_ARITH
  *   Set this option if you want to include the native arithmetic backends
  *   in your build.
  *
- * - MLK_CONFIG_MONOBUILD_WITH_NATIVE_FIPS202
+ * - MLK_CONFIG_USE_NATIVE_BACKEND_FIPS202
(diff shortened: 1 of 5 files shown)


examples/monolithic_build/mlkem_native_monobuild.c
```diff
@@ -69,13 +71,13 @@
 #include "mlkem/sampling.c"
 #include "mlkem/verify.c"
 
-#if !defined(MLK_CONFIG_MONOBUILD_CUSTOM_FIPS202)
+#if !defined(MLK_CONFIG_FIPS202_CUSTOM_HEADER)
```
mkannwischer, line 74: can we somewhere add an error if `MLK_CONFIG_FIPS202_CUSTOM_HEADER` is set,  but `MLK_CONFIG_FIPS202X4_CUSTOM_HEADER` isn't? I ran into this when doing the pqm4 integration.

mkannwischer (approved):
LGTM. Thanks @hanno-becker. 
One comment unrelated to your PR, but maybe you can staple that on this PR.

examples/monolithic_build/mlkem_native_monobuild.c
hanno-becker: Yes, but let's to that in a follow-up.

PR #1067, opened 2025-05-31
autogen: Check for prefixing of assembly labels

### LINKED ISSUE #1066: ASM: Make labels globally unique
Assembly files cannot be concatenated if there are colliding labels. Prefix all labels in assembly files to start with some file specific prefix.

hanno-becker:
* Based on #1063 
* Resolves #1066 

autogen: Automatically prefix assembly labels with function name 

This commit extends `autogen` to go through labels in assembly files
and prefix them with the name of the function that is defined in the
file. The script can also detect and fixup minor reordering mistakes
in the prefix. Still, some manual patching had to be some for the
AVX2 assembly.


scripts/autogen
```diff
@@ -1395,6 +1395,54 @@ def check_asm_register_aliases():
...
+        content = f.read()
+
+    # Find function symbol name
+    func_pattern = r"MLK_ASM_FN_SYMBOL\((.*)\)"
+    res = re.search(func_pattern, content)
+    if res is None:
+        raise Exception(f"Could not find function symbol in assembly file {filename}")
+    funcname = res.group(1)
+    lbl_prefix = funcname.replace("_asm", "") + "_"
+    lbl_pattern = r"^(\w+):"
```
mkannwischer, line 1415: Should this be "^\s*(\w+):" ?

scripts/autogen
```diff
@@ -1395,6 +1395,54 @@ def check_asm_register_aliases():
...
+        lbl_parts = lbl.split("_")
+        lbl_stripped = lbl
+        for i, p in enumerate(lbl_parts):
+            if p not in funcname_parts:
+                lbl_stripped = "_".join(lbl_parts[i:])
+                break
+        new_lbl = lbl_prefix + lbl_stripped
+        content = re.sub(rf"(^|\s){lbl}(:?)$", rf"\1{new_lbl}\2", content, flags=re.M)
+
+    update_file(filename, content, dry_run=dry_run)
```
mkannwischer: I don't see much value in auto-renaming those. We don't ever want that to happen as the inputs should already contain the namespace. I can only see this leading to confusion when adding further backends. Can we instead change this to a fail with a reasonable error "assembly label XXX is missing function-name prefix"? Or am I missing something here?

mkannwischer (changes requested):
Thanks @hanno-becker! The renaming looks good.

Two comments that we should discuss.

scripts/autogen
hanno-becker: I can do that, yes. It was mainly for me to systematically change the existing labels, which were basically all wrong. But now that it's done, for the future it's probably enough to just error out.

scripts/autogen
hanno-becker: Ok!

scripts/autogen
hanno-becker: I'll do this as a follow-up commit, so the history retains the adapting version.

hanno-becker: Adjusted

scripts/autogen
hanno-becker: Done

scripts/autogen
```diff
@@ -1416,34 +1416,18 @@ def gen_asm_loop_labels_for_file(filename, dry_run=False):
...
-        for i, p in enumerate(lbl_parts):
-            if p not in funcname_parts:
-                lbl_stripped = "_".join(lbl_parts[i:])
-                break
-        new_lbl = lbl_prefix + lbl_stripped
-        content = re.sub(rf"(^|\s){lbl}(:?)$", rf"\1{new_lbl}\2", content, flags=re.M)
-
-    update_file(filename, content, dry_run=dry_run)
+        if not lbl.startswith(lbl_prefix):
+            raise Exeception(
```
mkannwischer: typo: Should be `Exception`

mkannwischer (changes requested):

scripts/autogen
hanno-becker: done

PR #1068, opened 2025-06-01
autogen: Add monolithic assembly file

hanno-becker:
* Builds on #1067 
* Builds on #1063

Analogous to the SCU file examples/monolithic_build/mlkem_native_monobuild.c,
this commit extends autogen to emit a monolithic mlkem_native_monobuild.S
in the monolithic_build_multilevel_native example. This demonstrates that
all ASM files are concatenable into a single assembly unit. This relies on
the prework undefining register aliases, assembly directives, and making
ASM labels unique.


hanno-becker:
@mkannwischer We should wait until the CI on `main` is finished to not abort the HOL-Light CI.

mkannwischer (approved):

PR #1080, opened 2025-06-04
ACVP: Adjust to additionally support ACVTS data format

mkannwischer:
The ACVTS data structure is very slightly different from the sample files in the usnistgov/ACVP-Server Github repository:
The prompt consists of a 2-element list, where the first element is solely consisting of {"acvVersion": "1.0"} and the second element is the usual prompt containing the test values.

This commit extends the ACVP client to automatically detect that case and handle it properly.
The result is adjusted accordingly (getting the acvVersion appended).

I also disabled the alphabetic sorting of json keys as that just leads to confusion. We are already using the same order as the NIST testvectors, so there is no need to sort alphabetically.

### DIFF
--- a/test/acvp_client.py
+++ b/test/acvp_client.py
@@ -165,17 +165,29 @@ def runTestSingle(promptName, prompt, expectedResultName, expectedResult, output
     info(f"Running ACVP tests for {promptName}")
 
     assert expectedResult is not None or output is not None
+
+    # The ACVTS data structure is very slightly different from the sample files
+    # in the usnistgov/ACVP-Server Github repository:
+    # The prompt consists of a 2-element list, where the first element is
+    # solely consisting of {"acvVersion": "1.0"} and the second element is
+    # the usual prompt containing the test values.
+    # See https://pages.nist.gov/ACVP/draft-celi-acvp-ml-kem.txt for details.
+    # We automatically detect that case here and extract the second element
+    isAcvts = False
+    if type(prompt) is list:
+        isAcvts = True
+        assert len(prompt) == 2
+        acvVersion = prompt[0]
+        assert len(acvVersion) == 1
+        prompt = prompt[1]
+
     assert prompt["algorithm"] == "ML-KEM"
     assert prompt["mode"] == "encapDecap" or prompt["mode"] == "keyGen"
 
-    results = {
-        "vsId": prompt["vsId"],
-        "algorithm": "ML-KEM",
-        "mode": prompt["mode"],
-        "revision": prompt["revision"],
-        "isSample": prompt["isSample"],
-        "testGroups": [],
-    }
+    # copy top level fields into the results
+    results = prompt.copy()
+
+    results["testGroups"] = []
     for tg in prompt["testGroups"]:
         tgResult = {
             "tgId": tg["tgId"],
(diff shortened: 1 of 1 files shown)


test/acvp_client.py
```diff
@@ -165,6 +165,20 @@ def runTestSingle(promptName, prompt, expectedResultName, expectedResult, output
     info(f"Running ACVP tests for {promptName}")
 
     assert expectedResult is not None or output is not None
+
+    # The ACVTS data structure is very slightly different from the sample files
+    # in the usnistgov/ACVP-Server Github repository:
+    # The prompt consists of a 2-element list, where the first element is
+    # solely consisting of {"acvVersion": "1.0"} and the second element is
```
hanno-becker, line 172: Do you want to assert that `acvVersion` is indeed `{ "acvVersion" : "1.0" }`?

test/acvp_client.py
```diff
@@ -189,12 +203,16 @@ def runTestSingle(promptName, prompt, expectedResultName, expectedResult, output
...
+    if isAcvts is True:
+        results = [acvVersion, results]
+
     # Compare to expected results
     if expectedResult is not None:
         info(f"Comparing results with {expectedResultName}")
-        if json.dumps(results, sort_keys=True) != json.dumps(
-            expectedResult, sort_keys=True
-        ):
+        if json.dumps(results) != json.dumps(expectedResult):
```
hanno-becker, line 215: Hm, that seems a little dangerous? Does `json.dumps` make any guarantees about how a structure will be laid out?

mkannwischer: `json.dumps` uses `dict.iter` and since Python 3.7: `Dictionary order is guaranteed to be insertion order.`
https://docs.python.org/3.7/library/stdtypes.html#typesmapping
I'd like to strictly follow the order that NIST is using.

test/acvp_client.py
```diff
@@ -165,6 +165,20 @@ def runTestSingle(promptName, prompt, expectedResultName, expectedResult, output
     info(f"Running ACVP tests for {promptName}")
 
     assert expectedResult is not None or output is not None
+
+    # The ACVTS data structure is very slightly different from the sample files
```
hanno-becker, line 169: Is there a spec for ACVTS you can refer to here?

test/acvp_client.py
hanno-becker: 👍 Can you add a comment referencing that guarantee?

test/acvp_client.py
hanno-becker: (Or, at least assert that this is a 1-element dict with key `"acVVersion"`)

test/acvp_client.py
mkannwischer: added.

test/acvp_client.py
mkannwischer: I have aded an assertion that `len(acvVersion) == 1`. 
I wouldn't be surprised if they change the acvVersion without actually changing the data format.

test/acvp_client.py
```diff
@@ -165,6 +165,22 @@ def runTestSingle(promptName, prompt, expectedResultName, expectedResult, output
...
+    # See https://pages.nist.gov/ACVP/draft-celi-acvp-ml-kem.txt for details.
+    # We automatically detect that case here and extract the second element
+    isAcvts = False
+    if type(prompt) is list:
+        isAcvts = True
+        assert len(prompt) == 2
+        acvVersion = prompt[0]
+        assert len(acvVersion) == 1
+        prompt = prompt[1]
+
```
hanno-becker, line 183: There are a few discrepancies with the spec: 
* IsSample does not exist in the spec
* The response object in the spec only contains `acvVersion`, `vsId` and `testGroups`, not the other fields.

I'm not sure how to handle this. One could just keep the toplevel dict structure and merely change `testGroups`

test/acvp_client.py
mkannwischer: done.

test/acvp_client.py
mkannwischer: Yes, this is a good idea. I implemented that.

hanno-becker (approved):
LGTM. Thanks @mkannwischer !

PR #1085, opened 2025-06-13
Upgrade SLOTHY to the HEAD of its main branch, as of 13th June 2025.

### LINKED ISSUE #1084: Nix: Update SLOTHY to experiment with Neoverse_N1 model
SLOTHY has been updated add micro-architectural models for the instructions needed by mlkem-native for the neoverse-v1 target. We should update NIX to bring in this latest version of SLOTHY.

rod-chapman:
Fixes #1084 

This incorporates micro-arch models for all the instructions required by mlkem-native on the Neoverse-N1 core.

An actual attempt to apply SLOTHY optimizating all AArch64 assembly for Neoverse-V1 will follow in a separate PR.


mkannwischer:
SLOTHY did not finish in CI within one hour - on main it takes like 10 minutes. This is looking like a performance regression in SLOTHY that happened since May 12. There aren't that many changes that affect AArch64. https://github.com/slothy-optimizer/slothy/pull/212 is a candidate, but I don't currently see how that broke it. 
We probably have to bisect this.

mkannwischer:
Second attempt completed in 30 minutes. Still much slower than before.

mkannwischer:
I do not understand why, but everything seems alright now. Let's merge this and see if we run into the performance issues again at some point later.

mkannwischer (approved):

PR #1090, opened 2025-06-19
AWS-LC integration: Remove patch

hanno-becker:
The patch previously needed for integration of mlkem-native into AWS-LC has been merged upstream, and can be removed. The fixed test version for integration is bumped accordingly.


.github/workflows/integration-awslc.yml
```diff
@@ -42,9 +42,7 @@ jobs:
           commit: ${{ inputs.commit }}
       - name: Run importer
         run: |
-          cp integration/awslc/awslc.patch $AWSLC_DIR
           cd $AWSLC_DIR
```
mkannwischer: Remove this `cd`

hanno-becker: Hm, why?

hanno-becker: Done

PR #1094, opened 2025-06-22
simpasm: Add marker at start/end of core assembly

hanno-becker:
This commit modifies `scripts/simpasm` to emit textual markers at the beginning and end of the actual assembly. The purpose of those markers -- as indicated by comments also added -- is to facilitate basic assembly transformations, such as the removal or modification of headers, to suit other build infrastructures.


mkannwischer (commented):
The changes look good to me, but CI isn't happy yet. 

Out of interest: Could you elaborate what you use those markers for downstream?

hanno-becker:
It looks like this may not be needed for now.

PR #1097, opened 2025-06-24
AVX2 implementation of polyvec_basemul in assembly

dkostic:
This change first implements a slightly simpler basemul_avx2.
Then we implement polyvec_basemul for every level.


dev/x86_64/src/polyvec_basemul_acc_montgomery_cached_asm_k3.S
```diff
@@ -0,0 +1,160 @@
...
+vmovdqa (32*\iter + 16)*2(%rdi), _s1
+vpaddw _r0, _r1, _r0
+vpaddw _s0, _s1, _s0
+.endif
+
+vmovdqa _r0, (32*\iter +  0)*2(%rdi)
+vmovdqa _s0, (32*\iter + 16)*2(%rdi)
+/* Bounds: Since we are multiplying with signed canonical twiddles,
+ * each Montgomery multiplication has absolute value < q,
+ * and hence the coefficients of the output have absolute value < 2q. */
```
hanno-becker: Those comments need adjusting, I think. I don't think they were accurate before, but now we also need to take into account the accumulation.

As per the CBMC spec, what we know here is that `a`, `b` are < 4096 in absolute value, while we assume nothing about `c, d, dz` (e.g. 2^15 max absolute value). Then I think every individual Montgomery multiplication is bounded above by `3713` if I got the math right (referring to `poly.c` and `poly.h`). And then we accumulate up to 8 of those. Luckily, `8 * 3713 = 29704 < 2^15`.

dkostic: Good point! I fixed it in the latest update.

hanno-becker (approved):
This is great work, @dkostic -- thanks a lot.

PR #1100, opened 2025-06-27
Update the ACVP tests to 1.1.0.40 (add {en,de}capsulationKeyCheck)

### LINKED ISSUE #1091: Update to ACVP v1.1.0.40
https://github.com/usnistgov/ACVP-Server/releases/tag/v1.1.0.40

There is a significant change in this release regarding ML-KEM: 
```
Adds "encapsulationKeyCheck" and "decapsulationKeyCheck" as functions for ML-KEM Encap/Decap FIPS203 to exercise an implementation's capability to perform the Encapsulation Key Check in FIPS 203 Section 7.2 and the Decapsulation Key Check in FIPS 203 Section 7.3. These tests are only included if the appropriate function is present in the registration. They operate by providing a valid or invalid key and expecting the IUT to return a true for a valid key or false for an invalid key.
```

We should support the new tests.
This required updating our ACVP client as well as the test harness.

willieyz:
- Resolves #1091 

Update the ACVP test vector to 1.1.0.40
https://github.com/usnistgov/ACVP-Server/releases/tag/v1.1.0.40

Considerate about the following in the release note of 1.1.0.40:
```
Adds "encapsulationKeyCheck" and "decapsulationKeyCheck" as functions for ML-KEM Encap/Decap FIPS203 to exercise an implementation's capability to perform the Encapsulation Key Check in FIPS 203 Section 7.2 and the Decapsulation Key Check in FIPS 203 Section 7.3. These tests are only included if the appropriate function is present in the registration. They operate by providing a valid or invalid key and expecting the IUT to return a true for a valid key or false for an invalid key.
```

- Add new test cases to support `encapsulationKeycheck` and `decapsulationKeyheck`.
- Add new function `crypto_kem_check_pk`, `crypto_kem_check_sk` for checking.
- Add testPassed: false handling for the mismatch key length with corresponding security level


test/acvp_client.py
```diff
@@ -113,12 +117,13 @@ def run_encapDecap_test(tg, tc):
...
     elif tg["function"] == "decapsulation":
         acvp_bin = get_acvp_binary(tg)
+        dk_value = tc.get("dk", tg.get("dk"))
         acvp_call = exec_prefix + [
             acvp_bin,
             "encapDecap",
             "VAL",
             "decapsulation",
-            f"dk={tg['dk']}",
+            f"dk={dk_value}",
```
mkannwischer: please change this back.

mlkem/mlkem_native.h
```diff
@@ -130,6 +130,49 @@
...
+ * Specification: Implements @[FIPS203, Section 7.3, 'hash check']
+ *
+ **************************************************/
+
+/* Reference: Not implemented in the reference implementation @[REF]. */
+MLK_API_MUST_CHECK_RETURN_VALUE
+int MLK_API_NAMESPACE(check_sk)(
+    const uint8_t sk[MLKEM_SECRETKEYBYTES(MLK_CONFIG_API_PARAMETER_SET)]);
+
+
```
mkannwischer: Please remove this - we don't need this in the public API.

mlkem/src/kem.c
```diff
@@ -36,8 +36,8 @@
  * This is to facilitate building multiple instances
  * of mlkem-native (e.g. with varying security levels)
  * within a single compilation unit. */
-#define mlk_check_pk MLK_ADD_PARAM_SET(mlk_check_pk)
-#define mlk_check_sk MLK_ADD_PARAM_SET(mlk_check_sk)
+// #define mlk_check_pk MLK_ADD_PARAM_SET(mlk_check_pk)
+// #define mlk_check_sk MLK_ADD_PARAM_SET(mlk_check_sk)
```
mkannwischer: remove it.

mlkem/src/kem.c
```diff
@@ -50,8 +50,9 @@ __contract__(
 );
 #endif /* CBMC */
 
+
 /*************************************************
- * Name:        mlk_check_pk
+ * Name:        crypto_kem_check_pk
```
mkannwischer: Since these coments have moved to the kem.h, please remove them here.

test/acvp_mlkem.c
```diff
@@ -302,6 +328,46 @@ int main(int argc, char *argv[])
...
+        {
+          unsigned char ek[CRYPTO_PUBLICKEYBYTES];
+          /* Encapsulation only for "AFT" */
+          if (type != VAL)
+          {
+            goto encaps_usage;
+          }
+
+          /* Parse ek */
+          if (argc == 0 || decode_hex("ek", ek, sizeof(ek), *argv) != 0)
```
mkannwischer: This needs to be changed to also print testPassed=0 in case of failure.

test/acvp_client.py
```diff
@@ -131,6 +136,51 @@ def run_encapDecap_test(tg, tc):
...
+            err("FAIL!")
+            err(
+                f"{acvp_call} failed with error code {result.returncode}...tcID{tc["tcId"]}"
+            )
+            err(result.stderr)
+            exit(1)
+        # Extract results
+        for l in result.stdout.splitlines():
+            (k, v) = l.split("=")
+            results[k] = v
```
mkannwischer: you need to convert testPassed=0/1 to testPassed=False/True here.

test/acvp_client.py
willieyz: About this change, ACVP move the `dk` from test group `tg` to testcase `tc `in the 1.1.0.40 test vector for decapsulation, this is fallback workaround, I will add the TODO note, we remove this work around after the dk are all in test case `tc` in following version
```
The test group and test case formats for decapsulation test groups and test cases are updated so that dk is provided at the test case level rather than at the test group level.
IMPLICATIONS FOR ACVTS DEMO RELEASE: test vectors that were generated by releases prior to v1.1.0.40 with "isSample": false will fail on validation as a result of this update and will need to be regenerated using ACVTS release v1.1.0.40 or higher. CAVP will implement a workaround as part of the v1.1.0.40 release's deployment to Prod so that test vectors will not need to be regenerated on Prod.
```

test/acvp_mlkem.c
```diff
@@ -73,23 +79,22 @@ static int decode_hex(const char *prefix, unsigned char *out, size_t out_len,
...
   size_t prefix_len = strlen(prefix);
-
   /*
    * Check that hex starts with `prefix=`
    * Use memcmp, not strcmp
    */
   if (hex_len < prefix_len + 1 || memcmp(prefix, hex, prefix_len) != 0 ||
       hex[prefix_len] != '=')
   {
-    goto hex_usage;
```
hanno-becker: Why was this removed?

test/acvp_mlkem.c
```diff
@@ -248,8 +277,8 @@ int main(int argc, char *argv[])
       {
         case encapsulation:
         {
-          unsigned char ek[CRYPTO_PUBLICKEYBYTES];
```
hanno-becker, line 251: General note: Such renaming changes should be done separately.

test/acvp_mlkem.c
mkannwischer: NIST added tests with wrong key lengths, so we actually need to handle it when decoding fails. For consistency it was changed everywhere.

test/acvp_mlkem.c
mkannwischer: This renaming is needed because we switch from including mlkem_native.h to kem.h. 
What do you mean by separately? In its own commit?

test/acvp_mlkem.c
hanno-becker: It's not consistent at the moment? There is still `goto hex_usage` below. Let's keep the 1st and 3rd occasion, and only change the second which is about length encoding.

test/acvp_mlkem.c
hanno-becker: Yes. It's just a nit, not blocking.

willieyz: Thanks for pointing that out! 
I’ll keep it in mind and handle such changes in separate commits going forward.

hanno-becker (approved):
LGTM

mlkem/mlkem_native.h
mkannwischer, line 1: undo changes to mlkem_native.h please

mlkem/src/kem.c
```diff
@@ -128,17 +97,21 @@ static int mlk_check_sk(const uint8_t sk[MLKEM_INDCCA_SECRETKEYBYTES])
 
   mlk_hash_h(test, sk + MLKEM_INDCPA_SECRETKEYBYTES,
              MLKEM_INDCCA_PUBLICKEYBYTES);
+
```
mkannwischer: These whitespace changes are disstracting. Please remove them.

test/acvp_client.py
```diff
@@ -33,6 +33,10 @@
         f"{acvp_dir}/acvp_v1.1.0.39_keyGen_prompt.json",
         f"{acvp_dir}/acvp_v1.1.0.39_keyGen_expectedResults.json",
     ),
+    (
```
mkannwischer, line 32: As discussed offline, we want to keep 3 versions, so please remove `v1.1.0.36`.

test/acvp_client.py
```diff
@@ -33,6 +33,10 @@
...
+        f"{acvp_dir}/acvp_v1.1.0.40_keyGen_prompt.json",
+        f"{acvp_dir}/acvp_v1.1.0.40_keyGen_expectedResults.json",
```
mkannwischer, line 34: Please adjust the README.md in the acvp_data directory.

test/acvp_client.py
```diff
@@ -113,12 +121,15 @@ def run_encapDecap_test(tg, tc):
             results[k] = v
     elif tg["function"] == "decapsulation":
         acvp_bin = get_acvp_binary(tg)
+        # TODO: Remove the fallback workaround after update all test vector version
```
mkannwischer: Please be more specific here. 
```suggestion
        # TODO: Remove this fallback workaround. v.1.1.0.40 moved the dk from the 
        # tg to the tc. This can be removed when v1.1.0.39 is removed.
```

test/acvp_client.py
```diff
@@ -113,12 +121,15 @@ def run_encapDecap_test(tg, tc):
...
+        dk_value = tc.get("dk", tg.get("dk"))
         acvp_call = exec_prefix + [
             acvp_bin,
             "encapDecap",
             "VAL",
             "decapsulation",
-            f"dk={tg['dk']}",
+            # TODO: Remove the fallback workaround after update all test vector version
```
mkannwischer: remove this comment. one comment is sufficient.

test/acvp_mlkem.c
```diff
@@ -302,13 +331,60 @@ int main(int argc, char *argv[])
...
+          }
+
+          /* Parse ek */
+          if (decode_hex("ek", ek, sizeof(ek), *argv) != 0)
+          {
+            /*
+              "testPassed": false for
+              ACVP 1.1.0.40+ {en, de}capsulationKeyCheck test case:
+              key length mismatch with corresponding security level
+            */
```
mkannwischer:
```suggestion
            /*
              ACVP 1.1.0.40+ {en, de}capsulationKeyCheck test cases test keys of
              incorrect length. The mlkem-native API does not allow passing keys
              of incorrect length. We, hence, fail during decoding instead.
            */
```

test/acvp_mlkem.c
```diff
@@ -302,13 +331,60 @@ int main(int argc, char *argv[])
...
+          /* Encapsulation only for "VAL" */
+          if (type != VAL || argc == 0)
+          {
+            goto decapsulationKeyCheck_usage;
+          }
+
+          /* Parse dk */
+          if (decode_hex("dk", dk, sizeof(dk), *argv) != 0)
+          {
+            printf("testPassed=0\n");
```
mkannwischer, line 377:
```suggestion
            /*
              ACVP 1.1.0.40+ {en, de}capsulationKeyCheck test cases test keys of
              incorrect length. The mlkem-native API does not allow passing keys
              of incorrect length. We, hence, fail during decoding instead.
            */
            printf("testPassed=0\n");
```

mlkem/mlkem_native.h
willieyz: Done.

mlkem/src/kem.c
willieyz: Done

PR #1102, opened 2025-06-30
tests: Support stack usage measurement

hanno-becker:
Implements static and dynamic stack analysis using GCC -fstack-usage and valgrind massif.

Usage:
  make stack_512
  make run_stack_512
or
  ./scripts/tests stack [--peak-only --dump-massif]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,14 @@
 # Copyright (c) The mlkem-native project authors
 # SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
 
-.PHONY: func kat acvp \
-	func_512 kat_512 acvp_512 \
-	func_768 kat_768 acvp_768 \
-	func_1024 kat_1024 acvp_1024 \
-	run_func run_kat run_acvp \
-	run_func_512 run_kat_512 \
-	run_func_768 run_kat_768 \
-	run_func_1024 run_kat_1024 \
+.PHONY: func kat acvp stack \
+	func_512 kat_512 acvp_512 stack_512 \
+	func_768 kat_768 acvp_768 stack_768 \
+	func_1024 kat_1024 acvp_1024 stack_1024 \
+	run_func run_kat run_acvp run_stack \
+	run_func_512 run_kat_512 run_stack_512 \
+	run_func_768 run_kat_768 run_stack_768 \
+	run_func_1024 run_kat_1024 run_stack_1024 \
 	bench_512 bench_768 bench_1024 bench \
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -76,6 +76,39 @@ acvp_1024: $(MLKEM1024_DIR)/bin/acvp_mlkem1024
...
+	$(CC) $(CFLAGS) -DMLKEM_K=2 -Imlkem/src -o $@ $< $(BUILD_DIR)/libmlkem512.a $(call MAKE_OBJS, $(MLKEM512_DIR), $(wildcard test/notrandombytes/*.c))
+
+$(MLKEM768_DIR)/bin/stack_test768: test/stack_test.c $(BUILD_DIR)/libmlkem768.a $(call MAKE_OBJS, $(MLKEM768_DIR), $(wildcard test/notrandombytes/*.c))
+	$(Q)mkdir -p $(dir $@)
+	$(CC) $(CFLAGS) -DMLKEM_K=3 -Imlkem/src -o $@ $< $(BUILD_DIR)/libmlkem768.a $(call MAKE_OBJS, $(MLKEM768_DIR), $(wildcard test/notrandombytes/*.c))
+
+$(MLKEM1024_DIR)/bin/stack_test1024: test/stack_test.c $(BUILD_DIR)/libmlkem1024.a $(call MAKE_OBJS, $(MLKEM1024_DIR), $(wildcard test/notrandombytes/*.c))
+	$(Q)mkdir -p $(dir $@)
+	$(CC) $(CFLAGS) -DMLKEM_K=4 -Imlkem/src -o $@ $< $(BUILD_DIR)/libmlkem1024.a $(call MAKE_OBJS, $(MLKEM1024_DIR), $(wildcard test/notrandombytes/*.c))
+
```
mkannwischer: THis should move to components.mk and please align it it with the other rules theer.
Why do we need explicit rules for those targets? Can't we follow the same approach as for the other binaries?

Makefile
```diff
@@ -76,6 +76,39 @@ acvp_1024: $(MLKEM1024_DIR)/bin/acvp_mlkem1024
 	$(Q)echo "  ACVP       ML-KEM-1024:  $^"
 acvp: acvp_512 acvp_768 acvp_1024
 
+# Stack analysis targets - build with stack usage flags and stack test binaries
+stack_512: CFLAGS += -fstack-usage
```
mkannwischer: This should also move. Let's keep the root Makefile lean.

hanno-becker: Done

Makefile
hanno-becker: Done

PR #1104, opened 2025-07-03
nix: Introduce separate shells for cross compilation

### LINKED ISSUE #1050: Nix: Split up cross toolchains into multiple shells
Currently we have one nix shell that contains all the cross toolchains (x86_64-gcc aarch64-gcc riscv64-gcc riscv32-gcc ppc64le-gcc): 

https://github.com/pq-code-package/mlkem-native/blob/bb9f01345b343d33113ac257163436dcebd7f095/nix/util.nix#L39

https://github.com/pq-code-package/mlkem-native/blob/bb9f01345b343d33113ac257163436dcebd7f095/flake.nix#L97

While this is fine in CI as everything can be cached, this is undesirable locally for development.
We should split those shells up and have one shell per cross toolchain.

hanno-becker:
Previously, we used a single `ci-cross` nix shell for all cross-compilation tool chains. This shell takes a very long time to build and is therefore inconvenient to use for cross-compilation on a specific target.

This commit extends the nix flake by architecture specific ci cross shells:

- ci-cross-x86_64
- ci-cross-aarch64
- ci-cross-riscv64
- ci-cross-riscv32
- ci-cross-ppc64le
- ci-cross-aarch64_be

It also adjusts CI to use those shells.

Fixes #1050


.github/workflows/ci.yml
```diff
@@ -119,7 +119,7 @@ jobs:
       - name: build + test
         uses: ./.github/actions/multi-functest
         with:
-          nix-shell: ${{ matrix.target.mode == 'native' && 'ci' || 'ci-cross' }}
+          nix-shell: ${{ matrix.target.mode == 'native' && 'ci' || matrix.target.mode == 'cross-x86_64' && 'ci-cross-x86_64' || matrix.target.mode == 'cross-aarch64' && 'ci-cross-aarch64' || matrix.target.mode == 'cross-riscv64' && 'ci-cross-riscv64' || matrix.target.mode == 'cross-riscv32' && 'ci-cross-riscv32' || matrix.target.mode == 'cross-ppc64le' && 'ci-cross-ppc64le' || matrix.target.mode == 'cross-aarch64_be' && 'ci-cross-aarch64_be' || 'ci-cross' }}
```
mkannwischer: This is a little messy. Can we add this as a field in the target matrix, please?

nix/util.nix
```diff
@@ -101,6 +101,54 @@ rec {
...
+  _individual_toolchain = { name, cross_compilers }:
+    let
+      common_deps = builtins.attrValues
+        {
+          inherit (pkgs.python3Packages) sympy pyyaml;
+          inherit (pkgs)
+            gnumake
+            python3
+            qemu;
+        } ++ pkgs.lib.optionals (pkgs.stdenv.isDarwin) [ pkgs.git ];
```
mkannwischer, line 114: nit: If this is intended to be used only in CI (as the `ci_` prefix in the shell names indicates), then there is no need to support `git` within the nix shell.

hanno-becker: This is following the precedent of the `ci-cross` shell. I assume there was a reason why it was introduced, but don't recall. I'm not changing this for now.

.github/workflows/ci.yml
hanno-becker: Ok, done

nix/util.nix
mkannwischer: But that one is also used for the native toolchains right? So we need git there (at least if cross=false). Or i misunderstood.

Not important anyway.

hanno-becker: Ok, can we leave it as is then?

mkannwischer: Yes.

mkannwischer (approved):
Thanks @hanno-becker  for the changes! LGTM.

PR #1108, opened 2025-07-06
lint, autogen, format: Improve and unify UX

hanno-becker:
This commit makes a few non-functional changes to the scripts lint, format, and autogen:
- `autogen` prints high-level status updates in addition to the low-level per-file status updates
- `autogen` uses color coding like `format` does
- `lint` omits GH summary content if GITHUB_STEP_SUMMARY is not set.
- format uses color-coded error messages
- The color definitions across the scripts have been harmonized

The only functional changes are the following in `autogen`:
- We pass `dry_run` onto `update_via_copy`. This function would previously erroneously update files even when dry-run was enabled, instead of failing as expected. This was found while testing the non-functional changes above.
- We don't run the preprocessor auto-commenting when adding mini-bibliographies to source files. This previously cluttered the status output with `[if-else] FILENAME` message, misleading the user on what the current high-level task is. Now, auto-commenting is omitted during bibliography generation, and the low-level status lines read `[citations] ...` which makes more sense.


mkannwischer (approved):
Thanks @hanno-becker! LGTM!

PR #1111, opened 2025-07-08
TEST: Test macos-15

mkannwischer:
Next month Github is going to switch macos-latest from macos-14 to macos-15. This PR tests if this will cause any issues.

See https://github.com/actions/runner-images/issues/12520


mkannwischer:
Everything looks good. So I think it's safe to just stick with macos-latest which will at some point switch from 14 to 15. Okay, @hanno-becker?

hanno-becker:
@mkannwischer Ok!

PR #1112, opened 2025-07-08
Simplify gen_matrix() and poly_rej_uniform_x4()

rod-chapman:
Simplify gen_matrix() and poly_rej_uniform_x4()

1. Simplify poly_rej_uniform_x4() API to explicitly take 4 objects to write, rather than a slice of a larger array.
2. Update gen_matrix() in light of that.
3. Update test/bench_components_mlkem.c

Proof time of poly_rej_uniform_x4() reduces from 20s to 9s on Apple M1.
Proof time of gen_matrix() unchanged at 8s

Runtime performance benchmarks show no significant change.

### DIFF
--- a/mlkem/src/indcpa.c
+++ b/mlkem/src/indcpa.c
@@ -221,10 +221,9 @@ void mlk_gen_matrix(mlk_polymat a, const uint8_t seed[MLKEM_SYMBYTES],
   /* Sample 4 matrix entries a time. */
   for (i = 0; i < (MLKEM_K * MLKEM_K / 4) * 4; i += 4)
   {
-    uint8_t x, y;
-
     for (j = 0; j < 4; j++)
     {
+      uint8_t x, y;
       x = (i + j) / MLKEM_K;
       y = (i + j) % MLKEM_K;
       if (transposed)
@@ -239,11 +238,7 @@ void mlk_gen_matrix(mlk_polymat a, const uint8_t seed[MLKEM_SYMBYTES],
       }
     }
 
-    /*
-     * This call writes across mlk_polyvec boundaries for K=2 and K=3.
-     * This is intentional and safe.
-     */
-    mlk_poly_rej_uniform_x4(&a[i], seed_ext);
+    mlk_poly_rej_uniform_x4(&a[i], &a[i + 1], &a[i + 2], &a[i + 3], seed_ext);
   }
 
   /* For MLKEM_K == 3, sample the last entry individually. */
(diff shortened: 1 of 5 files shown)


mlkem/src/indcpa.c
```diff
@@ -239,31 +238,25 @@ void mlk_gen_matrix(mlk_polymat a, const uint8_t seed[MLKEM_SYMBYTES],
...
+  /* For MLKEM_K == 2 or 4, we're done.                             */
+  /* For MLKEM_K == 3, we have one more poly to sample with i == 8. */
+  if (MLKEM_K == 3)
+  {
+    mlk_assert(i == 8 && i / MLKEM_K == 2 && i % MLKEM_K == 2);
+    /* The final poly is on the leading diagonal of the matrix */
+    /* at coordinates [2,2], so the "transpose" flag here is   */
+    /* irrelevant.                                             */
+    seed_ext[0][MLKEM_SYMBYTES + 0] = 2;
+    seed_ext[0][MLKEM_SYMBYTES + 1] = 2;
```
hanno-becker: What are we gaining from this? I find the previous code more obvious.

rod-chapman: Well... it removes a path from the code, which I can only see as a win. Better comments explain what's going on. A small reduction in code size when K=2 or K=4 (assuming a compiler will remove the if statement entirely in those cases.) What's not to like?

rod-chapman:
For me, it's just much more obviously correct this way. It removes the need to understand (and the reliance on) how arrays are laid out in memory and the need to write over more than one array element.

hanno-becker (changes requested):
The change of signature `mlk_poly_rej_uniform_x4()` is fine to me if it helps further experimentation on #906. Please revert the other unrelated change -- I think it's a micro-optimization that hinders readability.

mlkem/src/indcpa.c
hanno-becker: I think we just have a different taste here; I like that with the current code it's obvious that the tail handling continues the pattern of the main loop. Also, it is easier to fall back to no-batching by commenting/guarding the batched loop and turning the `if(i < MLKEM_K * MLKEM_K)` into `while (i < MLKEM_K * MLKEM_K)`.

Happy to let @mkannwischer break the tie, but seeing that this is unrelated to #906, I prefer to leave it out.

rod-chapman:
I will disagree-and-commit on this issue. Reverted, squashed, commit message updated, and pushed.

mlkem/src/indcpa.c
rod-chapman: Agreed... I also find the repetition distasteful... but happy to go along with the majority voice.

hanno-becker (approved):

mkannwischer (approved):
Thanks @rod-chapman. The change as it stands right now makes sense to me.
It aligns the `poly_rej_uniform_x4` signature with the `poly_getnoise_etaX_x4` signature.

PR #1115, opened 2025-07-09
Switch mlk_polyvec and mlk_polymat to struct wrappers

hanno-becker:
- Change mlk_polyvec back to struct { mlk_poly vec[MLKEM_K]; }
- Change mlk_polymat to struct { mlk_polyvec vec[MLKEM_K]; }
- Update all function signatures to use pointer style
- Fix all implementations to use struct member access
- Update tests, benchmarks, and CBMC harnesses
- Add consistent const annotations

### DIFF
--- a/mlkem/src/indcpa.c
+++ b/mlkem/src/indcpa.c
@@ -296,9 +296,11 @@ void mlk_gen_matrix(mlk_polymat *a, const uint8_t seed[MLKEM_SYMBYTES],
  *              - mlk_polymat a: Input matrix. Must be in NTT domain
  *                  and have coefficients of absolute value < 4096.
  *              - mlk_polyvec v: Input polynomial vector. Must be in NTT
- *                  domain.
+ *                  domain and have coefficients of absolute value
+ *                  < MLK_NTT_BOUND.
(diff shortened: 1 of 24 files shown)


mkannwischer (commented):
Could you please elaborate why you would like to make this change?

hanno-becker:
@rod-chapman suggested that this unblocks #906. Adjusting it to check if that's true.

mkannwischer:
> @rod-chapman suggested that this unblocks #906. Adjusting it to check if that's true.

Should this be converted to a draft PR in that case?

hanno-becker:
@mkannwischer We could. One could make the argument, though, that the only reason to introduce `mlk_polymat/vec` as a array was to fix the UB in `gen_matrix()`, which is now moot given @rod-chapman's #1112.

rod-chapman:
Does this change improve performance and/or stability of proof on platforms that we care about (by which I really mean - AArch64/macOS, Intel/Linux, EC2 Graviton/Linux, and "whatever our CI runners are running..."?

I assume performance impact on generated code is negligible?

hanno-becker:
Closing this for now.

PR #1119, opened 2025-07-12
ACVP: Download testvectors transparently on-the-fly

mkannwischer:
This commit changes the acvp_client.py to automatically download the required
testvectors file for a given version of the ACVP Server.
The main benefit of this change is that changes to the testvectors (which have
been happening much more frequent than I expected) do not pollute our git
history to the same extend anymore.
The local copies of the ACVP files are removed.


test/acvp_client.py
```diff
@@ -3,49 +3,64 @@
...
+        "ML-KEM-encapDecap-FIPS203/prompt.json",
+        "ML-KEM-encapDecap-FIPS203/expectedResults.json",
+    ]
+
+    # Create directory structure
+    data_dir = Path(f"test/.acvp-data/{version}/files")
+    data_dir.mkdir(parents=True, exist_ok=True)
+
+    for file_path in files_to_download:
+        local_file = data_dir / file_path
```
hanno-becker, line 40: I didn't know that syntax existed.

test/acvp_client.py
```diff
@@ -3,49 +3,64 @@
...
+
+        if not local_file.exists():
+            url = f"{base_url}/{file_path}"
+            print(f"Downloading {file_path}...", file=sys.stderr)
+            try:
+                urllib.request.urlretrieve(url, local_file)
+                # Verify the file is valid JSON
+                with open(local_file, "r") as f:
+                    json.load(f)
+            except json.JSONDecodeError as e:
```
hanno-becker, line 51: Nit: `urllib.request.urlretrieve(url, local_file)` could be taken out of the scope of this `try ... except` block and get its own error message. It's not unthinkable that the URLs change or the server is down at some point.

hanno-becker: Ah, ignore. You have different `except` blocks anyway.

hanno-becker (approved):
One minor suggestion on error handling, otherwise looks very good to me. Thanks @mkannwischer !

test/acvp_client.py
mkannwischer: It does have its own except block? 
```py
            except Exception as e:
                print(f"Error downloading {file_path}: {e}", file=sys.stderr)
                local_file.unlink(missing_ok=True)
                return False
```

What are you missing from that error message?

hanno-becker: It's ok, I considered two `try` blocks for better separation, but you're right that with the first block confinded ty JSON errors, it should work as intended.

PR #1120, opened 2025-07-14
Namespacing: zetas -> mlk_zetas

mkannwischer:
I noticed that our zetas array in the C-backend does not contain any namespacing. This may clash in the case of single-compilation units builds, e.g. when built in a CU together with ML-DSA code.

Related: https://github.com/pq-code-package/mldsa-native/pull/359


hanno-becker:
Thanks for spotting this @mkannwischer !

Issue #1117, opened 2025-07-10
AArch64: Consider running Rejection Sampling through SLOTHY


mkannwischer:
Currently the only AArch64 assembly not superoptimized by SLOTHY is the rejection sampling. The only reason for that is that SLOTHY is lacking support for quite a few instructions in there. We should add support for those instructions and try if that improves performance.

hanno-becker:
Note however that the HOL Light proofs don't unroll the code so they will need rework if we use SW pipelining.

mkannwischer:
> Note however that the HOL Light proofs don't unroll the code so they will need rework if we use SW pipelining.

Noted. We'll get the performance w/ and w/o SW pipelining first. Then we can decide if either of them is worth consideration.

willieyz:
For this issue, I have found the missing instructions. Please see the list below:

- `sub buflen buflen #48`
- `bic tmp0.8h, #0xf0, lsl 8`
- `cmhi sign0.8h, mlkem_q.8h, val0.8h`
- `uaddlv s20, v4.8h`
- `fmov rec_idx_0_w, t0`
- `ldr q24, [x3, x12, lsl #4]`
- `cnt sign0.16b, sign0.16b`
- `tbl val0.16b, {val0.16b}, table0.16b`

I will create a separate PR in Slothy to add support for these instructions.
Once that's done, we can rerun the process to check for any remaining missing instructions or proceed with running rejection sampling through Slothy.

hanno-becker:
@willieyz What's the status here?

willieyz:
Hello @hanno-becker, here’s the current status of this issue:

1. We have added the instructions required for Rejection Sampling, see [SLOTHY PR#279](https://github.com/slothy-optimizer/slothy/pull/279).
2. As @mkannwischer mentioned in #1127, we are not considering running Rejection Sampling through SLOTHY for now, since there is no performance improvement without software pipelining. We plan to revisit this after reviewing our uArch model and identifying the cause of the mistake.

according to the above, I suggest we put this issue on hold and un-assign it for now?

hanno-becker:
Closing this for now.

Issue #48, opened 2024-06-04
Complete CI even when individual components fail


hanno-becker:
The current CI setup stops as soon as a single check fails. For example, if linting fails, it won't get to the KAT tests.

@potsrevennil Could you change this behavior so that all checks are run and a summary of failed/successful steps is printed in the end?

cothan:
I agree. Our CI become more complex, and nix setup time take a while as well. I think what @hanno-becker  said is the right approach.

PR #1127, opened 2025-07-23
AArch64: Add Rejection Sampling running through SLOTHY

willieyz:
- Related #1117 

This PR adds AArch64 assembly rejection sampling with superoptimization throught SLOTHY.

The previous AArch64 assembly rejection sampling was not superoptimized by SLOTHY due to a lack of support for the following instruction:
- https://github.com/pq-code-package/mlkem-native/issues/1117#issuecomment-3077889658
- `sub buflen buflen #48`
- `bic tmp0.8h, #0xf0, lsl 8`
- `cmhi sign0.8h, mlkem_q.8h, val0.8h`
- `uaddlv s20, v4.8h`
- `fmov rec_idx_0_w, t0`
- `ldr q24, [x3, x12, lsl #4]`
- `cnt sign0.16b, sign0.16b`
- `tbl val0.16b, {val0.16b}, table0.16b`

After we add above instructions with following draft PR in SLOTHY
- https://github.com/slothy-optimizer/slothy/pull/283
- https://github.com/slothy-optimizer/slothy/pull/279

Now that `mlkem/src/native/aarch64/src/rej_uniform_asm.S` is generated locally based on the local branch rebased on the two PRs above, we will try running it through CI to see if it improves performance.


hanno-becker:
@mkannwischer @willieyz What is the status of this?

mkannwischer:
> @mkannwischer @willieyz What is the status of this?

The status is that without software pipelining we do not see any improvement of performance (actually it makes performance slightly worse). Next step would be to debug our uArch model to see if this is caused by a mistake in the model. To use software pipelining one will have to rewrite the loop (and probably extend SLOTHY).
I won't get to this in the foreseeable future. Closing this for now.

Issue #1122, opened 2025-07-15
Consider updating bitwuzla 0.7.0 -> 0.8.2


mkannwischer:
#1121 suggests there may be problems with bitwuzla 0.8.x.

Bitwuzla [0.8.0](https://github.com/bitwuzla/bitwuzla/releases/tag/0.8.0) was released on May 23, 2025, but it had some (non-critical) test failures and, hence, never made it into nixpkgs.
Bitwuzla [0.8.1](https://github.com/bitwuzla/bitwuzla/releases/tag/0.8.1) was released on Jul 8, 2025 and is currently being pulled into nixpkgs: https://github.com/NixOS/nixpkgs/pull/421506.
We should try that once it lands in unstable. 
Bitwuzla [0.8.2](https://github.com/bitwuzla/bitwuzla/releases/tag/0.8.2) was released on Aug 5, 2025.

rod-chapman:
There are already soundness concerns raised with 0.8.1:
 https://github.com/bitwuzla/bitwuzla/issues/185

mkannwischer:
Okay. Sounds like we should skip this release then.

mpreiner:
Bitwuzla 0.8.2 was just released and fixes the mentioned issue.

hanno-becker:
Thanks for the update @mpreiner!

rod-chapman:
There are 52 functions in mlkem-native that use bitwuzla. I just tried them all with bitwuzla 0.7.0 and again with bitwuzla 0.8.2.

Result: all 52 proofs are OK with 0.8.2, with a very slight reduction in performance (3m5s real time versus 3m18s) on an 8-core macBook Pro.

hanno-becker:
That seems like it's within the normal performance variation.

PR #1128, opened 2025-07-24
[TEST]: Update bitwuzla to 0.8.1

mkannwischer:
This commit updates bitwuzla to 0.8.1.
As we have received some reports of potential problems with this bitwuzla version, this is mostly to make sure the proofs in mlkem-native still work. Resolves https://github.com/pq-code-package/mlkem-native/issues/1122

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -54,11 +54,11 @@
     },
     "nixpkgs-unstable": {
       "locked": {
-        "lastModified": 1748460289,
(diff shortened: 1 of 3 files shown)


mkannwischer:
An added benefit of this PR is that nixpkgs-unstable has updated CBMC to 6.7.1 meaning we do no longer have to build from source, but can instead use the binary cached in nixpkgs.

mkannwischer:
Skipping this version due to soundness concerns: https://github.com/pq-code-package/mlkem-native/issues/1122

Issue #1129, opened 2025-07-25
Implementation-defined signedness of "char" causes CBMC proof instability


rod-chapman:
I have found that CBMC proof on Graviton3/Ubuntu was not behaving same as on macOS and Intel machines. Analysis showed a difference in the generated GOTO files.

On macOS and Intel machines, we see "signedbv[8]" appearing in some statements. On Graviton/Linux, we see "unsignedbv[8]" on the same lines. This leads to differences in the generated SMT files and some proof instability.

Root cause is that plain "char" has implementation-defined signedness in C.

We have 2 instances of plain "char"  in the mlkem-native sources, in debug.[hc]. These should be made unambiguously "unsigned char".

CBMC can also find other instances of plain "char" in other header files, so we should use the -fsigned-char switch on goto-cc to force "signed char" semantics for _all_ platforms on which proofs are run. This requires a small change to proofs/cbmc/Makefile.common

hanno-becker:
I don't think this is a bug, but if it causes instability in CBMC, we should look into it. Thanks @rod-chapman !

rod-chapman:
It's a simple change. I will open a PR...

hanno-becker:
@rod-chapman No, please wait until we agree that we actually want this.

hanno-becker:
Closing this for now. @rod-chapman re-open if the issue persists and you think it is a problem with mlkem-native rather than CBMC.

PR #1130, opened 2025-07-25
Prevent signedness of "char" causing CBMC instability.

rod-chapman:
Prevent signedness of "char" causing CBMC instability.

1. Add -fsigned-char to goto-cc COMPILE_FLAGS in Makefile.common to force consistent proof behaviour on all platforms where plain "char" is unsigned by default (e.g. Graviton/Linux) AND where CBMC contract libraries contain use of plain "char *" type.

### DIFF
--- a/proofs/cbmc/Makefile.common
+++ b/proofs/cbmc/Makefile.common
@@ -311,7 +311,7 @@ CHECKFLAGS += $(CBMC_FLAG_UNSIGNED_OVERFLOW_CHECK)
 NONDET_STATIC ?=
 
 # Flags to pass to goto-cc for compilation and linking
-COMPILE_FLAGS ?= -Wall -Werror --native-compiler $(CC)
+COMPILE_FLAGS ?= -Wall -Werror -fsigned-char --native-compiler $(CC)
 LINK_FLAGS ?= -Wall -Werror
 EXPORT_FILE_LOCAL_SYMBOLS ?= --export-file-local-symbols
 


hanno-becker:
As mentioned on #1129  I don't (yet -- convince me!) agree that this is necessarily the right resolution. Is `char` is per-se evil? In the case of file names, I want whatever native character type the C implementation uses, but don't care about its signedness.

rod-chapman:
Plain "char" has implementation-defined semantics, so I would say it's "mostly evil" if it causes proof instability.

hanno-becker:
Why do debug functions enter the CBMC proofs in the first place?

rod-chapman:
It seems they don't (but you never know with CBMC - it does seem to generate dependencies in places you wouldn't expect...) In this case, the change to the declarations in debug.h isn't needed. The change to add -fsigned-char to the Makefile.common is enough to make the generated GOTO consistent on all platforms, because the latter changes the use of a "char *" cast in an actual assertion in a _body_ of a contract-implementing function...

In short, we could revert the change to debug.[hc] and keep the change to Makefile.common

rod-chapman:
OK... reverted, squashed and pushed. Let's see what CI says now.

hanno-becker:
@rod-chapman Can you provide more information on the underlying issue? How can one reproduce it? I don't like forcing choices that are implementation-defined (and variable in practice).

hanno-becker:
@rod-chapman Ping. Is this still relevant? Please provide further information or close.

rod-chapman:
Any change required is in CBMC and its runtime library sources, not mlkem-native, so closing this for now.

PR #1133, opened 2025-07-28
AArch64: Add YAML based documentation to assembly files

hanno-becker:
This commit adds documentation to all AArch64 assembly files describing purpose, C signature, and register ABI. The register ABI is written in YAML so it can be easily processed automatically; in particular, we will process it by an ABI checker to-be-implemented.


dev/aarch64_clean/src/intt.S
```diff
@@ -19,7 +19,36 @@
...
+ *     description: Twiddle factors for layers 1-5
+ *   x2:
+ *     type: buffer
+ *     size_bytes: 768
+ *     permissions: read-only
+ *     c_parameter: const int16_t twiddles56[384]
+ *     description: Twiddle factors for layers 6-7
+ * ```
+ *
+ * Stack Usage: 64 bytes for saving callee-saved NEON registers
```
mkannwischer: Nit: NEON -> Neon for consistency with the spelling in the rest of the repo.

dev/fips202/aarch64/src/keccak_f1600_x1_scalar_asm.S
```diff
@@ -8,6 +8,31 @@
...
+ *
+ * C Signature: void mlk_keccak_f1600_x1_scalar_asm(uint64_t state[25], const uint64_t rc[24])
+ *
+ * Register ABI:
+ * ```yaml
+ *   x0:
+ *     type: buffer
+ *     size_bytes: 200
+ *     permissions: read/write
+ *     c_parameter: uint64_t *state
```
mkannwischer:
```suggestion
 *     c_parameter: uint64_t state[25]
```
Same for other arguments in the fips202 backend.

mkannwischer (changes requested):
Thanks @hanno-becker! 
Please also update mlkem/src/native/aarch64/src/arith_native_aarch64.h with the explicit length.

Would it make sense to also use explit length in the autogenerated twiddle (aarch64_zetas.c)?

dev/aarch64_clean/src/intt.S
hanno-becker: Done

dev/fips202/aarch64/src/keccak_f1600_x1_scalar_asm.S
hanno-becker: Done

PR #1135, opened 2025-07-30
Add ABI checkers for x86_64, AArch64, PPC64le, Armv8.1-M

### LINKED ISSUE #884: Add ABI checker for assembly files
We had a near-miss in #850 where SLOTHY introduced the use of callee-saved registers, but they were not saved/restored. Luckily, the CI caught this, but only in a few tests compared to the overall number of jobs we are running. We need to do better at catching this.

One option is to have an explicit ABI-check for every function that's written in assembly: Set the callee-saved registers to random values, call the function, and check upon return that they are as before. Care has to be taken to set pointer-arguments correctly, but otherwise those ABI-checks don't need to do any other correctness check.

### LINKED ISSUE #1143: Add ABI checker for x86_64 assembly files
Same as https://github.com/pq-code-package/mlkem-native/issues/884, but for x86_64.

hanno-becker:
* Based on #1134
* Resolves #884
* Resolves #1143

```
    Add ABI checker for AArch64, x86_64 (SysV), PowerPC64, Armv8.1-M+MVE

    Verify that each assembly kernel preserves the callee-saved registers
    its platform calling convention requires:

    - AArch64 (AAPCS64):   x19-x28, x29/FP, lower 64 bits of d8-d15.
    - x86_64 (System V):   rbx, rbp, r12-r15. No SIMD register callee-saved.
    - PowerPC64 (ELFv2):   GPRs r14-r31, FPRs f14-f31, VRs v20-v31, CR2-CR4.
    - Armv8.1-M (AAPCS32): r4-r11 plus MVE Q4-Q7 (= D8-D15).

    A per-arch assembly call stub loads a random register state, calls the
    kernel, and captures the result; a checker then confirms the callee-saved
    set is intact. Pointer arguments are backed by correctly-sized buffers
    whose layout comes from a per-kernel YAML block; scripts/autogen turns
    that YAML into the per-kernel checks. A self-test of hand-written
    corrupters confirms the checker actually fires before kernel verdicts
    are trusted.

    Run via `make run_abicheck` or `scripts/tests abicheck`. It needs no
    library build, so `scripts/tests all` runs it by default.

    See test/abicheck/README.md for details.
```


hanno-becker:
@mkannwischer I think this is ready for a first pass from you. I'll also do another one, esp. on the PPC64le + Armv8.1-M callstubs which came from Claude -- but I think it's roughly in shape now and would appreciate your feedback.

scripts/simpasm
```diff
@@ -298,16 +298,42 @@ def simplify(logger, args, asm_input, asm_output=None):
...
-                "#ifdef __APPLE__",
-                f".global _{sym}",
-                f"_{sym}:",
-                "#else",
-                f".global {sym}",
-                f"{sym}:",
-                "#endif",
-                "",
-            ]
+            # Apple platforms exist only on AArch64 and x86_64; on those
```
hanno-becker, line 301: For the record: This is no longer strictly needed but was a fix that came out of the work when -- initially -- using `simpasm` to generate another bunch of .S files to be used by the ABI checker.

test/abicheck/x86_64/checks/check_nttunpack_avx2_asm.c
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright (c) The mlkem-native project authors
```
hanno-becker, line 2: For the record: I experimented consolidating all those .c files into invocations of a generic stub, but this ended up being pretty difficult to read. The current approach, while incurring a larger LOC count, seems better to me.

mkannwischer (approved):
This looks great! Thanks @hanno-becker for almost a year of work.

PR #1139, opened 2025-08-02
[TEST] CI: Update to Windows 2025

mkannwischer:
Github will switch windows-latest from Windows 2022 to Windows 2025 in September.
This PR is to make sure that our CI will not break.


mkannwischer:
Works fine. No need to change anything.

PR #1141, opened 2025-08-03
[TEST] Add gcc15

hanno-becker:


.github/workflows/ci.yml
```diff
@@ -228,6 +228,13 @@ jobs:
            c23: True
            opt: all
            examples: true
+         - name: gcc-15
+           shell: ci_gcc15
+           darwin: False
```
mkannwischer, line 234: Can you add a TODO here, so we do not forget to enable it once it works on Darwin?

mkannwischer (approved):
Thanks @hanno-becker!

.github/workflows/ci.yml
hanno-becker: Yes done

PR #1146, opened 2025-08-04
Makefile: Add compile-time compiler+host platform feature detection

### LINKED ISSUE #1144: Makefile: Add compile-time detection for compiler and host platform feature support
**Context:** The idea of `auto.mk` is to optimize the build for the platform on which mlkem-native is built. This is done by automatically adding various AVX2 feature flags.

**Issues:** There are issues at multiple levels.
1. On x86_64, `auto.mk` makes two _assumptions_ about the compilation host:
    i. It assumes that the host _compiler_ supports various feature flags: `-mavx2 -mbmi2 -mpopcnt -maes`
    ii. It assumes that the host _CPU_ supports the aforementioned AVX2 features.
2. On AArch64, `auto.mk` does not deal with the Armv8.4-a+SHA3 extension required for parts of the FIPS202 backend. However, the analogous way of dealing with this, by adding `-march=armv8.4-a+sha3` to `CFLAGS`, would suffer from the same issues as in point 1.

**Task:** 
1. Add a new `test/mk/compiler.mk` which detects whether the architectural features used by mlkem-native are supported by the host compiler. 
    i. For x86_64, detect whether the compiler supports AVX2, SSE2, and BMI2.
    ii. For AArch64, detect whether the compiler supports the SHA3 extension.

    The detected compiler support should be captured in _Makefile_ variables MLK_COMPILER_SUPPORTS_XXX. It **should not** be reflected in any CFLAGS. This is to avoid creating a dependency of the mlkem-native sources on our own build system.

     The feature detection for the compiler should be done by try-compilation.

2. Extend `test/mk/auto.mk` to check whether the _host CPU_ (the CPU on which mlkem-native is compiled) supports the aforementioned features. This should at least support any combination of {x86_64, AArch64} x {MacOS, Linux}. The results should be captured in MLK_HOST_SUPPORTS_XXX, following the same naming as MLK_COMPILER_SUPPORTS_XXX. Then, if for any given feature XXX, both MLK_COMPILER_SUPPORTS_XXX _and_ MLK_HOST_SUPPORTS_XXX are set, the CFLAGS should be amended with the appropriate architecture flag. The setting of `MLK_FORCE_XXX` directives should be kept. For platforms other than the ones listed above, the CFLAGS should not be further amended -- for those, one relies on the default compiler configuration being tailored to the target CPU.

3. Add a `make host_info` target to the main `Makefile` which summarizes the host-compiler/CPU feature support. This will require unconditionally importing `auto.mk` (unguarded by AUTO=0/1), but instead guarding the extension of the CFLAGS by `AUTO=1`.

hanno-becker:
Fixes #1144

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,8 @@
 	build test all \
 	clean quickcheck check-defined-CYCLES \
 	size_512 size_768 size_1024 size \
-	run_size_512 run_size_768 run_size_1024 run_size
+	run_size_512 run_size_768 run_size_1024 run_size \
+	host_info
 
 SHELL := /bin/bash
 .DEFAULT_GOAL := build
@@ -26,6 +27,8 @@ all: build
 W := $(EXEC_WRAPPER)
 
 include test/mk/config.mk
+include test/mk/compiler.mk
+include test/mk/auto.mk
 include test/mk/components.mk
 include test/mk/rules.mk
 
@@ -85,6 +88,15 @@ acvp_768:  $(MLKEM768_DIR)/bin/acvp_mlkem768
 acvp_1024: $(MLKEM1024_DIR)/bin/acvp_mlkem1024
 	$(Q)echo "  ACVP       ML-KEM-1024:  $^"
 acvp: acvp_512 acvp_768 acvp_1024
+
+ifeq ($(HOST_PLATFORM),Linux-aarch64)
+# valgrind does not work with the AArch64 SHA3 extension
+# Use armv8-a as the target architecture, overwriting a
+# potential earlier addition of armv8.4-a+sha3.
+$(MLKEM512_DIR)/bin/test_stack512:   CFLAGS += -march=armv8-a
+$(MLKEM768_DIR)/bin/test_stack768:   CFLAGS += -march=armv8-a
+$(MLKEM1024_DIR)/bin/test_stack1024: CFLAGS += -march=armv8-a
+endif
 
 stack_512: $(MLKEM512_DIR)/bin/test_stack512
 	$(Q)echo "  STACK      ML-KEM-512:   $^"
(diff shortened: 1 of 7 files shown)


test/mk/auto.mk
```diff
@@ -1,29 +1,113 @@
...
 endif
+endif # x86_64
+
+# AArch64 CFLAGS configuration
+ifeq ($(ARCH),aarch64)
+CFLAGS += -DMLK_FORCE_AARCH64
+
+# Add SHA3 flags only if both compiler and host support it
+ifeq ($(MK_COMPILER_SUPPORTS_SHA3)$(MK_HOST_SUPPORTS_SHA3),11)
+CFLAGS += -march=armv8.4-a+sha3
```
mkannwischer, line 87: I wonder if this overwrites user-provieded CFLAGS. For example, if I specifically call
```
 tests all --cflags="-march=armv8.4-a"
```
That ideally shouldn't be overwritten by the auto detection?

test/mk/compiler.mk
```diff
@@ -0,0 +1,65 @@
...
+#
+# Compiler feature detection for mlkem-native
+# This file detects whether the compiler supports various architectural features
+# used by mlkem-native through compile-time tests with C code containing inline assembly.
+#
+# Feature detection can be overridden by setting the corresponding variable on the command line:
+#   make MK_COMPILER_SUPPORTS_SHA3=0    # Disable SHA3 detection
+#   make MK_COMPILER_SUPPORTS_AVX2=0    # Disable AVX2 detection
+#   make MK_COMPILER_SUPPORTS_BMI2=0    # Disable BMI2 detection
+#   make MK_COMPILER_SUPPORTS_SSE2=0    # Disable SSE2 detection
```
mkannwischer, line 11: Do we want to prefix those with the architecture? It is not unimaginable that X86 introduces SHA3 instructions.

Makefile
```diff
@@ -176,6 +188,30 @@ run_size: \
...
+# Display host and compiler feature detection information
+# Shows which architectural features are supported by both the compiler and host CPU
+# Usage: make host_info [AUTO=0|1] [CROSS_PREFIX=...]
+host_info:
+	@echo "=== Host and Compiler Feature Detection ==="
+	@echo "Host Platform: $(HOST_PLATFORM)"
+	@echo "Target Architecture: $(ARCH)"
+	@echo "Compiler: $(CC)"
+	@echo "Cross Prefix: $(if $(CROSS_PREFIX),$(CROSS_PREFIX),<none>)"
+	@echo "AUTO: $(AUTO)"
```
mkannwischer, line 200: I must be doing something wrong here: 
```
$ nix develop #ci-cross-riscv-64
$ CROSS_PREFIX=riscv64-unknown-linux-gnu- make host_info
```

Gives me 
```
=== Host and Compiler Feature Detection ===
Host Platform: Linux-aarch64
Target Architecture: aarch64
Compiler: riscv64-unknown-linux-gnu-gcc
Cross Prefix: <none>
AUTO: 1

=== AArch64 Feature Support ===
SHA3: Host ❌ Compiler ❌
```

I am not sure how you want to support cross compilation here, but the Cross Prefix suggests that you do intend to support it in some way.

mkannwischer: Okay I found the issue - this works fine:
```
$ make host_info  CROSS_PREFIX=riscv64-unknown-linux-gnu-
```

and outputs
```
=== Host and Compiler Feature Detection ===
Host Platform: Linux-aarch64
Target Architecture: riscv64
Compiler: riscv64-unknown-linux-gnu-gcc
Cross Prefix: riscv64-unknown-linux-gnu-
AUTO: 1

=== Architecture Not Supported ===
No specific feature detection available for riscv64
```

But in CI we do set the CROSS_PREFIX in a environment variable, so that should be supported.

test/mk/auto.mk
mkannwischer: Okay, there seems to be no good way around this. Maybe we at least consistently set `AUTO=0` in CI when we set -march= manually?

mkannwischer (changes requested):
Thanks @hanno-becker. Mostly looks good. 
I ran into some problems when CROSS_PREFIX is passed as an environment variable (like we do in CI).

test/mk/compiler.mk
hanno-becker: We can add that when that happens.

test/mk/auto.mk
hanno-becker: Yes, when archflags are provided one shouldn't use auto. We could try to remove the `-march` flags from the CI -- it should not work correctly by itself... or so the theory.

test/mk/compiler.mk
mkannwischer: Okay.

Makefile
hanno-becker: Good spot.

hanno-becker: I can't reproduce this. @mkannwischer Did you run `make clean` before you did this?

hanno-becker: I think this is an issue with our auto-clean mechanism. It does not seem to work with CROSS_PREFIX when used as an environment variable.

hanno-becker: I created #1150 but this is unrelated to this PR.

mkannwischer: Thanks - that indeed explains it.

Issue #1148, opened 2025-08-07
Runtime dispatch based on CPU capabilities


dkostic:
**Platform**: 
Which platform does this concern?
- [ ] platform independent
- [x] aarch64
- [x] x86_64
- [x] rv64
- [ ] other

**Issue category**:

Which part(s) of mlkem-native does this issue concern:

- [x] core crypto: ML-KEM
- [ ] core crypto: FIPS202
- [ ] documentation
- [ ] integration
- [ ] CBMC
- [ ] CI
- [ ] Testing

**Describe the solution you'd like**
As part of the integration into AWS-LC we are trying to come up with a solution for runtime dispatching based on CPU capabilities. The CPUs that AWS-LC is built on and run on are not necessarily the same so we have to either generate only fully portable code or generate code that uses specific CPU extensions but checks at runtime if the extension is available and falls back to the portable code if not. In mlkem-native case this is an issue for x86 in particular, because the x86 native implementation requires AVX2 extension. But the same will apply to any native backend that might use an extension or a feature of a CPU that's not available on all CPUs for that particular architecture.

I implemented two proof-of-concepts how would that look like for mlkem-native: frontend and backend dispatch. Before going further with the implementation I wanted to ask for feedback and try to align on the way forward.

For the PoCs I reimplemented only one function (`mlk_poly_mulcache_compute`) but the same would have to be done for all other functions (and backends) that have both reference and native implementation. Here is the code:
- frontend dispatch: https://github.com/dkostic/mlkem-native/commit/473c33d4eab9d8cd49a66bf63cfe8f4d2983bad4
- backend dispatch: https://github.com/dkostic/mlkem-native/commit/e506c8985ce408a9f958cc514270d37f563ba868

The main commonality between the two approaches is a new function `mlk_is_native_capable` that is used to determine if the native or the reference code path should be followed. This function can be configured/customized by a user of mlkem-native in exactly the same way as `mlk_randombytes`. The main difference is where the dispatch is placed -- backend or frontend.

My preference is the _frontend_ dispatch. The main reason is that if we choose the backend dispatch then we would have to implement the dispatch in every supported backend (on top of the changes required in the frontend, see the examples for details). The frontend dispatch avoids this complexity and doesn't burden the backends.

I'm looking forward to any feedback you might have!

**Additional context**

mkannwischer:
Thanks @dkostic for bringing this up and for the PoC.

[...]

PR #1152, opened 2025-08-12
Runtime dispatch based on custom CPU capabilities function

dkostic:
Addresses #1148, and as agreed therein, the dispatch logic is implemented in the backends.


mlkem/src/config.h
```diff
@@ -311,6 +311,29 @@
...
+ *              own CPU capability function that can be used by the backend
+ *              to perform runtime dispatch between implementations that
+ *              require different CPU capabilities.
+ *
+ *              If this option is not set, mlkem-native expects a function
+ *              void randombytes(uint8_t *out, size_t outlen).
+ *
+ *              Set this option and define `mlk_randombytes` if you want to
+ *              use a custom method to sample randombytes with a different name
+ *              or signature.
```
mkannwischer: This needs to be adjusted.

mlkem/src/native/x86_64/meta.h
```diff
@@ -43,131 +44,275 @@ static MLK_INLINE int mlk_rej_uniform_native(int16_t *r, unsigned len,
...
 {
-  mlk_ntt_avx2((__m256i *)data, mlk_qdata.vec);
+  if (!mlk_is_native_capable())
+  {
+    return -1;
+  }
+  else
+  {
+    mlk_ntt_avx2((__m256i *)data, mlk_qdata.vec);
+    return 1;
```
mkannwischer: Is there a reason for using 1 as a success return code? I'm more used to using 0 to signal success.

dev/x86_64/meta.h
```diff
@@ -32,6 +32,7 @@
 #if !defined(__ASSEMBLER__)
 #include <string.h>
 #include "../../common.h"
+#include "../../native_capability.h"
 #include "src/arith_native_x86_64.h"
 
 static MLK_INLINE void mlk_poly_permute_bitrev_to_custom(int16_t data[MLKEM_N])
```
mkannwischer, line 38: Why is the mlk_nttunpack_avx2 call not guarded by `mlk_is_native_capable`?

mkannwischer: If AVX2 is not supported, it should be a no-op (because NTT and INTT won't be expecting the custom ordering anyway).

mkannwischer (changes requested):
Thanks @dkostic! This looks excellent. Some minor comments/questions.

Before we merge this, we should definitely add a test configuration that hardcodes `mlk_is_native_capable` to zero to test this in CI.
There are already some custom configuration tests [here](https://github.com/pq-code-package/mlkem-native/blob/51d641ad07f0df8b61a2aafbb553d67d1523e685/.github/workflows/ci.yml#L451), so this should be easy. I can help with adding that if you want.

mkannwischer:
CBMC proofs appear to blow up. That's something that also needs to be fixed before we can merge. 
@rod-chapman, could you maybe help with that?

mlkem/src/native/x86_64/meta.h
hanno-becker: The toplevel API uses `0` as the return code for success, which I'd suggest to stick with.

hanno-becker (changes requested):
Can we keep the original C functions as `foo_c()` and then have something like
```
void foo() {
   if (foo_native() == 0) {
      return;
   }
   foo_c();
}
```
for the dispatch? That modularizes the reasoning.

hanno-becker:
On the other hand, having a dedicated function for the dispatch creates a lot of churn for the CBMC proofs because a lot needs renaming, and additional proofs be created for wrappers. I am undecided.

Why are the CBMC proofs currently timing out? Can this be investigated first?

rod-chapman:
Strongly agree with Hanno. Please apply the following principles. Imagine we have a function called "foo" that needs to dispatch and _might_ need to fall back on the original C code at runtime.

1. Rename the original C code from foo() to foo_c(). You'll need to modify its proof - rename the proof directory, and update the Makefile, harness and so on. Everything else about foo_c() should be unchanged. Check that the CBMC proof of foo_c() still works. The big idea is is to leave the original reference code unmodified, and to strictly separate the dispatching logic from the functional code.
2. Add foo() as suggested by Hanno above.
3. foo_c() and foo_native() should have _identical_ CBMC contracts, except that foo_native can return "0" meaning "success" or "1" meaning "can't do that... please use the C back-end instead".

foo() should look something like this:
```
void foo() {
#if defined(MLK_USE_NATIVE_FOO)
   if (foo_native() == 0) {
      return;
   }
#endif
   foo_c();
}
```
so that platforms with some sort of native back-end will try that first, but platforms with NO native back-end will just call foo_c() straight away, right?

Does that make sense?

hanno-becker:
@rod-chapman This is what I mean in the review comment.

However, because it creates a fair amount of additional churn, I would suggest that we at least pick one of the more complex functions (NTT?) and fix the CBMC proof for the code _as it is_, to see if it is really an issue. Can you do that?

mkannwischer:
> 3. foo_c() and foo_native() should have _identical_ CBMC contracts, except that foo_native can return "0" meaning "success" or "1" meaning "can't do that... please use the C back-end instead".

We should stick to `-1` for the latter, because rejection sampling will return >= 0 in case of success.

> 
> Does that make sense?

@rod-chapman, @hanno-becker, can you elaborate a bit on what the benefit would be of splitting of the C part into it's own function? I see no immediate benefit of doing that and it is _a lot_ of work. For CBMC, it should look exactly the same? The resulting binary will look the same too.

rod-chapman:
Agreed on return codes - they should consistent with other functions.

rod-chapman:
I have corrected the proofs of proofs/cbmc/*native_aarch64 (9 functions). The harnesses needed to be updated to reflect that the called functions now return "int" rather than "void."

There are now at least 4 more functions (matvec_mul, poly_invntt_tomont_native, poly_ntt_native, and polyvec_basemul_montgomery_cached_native) to look at where the proofs are running, but blowing up in complexity.

rod-chapman:
I know what the problem is - at least with poly_ntt(). In short - when the call to poly_ntt_native() _fails_, we need to know that the polynomial p[] _hasn't_ changed at all, so that the subsequent loop invariant and the pre-condition on the call to ntt_layer() can be satisfied.  In short - poly_ntt_native() needs a stronger post-condition. Working on it now.

rod-chapman:
OK. I have fixed the "native" variant of the proof of mlk_poly_ntt(). See the most recent commit to see how it's done.

mlkem/src/config.h
dkostic: done

mlkem/src/native/x86_64/meta.h
dkostic: changed to -1 and 0 for fail/success.

dkostic:
@mkannwischer @hanno-becker I addressed your comments, and @rod-chapman helped out with CBMC. Let me know if you have any additional questions/concerns.

dkostic:
two issues with the CI (at the time of writing):
1. https://github.com/pq-code-package/mlkem-native/actions/runs/16950260497/job/48041457676?pr=1152. This one seems pretty random, not sure what's the problem.
2. https://github.com/pq-code-package/mlkem-native/actions/runs/16950260497/job/48041456645?pr=1152. This one also seems unrelated, but at least there is an error message "warning: download buffer is full; consider increasing the 'download-buffer-size' setting".

mlkem/src/compress.c
```diff
@@ -40,6 +39,15 @@ void mlk_poly_compress_d4(uint8_t r[MLKEM_POLYCOMPRESSEDBYTES_D4],
                           const mlk_poly *a)
 {
   unsigned i;
+#if defined(MLK_USE_NATIVE_POLY_COMPRESS_D4)
+  int ret;
+  mlk_assert_bound(a, MLKEM_N, 0, MLKEM_Q);
+  ret = mlk_poly_compress_d4_native(r, a->coeffs);
+  if (ret != -1)
```
rod-chapman: Hopefully my final comment.
You repeat this pattern:
```
if (ret != -1)
```
many times. I find the negative test here ("If not failure then...") far less easy to read and understand. I would naturally write:
```
if (ret == 0)
```
(i.e. "if success then...")

I have always found that positive-sense tests for "good thing" is easier to understand than negating a test for a bad thing. Is there some good reason for that pattern?

rod-chapman:
I ran all proofs with K=2,3,4 on Grv3/Ubuntu and all OK. The proof of `polyvec_basemul_acc_montgomery_cached_native` is a bit slow, but I can live with that for now. It will be improved when we come back to consider refining the output bounds of that function anyway.

mlkem/src/compress.c
dkostic: I followed the one existing example of "native-then-reference code" case: https://github.com/pq-code-package/mlkem-native/blob/main/mlkem/src/sampling.c#L128. But I guess we can do better here and add `NATIVE_FUNC_SUCCES/FAIL` macros and then check `ret == NATIVE_FUNC_SUCCES`. Would that work for you?

dkostic:
> Before we merge this, we should definitely add a test configuration that hardcodes mlk_is_native_capable to zero to test this in CI.

@mkannwischer I added the test. Could you please check if that's correct and the test actually run?

dkostic:
Everything is passing now except this one that keeps saying "download buffer is full": https://github.com/pq-code-package/mlkem-native/actions/runs/16975743408/job/48128706572?pr=1152. I tried restarting it twice, but no success so far.

dkostic:
@mkannwischer @hanno-becker @rod-chapman the CI is passing now and I think I addressed all your comments. Let me know if you have any additional questions/concerns!

hanno-becker (changes requested):
* I believe the CBMC proofs for `xxx_native()` must work with the contract for `mlk_is_native_capable`, otherwise the hardcoded default of `1` will render the `FAIL` branch unreachable.
* If possible, I would like to see the native API strengthened to include that for failure, the output buffer is unchanged. I think @rod-chapman you have experimented with this already?

Also, can we make the name of the native capability function more specific? I think there should be one backend capability function per feature (set) that the backend relies on? Starting with AVX2, we could have `mlk_avx2_capable()` or similar.

rod-chapman:
1. Yes - mlk_is_native_capable() needs a contract and must not be inlined to avoid the hard-coded value. Good catch.
2. I tried the "whole buffer is unchanged" thing yesterday but couldn't work out how to express that in CBMC contracts in way that worked. I will try again this morning, write a simple test case, and ask for help.
3. Function that (logically) return a Boolean value should return "1" for True and "0" for False in line with C's standard
encoding of the non-type bool.

dkostic:
@hanno-becker I tried to move the capability function to `sys.h` and tried different combinations of flags to make it work with all build flavors in `examples`, but I can't make it work. I think the main issue is this: https://github.com/pq-code-package/mlkem-native/blob/51d641ad07f0df8b61a2aafbb553d67d1523e685/scripts/autogen#L1260. Let me know if you have an idea how to make it work, otherwise, I'll revert back to a separate file.

dkostic:
> Also, can we make the name of the native capability function more specific? Should there not be one capability function per feature (set) that the backend relies on? E.g. for AVX2, we could have mlk_avx2_capable() or similar.

From our previous chats I understood that you don't want that level of detail in mlkem-native, but I can do it nevertheless. But let's see first if it's possible to have the functions in `sys.h` because that would simplify the implementation (otherwise, I would have to add a file per backend for capability functions).

hanno-becker:
@dkostic I am sorry if there has been a misunderstanding. I am indeed concerned about adding platform-specific logic for runtime capability detection to mlkem-native, but this is mostly about the implementation.

I am having a look at the issue you found with `sys.h`.

hanno-becker:
@dkostic I think the issue is fixed.

@rod-chapman Thanks for strengthening the specs. I added a macro to encapsulate the 'array unchanged' logic.

dev/x86_64/meta.h
hanno-becker: This has been addressed.

hanno-becker (approved):

PR #1156, opened 2025-08-19
Remove unnecessary usage of __m256i from x86 backend C code

dkostic:
The AVX2 functions are unnecessarily using `__m256i*` as function arguments. This makes the integration into aws-lc painful because: 1) aws-lc compiles C files _without_ platform specific flags like `-mavx2` so that the generated code is completely portable; 2) aws-lc supports compilers that _can't_ compile avx2 C code without the `-mavx2` flag. In this commit we simply refactor the code to remove all the unnecessary usage of "__m256i*". Note that `compress_avx2.c` still uses avx2 C intrinsics but that's not a problem for aws-lc integration at the moment because we will exclude that file for now.


mlkem/src/native/x86_64/src/consts.h
```diff
@@ -36,10 +36,9 @@
 #define MLK_AVX2_BACKEND_DATA_OFFSET_MULCACHE_TWIDDLES 640
 
 #ifndef __ASSEMBLER__
-#include "align.h"
-typedef MLK_ALIGNED_INT16(768) qdata_t;
+typedef int16_t qdata_t[768];
```
hanno-becker, line 39: Why was this struct wrapper removed?

dev/x86_64/src/consts.c
```diff
@@ -43,7 +42,7 @@
 #define MLK_AVX2_MASK 4095
 #define MLK_AVX2_SHIFT 32
 
-const qdata_t mlk_qdata = {{
+const __attribute__((aligned(32))) qdata_t mlk_qdata = {
```
hanno-becker: We should use the `MLK_ALIGN` primitive here

hanno-becker: Done

hanno-becker (approved):
LGTM

PR #1159, opened 2025-08-20
[TEST] Check performance impact of compress functions in AVX2

hanno-becker:


dkostic:
I tested this last week: https://github.com/pq-code-package/mlkem-native/pull/1149 :)

PR #1162, opened 2025-08-21
ASM: Add `L` prefix to all local labels

### LINKED ISSUE #1161: Prefix all internal assembly labels with L to indicate their local-ness
* Required for: #1160 

For every assembly file in mlkem-native, internal labels should be prefixed with `L`. This applies to both aarch64 and x86_64 backends, and all of clean, optimized, simplified, and raw versions of the assembly. Scripts referring to those labels need to be adjusted accordingly, e.g. the SLOTHY Makefile.

hanno-becker:
Fixes #1161


hanno-becker:
Unfortunately, this breaks `simpasm` because the symbolic branch target resolution no longer works.

PR #1166, opened 2025-08-22
Add CFI information to assembly

### LINKED ISSUE #1160: Call Frame Information (CFI) directives missing in assembly functions which breaks stack unwinding
**Describe the bug**
The assembly files are missing CFI directives which breaks stack unwinding. This causes back traces to be broken as the stack unwinding can't figure out what called the current function. This also breaks profiler's ability to attribute time spent in the function to the correct call stack. Broken profiling is the bigger concern for our customers because it impacts their ability to understand the impact of cryptographic operations in their application.

The problem appears to be worse on Arm because the SHA3 assembly is also missing CFI directives, we're working with s2n-bignum to fix that https://github.com/awslabs/s2n-bignum/issues/197, but you can see the mlkem-native code call stack is wrong, all of those functions are not called from bssl directly, they are called from SpeedKem:

Image

Image

**Platform**: All platforms with assembly implementations, at least aarch64 and x86_64

**Compiler**: All

**mlkem-native version**: main-branch

**How to reproduce**
Debug issue:
1. install GDB and build mlkem-native/aws-lc
2. run gdb
3. set a breakpoint in an assembly function: `break ntt_layer123_start`
4. Run the program and get a backtrace with `bt`
```
(gdb) bt
#0  ntt_layer123_start () at /home/ec2-user/aws-lc/crypto/fipsmodule/ml_kem/mlkem/native/aarch64/src/ntt.S:90
#1  0x00000000004fef98 in mlk_ntt_native (data=0xffffffffe4c0) at /home/ec2-user/aws-lc/crypto/fipsmodule/ml_kem/mlkem/../mlkem/native/aarch64/meta.h:30
#2  mlkem_poly_ntt (p=0xffffffffe4c0) at /home/ec2-user/aws-lc/crypto/fipsmodule/ml_kem/mlkem/poly.c:397
#3  mlkem512_polyvec_ntt (r=<optimized out>) at /home/ec2-user/aws-lc/crypto/fipsmodule/ml_kem/mlkem/poly_k.c:115
#4  mlkem512_indcpa_keypair_derand (pk=0xffffffffec60 "p\354\377\377\377\377", sk=0xffffffffde68 "\350\345n", coins=<optimized out>)
    at /home/ec2-user/aws-lc/crypto/fipsmodule/ml_kem/mlkem/indcpa.c:377
#5  0x0000000000000000 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

Profiling issue:
1. install perf and a profiler such as https://github.com/flamegraph-rs/flamegraph
2. build mlkem-native/aws-lc
3. profile the benchmark `flamegraph -- build/tool/bssl speed -filter ML-KEM`

**Expected behavior**
Stack unwinding works at all points in mlkem-native, back traces and profiling with existing tools work. Profiles show time spent with reasonable call stacks, e.g. main calls speed, speed calls mlkem high level api, high level api calls low level assembly functions.

**Additional context**
AWS-LC has a customer request https://github.com/aws/aws-lc/issues/2459 and we have been working to fix the our assembly files in PRs like https://github.com/aws/aws-lc/pull/2584.

hanno-becker:
* Resolves #1162 
* Resolves #1160


scripts/simpasm
```diff
@@ -297,27 +313,61 @@ def simplify(logger, args, asm_input, asm_output=None):
...
-            )
-            logger.error("I'll keep them around for you to have a look")
-            raise Exception("simpasm failed")
+
+        # When CFI is enabled, compare only the __text section content
+        if args.cfify:
+            logger.debug("Comparing __text section content for CFI comparison...")
+
+            # Extract __text section from both files
+            orig_text_cmd = [args.objdump, "-s", "--section=__text", tmp_objfile0]
```
hanno-becker: This needs to be improved. On Linux systems, the text section is `.text`.

hanno-becker: Done

proofs/hol_light/arm/mlkem/keccak_f1600_x4_v8a_v84a_scalar.S
```diff
@@ -200,6 +226,7 @@ keccak_f1600_x4_v8a_v84a_scalar_hybrid_initial:
         eor	x5, x5, x27
         shl	v0.2d, v31.2d, #0x3e
         ldr	x27, [sp, #0xd0]
+        .cfi_restore x27
```
mkannwischer: If I understand correctly then this means that x27 is restored to what it was prior to the function call, but that is not true?
Should the cfi directives be dropped for the local stack spilling and only be kept for the spilling of the callee-saved registers?

mkannwischer: If my understanding is correct: Does it hurt to keep it here?

mkannwischer (dismissed):

proofs/hol_light/arm/mlkem/keccak_f1600_x4_v8a_v84a_scalar.S
hanno-becker: > If I understand correctly then this means that x27 is restored to what it was prior to the function call, but that is not true?
> Should the cfi directives be dropped for the local stack spilling and only be kept for the spilling of the callee-saved registers?

I think you are right, good spot.

scripts/cfify
```diff
@@ -0,0 +1,280 @@
...
+                continue
+
+        result.append(line)
+        i += 1
+
+    return "\n".join(result)
+
+
+def main():
+    import re
```
mkannwischer: This is already imported at the start of the file?

mkannwischer (approved):
Thanks @hanno-becker! It is not pretty, but it does work well. All asm files are covered and the annotations look good!

I'd prefer an option where the annotations are added to the dev/* files and then pulled through the simpasm step, but that seems infeasible for now.

hanno-becker:
> I'd prefer an option where the annotations are added to the dev/* files and then pulled through the simpasm step, but that seems infeasible for now.

Me, too, and I hope we can get to that in time.

dkostic (approved):
I checked the x86 part, looks good to me!

PR #1178, opened 2025-09-05
x86 assembly: broadcast constants instead of loading from memory

dkostic:
Rewrite some of the x86 backend implementations to broadcast 
16-bit constants instead of loading them from memory.

This significantly simplifies the formal verification effort.


dkostic:
oh wow, I did not expect change in performance at all, let alone 70% slowdown. I'll investigate next week.

dkostic:
the huge performance regression of the first version of this PR was because of mixing VEX and non-VEX encoded instructions. All I needed to do to fix that is change `movd` to `vmovd`. These are equivalent instructions but `movd` is using legacy SSE encoding, while `vmovd` is using VEX (AVX2) encoding. If you are curious, here is more details: https://www.intel.com/content/dam/develop/external/us/en/documents/11mc12-avoiding-2bavx-sse-2btransition-2bpenalties-2brh-2bfinal-809104.pdf

dkostic:
@mkannwischer @hanno-becker I fixed the performance issues with this PR (see https://github.com/pq-code-package/mlkem-native/pull/1178#issuecomment-3357724030). Please let me know if you have any additional comments or concerns.

hanno-becker (changes requested):
Sorry for the delay @dkostic.

The change overall looks good to me, and is in line with the AArch64 backend. Thank you!

It looks like you can now remove a lot of constants from `consts.c`, or? What, if anything, needs to remain beyond the NTT / invNTT tables?

Also, can you squash the movd -> vmovd fixup into the main commit?

dkostic:
> Sorry for the delay @dkostic.
> 
> The change overall looks good to me, and is in line with the AArch64 backend. Thank you!
> 
> It looks like you can now remove a lot of constants from `consts.c`, or? What, if anything, needs to remain beyond the NTT / invNTT tables?
> 
> Also, can you squash the movd -> vmovd fixup into the main commit?

Thanks @hanno-becker ! Right, I was going to clean up the constants table next but since you already noticed it I did it in this PR. I also squashed the commits to have a single one.

dev/x86_64/src/consts.c
```diff
@@ -24,229 +24,23 @@
...
-/* check-magic: -10079 == signed_mod(MLK_AVX2_QINV*MLK_AVX2_FHI,2^16) */
-#define MLK_AVX2_FLO -10079
-/* check-magic: 1353 == pow(2, 32, MLKEM_Q) */
-#define MLK_AVX2_MONTSQHI 1353
-/* check-magic: 20553 == signed_mod(MLK_AVX2_QINV*MLK_AVX2_MONTSQHI,2^16) */
-#define MLK_AVX2_MONTSQLO 20553
-#define MLK_AVX2_MASK 4095
-#define MLK_AVX2_SHIFT 32
-
 MLK_ALIGN const int16_t mlk_qdata[768] = {
```
hanno-becker: Blocker: This needs to be adjusted.

There must be a linter that would catch this...? @mkannwischer

dev/x86_64/src/consts.h
```diff
@@ -21,19 +21,18 @@
...
-#define MLK_AVX2_BACKEND_DATA_OFFSET_REVIDXB 128
-#define MLK_AVX2_BACKEND_DATA_OFFSET_REVIDXD 144
-#define MLK_AVX2_BACKEND_DATA_OFFSET_ZETAS_EXP 160
-#define MLK_AVX2_BACKEND_DATA_OFFSET_16XSHIFT 624
-#define MLK_AVX2_BACKEND_DATA_OFFSET_MULCACHE_TWIDDLES 640
+#define MLK_AVX2_Q MLKEM_Q
+/* check-magic: -1044 == pow(2,16,MLKEM_Q) */
+#define MLK_AVX2_MONT -1044
+/* check-magic: -3327 == pow(MLKEM_Q,-1,2^16) */
+#define MLK_AVX2_QINV -3327
```
hanno-becker: Those are unused and can be removed

dev/x86_64/src/consts.h
```diff
@@ -21,19 +21,18 @@
...
+/* check-magic: 20159 == round(2^26/MLKEM_Q) */
+#define MLK_AVX2_V 20159
```
hanno-becker: This is only used in `compress_avx2.c` and should now be defined there?

hanno-becker (changes requested):
The length of the constant table hasn't been adjusted, and there are a few dead constants that should be removed or moved.

dev/x86_64/src/consts.c
dkostic: done

dev/x86_64/src/consts.h
dkostic: done

dev/x86_64/src/consts.h
dkostic: done

hanno-becker (approved):

PR #1181, opened 2025-09-08
Test: Switch from actions cache to buildjet cache

mkannwischer:
BuildJet offers 20 GiB cache for free, while Github actions only provides 10 GiB.


hanno-becker:
@mkannwischer (How) Do you want to validate further if the buildjet backend works correctly?

mkannwischer:
I have no bandwidth to consider this further for now. Closing.

PR #1188, opened 2025-09-12
Add unit test target

hanno-becker:
* Based on #1195 

This commit adds a new test type `unit` for unit tests,
accessible via `make [run_]unit_{512,768,1024}` and via
`tests unit ...` or `tests all --unit`.

While the expectation is that the unit test framework will
grow over time, we start with two things:

- We move the existing non-overflow tests for compression routines
  from test/test_mlkem.c -- which underlies the `func` test target --
  to the new unit test target.

- We add backend tests which, if an arithmetic backend is enabled,
  test that the backend behaves like the C default implementation
  (usually after modular reduction).

To enable the testing of static functions, we introduce the
keyword MLK_STATIC_TESTABLE. This unfolds to `static` by default,
but can be overwritten in the CFLAGS. For the unit test target,
it is simply set to the empty string, causing functions designated
as MLK_STATIC_TESTABLE to get external linkage.

Unfortunately, this requires a separate library built (rather than
reusing the library build used by the `func` target), but there does
not seem a clean way around this.

Over time, we can migrate functions currently marked `static` over
to MLK_STATIC_TESTABLE. We do not make an attempt here to do this,
but focus on the C backend functionality only.

Each backend unit test conducts both randomized and targeted tests:
It usually first tests conformance of C and native backend on some
simple inputs (such as all-0, or exactly one coefficient non-zero),
and then moves to a configurable number of random tests. Since the
backend may have different algorithms for modular arithmetic, we
cannot assume exact equality of outputs, and apply modular reduction
prior to comparison accordingly.


test/test_mlkem.c
```diff
@@ -147,71 +147,6 @@ static int test_invalid_ciphertext(void)
  * this ensures that no buffer overflow is happening. This is of interest
  * because the compressed buffers sometimes have unaligned lengths and
  * are therefore at risk of being overflowed by vectorized code. */
```
mkannwischer: Remove this comment

test/test_unit.c
```diff
@@ -0,0 +1,693 @@
...
+
+#include "notrandombytes/notrandombytes.h"
+
+#include "../mlkem/src/compress.h"
+#include "../mlkem/src/poly.h"
+#include "../mlkem/src/poly_k.h"
+
+#ifndef NUM_RANDOM_TESTS
+#ifdef MLKEM_DEBUG
+#define NUM_RANDOM_TESTS 1000
```
mkannwischer, line 17: What is the rationale behind doing fewer random tests in a debug build? I would expect the opposite?

test/test_unit.c
```diff
@@ -0,0 +1,693 @@
...
+
+  for (pos = 0; pos < MLKEM_N; pos += MLKEM_N / 8)
+  {
+    generate_i16_array_single(test_data, MLKEM_N, pos, 1);
+    CHECK(test_poly_reduce_core(test_data, "poly_reduce_single") == 0);
+  }
+
+  for (i = 0; i < NUM_RANDOM_TESTS; i++)
+  {
+    generate_i16_array_ranged(test_data, MLKEM_N, INT16_MIN, INT16_MAX);
```
mkannwischer: Can we stick to keeping the lower bound inclusive, and the upper bound exclusive as for the assertions?

test/test_unit.c
```diff
@@ -0,0 +1,693 @@
...
+  mlk_poly ref_poly;
+  int native_ret;
+
+  memcpy(native_result, input, MLKEM_N * sizeof(int16_t));
+  memcpy(ref_poly.coeffs, input, MLKEM_N * sizeof(int16_t));
+
+  native_ret = mlk_poly_tomont_native(native_result);
+  if (native_ret == MLK_NATIVE_FUNC_FALLBACK)
+  {
+    return 0; /* Skip if not supported */
```
mkannwischer: Can we output something in case the test is skipped? Just to catch it if we skip all the tests.

test/test_unit.c
```diff
@@ -0,0 +1,693 @@
...
+static int test_ntt_core(const int16_t *input, const char *test_name)
+{
+  int16_t native_result[MLKEM_N];
+  mlk_poly ref_poly;
+  int native_ret;
+
+  memcpy(native_result, input, MLKEM_N * sizeof(int16_t));
+  memcpy(ref_poly.coeffs, input, MLKEM_N * sizeof(int16_t));
+
+  native_ret = mlk_ntt_native(native_result);
```
mkannwischer: Should we add run-time assertions to the native functions to make sure the output bounds are also correct?

test/test_unit.c
mkannwischer: Actually should this ever occur? We check for support using the macros.

test/test_unit.c
hanno-becker: Good point. We should simply call the toplevel NTT because that will a) call the backend and b) include the bounds (in debug mode).

test/test_unit.c
hanno-becker: Yes, the runtime capability checks could fail, or for any other reason the input might be rejected.

test/test_unit.c
hanno-becker: Ok.

test/test_unit.c
hanno-becker: Because in debug build you have runtime assertions which make everything much slower.

test/test_mlkem.c
hanno-becker: Good spot

PR #1190, opened 2025-09-16
DO NOT MERGE: Add KyberSlash vulnerability to check ct tests

mkannwischer:


mkannwischer:
Since Github will eventually erase the logs, here are some screenshots: 

image
image

All tested versions of gcc (4.8, 4.9, 7, 11, 12, 13, 14, 15) lead to variable-latency errors on AArch64 and x86_64 when using -Os.
All tested versions of clang (14, 15, 16, 17, 18, 19, 20, 21) lead to variable-latency errors on AArch64 and x86_64 when using -Oz

PR #1191, opened 2025-09-16
Introduce explicit upper bounds on lengths of buffers

### LINKED ISSUE #1189: CBMC: Make buffer bounds explicit
Whenever `memory_no_alias` (`__CPROVER_is_fresh` under the hood) is used, CBMC implicitly assumes an internal bound on object sizes (~2^53 in our setup) even if no such bound is assumed in the code.

CBMC offers the option `--malloc-may-fail --malloc-fail-assert` to instead _assert_ upon use of `memory_no_alias`, the provided size does not exceed the maximum object size. This forces us to explicitly state upper bounds on object length, where they are not static.

The Makefile.common should be modified to always apply `--malloc-may-fail --malloc-fail-assert`, and the CBMC contracts adjusted if necessary. Note that even though the above parameters are documented for `cbmc`, it seems they need to be applied to `goto-instrument`, so best to amend both `CBMCFLAGS` and `GOTO_INSTRUMENT`.

rod-chapman:
Fixes #1189 

Introduce explicit upper bounds on lengths of buffers input and output buffers where appropriate.

Force CBMC to use --malloc-fail-assert for all proofs to remove assumption on buffer lengths.

### DIFF
--- a/mlkem/src/fips202/fips202x4.c
+++ b/mlkem/src/fips202/fips202x4.c
@@ -28,6 +28,7 @@ static void mlk_keccak_absorb_once_x4(uint64_t *s, uint32_t r,
                                       const uint8_t *in2, const uint8_t *in3,
                                       size_t inlen, uint8_t p)
 __contract__(
+  requires(inlen <= MLK_MAX_BUFFER_SIZE)
   requires(memory_no_alias(s, sizeof(uint64_t) * MLK_KECCAK_LANES * MLK_KECCAK_WAY))
   requires(r <= sizeof(uint64_t) * MLK_KECCAK_LANES)
   requires(memory_no_alias(in0, inlen))
@@ -78,7 +79,8 @@ static void mlk_keccak_squeezeblocks_x4(uint8_t *out0, uint8_t *out1,
                                         size_t nblocks, uint64_t *s, uint32_t r)
 __contract__(
     requires(r <= sizeof(uint64_t) * MLK_KECCAK_LANES)
-    requires(nblocks <= 8 /* somewhat arbitrary bound */)
+    requires(r == SHAKE128_RATE || r == SHAKE256_RATE)
+    requires(nblocks <= (MLK_MAX_BUFFER_SIZE / SHAKE256_RATE))
     requires(memory_no_alias(s, sizeof(uint64_t) * MLK_KECCAK_LANES * MLK_KECCAK_WAY))
     requires(memory_no_alias(out0, nblocks * r))
     requires(memory_no_alias(out1, nblocks * r))
--- a/mlkem/src/verify.h
+++ b/mlkem/src/verify.h
@@ -318,7 +318,8 @@ __contract__(ensures(return_value == (cond ? a : b)))
  *
  * Arguments:   const uint8_t *a: pointer to first byte array
  *              const uint8_t *b: pointer to second byte array
- *              size_t len:       length of the byte arrays
+ *              size_t len:       length of the byte arrays, upper-bounded
+ *                                to INT_MAX to control proof complexity
  *
  * Returns 0 if the byte arrays are equal, a non-zero value otherwise
  *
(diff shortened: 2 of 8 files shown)


hanno-becker (changes requested):
Thank you @rod-chapman for taking this on and identifying what needs adjusting.

I think it is a bit confusing to have a mix of multiple named constants and also hardcoded numerics.

Can you please introduce a single constant that is used whenever a contract works with an a priori unconstrained buffer length? For example, `MLK_MAX_BUFFER_SIZE`. It should be documented that it's to work around a CBMC limitation.

I believe proofs should work if this is defined as `2^52` on a 64-bit system, and `2^20` on a 32-bit system (though, as mentioned in #1192, we don't actually exercise that).

rod-chapman:
Imagine someone is running proofs on a 64-bit machine, but cross-compiling the code to a 32-bit machine where size_t is 32 bits. What should MLK_MAX_BUFFER_SIZE be in that case? I have seen customers come unstuck before with this kind of thing. I really don't like the idea that soundness of proofs depend on implementation-defined constants. I could define it as 2^20 once and for all. That should be fine for MLKEM, since we already know that 2048 (for input buffers) and 256 (for SHA3 output buffers) are fine.

hanno-becker:
> Imagine someone is running proofs on a 64-bit machine, but cross-compiling the code to a 32-bit machine where size_t is 32 bits. What should MLK_MAX_BUFFER_SIZE be in that case?

As mentioned, the constant would be different for 32-bit and 64-bit. I don't think we should use the 32-bit form unconditionally, because that poses a real restriction in the case of mldsa-native, where I don't want a separate constant.

> I really don't like the idea that soundness of proofs depend on implementation-defined constants.

I don't like it either, but we got to work with the limitations of CBMC.

mlkem/src/fips202/fips202x4.c
```diff
@@ -78,7 +79,8 @@ static void mlk_keccak_squeezeblocks_x4(uint8_t *out0, uint8_t *out1,
                                         size_t nblocks, uint64_t *s, uint32_t r)
 __contract__(
     requires(r <= sizeof(uint64_t) * MLK_KECCAK_LANES)
-    requires(nblocks <= 8 /* somewhat arbitrary bound */)
+    requires(r == SHAKE128_RATE || r == SHAKE256_RATE)
```
hanno-becker, line 82: Why was this changed? This does not seem necessary.

mlkem/src/fips202/fips202x4.c
```diff
@@ -78,7 +79,8 @@ static void mlk_keccak_squeezeblocks_x4(uint8_t *out0, uint8_t *out1,
                                         size_t nblocks, uint64_t *s, uint32_t r)
 __contract__(
     requires(r <= sizeof(uint64_t) * MLK_KECCAK_LANES)
-    requires(nblocks <= 8 /* somewhat arbitrary bound */)
```
hanno-becker, line 81: Why was this changed? This does not seem necessary.

mlkem/src/verify.h
```diff
@@ -338,9 +339,9 @@ __contract__(ensures(return_value == (cond ? a : b)))
 static MLK_INLINE uint8_t mlk_ct_memcmp(const uint8_t *a, const uint8_t *b,
                                         const size_t len)
 __contract__(
+  requires(len <= INT_MAX)
```
hanno-becker, line 342: Why is this not using `MLK_MAX_BUFFER_SIZE`?

hanno-becker (changes requested):
Noted two seemingly unrelated changes that should be reverted, I think. Also, one question about a deviation from the use of `MLK_MAX_BUFFER_SIZE` in the proof of `mlk_ct_memcmp`.

mlkem/src/fips202/fips202x4.c
rod-chapman: If the caller is only restricted to MLK_MAX_BUFFER_SIZE bytes, that's a lot more than 8 blocks I think. Proof fails otherwise.

mlkem/src/verify.h
rod-chapman: Because MLK_MAX_BUFFER_SIZE is 2^52(ish), so that means you'd have to use `size_t i;` rather than `unsigned i;` for the loop counter. If you make it `size_t` then the proof blows up. There are very few calls to this function with small, constant values for len (e.g. 32), so leaving the bound as INT_MAX and the loop counter as `unsigned` seems pragmatic.

hanno-becker: Ack!

mlkem/src/fips202/fips202x4.c
hanno-becker: Ack

hanno-becker (approved):

PR #1195, opened 2025-09-21
Hoist default C backend into separate functions

### LINKED ISSUE #1194: Hoist default C backend into separate functions
Each functionality that is currently replaceable by a custom backend is currently implemented in C in this fashion:
```c
if (foo_backend() == SUCCESS) {
   return;
}

/* Default C implementation */
```

This approach has two issues:
1. The Default C implementation cannot be easily tested independently. This is an issue for #1187, for example.
2. The CBMC proofs are more complex as they need to reason about both the dispatch logic and the default C implementation.

**Tasks:**
For every function in the arithmetic backend API (mlkem/src/native/api.h), locate its implementation in mlkem/src/* and the backend-frontend dispatch logic above. Then, assuming the function is called `foo`, create a function `foo_c()` for the default C implementation, and change the toplevel implementation of `foo()` to
```c
if (foo_backend(...) == SUCCESS) {
   return;
}

foo_c(...);
```

`foo_c` shuld have the seame CBMC contract as `foo()`. The documentation may just refer to that of `foo()`, though.

Finally, create a new CBMC proof harness for `foo_c`. It should be near identical to the proof harness for the previous `foo`, but remove anything related to the backend. Then, update the CBMC Makefile for `foo` to call `foo_c` by contract.

hanno-becker:
Previously, backend dispatch functions embedded their default C
implementations inline, mixing dispatch logic with algorithmic code.
This created two problems: (1) the C implementations could not be
tested independently from dispatch complexity, and (2) CBMC proofs
had to reason about both backend selection and correctness.

This change extracts the pure C implementations into separate _c
functions, enabling independent testing and simplified CBMC proofs.

Functions extracted:

Core polynomial operations:
- mlk_poly_tomont_c, mlk_poly_reduce_c, mlk_poly_mulcache_compute_c
- mlk_poly_ntt_c, mlk_poly_invntt_tomont_c
- mlk_poly_tobytes_c, mlk_poly_frombytes_c
- mlk_polyvec_basemul_acc_montgomery_cached_c

Compression operations:
- mlk_poly_compress_d4_c, mlk_poly_compress_d5_c
- mlk_poly_compress_d10_c, mlk_poly_compress_d11_c
- mlk_poly_decompress_d4_c, mlk_poly_decompress_d5_c
- mlk_poly_decompress_d10_c, mlk_poly_decompress_d11_c

Sampling operations:
- mlk_rej_uniform_c (merely renamed from mlk_rej_uniform_scalar,
  it was already hoisted out as a separate function)

The compression operations are an outlier because we don't yet
have separate CBMC proofs for the specific values of d. We do
hoist the C implementations into standalone functions, but do
not yet separate the CBMC proof of their correctness from the
dispatch logic.

* Resolves #1194


hanno-becker:
@mkannwischer @rod-chapman CBMC back down to 6-8min (was sometimes >1h before).

PR #1196, opened 2025-09-22
Add MacOS 26 CI

mkannwischer:


mkannwischer:
This was meant as a test. macos-26 is still in preview, so we should not merge it just yet. Closing for now.

PR #1197, opened 2025-09-24
Add make targets for building SMT files.

rod-chapman:
When investigating and/or debugging CBMC proof failures, it can be very useful to generate the SMT file(s), but without actually running the provers. It's also useful to generate the SMT files for one or all of Z3, Bitwuzla, or CVC5 for comparison.

Update Makefile.common for CBMC so that for a function x, new Makefile targets are:

smt: generates gotos/x.smt2 for prover (z3, bitwuzla, or cvc5) chosen by the user
     in the specific Makefile for that function.

smtz: generates gotos/x.smtz, but always for Z3, ignoring the choice in the Makefile
smtb: generates gotos/x.smtb, but always for Bitwuzla, ignoring the choice in the Makefile
smtc: generates gotos/x.smtc, but always for cvc5, ignoring the choice in the Makefile

smtall: generates all three targets smtz, smtb, smtc as above.


rod-chapman:
Proof Guide updated.

PR #1198, opened 2025-09-25
Makefile: Remove use of .WAIT primitive to support make <4.4

### LINKED ISSUE #1173: Consider supporting make < 4.4
We got this report in mldsa-native: https://github.com/pq-code-package/mldsa-native/issues/427

The same seems to be the case for mlkem-native: The current benchmarking make targets require make >=4.4. If there is an easy fix, we should support older make versions.

hanno-becker:
Previously, the Makefile used .WAIT to order the execution of the benchmarks, breaking compatibility with versions of `make` prior to 4.4.

This commit fixes this by serializing the benchmark execution explicitly, without relying on .WAIT.

* Resolves #1173


mkannwischer (approved):
LGTM. Thanks

Issue #1199, opened 2025-09-26
OQS CI: Constant-time tests are being skipped


mkannwischer:
Currently, our OQS tests in CI look like this:
```
============================= test session starts ==============================
platform linux -- Python 3.12.3, pytest-7.4.4, pluggy-1.4.0
rootdir: /home/runner/work/mlkem-native/mlkem-native/libOQS
collected 1466 items / 1421 deselected / 45 selected

tests/test_acvp_vectors.py .........                                     [ 20%]
tests/test_alg_info.py ...                                               [ 26%]
tests/test_cmdline.py ...                                                [ 33%]
tests/test_code_conventions.py ...                                       [ 40%]
tests/test_constant_time.py sss                                          [ 46%]
tests/test_distbuild.py sss                                              [ 53%]
tests/test_kat.py ...                                                    [ 60%]
tests/test_kat_all.py ...                                                [ 66%]
tests/test_leaks.py ...                                                  [ 73%]
tests/test_mem.py ...                                                    [ 80%]
tests/test_speed.py ...                                                  [ 86%]
tests/test_wycheproof_vectors.py ......                                  [100%]

=============== 39 passed, 6 skipped, 1421 deselected in 13.54s ================
```
i.e., we are skipping the constant-time test. We should enable them.

PR #1200, opened 2025-09-26
OQS CI: Enable constant-time tests

mkannwischer:
Currently our liboqs integration tests are skipping the oqs constant time tests. There is no reason to skip those.

This commit adds
-DCMAKE_BUILD_TYPE=Debug -DOQS_ENABLE_TEST_CONSTANT_TIME=ON which enables the constant-time tests.

 - Resolves https://github.com/pq-code-package/mlkem-native/issues/1199

### DIFF
--- a/.github/actions/config-variations/action.yml
+++ b/.github/actions/config-variations/action.yml
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
 
-name: 'Cuatom config tests'
+name: 'Custom config tests'
 description: 'Build and test mlkem-native with various custom configs'
 inputs:
   gh_token:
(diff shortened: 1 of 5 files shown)


mkannwischer:
https://github.com/pq-code-package/mlkem-native/actions/runs/18054655007 shows that this does not work as expected - removing a declassification causes `tests/test_leaks.py` to fail, but `test_constant_time.py` still happily passes.
I need to look into this in more detail.

mkannwischer:
> https://github.com/pq-code-package/mlkem-native/actions/runs/18054655007 shows that this does not work as expected - removing a declassification causes `tests/test_leaks.py` to fail, but `test_constant_time.py` still happily passes. I need to look into this in more detail.

I looked into it. It turns out that the whitelist in OQS still matches the functions that were tested here.
When the whitelist is removed from OQS, it fails as it should:
https://github.com/pq-code-package/mlkem-native/actions/runs/18152553215

I added that removal to the CI workflow.

hanno-becker:
@mkannwischer Can you open an issue so we don't forget to remove the patch once the fix has been merged?

hanno-becker (approved):

PR #1202, opened 2025-09-29
HOL-Light: Add target to update byte code in proof scripts

### LINKED ISSUE #1180: HOL-Light: Avoid manual patchup of byte code and disassembly in proof scripts
Currently, the *.ML HOL-Light proof scripts embed the byte code and disassembly of the target object file. Both have to be updated manually upon changes to the code such as re-running SLOTHY. This is tedious and unnecessary, seeing that all proofs are agnostic to specific instruction ordering.

Task: Either remove the embedding of the byte code in the *.ML file, or extend `autogen` to generate it automatically using `print_literal_from_elf` from within the HOL-Light shell.

hanno-becker:
* Resolves #1180 

This commit adds the `dump_bytecode` target to
`proofs/hol_light/arm/Makefile`, dumping the bytecode of all AArch64 object files.

The output of this target is then used by the new command

```
autogen --update-hol-light-bytecode
```

to update the bytecode embedded in the various HOL-Light proof scripts.

Note that this is purely convenience, not a soundness issue: The proof scripts fail at runtime if the bytecode is not up to date. Yet, any change to the assembly previously necessitated a manual extraction of the bytecode in the format expected by HOL-Light, which is cumbersome. Now, one only needs to run the above `autogen` command once to update the byte code in the scripts.

The HOL-Light CI is extended to run
`autogen --update-hol-light-bytecode --dry-run` to check that the byte code is up to date, before trying to run the proofs (which would also fail, but potentially much later).


scripts/autogen
```diff
@@ -2670,6 +2670,81 @@ def gen_citations(dry_run=False):
     gen_bib_file(bibliography, dry_run=False)
 
 
+def extract_bytecode_from_output(output_text):
+    """Extract bytecode sections from test.ml output into a dictionary."""
```
mkannwischer, line 2674:
```suggestion
    """Extract bytecode sections from dump_bytecode.ml output into a dictionary."""
```

scripts/autogen
```diff
@@ -2670,6 +2670,81 @@ def gen_citations(dry_run=False):
...
+    bytecode_dict = {}
+
+    # Find all bytecode sections using a robust approach
```
mkannwischer, line 2677: AI?

scripts/autogen
```diff
@@ -2670,6 +2670,81 @@ def gen_citations(dry_run=False):
...
+    lines = output_text.split("\n")
+    i = 0
+
+    while i < len(lines):
+        line = lines[i]
+        if "=== bytecode start: mlkem/" in line and ".o" in line:
+            # Extract filename
+            match = re.search(r"mlkem/([^/\s]+?)\.o", line)
```
mkannwischer, line 2685: just match the regex alerady in the first if?

hanno-becker:
@mkannwischer I'll address those in a follow-up to avoid re-running the entire HOL-Light CI.

PR #1203, opened 2025-09-29
Fix bibliography, add all authors for mupq

willieyz:
- This is a samll fix about bibliography, this PR add all authors for `mupq` instead of just use the  `et.al`
- see https://github.com/pq-code-package/mldsa-native/pull/482#discussion_r2381480490


mkannwischer (approved):
LGTM. THanks @willieyz.

PR #1212, opened 2025-10-02
Fix typo

willieyz:
- We found this typo during mldsa-native PR#489, 
- Fixing it with this PR, see: https://github.com/pq-code-package/mldsa-native/pull/489#discussion_r2396907530


mkannwischer:
I included that fix in #1200 to save some CI time.

PR #1216, opened 2025-10-04
API: Add check_sk and check_pk to external API

hanno-becker:


mkannwischer (approved):
LGTM. Thanks @hanno-becker!

PR #1218, opened 2025-10-06
HOL-Light: Avoid hardcoded lengths in HOL-Light proof scripts

### LINKED ISSUE #1217: HOL-Light: Remove hardcoded PC offsets
The HOL-Light proof scripts currently hardcode various PC offsets related to the length of the byte code overall, the preamble (stack spill) and postamble (stack restore). Adjusting those offsets is trivial but combersome, and should be automated.

Introduce definitional wrappers for those offsets so that when the total length of the code changes, but the length of preamble/postamble doesn't, no changes to the proof scripts are necessary. This will further simplify any experimentation with the code, such as #1088.

hanno-becker:
* Resolves #1217 

Previously, the HOL-Light proof scripts contained various hardcoded constants related to the length of the code under proof, as well as the lengths of preamble and postamble (stack spill and restore).

This commit replaces those hardcoded numerals with symbolic definitions. As a result, SLOTHY re-optimization should now no longer require a proof script change beyond the update of the byte code which is done by `autogen` already.


mkannwischer (approved):
Thanks @hanno-becker! Very welcome improvement!

PR #1220, opened 2025-10-07
Add initial support for Armv8.1-M+MVE targets

willieyz:
- This PR porting following changes in following PR from `mldsa-native` to `mlkem-native`, adds initial support for baremetal targets.
  - https://github.com/pq-code-package/mldsa-native/pull/462 

- General build system changes:
  - Add separate LDFLAGS (previously only CFLAGS for both compilation and linking)
  - Add EXTRA_MAKEFILE to include custom makefiles with platform-specific configuration
  - Add EXTRA_SOURCES to compile and link additional platform files
  - Fix ACVP tests to respect EXEC_WRAPPER
- Baremetal support for MPS3 AN547 (Arm Cortex-M55):
  - Add nix shell (arm-embedded) with arm-none-eabi-gcc, qemu, and platform support files from pqmx
  - Add test/baremetal/platform/m55-an547/platform.mk with compiler flags, and build configuration
  - Add exec_wrapper.py to pass test arguments via QEMU
  - Add CI workflow for functional, KAT, and ACVP tests on AN547

- Note that: unlike `mldsa-native`, `mlkem-native` auto-detects the host architecture (AArch64) during cross-compilation, incorrectly setting `MLK_FORCE_AARCH64` and breaking Cortex-M55 baremetal builds, we added `extra_args: --no-auto` in `.github/workflows/baremetal.yml` to disable auto-detection in baremetal build + test.

### DIFF
--- a/test/baremetal/platform/m55-an547/platform.mk
+++ b/test/baremetal/platform/m55-an547/platform.mk
@@ -59,4 +59,7 @@ LDFLAGS += \
 
 # Extra sources to be included in test binaries
 EXTRA_SOURCES = $(wildcard $(M55_AN547_PATH)/*.c)
+# The CMSIS files fail compilation if conversion warnings are enabled
+EXTRA_SOURCES_CFLAGS = -Wno-conversion -Wno-sign-conversion
+
 EXEC_WRAPPER := $(realpath $(PLATFORM_PATH)/exec_wrapper.py)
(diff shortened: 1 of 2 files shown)


willieyz:
> > Note that: unlike mldsa-native, mlkem-native auto-detects the host architecture (AArch64) during cross-compilation, incorrectly setting MLK_FORCE_AARCH64 and breaking Cortex-M55 baremetal builds, we added extra_args: --no-auto in .github/workflows/baremetal.yml to disable auto-detection in baremetal build + test.
> 
> Why don't we force `AUTO=0` (in the makefile) if a cross-prefix is set?

Hello, @hanno-becker, thank you for your review!
Now I add the following in `config.mk` instead of using the `extra_args`:
https://github.com/pq-code-package/mlkem-native/blob/0ce68798341e2a9caf5d9b5ddf08120a09985771/test/mk/config.mk#L80-L85

hanno-becker:
@mkannwischer Do you agree with forcing `AUTO=0` for cross-compilation?

hanno-becker:
> @mkannwischer Do you agree with forcing AUTO=0 for cross-compilation?

Hm, I'm no longer sure about this. We rely on AUTO e.g. #1037

hanno-becker:
I removed the disabling of `AUTO` for cross-compilation again. What was instead needed is detecting the cross compilation target in `config.mk`. See the latest commit.

hanno-becker:
@mkannwischer Please see the latest additions.

hanno-becker (approved):

mkannwischer (approved):

Issue #1229, opened 2025-10-15
Consider resolving `-Wimplicit-int-conversion` and `-Wsign-conversion` warnings


mkannwischer:
When compiling mlkem-native with `-Wimplicit-int-conversion` or `-Wsign-conversion` there are a lot of warnings. A potential consumer of mlkem-native ([OpenTitan](https://github.com/zerorisc/expo)) compiles with both these warnings and it would make integration easier if we fix those warnings.

PR #1230, opened 2025-10-15
Fix `-Wconversion` warnings

mkannwischer:
This commit enables further compiler warnings and adjusts the code
triggering the warnings. In some cases, we can work around the warnings
by tightening input/output types in function signatures. In other cases,
we explain in a comment why the truncation does not alter the value.
Finally, in some cases, the truncation can actually alter the value,
and we rely on signed->unsigned and unsigned->signed integer conversions.

Note that CBMC already checks that integer conversion does not alter
values, so the comments are a helpful indicator that we thought about
this, but not critical.
Resolves https://github.com/pq-code-package/mlkem-native/issues/1229

### DIFF
--- a/examples/basic/Makefile
+++ b/examples/basic/Makefile
@@ -59,6 +59,7 @@ CFLAGS := \
 	-Werror \
 	-Wpointer-arith \
 	-Wredundant-decls \
+	-Wconversion \
 	-Wno-long-long \
 	-Wno-unknown-pragmas \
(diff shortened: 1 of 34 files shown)


hanno-becker:
I believe CBMC already checks that those (so far) implicit truncations don't actually truncate. Hence, they should all be safe to be introduced. Can you check?

hanno-becker:
Oddly enough, this is causing a slight slowdown..

PR #1231, opened 2025-10-15
Add MLK_CONFIG_SERIAL_FIPS202_ONLY option

mkannwischer:
Currently rejection sampling is performed using 4-way batched Keccak
regardless whether that is beneficial or not.
That approach requires to keep around 4 Keccak states in memory.
This approach is incompatible with using a Keccak accelerator where there is
only a single state and it lives inside of the accelerator.

This commit adds an option MLK_CONFIG_SERIAL_FIPS202_ONLY that switches
to serial processing for the rejection sampling.
This enables offloading all hashing operations in mlkem-native to a single
accelerator.


mlkem/src/indcpa.c
```diff
@@ -216,14 +216,15 @@ void mlk_gen_matrix(mlk_polymat a, const uint8_t seed[MLKEM_SYMBYTES],
     mlk_memcpy(seed_ext[j], seed, MLKEM_SYMBYTES);
   }
 
+#if !defined(MLK_CONFIG_SERIAL_FIPS202_ONLY)
```
hanno-becker, line 234: Can we not just change the implementation of `mlk_poly_rej_uniform_x4` to call 4x `mlk_poly_rej_uniform` if `MLK_CONFIG_SERIAL_FIPS202_ONLY` is set? This way, we don't need to touch the CBMC proof for `mlk_gen_matrix`, and we only need to show the 4x `mlk_poly_rej_uniform` indeed satisfies the spec for `mlk_poly_rej_uniform_x4`.

hanno-becker (changes requested):
I support the introduction of this option, but I think the implementation and adjustment of CBMC proofs would be simpler if we had the compile-time dispatch in `mlk_poly_rej_uniform_x4`, as mentioned in the comment.

We should also introduce another example where this option is used alongside a custom FIPS202 implementation which maintains internal static (!) state and fails if more than one call-chain is active at once.

mlkem/src/indcpa.c
hanno-becker: No longer sure on second thought since the present approach would remove the additional code for the batched sampling loop.

hanno-becker:
One should be able to test this by detecting in #1233 whether `MLK_CONFIG_SERIAL_FIPS202_ONLY` is set, and if so:
* Use a `static` (!) Keccak state, typedef'ing the state type used by the frontend to a dummy struct
* Initializing said `static` state in state `RESET`
* Checking in `init` that the incoming state is `RESET`.

PR #1239, opened 2025-10-19
Introduce helper functions for signed<->unsigned conversions

hanno-becker:


mlkem/src/poly.h
```diff
@@ -103,8 +75,7 @@ __contract__(
   const uint32_t QINV = 62209;
 
   /* Compute a*q^{-1} mod 2^16 in unsigned representatives. */
-  const uint16_t a_reduced = (uint16_t)(a & (int32_t)UINT16_MAX);
-
+  const uint16_t a_reduced = mlk_cast_int16_to_uint16(a);
```
mkannwischer: should this be using the int32 variant?

hanno-becker: Good catch, fixed.

I am baffled by what `-Wconversion` does warn about, and what not. This should definitely have triggered a warning/error?

PR #1240, opened 2025-10-20
FIPS202: Provide alignment constraints for output buffers

hanno-becker:
mlkem-native will only call the FIPS202 API with aligned buffers, which can be useful for users providing their own FIPS202 implementation.

This commit extends the documentation and CBMC contracts accordingly, and adds debug assertions for checking alignment. Note that CBMC does, to the best of my knowledge, not support talking about absolute pointer alignment, so we cannot rely on proof here.

This is relevant for https://github.com/zerorisc/expo/pull/97/

### DIFF
--- a/mlkem/mlkem_native.S
+++ b/mlkem/mlkem_native.S
@@ -280,8 +280,10 @@
 #undef mlk_assert
 #undef mlk_assert_abs_bound
 #undef mlk_assert_abs_bound_2d
+#undef mlk_assert_alignment
 #undef mlk_assert_bound
 #undef mlk_assert_bound_2d
+#undef mlk_debug_check_alignment
 #undef mlk_debug_check_assert
 #undef mlk_debug_check_bounds
 /* mlkem/src/poly.h */
--- a/mlkem/mlkem_native.c
+++ b/mlkem/mlkem_native.c
@@ -269,8 +269,10 @@
 #undef mlk_assert
 #undef mlk_assert_abs_bound
 #undef mlk_assert_abs_bound_2d
+#undef mlk_assert_alignment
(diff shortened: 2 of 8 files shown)


hanno-becker:
@mkannwischer Unfortunately, one cannot fully discharge those constraints, as sometimes one hashed straight into a buffer provided at the toplevel API. One would need to either a) add alignment constraints at the toplevel, b) add a copy at the toplevel, c) avoid the alignment constraints altogether.

We should definitely extend our testing to exercise unaligned buffers at the toplevel.

mkannwischer:
> @mkannwischer Unfortunately, one cannot fully discharge those constraints, as sometimes one hashed straight into a buffer provided at the toplevel API. One would need to either a) add alignment constraints at the toplevel, b) add a copy at the toplevel, c) avoid the alignment constraints altogether.
> 
> We should definitely extend our testing to exercise unaligned buffers at the toplevel.

That is indeed unfortunate. Do you still want to merge this PR now? It gives the false impression that FIPS202 can assume alignment of the output.

hanno-becker:
> Do you still want to merge this PR now? 

No, definitely not in this form.

hanno-becker:
>@mkannwischer Unfortunately, one cannot fully discharge those constraints, as sometimes one hashed straight into a buffer provided at the toplevel API. One would need to either a) add alignment constraints at the toplevel, b) add a copy at the toplevel, c) avoid the alignment constraints altogether.

This is blocking this change.

> We should definitely extend our testing to exercise unaligned buffers at the toplevel.

This is already the case https://github.com/pq-code-package/mlkem-native/blob/main/test/test_mlkem.c#L61

Closing this for now.

Issue #1242, opened 2025-10-21
CI: Test OpenTitan integration


mkannwischer:
I have recently integrated mlkem-native into OpenTitan: https://github.com/zerorisc/expo/pull/97

We should test that integration in a similar way as we test the OQS and AWS-LC integration in CI to make sure our changes do not break the integrations.
It boils down to running 

```
./bazelisk.sh test --test_output=streamed --disk_cache=~/bazel_cache //sw/device/tests/crypto:mlkem_functest_sim_verilator
```

after updating the bazel dependency on mlkem-native.

PR #1243, opened 2025-10-21
CI: Add OpenTitan integration test 

mkannwischer:
This commit adds an OpenTiten integration test to our CI which clones
OpenTitan, updates the mlkem-native dependency to the current commit, and runs
the ML-KEM tests using verilator.

- Resolves https://github.com/pq-code-package/mlkem-native/issues/1242

OpenTitan requires an old version of verilator (4.210) which is compiled from
source taking around 5-10 minutes. We cache it (80 MiB) in the Github cache
to speed-up CI.

Another expensive step is the initial compilation of the hardware simulation.
Verilator commonly takes 10-15 minutes. Usually this is cached in the bazel
cache such that subsequent tests are much faster.
I considered caching this which would speed-up CI by around 10-15 minutes,
but require 3 GiB of caching. We are already tight in Github cache, so at this
point, I don't think we can afford it.

### DIFF
--- a/.github/actions/setup-opentitan/action.yml
+++ b/.github/actions/setup-opentitan/action.yml
@@ -1,0 +1,85 @@
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+name: Setup OpenTitan
+description: Setup OpenTitan for testing
+
+inputs:
+  expo-repository:
+    description: 'Expo repository to clone'
+    required: false
+    default: 'https://github.com/zerorisc/expo'
+  expo-commit:
+    description: 'Expo commit to checkout'
+    required: false
+    default: 'master'
+
+runs:
+  using: composite
+  steps:
+    - name: Fetch expo repository
+      shell: bash
+      run: |
+        # Ensure HOME is set for self-hosted runners
+        echo "HOME=${HOME:-/home/runner}" >> $GITHUB_ENV
+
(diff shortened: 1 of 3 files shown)


mkannwischer:
This does seem to almost work the way it should now. TODOs are: 
1) Figure out why the smoketest still takes so long even with the bazel cache
- The reason was that one also has to cache `~/.cache/bazel` - that is 3 GiB though, so it does not appear feasible. 
2) Figure out if it is worth to switch to a smaller/larger EC2 box. 
- I saw barely a difference, so I switched to 2xlarge
3) Re-add the dependency on base
- done
4) Make sure that the CI actually fails if ML-KEM breaks.
- done

hanno-becker (approved):
Thank you @mkannwischer!

I think we might want to reduce the number of functests run to speed this up (I had the same issue in #1245), but this can be a follow-up.

mkannwischer:
> Thank you @mkannwischer!
> 
> 
> 
> I think we might want to reduce the number of functests run to speed this up (I had the same issue in #1245), but this can be a follow-up.

Thanks for the review!
Well, this is only a single run. If you want to reduce you either need to split randomized/derand tests or split by parameter set. I had briefly experimented with splitting it into 6 separate tests and let bazel run it in parallel - but that did not result in any wall clock speed up on my laptop.

PR #1244, opened 2025-10-21
HOL-Light: Allow interactive use of hol.sh in the hol_light shell

hanno-becker:
Previously, while compiling and running the HOL-Light proofs in `nix develop .#hol_light` worked, trying to start the interactive shell `hol.sh` for proof development would fail. This required proof developers to install HOL-Light and s2n-bignum locally, both inconvenient and error proone because one would need to keep that local version and the nix version in sync.

This commit adjusts the nix flake to setup HOL-Light in such a way that hol.sh can be used interactively inside the `hol_light` shell.

Now, when you enter the shell via `nix develop .#hol_light`, the interactive shell `hol.sh` is in the PATH, and within it, both s2n-bignum and proofs/hol_light/arm are added to HOL-Light's own search path (`load_path`). In particular, you can do

```
  needs "arm/proofs/base.ml";;
```

to load the s2n-bignum base sources, or

```
  needs "proofs/mlkem_poly_tobytes.ml";;
```

or similar, to test a proof script.

The CI is extended to exercise the interactive shell, running one of the faster proofs -- mlkem_poly_tobytes.ml -- interactively.

### DIFF
--- a/.github/workflows/hol_light.yml
+++ b/.github/workflows/hol_light.yml
@@ -49,6 +49,22 @@ jobs:
           nix-shell: 'hol_light'
           script: |
             autogen --update-hol-light-bytecode --dry-run
+  hol_light_interactive:
+    name: HOL-Light interactive shell test
+    runs-on: pqcp-arm64
+    needs: [ hol_light_bytecode ]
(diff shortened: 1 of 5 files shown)


hanno-becker:
Thank you @mkannwischer, it should be fixed now.

PR #1245, opened 2025-10-22
Add baremetal environment for 16-bit testing

hanno-becker:
The primary purpose of this PR is to confirm and test that mlkem-native makes no assumption about the integer width of the underlying C implementation; in particular, that it still works with 16-bit `int`/`unsigned`. To this end, a baremetal environment is added using `avr-gcc` to compile for AVR and simulate the code using `simavr`. Since we currently use too much stack for any MCU supported by `simavr`, we patch it to provide 32K instead of 16K RAM to Atmega128rfr2. 

**Note:** This PR is not meant to suggest mlkem-native is ready for, nor that it is or will ever be a goal to make it ready for, low-end MCUs. It mainly serves as a forcing function to ensure that we don't rely on specific integer widths.

### DIFF
--- a/test/gen_KAT.c
+++ b/test/gen_KAT.c
@@ -14,7 +14,10 @@
 #include <io.h>
 #endif
 
-#define NTESTS 100
+/* Note that the entire output of this application feeds into
+ * a hash compared with those in META.yml. If you want to reduce
+ * the number of KAT tests, META.yml needs adjusting as well. */
+#define NTESTS_KAT 100
 
 #define CHECK(x)                                              \
   do                                                          \
@@ -64,7 +67,7 @@ int main(void)
 
   mlk_shake256(coins, sizeof(coins), seed, sizeof(seed));
 
-  for (i = 0; i < NTESTS; i++)
+  for (i = 0; i < NTESTS_KAT; i++)
   {
     mlk_shake256(coins, sizeof(coins), coins, sizeof(coins));
 
--- a/test/test_mlkem.c
+++ b/test/test_mlkem.c
@@ -10,8 +10,8 @@
 #include "../mlkem/mlkem_native.h"
 #include "notrandombytes/notrandombytes.h"
 
-#ifndef NTESTS
-#define NTESTS 1000
+#ifndef NTESTS_FUNC
+#define NTESTS_FUNC 1000
 #endif
 
 #define CHECK(x)                                              \
@@ -168,7 +168,7 @@ int main(void)
    * Normally, you would want to seed a PRNG with trustworthy entropy here. */
   randombytes_reset();
 
-  for (i = 0; i < NTESTS; i++)
+  for (i = 0; i < NTESTS_FUNC; i++)
   {
     CHECK(test_keys() == 0);
     CHECK(test_keys_unaligned() == 0);
--- a/.github/actions/ct-test/action.yml
+++ b/.github/actions/ct-test/action.yml
@@ -40,4 +40,4 @@ runs:
         run: |
           make clean
           # Disable the AArch64 SHA3 extension as it's not yet supported by valgrind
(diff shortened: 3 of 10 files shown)


nix/avr/simavr-uart-output-fix.patch
```diff
@@ -0,0 +1,22 @@
...
+ 			p->stdio_out = malloc(maxsize);
+-		p->stdio_out[p->stdio_len++] = v < ' ' ? '.' : v;
+-		p->stdio_out[p->stdio_len] = 0;
+ 		if (v == '\n' || p->stdio_len == maxsize) {
+ 			p->stdio_len = 0;
+ 			AVR_LOG(avr, LOG_OUTPUT,
+ 					FONT_GREEN "%s\n" FONT_DEFAULT, p->stdio_out);
++		} else {
++			p->stdio_out[p->stdio_len++] = v;
++			p->stdio_out[p->stdio_len] = 0;
```
hanno-becker, line 21: Should have mentioned this: The UART output of `simavr` adds unwanted line breaks and dots `.` at every line break, breaking the KAT test.

test/test_mlkem.c
```diff
@@ -10,8 +10,8 @@
 #include "../mlkem/mlkem_native.h"
 #include "notrandombytes/notrandombytes.h"
 
-#ifndef NTESTS
-#define NTESTS 1000
+#ifndef NTESTS_FUNC
```
mkannwischer, line 13: This has an unintended side-effect on the constant-time testing.
Please also change the `-DNTESTS` there if you really need to rename to `NTESTS_FUNC` - not sure why that is needed at all.

test/gen_KAT.c
```diff
@@ -14,7 +14,7 @@
 #include <io.h>
 #endif
 
-#define NTESTS 1000
+#define NTESTS 100
```
mkannwischer: We could consider splitting this into two tests: One with fewer iterations, one with more. 
In SUPERCOP this is called `checksumsmall` and `checksumbig`.

nix/avr/simavr-32kb-ram.patch
```diff
@@ -0,0 +1,15 @@
...
+--- a/simavr/cores/avr/iom128rfr2.h
++++ b/simavr/cores/avr/iom128rfr2.h
+@@ -6300,8 +6300,8 @@
+ 
+ #define SPM_PAGESIZE                    (256)
+ #define RAMSTART                        (0x200)
+-#define RAMSIZE                         (0x4000)
+-#define RAMEND                          (0x41FF)
++#define RAMSIZE                         (0x8000)
++#define RAMEND                          (0x81FF)
```
mkannwischer: Have you tried increasing this more? If we can go to 192 KB / 256 KB, we could easily do the same for mldsa-native.

mkannwischer (changes requested):
Thanks @hanno-becker! Awesome that you also still got the ACVP tests to work.

Currently there is one blocker: You unintentionally increased the number of tests in the constant-time testing.
~I'd appreciate if we could try increasing the RAM a bit more so that we can also integrate this into mldsa-native.~ Stupid me - we only have 16-bit addresses. Nevermind.

test/test_mlkem.c
hanno-becker: The risk with NTESTS is that you overwrite it for gen_KAT, and then the test fails. It doesn't happen for CT because you _only_ run functests. But if you have a single CFLAGs applied to all of KAT, func, ACVP, then you don't have the control to run fewer functests, but (necessarily) as many KAT tests.

hanno-becker: Fixed

test/gen_KAT.c
hanno-becker: Good idea. https://github.com/pq-code-package/mlkem-native/issues/1247

PR #1246, opened 2025-10-22
[TEST] Test on Arm MPS3 FPGA

mkannwischer:


hanno-becker:
@mkannwischer Thank you for investigating this. It would be great to have the benchmark results reported alongside the existing benchmarks, so it will be easy to follow along how the Keccak and NTT optimizations for MVE affect performance.

hanno-becker:
Closing this for lack of activity, plus we are not allowed by the EULA to publish any performance data without permission by Arm.

PR #1250, opened 2025-10-23
CI: Switch to r8g.xlarge for CBMC proofs

hanno-becker:
The r instances have a higher Memory/vCPU ratio and are therefore
more suited for the memory-hungry CBMC tests. Specifically, r8g.xlarge
has the same RAM as c7g.4xlarge, a newer generation Graviton core,
and at the time of writing currently comes at a lower price point.

Therefore, this commit modifies the CBMC CI job to use r8g.xlarge
instances instead of c7g.4xlarge.

### DIFF
--- a/.github/workflows/cbmc.yml
+++ b/.github/workflows/cbmc.yml
@@ -17,9 +17,9 @@ jobs:
     uses: ./.github/workflows/ci_ec2_reusable.yml
     with:
       name: CBMC (MLKEM-512)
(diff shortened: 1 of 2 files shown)


hanno-becker:
@mkannwischer I think this could be a saving. See https://aws.amazon.com/ec2/pricing/on-demand/ for a cost-comparison between c7g.4xlarge (previously) and r8g.xlarge (proposed).

PR #1254, opened 2025-10-24
README: Add `Supported platforms` section

hanno-becker:


hanno-becker:
This should be updated and also consolidated with mldsa-native. We should have uniform README.md's across those projects. I'll update.

PR #982, opened 2025-04-29
Relicense mlkem-native as `Apache-2.0 OR ISC OR MIT`

hanno-becker:
Consent to the relicensing has been obtained from all mlkem-native contributors as individual commits to RELICENSE.md. See https://github.com/pq-code-package/mlkem-native/pull/967

PR #1256, opened 2025-10-25
Correct example Makefile licenses to Apache-2.0 OR ISC OR MIT

mkannwischer:
The example Makefiles were incorrectly listing a CC-BY license that we commonly use for documentation, but as they are code they should be Apache-2.0 OR ISC OR MIT.
This commit relicenses them as Apache-2.0 OR ISC OR MIT.


hanno-becker:
I'll re-do the changes to the Makefiles based on the state https://github.com/pq-code-package/mlkem-native/pull/982/

PR #1257, opened 2025-10-26
Relicense and clean up example Makefiles

hanno-becker:
This PR relicenses the Makefiles in `examples/*` to `Apache-2.0 OR ISC OR MIT`. It was an oversight that this was not done in #982. This PR reverts the Makefiles to the state they were in at #982, relicenses them (based on the consent given at the time), and then applies various fixes and improvements atop to get to a working set of makefiles again.

### DIFF
--- a/examples/basic/Makefile
+++ b/examples/basic/Makefile
@@ -1,4 +1,5 @@
-# (SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT)
+# Copyright (c) The mlkem-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
 
 .PHONY: build run clean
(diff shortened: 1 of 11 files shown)


hanno-becker:
I see that the `size` target is no longer present in the examples. On the other hand, this wasn't exercised in testing anyway so far.

examples/basic/Makefile
```diff
@@ -1,31 +1,62 @@
-# (SPDX-License-Identifier: CC-BY-4.0)
+# (SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT)
```
mkannwischer: Please add the copyright notice as well. Why is is in `()`?

hanno-becker: Done

PR #1258, opened 2025-10-26
CI/OT: Switch to c7i.2xlarge

hanno-becker:
At present, those runners come at a similar price point, but should be faster.


hanno-becker:
30min instead of 48min. Worth it, I think. cf https://aws.amazon.com/ec2/pricing/on-demand/

hanno-becker:
I was wrong. It's basically the same cost per run, but a few min faster. So we can still merge it.

PR #1266, opened 2025-11-01
Reduce runtime of autogen

hanno-becker:
Various smaller improvements to speed up `scripts/autogen`.

On my local machine: 40s -> 16s

### DIFF
--- a/scripts/autogen
+++ b/scripts/autogen
@@ -1445,8 +1445,9 @@ def get_defines(all=False):
     else:
         files = get_header_files() + get_c_source_files() + get_asm_source_files()
 
-    for c in files:
-        yield from get_defines_from_file(c)
+    with ThreadPoolExecutor() as executor:
+        for results in executor.map(get_defines_from_file, files):
+            yield from results
 
 
 def get_checked_defines():


scripts/autogen
```diff
@@ -3133,15 +3213,27 @@ def _main():
...
     high_level_status("Checked macro typos")
 
-    gen_test_configs(args.dry_run)
+    gen_test_configs()
     high_level_status("Generated test configs")
 
+    gen_preprocessor_comments()
+    high_level_status("Generated preprocessor comments")
+
+    finalize()
```
mkannwischer: Why is there no args.dry_run here? 
I don't understand how the linting works. I thought it will fail in the finalize step, but that does not seem to be the case.

mkannwischer (changes requested):

scripts/autogen
hanno-becker: Argh, thank you for catching this. This is crucial.

hanno-becker: Fixed.

hanno-becker:
@mkannwischer Good catch. It should be addressed now.

mkannwischer (approved):
Thanks for eliminating the race conditions. Works nicely now.

PR #1275, opened 2025-11-04
autogen/simpasm: Add support for Intel syntax

hanno-becker:
This commit extends scripts/simpasm to support the emission of x86_64 assembly in Intel syntax via --x86-64-syntax intel.

The CI is extended to exercise this on an x86_64 runner and run the full tests afterwards.

This can be useful for consuming libraries wishing to use a different syntax than what mlkem-native uses. A similar approach may be usable in the future to add assembly suitable for the Windows assembler.


mkannwischer (approved):
Nice, thanks @hanno-becker! Tested on my x86 machine - works like a charm!

PR #1276, opened 2025-11-05
ci_ec2_reusable: Set ec2-volume-size also on re-try

mkannwischer:
https://github.com/pq-code-package/mlkem-native/pull/1250 added a volume size argument to ci_ec2_reusable, but it is only used for the first attempt to start the instance.
If that fails, the retry will fall back to the default volume size which may very well be 8 GiB which is doomed to fail.


mkannwischer:
I included the commit in #1277

PR #1278, opened 2025-11-05
autogen: Only consider tracked files

hanno-becker:
This commit further improves the performance of scripts/autogen by having it consider only files tracked by git.

Originally, this was not possible since we could create new files directly in the file system but wanted later steps in autogen to pick them up for further modification.

Now that we hold back any FS write operations until the end of autogen and instead work on an internal file cache, this problem goes away, and we can use 'git ls-files' exclusively for file listing.

A slight nuisance arises as we can no longer use pathlib.Path().glob() for pattern matching, and the corresponding operation PurePath.full_match seems to available since Python 3.13 only. We hence opt for a slightly hacky implementation which manually transfers the glob pattern into a regexp first.

### DIFF
--- a/scripts/autogen
+++ b/scripts/autogen
@@ -1405,18 +1405,24 @@ def get_files(pattern, strip_mlkem=False):
     def normalize(f):
         return f.removeprefix("mlkem/") if strip_mlkem else f
 
-    fs_files = {
-        str(p)
-        for p in pathlib.Path().glob(pattern)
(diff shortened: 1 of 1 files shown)


mkannwischer (approved):
Thanks @hanno-becker. LGTM and we see good results in mldsa-native: https://github.com/pq-code-package/mldsa-native/pull/623

PR #1279, opened 2025-11-05
Upgrade to CBMC 6.8.0

rod-chapman:
This PR updates the NIX config to use CBMC 6.8.0 which was tagged on 5th November 2025.


rod-chapman:
Significant changes:
1. Much reduced noise on stdout when running locally. GitHub runner logs will be far more readable.
2. Addition of `--external-smt2-solver` switch to allow selection of a wrapper script that wraps a particular solver and could add particular command-line switches for a particular proof.
3. Simpler modelling of array objects, hopefully leading to performance and stability improvements.

PR #1284, opened 2025-11-06
autogen: Fix missing indirection via file cache

hanno-becker:
Symptom:

```
  rm mlkem/src/native/aarch64/src/arith_native_aarch64.h
  autogen
```

fails with

```
  FileNotFoundError: [Errno 2] No such file or directory: \
     'mlkem/src/native/aarch64/src/arith_native_aarch64.h'
```

Context:

A previous commit has introduced an internal file cache to autogen.
During main operation of autogen, all file read/write operations
should go through this in-memory cache.

Issue:

Two instance of direct FS reads via `open(...)` were overlooked
when this change was introduced, leading to an attempt to load
a file from the file system which only exists in the cache.

This commit fixes this by calling the file cache `read_file()`
operation intead.

### DIFF
--- a/mlkem/src/config.h
+++ b/mlkem/src/config.h
@@ -8,7 +8,8 @@
  *
  * - [FIPS140_3_IG]
  *   Implementation Guidance for FIPS 140-3 and the Cryptographic Module
- *   Validation Program National Institute of Standards and Technology
+ *   Validation Program
+ *   National Institute of Standards and Technology
  *   https://csrc.nist.gov/projects/cryptographic-module-validation-program/fips-140-3-ig-announcements
  */
 
--- a/examples/basic_deterministic/mlkem_native/custom_no_randomized_config.h
+++ b/examples/basic_deterministic/mlkem_native/custom_no_randomized_config.h
@@ -8,7 +8,8 @@
  *
  * - [FIPS140_3_IG]
  *   Implementation Guidance for FIPS 140-3 and the Cryptographic Module
- *   Validation Program National Institute of Standards and Technology
+ *   Validation Program
+ *   National Institute of Standards and Technology
  *   https://csrc.nist.gov/projects/cryptographic-module-validation-program/fips-140-3-ig-announcements
  */
 
--- a/examples/custom_backend/mlkem_native/custom_config.h
+++ b/examples/custom_backend/mlkem_native/custom_config.h
@@ -8,7 +8,8 @@
  *
  * - [FIPS140_3_IG]
  *   Implementation Guidance for FIPS 140-3 and the Cryptographic Module
- *   Validation Program National Institute of Standards and Technology
+ *   Validation Program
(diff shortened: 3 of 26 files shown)


integration/liboqs/ML-KEM-768_META.yml
```diff
@@ -77,11 +77,11 @@ implementations:
   signature_dec: PQCP_MLKEM_NATIVE_MLKEM768_AARCH64_dec
   sources: integration/liboqs/config_aarch64.h integration/liboqs/fips202_glue.h integration/liboqs/fips202x4_glue.h
     mlkem/src/cbmc.h mlkem/src/common.h mlkem/src/compress.c mlkem/src/compress.h
-    mlkem/src/debug.c mlkem/src/debug.h mlkem/src/indcpa.c mlkem/src/indcpa.h mlkem/src/kem.c
-    mlkem/src/kem.h mlkem/src/native/api.h mlkem/src/native/meta.h mlkem/src/params.h
-    mlkem/src/poly.c mlkem/src/poly.h mlkem/src/poly_k.c mlkem/src/poly_k.h mlkem/src/randombytes.h
-    mlkem/src/sampling.c mlkem/src/sampling.h mlkem/src/symmetric.h mlkem/src/sys.h
-    mlkem/src/verify.c mlkem/src/verify.h mlkem/src/zetas.inc mlkem/src/native/aarch64
+    mlkem/src/config.h.new mlkem/src/debug.c mlkem/src/debug.h mlkem/src/indcpa.c
```
mkannwischer: There should be no `config.h.new` in here

This file should not be changed at all.

mlkem/src/config.h
```diff
@@ -8,7 +8,8 @@
  *
  * - [FIPS140_3_IG]
  *   Implementation Guidance for FIPS 140-3 and the Cryptographic Module
- *   Validation Program National Institute of Standards and Technology
+ *   Validation Program
+ *   National Institute of Standards and Technology
```
mkannwischer, line 12: Why is this new line introduced? unintentional change?

mkannwischer: Nevermind - this is actually fine.

integration/liboqs/ML-KEM-768_META.yml
hanno-becker: Thanks! That was a bit careless, apologies.

hanno-becker: I opened https://github.com/pq-code-package/mlkem-native/issues/1286 to track the underlying issue, which is unrelated to the recent changes.

mkannwischer: Good to know that the oqs import apparently does not fail if you include files that dont exist.

hanno-becker:
Aborting CI until this is rebase atop `main` after #1282

PR #1288, opened 2025-11-06
Stabilizes proof of polyvec_add() and demo use of external-smt2-solver

### LINKED ISSUE #1287: CBMC: Revisit polyvec_add
PR #1282 has introduced a proof-time regression for `polyvec_add`, which is not the most expensive proof with ~4min for MLKEM-1024 in CI.

Tasks (in this order):
1. Test whether different solver / solver parameters can improve proof time.
2. If the answer to 1. is "no" (and ideally only then), add the loop invariant and see if that helps.

rod-chapman:
Fixes #1287

This PR improves performance and stability of the proof of polyvec_add() through use of the new CBMC --external-smt2-solver switch.

This also serves as a demonstration of how the new switch can be used, and we update the Proof Guide document accordingly.

### DIFF
--- a/proofs/cbmc/polyvec_add/Makefile
+++ b/proofs/cbmc/polyvec_add/Makefile
@@ -48,7 +48,7 @@ FUNCTION_NAME = mlk_polyvec_add
 # EXPENSIVE = true
 
 # This function is large enough to need...
-CBMC_OBJECT_BITS = 8
+CBMC_OBJECT_BITS = 10
 
 # If you require access to a file-local ("static") function or object to conduct
 # your proof, set the following (and do not include the original source file


proofs/cbmc/lib/z3_bv_sort
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# Copyright (c) The mlkem-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+z3 rewriter.bv_sort_ac=true "$@"
```
hanno-becker, line 4: A comment on what this is doing would be helpful. Can be a follow-up so we don't need to re-run the whole CI.

proofs/cbmc/lib/z3_smt_only
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# Copyright (c) The mlkem-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+z3 tactic.default_tactic=smt "$@"
```
hanno-becker, line 4: Same here

proofs/cbmc/lib/z3_bv_sort
rod-chapman: Agreed. The Z3 "-pp" option says that this option:
```
sort the arguments of all AC operators
```
which doesn't help much. I can only ask...

proofs/cbmc/lib/z3_smt_only
rod-chapman: This one I understand a bit more. The default tactic is "simplify THEN smt", but the "simplify" tactic has a bad habit of diverging or making things worse, so sometimes it's better to switch it off and run the SMT tactic only. Hence "smt_only".

mkannwischer:
I experimented a little with this. Increasing `CBMC_OBJECT_BITS` to 10 in this PR seems to further improve performance:

| OBJECT_BITS | Proof Runtime (c7g.4xlarge) |
| -------------- | ---------------------------- |
| 8 (current)      | 121s                                          |
| 9                      | 110s                                          |
| 10                    | 72s                                            |
| 11                    | 391s                                          |

mkannwischer:
Another data point: In mldsa-native for K=4, the polyveck_add proof (using invariants) takes only 31 seconds. And that one has a much stronger contract than what we need in mlkem-native: 

```c
void mld_polyveck_add(mld_polyveck *u, const mld_polyveck *v)
__contract__(
  requires(memory_no_alias(u, sizeof(mld_polyveck)))
  requires(memory_no_alias(v, sizeof(mld_polyveck)))
  requires(forall(k0, 0, MLDSA_K, forall(k1, 0, MLDSA_N, (int64_t) u->vec[k0].coeffs[k1] + v->vec[k0].coeffs[k1] < REDUCE32_DOMAIN_MAX)))
  requires(forall(k2, 0, MLDSA_K, forall(k3, 0, MLDSA_N, (int64_t) u->vec[k2].coeffs[k3] + v->vec[k2].coeffs[k3] >= INT32_MIN)))
  assigns(object_whole(u))
  ensures(forall(k4, 0, MLDSA_K, forall(k5, 0, MLDSA_N, u->vec[k4].coeffs[k5] == old(*u).vec[k4].coeffs[k5] + v->vec[k4].coeffs[k5])))
  ensures(forall(k6, 0, MLDSA_L,
                array_bound(u->vec[k6].coeffs, 0, MLDSA_N, INT32_MIN, REDUCE32_DOMAIN_MAX)))
);
```

proofs/cbmc/lib/z3_smt_only
hanno-becker: Ok, it would be useful to explain that in a follow-up.

hanno-becker:
Thanks @mkannwischer @rod-chapman. Switched to OBJECT_BITS=10 and removed the nix modification. Good to go.

rod-chapman:
I like the simplifications, and not having to modify the NIX config.

PR #1295, opened 2025-11-10
Replace `/bin/bash` by `/usr/bin/env bash`

nmouha:
The mlkem-native package fails to build on NixOS, because NixOS does not provide `/bin/bash` by default.

On NixOS, the recommended approach is to use `/usr/bin/env bash` because this makes it easier to install different versions of `bash` on the same system. (By letting `PATH` decide which binary to execute.) Note that `/usr/bin/env bash` is also the recommended approach on MacOS to use a newer `bash` version than the system-provided one.

I see that `/usr/bin/env bash` and `/usr/bin/env python3` are already used in 18 locations (see for example: [proofs/cbmc/list_proofs.sh](https://github.com/pq-code-package/mlkem-native/blob/main/proofs/cbmc/list_proofs.sh)). Therefore, I'm proposing to replace the remaining instances of `/bin/bash` to `/usr/bin/env bash`. So, this change not only fixes builds on NixOS, but is also more consistent.

### DIFF
--- a/META.sh
+++ b/META.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
 
 # Helper script to query META.yml
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@
 	run_size_512 run_size_768 run_size_1024 run_size \
 	host_info
 
-SHELL := /bin/bash
+SHELL := /usr/bin/env bash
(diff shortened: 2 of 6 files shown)


hanno-becker:
Thank you very much @nmouha. Unfortunately, our CI does currently not run fully on PRs from forks, so I opened #1297 and also added a test reproducing the issue.

hanno-becker:
Merged via #1297. Thanks again @nmouha!

nmouha:
Thank you @hanno-becker!

PR #1297, opened 2025-11-10
Add NixOS build+test to CI and fix bash invocations

hanno-becker:
* Add basic CI test for using mlkem-native with NixOS, reproducing bug found by @nmouha in #1295
* Cherry-pick @nmouha's patch from #1295 here, to ensure full CI passes (does currently not work from a fork).

### DIFF
--- a/META.sh
+++ b/META.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
 
 # Helper script to query META.yml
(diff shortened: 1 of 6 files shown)


nmouha:
Why not use `nix-shell -p` instead of `nix-env -i`?

The following is much faster and should work on any NixOS system:

```
nix-shell -p python3 gcc gnumake perl
cd mlkem-native
make build
make test
```

hanno-becker:
Thank you @nmouha, that helped!

nmouha:
Great! Thank you, @hanno-becker and @mkannwischer!

PR #1301, opened 2025-11-12
Use upstream instead of local version of z3 4.15.3 [full CI]

mkannwischer:
Running full CI on #1296.


mkannwischer:
Cancelled this as we need the CI runners in mldsa-native right now. But I think this is ready to be merged.

PR #1303, opened 2025-11-12
HOL-Light: Add support for cross-compilation of target object files

hanno-becker:
* Based on #1313


hanno-becker:
Redundant after https://github.com/pq-code-package/mlkem-native/pull/1444

PR #1312, opened 2025-11-19
SLOTHY CI: Do not use custom AMI 

mkannwischer:
This commit switches away from a custom AMI and instead specifies the volume size which should achieve the same.

Two small changes stapled onto this:
 - Correct the used SLOTHY models in the READMEs
 - fix a typo in autogen


hanno-becker (approved):
Looks good on paper; let's see what the CI says.

Issue #1307, opened 2025-11-13
Assembly: Use consistent syntax for macro definitions and invocations


hanno-becker:
* Relates to: https://github.com/pq-code-package/mlkem-native/issues/1304

Most assemblers seems to support `.macro` definitions/invocations with and without commas between parameters. Some, however, require a comma. In the code base, we mix the two styles on the codebase -- sometimes even in the same file. 

**Task:**
- Add a job to `autogen` which identifies macro definitions and invocations in `*.S` files and modifies them to use a comma as the separator between arguments.
- Re-run `autogen`.

hanno-becker:
I get the impression that some older assemblers only support `,` separated arguments, so we should probably go for that. This may be related to #1304

PR #1313, opened 2025-11-19
Use consistent syntax for macro definitions and invocations

hanno-becker:
Most assemblers seems to support .macro definitions/invocations with and without commas between parameters. Some, however, require a comma. In the code base, we mix the two styles on the codebase -- sometimes even in the same file.

This commit adjusts `autogen` to automatically adjust definitions and invocations of assembly macros to use comma-separated arguments.

* Resolves https://github.com/pq-code-package/mlkem-native/issues/1307


scripts/autogen
```diff
@@ -2050,6 +2054,85 @@ def check_asm_loop_labels():
...
+    update_file(filename, "\n".join(new_lines))
+
+
+def normalize_asm_macro_syntax():
+    """Normalize macro syntax in all assembly files"""
+    # Operate on assembly files in dev/ only. The ones in mlkem/ are autogenerated.
+    files = list(filter(lambda s: s.startswith("dev/"), get_asm_source_files()))
+    files += list(filter(lambda s: s.endswith(".inc"), get_files("dev/**/*.inc")))
+    with ThreadPoolExecutor() as executor:
+        results = list(executor.map(normalize_asm_macro_syntax_for_file, files))
```
mkannwischer: nit: remove the results here

scripts/autogen
```diff
@@ -2050,6 +2054,85 @@ def check_asm_loop_labels():
...
+
+    # First pass: collect macro names
+    macro_names = set()
+    for line in lines:
+        macro_match = _RE_MACRO_DEF.match(line)
+        if macro_match:
+            macro_names.add(macro_match.group(1))
+
+    # Second pass: normalize syntax
+    modified = False
```
mkannwischer: nit: motified is never used

mlkem/src/native/x86_64/src/fq.inc
mkannwischer, line 1: Can we remove those inc files in mlkem/src/native/x86_64/src? They are not used anyway.

hanno-becker: Not so easily, unfortunately, because we use `.inc` both for C and ASM includes.

scripts/autogen
hanno-becker: Fixed

scripts/autogen
hanno-becker: Removed here and in a few other places

mlkem/src/native/x86_64/src/fq.inc
hanno-becker: Let's do this in a follow-up.

Issue #1315, opened 2025-11-20
Passing CFLAGS as make argument results in KAT failure


mkannwischer:
We usually pass CFLAGS as environment variables:
```
CFLAGS="-std=c11" make quickcheck
```
works perfectly fine

However, it should also work to pass it as an argument to make:
```
make quickcheck CFLAGS="-std=c11"
```
Currently this leads to an obscure error when running the ML-KEM-512 testvectors

hanno-becker:
@mkannwischer OOI, can you say more about _why_
```
make quickcheck CFLAGS="-std=c11"
```
fails currently?

PR #1316, opened 2025-11-20
Make: Support CFLAGS as both env var and make argument

mkannwischer:
We usually pass arguments to make through environment variables: CFLAGS="-std=c11" make quickcheck

It currently does not work to pass it as a make argument: make quickcheck CFLAGS="-std=c11"

This commit fixes that and extends the C90 CI to test both variants.

- Resolves https://github.com/pq-code-package/mlkem-native/issues/1315

### DIFF
--- a/test/mk/config.mk
+++ b/test/mk/config.mk
@@ -30,7 +30,8 @@ CC_AR  := $(CROSS_PREFIX)$(CC_AR)
 #################
 # Common config #
 #################
-CFLAGS := \
+
+override CFLAGS := \
 	-Wall \
 	-Wextra \
 	-Werror=unused-result \


.github/workflows/base.yml
```diff
@@ -152,11 +152,17 @@ jobs:
...
+          make clean >/dev/null
+          OPT=1 CFLAGS="-std=c90" make quickcheck
+      - name: make quickcheck (CFLAGS as make argument)
         run: |
-          OPT=0 CFLAGS=-std=c90 make quickcheck
           make clean >/dev/null
-          OPT=1 CFLAGS=-std=c90 make quickcheck
+          OPT=0 make quickcheck CFLAGS="-std=c90"
+          make clean >/dev/null
+          OPT=1 make quickcheck CFLAGS="-std=c90"
```
hanno-becker: I'm not sure this test is worth having. If we do want it, we should have a means to check if the CFLAGS actually were applied, which we don't have at the moment?

test/mk/config.mk
```diff
@@ -30,7 +30,8 @@ CC_AR  := $(CROSS_PREFIX)$(CC_AR)
 #################
 # Common config #
 #################
-CFLAGS := \
+
+override CFLAGS := \
```
hanno-becker, line 34: There is value in the previous version since it allows you to completely specify CFLAGS rather than merely amending the default CFLAGS. However, I agree that env vs Make variable is not a good way to do this. Have you checked if we anywhere pass CFLAGS via Makefile arguments in the CI?

.github/workflows/base.yml
mkannwischer: Well this test would have caught this problem much earlier.
But, okay - removed.

test/mk/config.mk
mkannwischer: Yes, I checked that CFLAGS was never used as a Makefile argument in CI - it would have immediately resulted in a KAT failure as far as I can see. 


Completely setting CFLAGS is incompatible with our current build system as we rely on setting `MLK_CONFIG_PARAMETER_SET` internally. 

If you meant only
```
override CFLAGS := \
	-Wall \
	-Wextra \
	-Werror=unused-result \
	-Wpedantic \
	-Werror \
	-Wmissing-prototypes \
	-Wshadow \
	-Wpointer-arith \
	-Wredundant-decls \
	-Wconversion \
	-Wsign-conversion \
	-Wno-long-long \
	-Wno-unknown-pragmas \
	-Wno-unused-command-line-argument \
	-O3 \
	-fomit-frame-pointer \
	-std=c99 \
	-pedantic \
	-MMD \
	$(CFLAGS)
```

Most of these flags can be cancelled by their opposites/alternative values (e..g, -Wno-error). Notable exceptions are apparently `-Wall` and `-Wextra` for which you can't just turn off all covered warnings in one flags (one would have to turn them off individually).
Another exception is apparently `-MMD` which can't be undone in a later flag - why do we need that anyway?

Seems very unlikely that this ever becomes a problem though at least with the current flags. I'd not worry about this now.

hanno-becker: > Most of these flags can be cancelled by their opposites/alternative values (e..g, -Wno-error). 

This assumes that the compiler knows about them in the first place. I seem to remember that at least compcert didn't like a few of our standard options (and failed because of that).

mkannwischer: Very good point! I did not think of this. Can we open another issue for this and explore how this can best be solved?

hanno-becker: Ok!

hanno-becker (approved):

PR #1317, opened 2025-11-20
tests: Run check-namespace script also for tests all

mkannwischer:
Currently the check-namespacing script is only run when tests func --check-namespace
is called

For
tests all, the --check-namespace has no effect.

This commit fixes that by adding the namespace check to the all test as well.

This is important as we only run the all target in CI.


hanno-becker (approved):
Good catch

PR #1321, opened 2025-11-21
CBMC: Allow specification of per-proof timeouts

hanno-becker:
This commit adds support for per-proof timeouts in CBMC proofs to prevent CI from running for hours on failing or slow proofs. The implementation leverages Litani's existing timeout mechanism via the CBMC_TIMEOUT environment variable.

Add --per-proof-timeout argument to run-cbmc-proofs.py (default: 1800s) Pass timeout to Litani via CBMC_TIMEOUT environment variable Modify lib/summarize.py to detect timeout_reached flag and display "timeout" instead.  Add --per-proof-timeout argument to tests cbmc command Set explicit 30-minute timeout in CI workflow

* Port of https://github.com/pq-code-package/mldsa-native/pull/734

### DIFF
--- a/.github/actions/cbmc/action.yml
+++ b/.github/actions/cbmc/action.yml
@@ -53,5 +53,5 @@ runs:
         shell: ${{ env.SHELL }}
         run: |
           echo "::group::cbmc_${{ inputs.mlkem_k }}"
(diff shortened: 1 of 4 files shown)


mkannwischer (approved):
LGTM.

PR #1325, opened 2025-11-28
config.h: Add missing MLK_CONFIG_NO_ASM_VALUE_BARRIER

willieyz:
- During the review of https://github.com/pq-code-package/mldsa-native/pull/745#pullrequestreview-3517898391, we discovered that the macro MLK_CONFIG_NO_ASM_VALUE_BARRIER is missing in mlkem-native, adde it and removed the corresponding exception in autogen.
- We also added the missing FIPS203 citation in this PR.


mlkem/src/config.h
```diff
@@ -452,6 +458,20 @@
...
+/******************************************************************************
+ * Name:        MLK_CONFIG_NO_ASM_VALUE_BARRIER
+ *
+ * Description: If this option is set, mlkem-native will be built without
+ *              use of native code or inline assembly for value barriers.
+ *
+ *              By default, inline assembly (if available) is used to implement
+ *              value barriers.
+ *              Without inline assembly, mlkem-native will use a global volatile
+ *              'opt blocker' instead; see ct.h.
```
mkannwischer: There is no ct.h in mlkem-native. It's called verify.h

willieyz: Hello, @mkannwischer , thank you for your review,
I had remove the ct.h and replace with verify.h, also check for other place, this is the only miss copy error in mlkem about ct.h, thank you!

Issue #1300, opened 2025-11-11
Add shellcheck as a static analysis tool for shell scripts


L-series:
As a part of the static analysis toolset it would be iteresting to include something like [shellcheck](https://github.com/koalaman/shellcheck) which is a linter for shell scripts. 

Here is a snippet of some suggestions when calling shellcheck on the `META.sh` script which is used in testing:

```
In META.sh line 18:
  grep $1 -A 1 |
       ^-- SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
  grep "$1" -A 1 |


In META.sh line 19:
  grep $2 |
       ^-- SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
  grep "$2" |


In META.sh line 23:
if (which yq 2>&1 >/dev/null); then
             ^--^ SC2069 (warning): To redirect stdout+stderr, 2>&1 must be last (or use '{ cmd > file; } 2>&1' to clarify).


In META.sh line 24:
  QUERY=".implementations | .[] | select(.name==\"$1\") | .\"$2\""
         ^-- SC2089 (warning): Quotes/backslashes will be treated literally. Use an array.


In META.sh line 25:
  echo "cat $META | yq "$QUERY" -r"
                        ^----^ SC2027 (warning): The surrounding quotes actually unquote this. Remove or escape them.
                        ^----^ SC2090 (warning): Quotes/backslashes in this variable will not be respected.
                        ^----^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
  echo "cat $META | yq ""$QUERY"" -r"


In META.sh line 28:
  if [[ $VAL_JQ != $VAL ]]; then
                   ^--^ SC2053 (warning): Quote the right-hand side of != in [[ ]] to prevent glob matching.


In META.sh line 44:
  echo $VAL
       ^--^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
  echo "$VAL"
```

Moreover, one can use shellcheck as an aid to make their scripts fully POSIX compliant. Practically, this adds support for systems using `sh`, `dash` or `ksh` as their shell (this also removes bash as a project dependency). Here is a snippet of `shellcheck -s sh` on the same file:

```
In META.sh line 28:
  if [[ $VAL_JQ != $VAL ]]; then
     ^-------------------^ SC3010 (warning): In POSIX sh, [[ ]] is undefined.
[...]

PR #1329, opened 2025-12-01
linting: add shellcheck support for shell script linting

L-series:
This pr introduces linting for shell scripts. It also fixes the warnings that the linter brought up initally. 

https://github.com/pq-code-package/mlkem-native/issues/1300


scripts/format
```diff
@@ -29,29 +27,33 @@ error()
...
-if ! command -v shfmt 2>&1 >/dev/null; then
+if ! command -v shfmt >/dev/null 2>&1; then
   error "shfmt not found. Are you running in a nix shell? See BUILDING.md."
   exit 1
 fi
-shfmt -s -w -l -i 2 -ci -fn $(shfmt -f $(git grep -l '' :/))
+
+# Find all scripts in the repo
+ALL_FILES=$(git grep -l '' :/)
+SHELL_SCRIPTS=$(echo "$ALL_FILES" | xargs shfmt -f)
```
hanno-becker: What about this:
```bash
SHELL_SCRIPTS=$(git grep -l '' :/ | sed "s/^/'/;s/$/'/" | tr '\n' ' ' | xargs -L 1 shfmt -f)
```
This produces a newline separated list of shell script filesnames while invoking `shfmt -f` only once. 
```bash
SHELL_SCRIPTS=$(git grep -l '' :/ | xargs -I {} shfmt -f '{}')
```
achieves the same thing but is much slower because it invokes `shfmt -f` once per file.

And then you can do 
```bash
echo "$SHELL_SCRIPTS" | xargs -L 1 shfmt -s -l -i 2 -ci -fn
```
for the formatting, I believe (due to the small number of shell scripts, it doesn't matter that there are multiple invocations here, though you could do the same munging as above to reduce it to one invocation).

L-series: Hmm, personally I prefer the second alternative because it is somehow easier to understand. It comes at a slight performance cost yes, however it is marginal in the grand scheme. 

First implementation :
```
git grep -l '' :/  0,01s user 0,02s system 254% cpu 0,012 total
sed "s/^/'/;s/$/'/"  0,00s user 0,00s system 29% cpu 0,011 total
tr '\n' ' '  0,00s user 0,00s system 11% cpu 0,011 total
xargs -L 1 shfmt -f  0,01s user 0,01s system 30% cpu 0,063 total
```
Second implementation: 
```
git grep -l '' :/  0,01s user 0,02s system 251% cpu 0,011 total
xargs -I {} shfmt -f '{}'  0,48s user 1,01s system 115% cpu 1,285 total
```

hanno-becker: Yes but in the second implementation, `xargs -I {} shfmt -f '{}'` is called once per file; on my system,
```bash
echo "$SHELL_SCRIPTS" | xargs -L 1 shfmt -s -l -i 2 -ci -fn
```
takes >10s, adding to the already slow `format`. How long does it take on your's? 

While clumsy,
```bash
SHELL_SCRIPTS=$(git grep -l '' :/ | sed "s/^/'/;s/$/'/" | tr '\n' ' ' | xargs -L 1 shfmt -f)
```
runs in 0.3s because it only invokes `shfmt` once.

I've noted before that my shell seems to be rather slow, not sure why, so it may be specific to my setup -- please check.

L-series: On my end the first implementation takes only around half a second. I wonder why you are seeing times of 10+ sec. For reference I'm on an 8 core laptop running mainline linux kernel.

hanno-becker: Apologies, I miswrote. It's
```bash
SHELL_SCRIPTS=$(git grep -l '' :/ | xargs -I {} shfmt -f '{}')
```
that takes a long time because it sequentially spawns so many subprocesses (750). How long does it take for you? Even with `-P $(nproc)` is ~3s for me.

The following works, too, and is a bit cleaner than the `sed+tr` hack:
```bash
SHELL_SCRIPTS=$(git grep -l '' :/ | xargs printf "'%s' " | xargs -L 1 shfmt -f)
```

I don't know why it's slower on my end, but bottomline (until I understand it) is that the above makes `format` far slower than what it already is, which I would really like to avoid as I use it all the time in day-to-day development.

cc @mkannwischer

L-series: Hmm, that's quite odd. Personally I'm fine with either one, I agree that we should avoid unnecessary slowdowns if possible. 

Here are benchmarks on my pc:
```
git grep -l '' :/  0.00s user 0.02s system 291% cpu 0.008 total
xargs -I {} shfmt -f '{}'  0.25s user 0.57s system 108% cpu 0.750 total
```
and laptop:
```
git grep -l '' :/  0,01s user 0,02s system 227% cpu 0,013 total
xargs -I {} shfmt -f '{}'  0,45s user 0,91s system 116% cpu 1,167 total
```

hanno-becker: It may be an oddity on my system, but until I find out, I'd prefer if we could use the faster version
```
SHELL_SCRIPTS=$(git grep -l '' :/ | xargs printf "'%s' " | xargs -L 1 shfmt -f)
```
which is still fairly readable I'd say. Can you update the PR accordingly? Would be good to get it over the line.

L-series: Pushed the changes, everything should be in line with mldsa!

scripts/lint
```diff
@@ -106,15 +104,36 @@ gh_error_simple()
...
+  fi
+
+  checkerr "Lint shellcheck" "$(echo $SHELL_SCRIPTS | xargs shellcheck --severity=warning)"
+}
+
 # Formatting
 SUCCESS=true
 
+# Get list of shell scripts and .c/.h files for linting
+ALL_C_FILES=$(git ls-files ":/*.c" ":/*.h")
```
hanno-becker: While we are at it, can we align this with `scripts/format` and filter on non-symlinks? Otherwise, we're re-checking the same files over an over again, making `lint` notably slower than `format`.
```bash
ALL_C_FILES=$(git ls-files ":/*.c" ":/*.h" | xargs -P "$nproc$" -I {} sh -c 'test ! -L "{}" && echo "{}"')
```
nproc needs copying from `format` I think.

Or, perhaps better, write the format dry-run as an explicit loop as in the other linting steps, which already check for symlinks.

hanno-becker (changes requested):
Thank you for your patience, @L-series. I noticed one major inefficiency in `lint` which we might want to address along the way; see comment.

scripts/lint
L-series: Sounds good! I changed it to use a loop as per the other linting steps.

L-series:
@hanno-becker No worries! I made the changes to `lint` and now the runtime is comparable to `format`.

hanno-becker (approved):
LGTM

mkannwischer (approved):

Issue #1338, opened 2025-12-03
Stack usage is much higher than mlkem-c-embedded


davidchisnall:
I have been trying these repositories to get some rough performance estimates for the performance of CHERIoT cores for PQC.  I initially started with mlkem-c-embedded and the stack usage peaked at around 4 KiB.  Switching to mlkem-native, it gets closer to 15 KiB.  The performance is within a rounding error of the same (a few percent either way, but for this kind of testing well within error margins).

hanno-becker:
Thank you @davidchisnall! How much stack do you have?

davidchisnall:
Most systems use 1-4 KiB stacks per thread.  For TLS, we typically have 6 KiB because BearSSL does so love to do stack allocations.  I had to modify our RTOS slightly to support stacks as big as this needs (which was something I meant to do anyway, so a good motivation).  

In my ideal world, you'd be able to pass in the space for the intermediate computations so that we could allocate it on the heap, which means the memory is reclaimed as soon as we're done with it and can use it for something else (we have hardware-enforced temporal safety for the heap, so confidentiality is not a concern).

hanno-becker:
If at least the largest internal structure allocations (e.g
https://github.com/pq-code-package/mlkem-native/blob/49caebbb19a2ef2e593aff79d34fa2e76560e83e/mlkem/src/indcpa.c#L468) were (configurably) heap allocated, would that do something for you?

This is something we could achieve fairly easily I think.

davidchisnall:
With the disclaimer that I have about an hour and a half's experience with this codebase:

Yes, that function, `mlk_indcpa_dec`, and `mlk_indcpa_keypair_derand` appear to be responsible for a lot of the stack allocations.  If there were a way of passing in the space for the big `mlk_polyvec`, `mlk_poly`, and  `mlk_polyvec_mulcache` structures that these use then that would probably help a lot.

hanno-becker:
Do you need to pass in the space or is it enough to make allocation flexible as suggested in https://github.com/pq-code-package/mldsa-native/issues/763? That would allow you to remap it to the heap.

davidchisnall:
That would work as long as we could pass in a context parameter that is passed to the `MLD_ALLOC` and `MLD_FREE` macros.  We'd like to be able to build the crypto parts as a shared library and CHERIoT shared libraries are stateless and do not hold a capability that authorises allocation on their own, they must have one passed in by the calling compartment.

[...]

PR #1339, opened 2025-12-04
Add context parameter

davidchisnall:
Following on from the discussion in #1338, this provides a first cut at adding an optional context parameter.  This is two commits:

 - The first exposes the context parameter in the public APIs.
 - The second plumbs it through to some of the internal places that use big stack allocations.

This should serve as a foundation for allowing stack allocations to be provided from some external space (heap allocations or otherwise).

### DIFF
--- a/mlkem/src/common.h
+++ b/mlkem/src/common.h
@@ -183,10 +183,10 @@
   (arg0, arg1, arg2, arg3, context)
 #else
 #define MLK_CONTEXT_PARAMETERS_0() ()
-#define MLK_CONTEXT_PARAMETERS_1(arg0) (arg0)
-#define MLK_CONTEXT_PARAMETERS_2(arg0, arg1) (arg0, arg1)
-#define MLK_CONTEXT_PARAMETERS_3(arg0, arg1, arg2) (arg0, arg1, arg2)
-#define MLK_CONTEXT_PARAMETERS_4(arg0, arg1, arg2, arg3) \
+#define MLK_CONTEXT_PARAMETERS_1(arg0, context) (arg0)
+#define MLK_CONTEXT_PARAMETERS_2(arg0, arg1, context) (arg0, arg1)
+#define MLK_CONTEXT_PARAMETERS_3(arg0, arg1, arg2, context) (arg0, arg1, arg2)
+#define MLK_CONTEXT_PARAMETERS_4(arg0, arg1, arg2, arg3, context) \
   (arg0, arg1, arg2, arg3)
 #endif
 


mlkem/mlkem_native.h
```diff
@@ -128,6 +128,23 @@
 #define MLK_API_MUST_CHECK_RETURN_VALUE
 #endif
 
+/*
+ * If the integration wants to provide a context parameter for use in
+ * platform-specific hooks, then it should define this parameter.
+ */
+#ifdef MLK_CONFIG_INTEGRATION_REQUIRES_CONTEXT
+#ifndef MLK_CONTEXT_PARAMETER_TYPE
+#define MLK_CONTEXT_PARAMETER_TYPE void
```
hanno-becker: I forgot to reply to your comment on this in #1338, but I think you were right that we can always use `void` here and leave it to the user to cast appropriately.

mlkem/src/kem.c
```diff
@@ -305,7 +305,7 @@ int crypto_kem_dec(
     return -1;
   }
 
-  mlk_indcpa_dec(buf, ct, sk);
+  mlk_indcpa_dec(buf, ct, sk MLK_CONTEXT_PARAMETER);
```
hanno-becker: The mechanism itself is OK for me, but I think this `,`-less way of passing the optional context is extremely confusing.

One could consider something like
```c
mlk_indcpa_dec MLK_CTXT(buf, ct, sk);
```
or
```c
mlk_indcpa_dec (MLK_CTXT(buf, ct, sk));
```
for applications, and
```c
void mlk_indcpa_dec MLK_CTXT_DECL(args....)
```
for declarations (like an attribute), where `MLK_CTXT / MLK_CTXT_DECL` are variadic macros that optionally append `ctxt` / `void *ctxt` at the end?

Happy about other ideas (cc @mkannwischer)

hanno-becker: Another idea:

Write the code as-if the ctxt argument was unconditional, e.g.
```c
void mlk_indcpa_dec
  (void *ctxt,
   uint8_t m[MLKEM_INDCPA_MSGBYTES],
   const uint8_t c[MLKEM_INDCPA_BYTES],
   const uint8_t sk[MLKEM_INDCPA_SECRETKEYBYTES])
```
and
```c
mlk_indcpa_dec(ctxt, buf, ct, sk);
```
but do something like
```c
#define mlk_indcpa_dec MLK_NAMESPACE_K(dec) MLK_HAS_CTXT
```
where either
```c
#define MLK_HAS_CTXT(...) (__VA_ARGS__)
```
or
```c
#define MLK_HAS_CTXT(ctxt,...) (__VA_ARGS__)
```

hanno-becker: And for the top-level `mlkem_native.h` header, we can just do the obvious thing to make it least confusing:
```c
int MLK_API_NAMESPACE(keypair_derand)(
#if defined(MLK_CONFIG_API_WITH_CONTEXT)
    void *ctxt,
#endif
    uint8_t pk[MLKEM_PUBLICKEYBYTES(MLK_CONFIG_API_PARAMETER_SET)],
    uint8_t sk[MLKEM_SECRETKEYBYTES(MLK_CONFIG_API_PARAMETER_SET)],
    const uint8_t coins[2 * MLKEM_SYMBYTES]);
```

mlkem/mlkem_native.h
davidchisnall: This macro is used in only one place, so it seemed low cost to add here in exchange for avoiding some casts elsewhere.

mlkem/src/kem.c
davidchisnall: This proposal moves the context parameter to the first argument.  I'm happy with that, but I assumed the APIs had been designed to have consistent parameter orders and take advantage of in-place forwarding for values that are passed down the stack.  If that's not the case, the change is easy.  Placing it last avoids disrupting argument layout which I would expect to save a few move instructions.

I had another version that had a macro that expanded to the argument if the context was provided or not otherwise, but it ends up being more verbose.

The proposal to use `__VA_ARGS__` would likely be rejected by your linters, since that's a C99 addition (it's been over 25 years since I worked on a project with a C dialect lower than C99, and these days most of the C I touch defaults to C23, since it's easier to support than C89 due to removing problematic K&R declaration, but your policy says C90 so I'm trying to abide by that).

hanno-becker: Hmm... crumbs, I had forgotten about `__VA_ARGS__` not being C90.

hanno-becker: I am out of good ideas. I also don't want the search for an 'elegant' solution within the constraints of C90 block progress towards the actual functional issue you have.

To progress experimentation, we could for now just add `void *ctxt` unconditionally to all relevant functions. At the top-level, we'd add an `_ext` suffix to the new API, and re-define the old API by forwarding NULL as the context. It would be interesting to see in this case whether the compiler is able to realize that `ctxt` is unused everywhere in the default case of stack allocation -- at least in the usecase where the library is built from the monolithic `mlkem_native.c`, where all but the top-level API have internal linkage.

hanno-becker: I suppose one can still do https://github.com/pq-code-package/mlkem-native/pull/1339#discussion_r2591481534 if one has a few `MLK_PARAMS_WITH_CTXT_N` macros which do the ctxt-contraction for a fixed number of parameters. The pain of looking up the # of parameters and writing
```c
#define mlk_indcpa_dec MLK_NAMESPACE_K(dec) MLK_PARAMS_WITH_CTXT_3
```
is perhaps acceptable.

davidchisnall: I don't really like adding an unused parameter because that will hurt code generation on all embedded platforms that don't use this feature and aren't using link-time optimisation.

davidchisnall: I'll see if I can come up with a cleaner solution.

hanno-becker: What do you think about https://github.com/pq-code-package/mlkem-native/pull/1339#discussion_r2592426015 ? That still has the benefits of https://github.com/pq-code-package/mlkem-native/pull/1339#discussion_r2591481534

davidchisnall: It's probably the least-bad option so far, but it is still annoying to have to write a different macro based on the number of arguments.

mkannwischer: I also can't think of a better solution right now.
I'd be okay with adding that if that solves your problem, @davidchisnall. Do you want to give it a try?

hanno-becker: @davidchisnall Do you still have bandwidth to do this?

davidchisnall: Not right now, I'm on holiday for Christmas, but I should when I get back.

hanno-becker: Have a good break

hanno-becker:
@davidchisnall What are your plans for this?

davidchisnall:
Just catching up after Christmas.  I hope to make a start on this either today or tomorrow.

davidchisnall:
I've rebased it and plumbed through the context parameter to everything that calls `MLK_ALLOC`.  I have tested it in a configuration that does *not* use the context parameter, I'm now going to make it use the context in a custom `MLK_ALLOC` implementation.

davidchisnall:
And now tested using heap memory, with this in the platform config file:

```c
#define MLK_CONTEXT_PARAMETER_TYPE AllocatorCapability
#define MLK_CONFIG_INTEGRATION_REQUIRES_CONTEXT

// TODO: The timeout should probably be passed in as part of the context, but
// for now just fail instantly if there's a failure.
#define MLK_CUSTOM_ALLOC(value, type, count, context)                          \
	Timeout t     = {0, 0};                                                    \
	type   *value = heap_allocate_array(&t, context, sizeof(type), count, 0);     \
	value         = cheri_is_valid(value) ? value : NULL;

#define MLK_CUSTOM_FREE(value, type, count, context) \
	heap_free(context, value)
```

The results of a tiny toy benchmark before:

```console
PQC benchmark: Starting: stack used: 224 bytes, cycles elapsed: 38
PQC benchmark: Generated ML-KEM key pair: stack used: 14272 bytes, cycles elapsed: 4603260
PQC benchmark: Encrypted secret pair with ML-KEM: stack used: 17408 bytes, cycles elapsed: 1737502
PQC benchmark: Decrypted secret pair with ML-KEM: stack used: 18464 bytes, cycles elapsed: 2142933
PQC benchmark: Compared results successfully for ML-KEM: stack used: 224 bytes, cycles elapsed: 413
```

And after:

```console
PQC benchmark: Starting: stack used: 224 bytes, cycles elapsed: 38
PQC benchmark: Generated ML-KEM key pair: stack used: 224 bytes, cycles elapsed: 4109196
PQC benchmark: Encrypted secret pair with ML-KEM: stack used: 224 bytes, cycles elapsed: 1772934
PQC benchmark: Decrypted secret pair with ML-KEM: stack used: 224 bytes, cycles elapsed: 2211146
PQC benchmark: Compared results successfully for ML-KEM: stack used: 224 bytes, cycles elapsed: 413
```

For comparison, these were the numbers from before Christmas:

```console
PQC benchmark: Starting: stack used: 224 bytes, cycles elapsed: 41
PQC benchmark: Generated ML-KEM key pair: stack used: 14304 bytes, cycles elapsed: 5143987
PQC benchmark: Encrypted secret pair with ML-KEM: stack used: 17440 bytes, cycles elapsed: 1773235
PQC benchmark: Decrypted secret pair with ML-KEM: stack used: 18464 bytes, cycles elapsed: 2176226
PQC benchmark: Compared results successfully for ML-KEM: stack used: 224 bytes, cycles elapsed: 414
```

Adding some debug logging to the allocation calls, I see that it is allocating some big objects:

```console
PQC benchmark: Starting: stack used: 224 bytes, cycles elapsed: 38
mlkem: Heap allocation of 64 bytes: 0x20061410 (v:1 0x20061410-0x20061450 l:0x40 o:0x0 p: G RWcgm- -- ---)
mlkem: Heap allocation of 10112 bytes: 0x20061480 (v:1 0x20061480-0x20063c00 l:0x2780 o:0x0 p: G RWcgm- -- ---)
PQC benchmark: Generated ML-KEM key pair: stack used: 224 bytes, cycles elapsed: 3608373
mlkem: Heap allocation of 32 bytes: 0x20061458 (v:1 0x20061458-0x20061478 l:0x20 o:0x0 p: G RWcgm- -- ---)
mlkem: Heap allocation of 128 bytes: 0x20063c08 (v:1 0x20063c08-0x20063c88 l:0x80 o:0x0 p: G RWcgm- -- ---)
mlkem: Heap allocation of 2688 bytes: 0x20063c90 (v:1 0x20063c90-0x20064710 l:0xa80 o:0x0 p: G RWcgm- -- ---)
mlkem: Heap allocation of 13088 bytes: 0x20064740 (v:1 0x20064740-0x20067a60 l:0x3320 o:0x0 p: G RWcgm- -- ---)
PQC benchmark: Encrypted secret pair with ML-KEM: stack used: 224 bytes, cycles elapsed: 1797605
mlkem: Heap allocation of 1248 bytes: 0x20067a68 (v:1 0x20067a68-0x20067f48 l:0x4e0 o:0x0 p: G RWcgm- -- ---)
mlkem: Heap allocation of 32 bytes: 0x20064718 (v:1 0x20064718-0x20064738 l:0x20 o:0x0 p: G RWcgm- -- ---)
mlkem: Heap allocation of 4864 bytes: 0x20061480 (v:1 0x20061480-0x20062780 l:0x1300 o:0x0 p: G RWcgm- -- ---)
mlkem: Heap allocation of 13088 bytes: 0x20067f60 (v:1 0x20067f60-0x2006b280 l:0x3320 o:0x0 p: G RWcgm- -- ---)
```

Unfortunately, a cross-compartment call (including the heap allocator call) resets the stack high-water mark to the current amount, so these numbers are not meaningful anymore (they record the amount of stack usage at the point where `MLK_CUSTOM_FREE` is called).  

If I add a log of the stack high-water mark in `MLK_CUSTOM_FREE`, I get:

4,160 bytes used in key generation (down from 14,272), 4,256 in encryption (from 17,408), and 4,864 in decryption (from 18,464).  So a lot has moved to the heap, but this still requires a moderately large stack (though on the same order as a thread that does TLS now).

It's a bit annoying that `MLK_FREE` does a null check and a zeroise call.  Both of these are also done internally by our `heap_free` function and zeroing these huge allocations is not exactly free on a microcontroller, so doing it twice is a bit wasteful.

mkannwischer:
Thanks @davidchisnall for the update and all the data! Very useful!

> 4,160 bytes used in key generation (down from 14,272), 4,256 in encryption (from 17,408), and 4,864 in decryption (from 18,464). So a lot has moved to the heap, but this still requires a moderately large stack (though on the same order as a thread that does TLS now).

I opened https://github.com/pq-code-package/mlkem-native/issues/1462 to use `MLK_ALLOC` for more buffers. 
`poly_rej_uniform_x4` accounts for 3 KiB. If you set `MLK_CONFIG_SERIAL_FIPS202_ONLY`, 2 KiB should go away.

davidchisnall:
Thanks, I haven't gone through the configuration options yet, I was firmly in the 'make it work' stage.  With that set, I see (all tests are run with key size 786, on a cycle-accurate Verilator simulation of the CHERIoT Ibex core):

 - *Generate the key* Time: 4,109,196 -> 3,223,120 cycles.  Stack: 4,128 -> 1,920 bytes
 - *Generate and encrypt secret* Time: 1,772,934 -> 1,765,174 cycles. Stack: 4,224 -> 2016 bytes
 - *Decrypt secret* Time: 2,211,146 -> 2,204,957 cycles. Stack:  4,160 -> 1,952 bytes.

So that's a pretty big improvement in stack usage *and* it's faster on a microcontroller core.  So definitely keeping that in the default configuration!

davidchisnall:
The other things I'm slightly unhappy with is that we don't have a way of running cleanup in cases where the code crashes (which can happen if we run out of stack).  The fact that `MTK_FREE` is implemented with a `do {...} while` block means that the calls to the custom alloc and free functions are not in the same scope and so we can't just add our cleanup macros in there (it would be nice if we could completely replace `MTK_FREE`, since it does redundant work).

If I patch `MTK_FREE` to simply call `MTK_CUSTOM_FREE` then I can make this work.  With that additional patch, I can use our cleanup macros and try running with a tiny stack and, when we run out of stack, all of my heap allocations are freed and the top-level functions return failure.

Is there a use case for someone wanting a custom `MTK_FREE` and not being able to add their own `explicit_bzero` call or similar before the call to `free`?

hanno-becker:
@davidchisnall There are two reason why `MLK_FREE` wraps around `MLK_CUSTOM_FREE`: a) To ensure that zeroization happens. b) To work around the nuisance of custom code in the `mlkem_native_config.h` using functions from `verify.h` (specifically, `mld_zeroize`), creating a circular dependency (we touched on this before, I believe).

I agree with you that with proper documentation warning, a) is not a strong argument. b) is certainly not strong either, but it needs some thought how we can allow customization of `MLK_FREE` that can reuse `mld_zeroize`. If that's solved, I am happy to do what you propose and simply allow customizing all of `MLK_FREE`. If we allowed ourselves to drop zeroization from the (test-only) custom implementations using the heap and the bump allocator, we could even do it immediately -- but that would set a bad example.

hanno-becker:
> nuisance of custom code in the mlkem_native_config.h using functions from verify.h (specifically, mld_zeroize), creating a circular dependency (we touched on this before, I believe).

... this is only an issue if one wants to inline `mld_zeroize` into `MLK_FREE` within `mlkem_native_config.h` (as currently done in `custom_heap_alloc_config.h`). If you just call your own `free` wrapper with external linkage, the source file defining that wrapper can still include `verify.h` and refer to `mld_zeroize`. That may be good enough to adjust the existing heap and bump-allocator tests?

I don't have time to do this right now, but bottomline is: I am happy to do what you want, so long as we adjust the heap and bump-allocator tests/configs to include the zeroization via `mld_zeroize`. If you have time, feel free to open a PR.

davidchisnall:
So the main use case, if I understand correctly, is integrations that want to have a custom allocator but do *not* have a custom equivalent of `memset_s` or similar?  If you have a custom free but your platform has `memset_s`, `explicit_bzero` or similar, then you don't need to call `mlk_zeroize` because you have your own implementation that you can use (and which is probably better optimised and so you'll want to use it in place of `mlk_zeroize`).  That seems fairly niche, but maybe there are some examples?  Would moving `mlk_zeroize`'s prototype the public header solve this?

hanno-becker:
@davidchisnall `mlk_zeroize` is already customizable by the user. I wanted the convenience of `MLK_FREE` being able to refer to that potentially customized `mlk_zeroize` so there would only be one place where a custom zeroization function would need to be plugged in. Admittedly, referencing a custom zeroize twice is only a mild inconvenience; and, as mentioned, you can still refer to `mlk_zeroize` in a custom-`MLK_FREE` so long as it's not inlined into the config.

mlkem/mlkem_native.h
davidchisnall: Our use case actually wants to have the `__sealed_capability` qualifier on this, so allowing the user to specify the type turns out to be useful.

mlkem/src/kem.c
davidchisnall: The current version uses the `_n` macro proposal.  It's a much smaller and more readable change than I expected.

hanno-becker: @davidchisnall The current proposal is different from what we considered above (cf https://github.com/pq-code-package/mlkem-native/pull/1339#discussion_r2591481534), which was
```c
#define mlk_indcpa_dec MLK_NAMESPACE_K(dec) MLK_PARAMS_WITH_CTXT_3
```
etc., where `MLK_PARAMS_WITH_CTXT_3(ctxt,a,b,c)` is either `(a,b,c)` or `(c,a,b,c)`.

Then, in the bulk of the code, you can add the context as if it was unconditional, making the change less intrusive.

Am I overlooking something?

davidchisnall: I'm not sure I completely understand that proposal.  The macro is defined with the macro arguments as another macro?  Is that legal in C preprocessor, I thought the macro arguments were parsed before macro expansion?

hanno-becker: Here's an example of what I had in mind (https://godbolt.org/z/8bYsfE5cG):
```c
#include <stdint.h>

#define MLK_USE_CONTEXT

// Namespace helper
#define MLK_NAMESPACE_K(name) mlkem_##name

// Parameter macros with optional context
#ifdef MLK_USE_CONTEXT
  #define MLK_PARAMS_WITH_CTXT_3(ctxt, a, b, c) (ctxt, a, b, c)
  #define MLK_PARAMS_WITH_CTXT_2(ctxt, a, b)    (ctxt, a, b)
#else
  #define MLK_PARAMS_WITH_CTXT_3(ctxt, a, b, c) (a, b, c)
  #define MLK_PARAMS_WITH_CTXT_2(ctxt, a, b)    (a, b)
#endif

// Define the API function name as a macro
#define mlk_foo MLK_NAMESPACE_K(foo) MLK_PARAMS_WITH_CTXT_3
#define mlk_bar MLK_NAMESPACE_K(bar) MLK_PARAMS_WITH_CTXT_2

void mlk_bar(void *ctxt, uint8_t *arg0, void* arg1)
{
    // nothing for now
}

void mlk_foo(void *ctxt, uint8_t *arg0, int arg1, void *arg2)
{
  /* pass through ctxt */
  mlk_bar(ctxt, arg0, arg2);
}
```

hanno-becker: This is for the internal source. In the top-level API, I would suggest to make the absence/presence of the context obvious through
```c
void MLK_API_NAMESPACE(foo)(
    #if defined(MLK_USE_CONTEXT)
    void *ctxt,
    #endif
    uint8_t *arg0, int arg1, void *arg2
);```

Comment (davidchisnall) [reply mlkem/src/kem.c]: I see.  I'll have a go at that approach.  I don't really like the inline `#ifdef`s elsewhere in the top level (I'm not sure about this project's style requirements, I normally work with a coding style that says inline `#ifdef`s should be avoided and all `#ifdef`s should be self contained).

You're also putting the context parameter first.  I don't have a very strong objection here, but (as I mentioned previously) it means that you're shuffling parameters around, which means any optimisation that you've done there is gone (I'm not sure if you have?).  For a C++ wrapper it would be nicer if the context parameter were last because then we can default it (for most of our users, the default would be their malloc capability).

Comment (hanno-becker) [reply mlkem/src/kem.c]: I don't feel strongly about 
```c
void MLK_API_NAMESPACE(foo)(
    #if defined(MLK_USE_CONTEXT)
    void *ctxt,
    #endif
    uint8_t *arg0, int arg1, void *arg2
);
```
vs.
```c
#if defined(MLK_USE_CONTEXT)
void MLK_API_NAMESPACE(foo)(
    void *ctxt,
    uint8_t *arg0, int arg1, void *arg2
);
#else
void MLK_API_NAMESPACE(foo)(
    uint8_t *arg0, int arg1, void *arg2
);
#endif
```
Either way is fine by me. The header is already a bit clumsy to read because of the namespacing macros, so this ifdef-split doesn't make it much worse. I mainly care about keeping the source files readable, and like the above proposal for that reason. 

@mkannwischer What about you?

Regarding the context at front vs. back, I don't feel strongly either. I personally find it easier to read if the context is at the front, but that is just taste. If there are technical reasons to prefer it at the back, that's OK for me as well. We did not (yet) do any optimizations of parameters that would be broken by the insertion of the context.

davidchisnall: > I don't feel strongly about ...

Both of those would be violations of the style guides I normally use.  This is why the `DECLARE` macros were there: so that the conditional code is in the definition of macros, not interleaved with source.  If you really prefer that for this project, I can do that, but I avoid it because it tends to make code harder to read.  This is particularly true in this project where the clang-format rules don't intent preprocessor things (note that your example above is actually invalid C, preprocessor directives must start with a hash at the beginning of the line, but you can add indentation space between it and the next token.  This actually matters in the case where you have a preprocessor macro with a parameter called `define` or `else` that you stringify, though most other cases are unambiguous).

If the internal APIs use something like `MLK_PARAMS_WITH_CTXT_n` in their declarations, then they end up needing `MLK_CONTEXT_PARAMETER_TYPE` there, which adds a bit of clutter, but is mostly fine.  Both mean that the way that the functions are declared in the public and internal headers is different, which I would have considered a maintenance problem, but if you're happy with that then I can do it.

hanno-becker: I am not too fond of the `DECLARE` macro style -- because the user has to look up what the macro actually means -- and prefer either of the two approaches above for the public API, which to me are more explicit. I think this is ultimately taste. @mkannwischer What do you think?

mkannwischer: I agree with @hanno-becker here. Readability of `mlkem_native.h` is not a primary concern right now, but readability of the sources is. I prefer going with Hanno's [proposal](https://github.com/pq-code-package/mlkem-native/pull/1339#discussion_r2672480032).
Between Hanno's two variants for the top-level, I'd go for the [more compact one](https://github.com/pq-code-package/mlkem-native/pull/1339#discussion_r2672504511) for now. We can consider an overhaul of the `mlkem_native.h` at a later point. 

I don't feel strongly about having the context first or last. I think they are equally easy to read. Having it first would make it possible to switch to the variadic macros should we eventually give up on C90. Having it last enables the optimizations described by @davidchisnall. Both fine for me.

davidchisnall: I've pushed a new version that hopefully does what you wanted.  If you think this is more maintainable, then I'm happy with it.

mlkem/mlkem_native.h
```diff
@@ -181,6 +181,21 @@
...
+ */
+#ifdef MLK_CONFIG_INTEGRATION_REQUIRES_CONTEXT
+#ifndef MLK_CONTEXT_PARAMETER_TYPE
+#define MLK_CONTEXT_PARAMETER_TYPE void *
+#endif
+#else
+#ifdef MLK_CONTEXT_PARAMETER_TYPE
+#error MLK_CONTEXT_PARAMETER_TYPE is defined but MLK_CONFIG_INTEGRATION_REQUIRES_CONTEXT is not.
+#endif
+#endif
```
hanno-becker: As a sanity check on the config, I think this should go in `common.h` after the inclusion of `config.h`.

mlkem/mlkem_native_config.h
```diff
@@ -655,6 +655,26 @@
...
+ * Name:        MLK_CONFIG_INTEGRATION_REQUIRES_CONTEXT
+ *
+ * Description: Set this to add a context parameter that is provided to public
+ *              API functions and is then available in custom callbacks.
+ *
+ *****************************************************************************/
+/* #define MLK_CONFIG_INTEGRATION_REQUIRES_CONTEXT */
+
+/******************************************************************************
+ * Name:        MLK_CONTEXT_PARAMETER_TYPE
```
hanno-becker: So far, all configuration options start with `MLK_CONFIG`, so I'd prefer this to be named `MLK_CONFIG_CONTEXT_PARAMETER_TYPE`.

mlkem/mlkem_native_config.h
```diff
@@ -655,6 +655,26 @@
  *****************************************************************************/
 /* #define MLK_CONFIG_SERIAL_FIPS202_ONLY */
 
+/******************************************************************************
+ * Name:        MLK_CONFIG_INTEGRATION_REQUIRES_CONTEXT
```
hanno-becker, line 659: Nit: Rename as `MLK_CONFIG_CONTEXT_PARAMETER_REQUIRED`.

mlkem/mlkem_native_config.h
```diff
@@ -655,6 +655,26 @@
...
+/******************************************************************************
+ * Name:        MLK_CONTEXT_PARAMETER_TYPE
+ *
+ * Description: Set this to define a custom type for the context parameter used
+ *              by MLK_CONFIG_INTEGRATION_REQUIRES_CONTEXT.  This should be the
+ *              type of the pointee: the context parameter is always a pointer.
+ *              If this is not set, the context parameter will be void*.
+ *
+ *****************************************************************************/
+/* #define MLK_CONTEXT_PARAMETER_TYPE void */
```
hanno-becker:
```suggestion
/* #define MLK_CONTEXT_PARAMETER_TYPE void * */
```

mlkem/src/common.h
```diff
@@ -186,16 +213,32 @@
...
+#define MLK_CALL_CUSTOM_FREE(v, T, N, context) \
+  MLK_CUSTOM_FREE(v, T, N, context)
+#else
+#define MLK_CALL_CUSTOM_ALLOC(v, T, N, context) \
+  MLK_CUSTOM_ALLOC(v, T, N) 
+#define MLK_CALL_CUSTOM_FREE(v, T, N, context) \
+  MLK_CUSTOM_FREE(v, T, N)
+#endif
+
+#define MLK_ALLOC(v, T, N) MLK_CALL_CUSTOM_ALLOC(v, T, N, context)
```
hanno-becker: Can one not use this?
```c
#define MLK_ALLOC MLK_CUSTOM_ALLOC MLK_CONTEXT_PARAMETERS_3
```

mlkem/mlkem_native.h
hanno-becker: I think we can require the user to provide `MLK_CONFIG_CONTEXT_PARAMETER_TYPE` whenever `MLK_CONFIG_CONTEXT_PARAMETER_REQUIRED` is set, and not provide `void *` as a default. This way, we don't need the logic of setting it to `void *` here, and the sanity check `defined(MLK_CONFIG_CONTEXT_PARAMETER_TYPE) == defined(MLK_CONFIG_CONTEXT_PARAMETER_REQUIRED)` can go into `common.h`.

mlkem/src/common.h
hanno-becker: Hmm, apparently not; pity.

davidchisnall: No, see the comment above.  Macro expansion is single pass, unfortunately.

mlkem/mlkem_native.h
davidchisnall: I've removed the copy here and simplified the check in common.h as suggested.

mlkem/mlkem_native_config.h
davidchisnall: Fixed.

mlkem/mlkem_native_config.h
davidchisnall: Fixed

davidchisnall: (Also fixed the comment, which is now stale).

davidchisnall:
Ooops, I messed up some of the global rename and somehow also undid something that was hiding a merge error.  Should both be fixed now.

mlkem/src/common.h
hanno-becker: Interesting... unexpected that this works for `#define mlk_foo MLK_NAMESPACE(foo) MLK_CONTEXT_PARAMETERS_n`(where the call to `MLK_CONTEXT_PARAMETERS_n` can only be recognized after expansion `mlk_foo`) but not for `#define MLK_ALLOC MLK_CUSTOM_ALLOC MLK_CONTEXT_PARAMETERS_3`. Seems to be due to left-to-right scanning of the preprocessor? 

This is a minimal example:
```c
#define INNER(a, b, c) (a, b, c)
#define OUTER(x, y, z) x + y + z
#define COMBINED OUTER INNER

int main() {
    int r = COMBINED(1, 2, 3);
}
```

davidchisnall: Yup, I was surprised that the first worked (which is why I didn't understand the suggestion originally, and had to remind myself how C preprocessor execution works.  I've mostly replaced preprocessor use with C++ constexpr in code I maintain recently, so this was all paged out of my brain).

hanno-becker: One layer of indirection solves this
```c
#define INNER(a, b, c) (a, b, c)
#define OUTER(x, y, z) x + y + z
#define APPLY(f, args) f args
#define COMBINED(a, b, c) APPLY(OUTER, INNER(a, b, c))

int main() {
    int r = COMBINED(1, 2, 3);
}
```
but in the concrete case at hand, it's not very palatable either: 
```c
#define MLK_APPLY(f, args) f args
#define MLK_ALLOC(ctxt, v, T, N) MLK_APPLY(MLK_CUSTOM_ALLOC, MLK_CONTEXT_PARAMETERS_3(ctxt, v, T, N))
```

@davidchisnall wdyt?

davidchisnall: I have no strong preferences, happy either way.

hanno-becker: Let's do
```c
#define MLK_APPLY(f, args) f args
#define MLK_ALLOC(ctxt, v, T, N) MLK_APPLY(MLK_CUSTOM_ALLOC, MLK_CONTEXT_PARAMETERS_3(ctxt, v, T, N))
```
it's a bit more compact and avoids another `#ifdef` on `MLK_CONFIG_INTEGRATION_REQUIRES_CONTEXT`

davidchisnall: Three variants later, I think I have done what you wanted.  I think this one is now simpler than my original version and means that it's a small change for a future version can do:

```c
#define MLK_FREE(v, T, N) \
  MLK_APPLY(MLK_CUSTOM_FREE, MLK_CONTEXT_PARAMETERS_3(v, T, N, context))
```

If the zeroing is moved to the custom-free's responsibility.

hanno-becker: I agree

hanno-becker: Hmm, sorry, another question.

Now we have the context parameter explicit in the function declarations and definitions in `{kem,indcpa}.{c,h}`, but it is implicit for `MLK_ALLOC` which is still invoked as `MLK_ALLOC(v, T, N)`. I think we should either a) make the context parameter implicit in the functions as well (by auto-adding it as part of MLK_CONTEXT_PARAMETERS_n), in which case `{kem,indcpa}.{c,h}` change hardly at all or b) make it explicit in the `MLK_ALLOC/MLK_FREE` calls as well.

hanno-becker: No, option a) doesn't work, because the shape of the context parameter depends on whether it's a declaration or call. So I think we should go with b) and make the context parameter explicit in `MLK_ALLOC/MLK_FREE` as well, so the code reads as-if the context was unconditionally present.

davidchisnall: I wondered about that.  Happy to make that change.

davidchisnall: Done

davidchisnall:
I believe all review comments are now addressed.

mlkem/src/common.h
```diff
@@ -164,18 +164,49 @@
...
+  (arg0, arg1, arg2, context)
+#define MLK_CONTEXT_PARAMETERS_4(arg0, arg1, arg2, arg3, context) \
+  (arg0, arg1, arg2, arg3, context)
+#else
+#define MLK_CONTEXT_PARAMETERS_0() ()
+#define MLK_CONTEXT_PARAMETERS_1(arg0) (arg0)
+#define MLK_CONTEXT_PARAMETERS_2(arg0, arg1) (arg0, arg1)
+#define MLK_CONTEXT_PARAMETERS_3(arg0, arg1, arg2) (arg0, arg1, arg2)
+#define MLK_CONTEXT_PARAMETERS_4(arg0, arg1, arg2, arg3) \
+  (arg0, arg1, arg2, arg3)
```
hanno-becker:
```suggestion
#define MLK_CONTEXT_PARAMETERS_1(arg0, context) (arg0)
#define MLK_CONTEXT_PARAMETERS_2(arg0, arg1, context) (arg0, arg1)
#define MLK_CONTEXT_PARAMETERS_3(arg0, arg1, arg2, context) (arg0, arg1, arg2)
#define MLK_CONTEXT_PARAMETERS_4(arg0, arg1, arg2, arg3, context) \
  (arg0, arg1, arg2, arg3)
```

hanno-becker:
We still need to work through a test issue: Some tests (at least `stack` and `acvp`, cf. `make stack_512` or `make acvp`) don't use `mlkem_native.h`, but `kem.h`, and the invocations to the public API are therefore expecting a context (macro-)parameter. Ideally, those tests should work directly with `mlkem_native.h`. I need to look more closely, but am happy to be beaten to it.

mlkem/src/kem.c
```diff
@@ -135,8 +137,9 @@ __contract__(
 
 /* Reference: Not implemented in the reference implementation @[REF]. */
 MLK_MUST_CHECK_RETURN_VALUE
-static int mlk_check_pct(uint8_t const pk[MLKEM_INDCCA_PUBLICKEYBYTES],
-                         uint8_t const sk[MLKEM_INDCCA_SECRETKEYBYTES])
+static int(lk_check_pct, uint8_t const pk[MLKEM_INDCCA_PUBLICKEYBYTES],
```
hanno-becker, line 140: This seems to be a remnant from the previous version

hanno-becker: The other definition of `mlk_check_pct` also needs a `MLK_CONFIG_CONTEXT_PARAMETER_TYPE context`

hanno-becker:
I experimented a bit in #1467 and added a few smaller fixes. Can you take a look and add them here if you agree? It looks like at least the base tests pass. (The main purpose of #1467 is to run `autogen` [which needs CI] and the full CI [which unfortunately needs a PR from the main repository], otherwise everything should happen here)

davidchisnall:
The fixes all look correct to me, thank you!

hanno-becker:
Merged via #1467. Thank you @davidchisnall !

davidchisnall:
FYI: I've pushed [updates to the CHERIoT library builds for this and mldsa-native to use heap allocation](https://github.com/CHERIoT-Platform/cheriot-pqc/commit/190274061fb7100f226c0545f9e360f35080e752).  The example now uses a 4 KiB stack, which is far more friendly on embedded systems.

hanno-becker:
@davidchisnall This is great to see, thank you

PR #1342, opened 2025-12-05
Make value barrier volatile

hanno-becker:
This commit modifies the definition of the standard value barrier by adding a `volatile` qualifier, turning it into a side-effect which the compiler cannot elide or reorder with other side-effects. Without `volatile`, there is risk the compiler could correlate input and output merely from the fact that the value barrier is deterministic. We have not seen this happening, but it appears safer nonetheless to add a volatile qualifier.

This would be a deviation from OpenSSL, which also uses a non-volatile value barrier.

### DIFF
--- a/mlkem/src/verify.h
+++ b/mlkem/src/verify.h
@@ -114,21 +114,21 @@ __contract__(ensures(return_value == b)) { return (b ^ mlk_ct_get_optblocker_u8(
 static MLK_INLINE uint32_t mlk_value_barrier_u32(uint32_t b)
 __contract__(ensures(return_value == b))
 {
-  __asm__("" : "+r"(b));
+  __asm__ volatile("" : "+r"(b));
   return b;
 }
 
 static MLK_INLINE int32_t mlk_value_barrier_i32(int32_t b)
 __contract__(ensures(return_value == b))
 {
-  __asm__("" : "+r"(b));
+  __asm__ volatile("" : "+r"(b));
   return b;
 }
 
 static MLK_INLINE uint8_t mlk_value_barrier_u8(uint8_t b)
 __contract__(ensures(return_value == b))
 {
-  __asm__("" : "+r"(b));
+  __asm__ volatile("" : "+r"(b));
   return b;
 }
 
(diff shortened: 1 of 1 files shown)


hanno-becker:
@mkannwischer WDYT

hanno-becker:
mldsa-native port is already here: https://github.com/pq-code-package/mldsa-native/pull/772

hanno-becker:
While at it, the OpenSSL value barrier is slightly different from our's:
```c
static ossl_inline unsigned int value_barrier(unsigned int a)
{
#if !defined(OPENSSL_NO_ASM) && defined(__GNUC__)
    unsigned int r;
    __asm__("" : "=r"(r) : "0"(a));
#else
    volatile unsigned int r = a;
#endif
    return r;
}
```
vs.
```c
static MLK_INLINE uint32_t mlk_value_barrier_u32(uint32_t b)
__contract__(ensures(return_value == b))
{
  __asm__ volatile("" : "+r"(b));
  return b;
}
```

I don't think the separation of input and output is necessary here, but @mkannwischer have another look as well please.

hanno-becker:
> I noticed we are using __asm__ __volatile__ in the zeroize rather than __asm__ volatile. Can we pick one?
__asm__ __volatile__ seems more consistent.

Done.

mkannwischer:
> While at it, the OpenSSL value barrier is slightly different from our's:
> 
> ```c
> static ossl_inline unsigned int value_barrier(unsigned int a)
> {
> #if !defined(OPENSSL_NO_ASM) && defined(__GNUC__)
>     unsigned int r;
>     __asm__("" : "=r"(r) : "0"(a));
> #else
>     volatile unsigned int r = a;
> #endif
>     return r;
> }
> ```
> 
> vs.
> 
> ```c
> static MLK_INLINE uint32_t mlk_value_barrier_u32(uint32_t b)
> __contract__(ensures(return_value == b))
> {
>   __asm__ volatile("" : "+r"(b));
>   return b;
> }
> ```
> 
> I don't think the separation of input and output is necessary here, but @mkannwischer have another look as well please.

I don't see a difference in practice either.

PR #1344, opened 2025-12-05
Use `MLK_CONFIG_EXTERNAL_API_QUALIFIER` in the public header.

### LINKED ISSUE #1343: `MLK_CONFIG_EXTERNAL_API_QUALIFIER` / `MLK_EXTERNAL_API` not used in public header
`MLK_CONFIG_EXTERNAL_API_QUALIFIER` can be set to define a value for `MLK_EXTERNAL_API`.  This is useful for controlling visibility across shared-library boundaries, but it is not used in the public header, which can lead to incorrect code generation at the call site if this macro is actually used.

davidchisnall:
Fixes #1343


mlkem/mlkem_native.h
```diff
@@ -128,6 +128,13 @@
 #define MLK_API_MUST_CHECK_RETURN_VALUE
 #endif
 
+
+#if !defined(MLK_CONFIG_EXTERNAL_API_QUALIFIER)
+#define MLK_EXTERNAL_API
+#else
+#define MLK_EXTERNAL_API MLK_CONFIG_EXTERNAL_API_QUALIFIER
+#endif
```
hanno-becker: At the moment, we keep the API-config `MLK_CONFIG_API_XXX` standalone and separate from the build-config in `config.h` (and depending headers such as `sys.h` or `common.h`); the hope was that it would make it easier to distribute mlkem-native as a shared library. Maybe that's a bad choice and we need to revisit (I've tried in the past to make this easier to use, but without success -- happy about ideas..), but until we change it consistently, we should follow the existing pattern, which would be:
- Document a new `MLK_CONFIG_API_QUALIFIER [optional]` at the top of `mlkem_native.h`.
- Derive `MLK_API_QUALIFIER` from this and use it in the API declarations.
- At the bottom of `common.h`, add `#define MLK_CONFIG_API_QUALIFIER MLK_EXTERNAL_API` to align the API-config with the build-config -- this is merely for an internal API consistency check where we include `mlkem_native.h` and `kem.h` together.

davidchisnall: For what it's worth, as someone trying to build a library, this separation increases the amount of work I need to do.  

The easiest things to work with are ones that just require me to provide a single header describing the integration and include that in both external and internal-facing headers.

The existing config.h is a bit frustrating because it has a bunch of config values that I want, but it's also the only place I can inject an include of a system header needed for the randombytes implementation.

hanno-becker: @davidchisnall Thank you, it's useful to get this feedback; sorry for the pain. What would you concretely suggest?

One way `mlkem_native.h` would be easier to use is to add
```c
#define MLK_CONFIG_API_PARAMETER_SET MLK_CONFIG_PARAMETER_SET
#define MLK_CONFIG_API_NAMESPACE_PREFIX \
  MLK_ADD_PARAM_SET(MLK_CONFIG_NAMESPACE_PREFIX)
#if defined(MLK_CONFIG_EXTERNAL_API_QUALIFIER)
#define MLK_CONFIG_API_QUALIFIER MLK_CONFIG_EXTERNAL_API_QUALIFIER
#endif
```
to `mlkem_native.h` (and removing it from the bottom of `common.h`). Then, if you specify the config via CFLAGS, the same CFLAGS will work for the `mlkem_native.h`. Similarly, if you have the full `config.h` at hand, you can include that before including `mlkem_native.h`, and you also get the right API without having to explicitly duplicate the shared parts of the config.

davidchisnall: It's easier if you aren't stuck with old C, but in my ideal world:

 - The public header would be included in all compilation units.
 - The public header would have something like this at the top:

```c
#if __has_include(<mlk_platform_config.h>)
#    include <mlk_platform.config.h>
#endif
```

Then all of my configuration would go in that file and it would be included if it existed.  With C90, you don't have `__has_include`, so you could make that a mandatory include file (which could be empty).  That's the approach taken by some other projects using older C dialects.

 - Don't have different macro names for internal and external settings.
 - Do define a macro (e.g. `MLK_BUILDING_INTERNAL`) when building the library (so things like Windows dllimport / dllexport can work nicely).
 - Provide an example `mlk_platform.config.h` that shows all of the things I can set, so all I need to do is copy that file, tweak it, and then add it in an include path.

hanno-becker: > Provide an example mlk_platform.config.h that shows all of the things I can set, so all I need to do is copy that file, tweak it, and then add it in an include path.

`config.h` shows all things you can change to configure the build. Many of them affect the implementation only (e.g. whether to use a native backend), not the API, so it seems wrong to include it in the external-facing header. One could try to make the external-facing config a subset of the build-config, but I don't see yet how one would end up with just a single header.

davidchisnall: The things in `config.h` are quite difficult to use.  For example, `MLK_CONFIG_CUSTOM_ZEROIZE` and `MLK_CONFIG_CUSTOM_RANDOMBYTES` can each be set to provide a custom function to call.  I can set `DMLK_CONFIG_CUSTOM_ZEROIZE` and `mlk_zeroize=explicit_bzero` because you happen to include the header that defines `explicit_bzero`.  But I can't do the same with `mlk_randombytes` because you don't include the header where we define the relevant function.  So I need to use `MLK_CONFIG_FILE`.  But then your `config.h` is not included.  And this means that I need to either define those things in my config file or include yours.  So I end up with an `mlkconfig.h` that looks like this:

```c
#include <randombytes.h>
#include "mlkem-native/mlkem/src/config.h"
```

And I also now have a header that wraps your public header that looks like this:

```c
#pragma once
#include <cdefs.h>
#define MLK_CONFIG_API_QUALIFIER __cheriot_libcall
__BEGIN_DECLS
#include "mlkem-native/mlkem/mlkem_native.h"
__END_DECLS
```

The nicest example of a library that's easy to integrate that I've seen is Microvium, where I just provide a `microvium_port.h` and don't require any extra compiler flags anywhere.  Compiling their .c file or including their .h file just works.  Their .h file include the `_platform.h` one automatically and everything just works.

hanno-becker: Thank you for the details and the pointer to Microvium.

The analogue of `microvium_port.h` would be `config.h`, I think. So you would create your own copy of `config.h`, adapt it as needed, and direct the compiler to it via `-DMLK_CONFIG_FILE`. If you do that, then I think all configuration should happen in that file, and not via further CFLAGS -- e.g. rather than `-D MLK_CONFIG_CUSTOM_ZEROIZE` you'd just modify your copy of the `config.h`; same for the other customizations.

At least as far as the build is concerned, would that not be similarly convenient, or what am I missing?

For the external header, we can check if `MLK_CONFIG_FILE` is defined and, if so, include it in `mlkem_native.h`. My primary concern here, as mentioned above, is that then the external-facing header depends on a configuration file which includes implementation-details that are irrelevant for the external API. Is that not the case for Microvium?

davidchisnall: The difference is that:

 - Everything that includes the public header needs to have `MLK_CONFIG_FILE` defined, so that's a configuration I have to propagate to library consumers.  The easiest way of doing this is to wrap the header and tell people not to use the one that they might be used to including in other contexts (or do some `#include_next` hackery).
 - My configuration is still split between the build system (defining `MLK_CONFIG_FILE`) and the config file.

I'm not that concerned about private config leaking into public things for three reasons:

 - Everything is `MLK_` prefixed, so it's unlikely to cause problems.
 - Sometimes it's useful to be able to query the values used internally in an API consumer (at least in `_Static_assert` or similar, to make sure the library really was built as you expect).
 - If you do the`MLK_BUILDING_INTERNAL` thing then I can guard anything I actively want to hide from API consumers in my platform header with a simple `#ifdef`.

hanno-becker: > Everything that includes the public header needs to have MLK_CONFIG_FILE defined, so that's a configuration I have to propagate to library consumers. The easiest way of doing this is to wrap the header and tell people not to use the one that they might be used to including in other contexts (or do some #include_next hackery).

I don't yet follow this point. Even if we had `__has_include` and could use that instead of `MLK_CONFIG_FILE` (with some fixed filename like `mlkem_native_port.h`), you'd still need to pass a suitable `-I ...` to all consumers, which isn't any more or less convenient than passing `-DMLK_CONFIG_FILE=...`?

davidchisnall: When I package the library, I control the locations of the include files.  I can put the config file in the same location as the public header.

davidchisnall: (Unrelated to this tangent, I believe I fixed the issues in the original comment)

hanno-becker: Let me try to summarize what I understood could be improved. 

1. Move `mlkem/src/config.h` to `mlkem/mlkem_native_config.h`.
2. In internal headers, default to `#include "mlkem_native_config.h"`, unless `MLK_CONFIG_FILE` is set, in which case that is included. Perhaps we can get rid of the latter at some point, but it's convenient to keep it so we don't have to change all tests that rely on it.
3. Same for `mlkem_native.h`: Include `mlkem_native_config.h` or `MLK_CONFIG_FILE`, whatever is used. Then rely on the relevant parts of the config to define the external-facing API, rather than having separate configuration options.
4. Potentially add `MLK_BUILD_INTERNAL` which configs can use to guard internal-only configuration, and which is set when `common.h` includes the config, but not when `mlkem_native.h` does it.

With this, both for building and consuming the library, by default you only need to make sure that your adjusted copy of `mlkem_native_config.h` is in the include path.

Did I forget something? Otherwise, I'll give this a try at some point.

hanno-becker:
@davidchisnall Can you rebase atop `main` and add the fixes from https://github.com/pq-code-package/mlkem-native/pull/1351? There's a mismatch between committer and author e-mail which breaks the DCO check when I rebase your commit.

hanno-becker:
Done as part of #1352

PR #1348, opened 2025-12-06
Use a single configuration for internal and external headers

hanno-becker:
Following the conversation with @davidchisnall in #1344, this draft PR explores the use of a single configuration file for both internal files and the external-facing header`mlkem_native.h`.


hanno-becker:
@davidchisnall I'd be glad if you could try this and see if it makes your integration smoother. 

cc @mkannwischer

hanno-becker:
As it stands, this is a breaking change (as witnessed e.g. by the breaking AWS-LC integration) that could only be merged as part of a v2 release.

If we keep `mlkem_native_config.h` as `src/config.h` as done previously and add a 'backwards compatibility mode' in `mlkem_native.h` (by checking if the now-removed `MLK_CONFIG_API_XXX` are set), we might be able to keep this compatible, though.

hanno-becker:
Closing in favor of the (hopefully) API-compatible #1352

Issue #1349, opened 2025-12-07
Makefile: Config tracking is too slow


hanno-becker:
When doing `make` or even just `make clean`, I'm waiting for 2-3s for `GEN     test/build/config.mk` to appar, indicating progress on the tracking of modified configuration options. This should be instantaneous.

**Task:**
- Understand why the preparation of `test/build/config.mk` takes so long, and speed it up.

PR #1350, opened 2025-12-07
Speed up `make`

hanno-becker:
* Resolves https://github.com/pq-code-package/mlkem-native/issues/1349

A plain `make clean` would previously take >3s, primarily because of an unnecessary use of dozens of shell invocations that could be replaced by a mere `subst` inside `make`.

Also, skip the bulk of the makefiles when cleaning, and align the output format of `make clean` with that of other targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,11 +31,16 @@ endif
 
 W := $(EXEC_WRAPPER)
 
+BUILD_DIR ?= test/build
+
+# Skip includes for clean target
+ifneq ($(MAKECMDGOALS),clean)
 include test/mk/config.mk
 include test/mk/compiler.mk
 include test/mk/auto.mk
 include test/mk/components.mk
 include test/mk/rules.mk
+endif
 
 quickcheck: test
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -228,17 +233,29 @@ else
...
+
+.PHONY: $(EXAMPLE_CLEAN_TARGETS)
+
+$(EXAMPLE_CLEAN_TARGETS): clean-%:
+	@echo "  CLEAN   $*"
+	-@$(MAKE) clean -C $* >/dev/null 2>&1
+
+clean: $(EXAMPLE_CLEAN_TARGETS)
+	@echo "  RM      *.gcno *.gcda *.lcov *.o *.so"
+	-@$(RM) -rf *.gcno *.gcda *.lcov *.o *.so
```
mkannwischer: Can this be removed? There should be no such files outside of the BUILD_DIR?

hanno-becker: There shouldn't, yes.

PR #1351, opened 2025-12-07
[CI] Use `MLK_CONFIG_EXTERNAL_API_QUALIFIER` in the public header.

hanno-becker:
Copy of #1344 from a local branch to run full CI, plus minor fixes.


hanno-becker:
Done as part of #1352

Issue #1274, opened 2025-11-04
Do not require fips202x4.h when using `MLK_CONFIG_SERIAL_FIPS202_ONLY`


mkannwischer:
Currently, `MLK_CONFIG_SERIAL_FIPS202_ONLY` only disables using the 4-way batched shake128 for rejection sampling. However, we still use `shake256x4` for sampling error/secret vectors.
While this is not big problem (as shake256x4 can easily be mapped to serial computation by including a custom fips202x4.h), it still forces a consumer requiring `MLK_CONFIG_SERIAL_FIPS202_ONLY` to provide a custom fips202x4.h header (see e.g., the [current OpenTitan integration](https://github.com/zerorisc/expo/blob/master/sw/device/lib/crypto/impl/mlkem/fips202x4_glue.h)). 

It would be more convienent to not include fips202x4 at all if `MLK_CONFIG_SERIAL_FIPS202_ONLY` is set and handle it internally instead.

I implemented that in mldsa-native (as it all fips202x4 APIs there are incompatible with  `MLK_CONFIG_SERIAL_FIPS202_ONLY` anyway) here: https://github.com/pq-code-package/mldsa-native/pull/558

mkannwischer:
Once implemented, we should not forget to adjust the `bring_your_own_fips202_static` example - particularly the README.

PR #1352, opened 2025-12-07
Use a single configuration file for internal and external headers 

hanno-becker:
Variant of #1348 which aims to be API-compatible so we can merge it prior to v2.

-------

Previously, mlkem_native.h required users to set MLK_CONFIG_API_* macros
before including it, separate from the MLK_CONFIG_* macros used for the
build configuration. This distinction between internal and external
configuration it confusing.

Now, mlkem_native.h derives all API settings from the same
build configuration file used internally. Also, the configuration
file is moved from `mlkem/src/config.h` to `mlkem/mlkem_native_config.h`.

- The legacy configuration via MLK_CONFIG_API_XXX still works for
  backwards compatibility, but will be removed in v2.

- New Config Options:
  * MLK_CONFIG_MULTILEVEL_BUILD: Explicit flag for multi-level builds

    This makes it simpler to get headers for multi-level builds.

    In the common case, the configs used in a multi-level build are
    identical except for the setting of MLK_CONFIG_PARAMETER_SET and
    MLK_CONFIG_MULTILEVEL_{NO,WITH}_SHARED. However, the latter only
    affect the build, not the API.

    With MLK_CONFIG_MULTILEVEL_BUILD exposed as a separate option,
    mlkem_native.h can use it to determine whether to suffix the
    namespace prefix with the parameter set (512/768/1024) or not,
    so no adaptation of MLK_CONFIG_MULTILEVEL_{NO,WITH}_SHARED is
    needed. Instead, a multi-level header simply becomes:

    ```c
    #define MLK_CONFIG_PARAMETER_SET 512
    #include "mlkem_native.h"
    #undef MLK_CONFIG_PARAMETER_SET
    #undef MLK_H

    #define MLK_CONFIG_PARAMETER_SET 768
    #include "mlkem_native.h"
    #undef MLK_CONFIG_PARAMETER_SET
    #undef MLK_H

    #define MLK_CONFIG_PARAMETER_SET 1024
    #include "mlkem_native.h"
    #undef MLK_CONFIG_PARAMETER_SET
    #undef MLK_H
    ```

    For backwards compatibility, MLK_CONFIG_MULTILEVEL_BUILD is not
    used in the build, which continues to detect multilevel builds
    via the existing options MLK_CONFIG_MULTILEVEL_{WITH,NO}_SHARED.

  * MLK_CONFIG_NO_SUPERCOP: We had this before in mlkem_native.h as
    MLK_CONFIG_API_NO_SUPERCOP; it's now in the config and documented.

  * MLK_CONFIG_CONSTANTS_ONLY: We had this before in mlkem_native.h as
    MLK_CONFIG_API_CONSTANTS_ONLY; it's now in the config and documented.

- Build-Internal vs API Config Separation: Config file now guards
  build-only options with #if defined(MLK_BUILD_INTERNAL).

- Example Directory Restructuring: All examples refactored with mlkem/
  subdirectories renamed to mlkem_native/ and CFLAGS configuration
  replaced by custom configs.
  Updated READMEs to reflect new paths and clarify which options are
  set in config files vs passed via CFLAGS.


hanno-becker:
@davidchisnall @mkannwischer I'd be glad if you could take a look. The main goal here is to make integration simpler which @davidchisnall reported to be cumbersome in the previous split-config approach.

mlkem/mlkem_native_config.h
```diff
@@ -73,6 +75,95 @@
...
+/******************************************************************************
+ *
+ * Build-only configuration options
+ *
+ * The remaining configurations are build-options only.
+ * They do not affect the API described in mlkem_native.h.
+ *
+ *****************************************************************************/
+
+#if defined(MLK_BUILD_INTERNAL)
```
mkannwischer, line 166: Is this pre-processor guard strictly needed, or is this just to not pollute the consuming CU with irrelevant pre-processor macros?

mlkem/mlkem_native.h
```diff
@@ -56,14 +66,9 @@
  *   MLK_CONFIG_API_PARAMETER_SET or MLK_CONFIG_API_NAMESPACE_PREFIX,
  *   nor include a configuration.
  *
- * # Multi-level builds
+ * - MLK_CONFIG_API_QUALIFIER [optional]
```
mkannwischer, line 69: `MLK_CONFIG_API_QUALIFIER` is not a valid configuration option. Did you want to add that? Or do you want to rename here to the pre-existing option `MLK_CONFIG_EXTERNAL_API_QUALIFIER`

examples/bring_your_own_fips202_static/README.md
```diff
@@ -1,28 +1,56 @@
...
+- Your application already has a FIPS-202 software/hardware implementation you want to reuse
+- Your FIPS-202 implementation does not support multiple active SHA3/SHAKE computations.
 
 ## Components
 
-An application using mlkem-native with a custom FIPS-202 implementation needs the following:
+1. Arithmetic part of mlkem-native: [`mlkem/src/`](../../mlkem/src) (excluding `fips202/`)
+2. A secure random number generator implementing [`randombytes.h`](../../mlkem/src/randombytes.h)
+3. Custom FIPS-202 implementation with header compatible with [`fips202.h`](../../mlkem/src/fips202/fips202.h)
+4. Stub `fips202x4.h` header (the 4x functions are unused in this mode)
```
mkannwischer: This isn't entirely true. `mlk_shake256x4` is still used.
I have opened https://github.com/pq-code-package/mlkem-native/issues/1274 for that. Once that is implemented one does not need to provide the `fips202x4.h` header anymore.

examples/monolithic_build/README.md
```diff
@@ -1,20 +1,46 @@
 [//]: # (SPDX-License-Identifier: CC-BY-4.0)
 
-# Single-level mlkem-native in a single compilation unit
+# Monolithic Build (C Backend)
 
-This directory contains a minimal example for how to build a single instance of mlkem-native in a single compilation
-unit. Only the C-backend is exercised.
+This directory contains a minimal example for building mlkem-native as a single compilation unit
+using the auto-generated `mlkem_native.c` file.
```
mkannwischer: Clarify that this is only a single parameter set.

examples/monolithic_build/README.md
```diff
@@ -1,20 +1,46 @@
...
-The auto-generated source file [mlkem_native.c](mlkem_native/mlkem_native.c) includes all mlkem-native C source
-files. Moreover, it clears all `#define`s clauses set by mlkem-native at the end, and is hence amenable to multiple
-inclusion in another compilation unit. It exposes the API [mlkem_native.h](mlkem_native/mlkem_native.h).
+## Use Case
 
-The configuration file [mlkem_native_config.h](mlkem_native/mlkem_native_config.h) sets
-`MLK_CONFIG_INTERNAL_API_QUALIFIER` to `static`, making all internal functions static for the single-CU build.
+Use this approach when:
+- You want the simplest possible build integration (one `.c` file)
+- You're using only C (no native backends)
```
mkannwischer, line 13: Add `- You need only one ML-KEM security level (512, 768, or 1024)`

examples/bring_your_own_fips202_static/README.md
```diff
@@ -1,28 +1,56 @@
...
-Build this example with `make build`, run with `make run`.
+```bash
+make build   # Build the example
+make run     # Run the example
+```
+
+## Warning
+
+The `randombytes()` implementation in `test_only_rng/` is for TESTING ONLY.
+You MUST provide a cryptographically secure RNG for production use.
```
mkannwischer, line 53: Unify this warning in all examples.

examples/monolithic_build_multilevel_native/README.md
```diff
@@ -1,38 +1,50 @@
...
-The manually written source file [mlkem_native_all.c](mlkem_native_all.c) includes
-[mlkem_native.c](mlkem_native/mlkem_native.c) three times, each time using the fixed config
-[multilevel_config.h](multilevel_config.h), but changing the security level (specified
-by `MLK_CONFIG_PARAMETER_SET`) every time. For each inclusion, it sets `MLK_CONFIG_FILE`
-appropriately first, and then includes the monobuild:
-```C
-/* Three instances of mlkem-native for all security levels */
+Use this approach when:
+- You need all ML-KEM security levels in one application
+- You want optimal performance via native assembly (AArch64/AVX2)
```
mkannwischer: This isn't limited to AArch64 and AVX2

examples/monolithic_build_multilevel_native/README.md
```diff
@@ -1,38 +1,50 @@
...
+- You want maximum code sharing and LTO optimization
```
mkannwischer: How is this related to LTO?

mkannwischer (changes requested):
Thanks @hanno-becker. Seems to work like a charm.
I have a couple of comments/questions.

mlkem/mlkem_native_config.h
hanno-becker: The latter, yes

mlkem/mlkem_native.h
hanno-becker: Yes, I wanted to add that -- it follows the existing pattern. So long as we keep the `MLK_CONFIG_API_XXX` option for backwards compatibility, it should accurately reflect the relevant parts of the `config.h` configuration. As @davidchisnell noted in #1343 there was nothing corresponding to `MLK_CONFIG_EXTERNAL_API_QUALIFIER`.

examples/bring_your_own_fips202_static/README.md
hanno-becker: Ack

examples/monolithic_build/README.md
hanno-becker: Done

examples/monolithic_build/README.md
hanno-becker: Done

examples/bring_your_own_fips202_static/README.md
hanno-becker: Done

examples/monolithic_build_multilevel_native/README.md
hanno-becker: Fixed

examples/monolithic_build_multilevel_native/README.md
hanno-becker: Sorry, oversight. Fixed.

examples/bring_your_own_fips202_static/README.md
```diff
@@ -1,28 +1,55 @@
...
+- `MLK_CONFIG_FIPS202X4_CUSTOM_HEADER`: Path to stub `fips202x4.h`
 
-1. Arithmetic part of the mlkem-native source tree: [`mlkem/src/`](../../mlkem/src)
-2. A secure pseudo random number generator, implementing [`randombytes.h`](../../mlkem/src/randombytes.h).
-2. A custom FIPS202 with `fips202.h` header compatible with [`mlkem/src/fips202/fips202.h`](../../mlkem/src/fips202/fips202.h).
-   The FIPS202x4 header `fips202x4.h` is unused with `MLK_CONFIG_SERIAL_FIPS202_ONLY` and can be filled with stubs.
-3. The application source code
+Your custom FIPS-202 implementation must provide:
+- `mlk_shake128_absorb_once()`, `mlk_shake128_squeezeblocks()`, `mlk_shake128_release()`
+- `mlk_shake256()`, `mlk_sha3_256()`, `mlk_sha3_512()`
```
mkannwischer, line 33: `mlk_shake256x4` is missing from this list.

hanno-becker: Good spot! Fixed

hanno-becker:
@davidchisnall This is ready for merge now. Do you have time to give it a try?

davidchisnall:
Thanks for this, I can confirm that it works nicely for my use case.  #1351 is not necessary after it.  Doing the same in mldsa would be amazing.

hanno-becker:
@davidchisnall Thank you for confirming, yes, we'll do the same for mldsa-native.

Out of interest, which repository are you targeting for your integration? https://github.com/CHERIoT-Platform/cheriot-rtos?

davidchisnall:
Once it's a bit cleaner, I'll create a new one for PQC libraries.

hanno-becker:
> Doing the same in mldsa would be amazing.

@mkannwischer is doing this here https://github.com/pq-code-package/mldsa-native/pull/782. Should hopefully be ready some point today.

hanno-becker:
@davidchisnall This is merged in mldsa-native as well, now. If you are interested in mldsa-native, I assume that the memory usage is a problem there as well (both the fact that it's still statically stack-allocated, and the mere amount of RAM). Feel free to open an issue so the desire for optimizations in this regard is publicly documented. Focusing on verification first, @mkannwischer has already made significant reductions in https://github.com/pq-code-package/mldsa-native/pull/751; this is still experimental, but knowing that this would be useful for your use case would provide good impetus to continue with it.

PR #1358, opened 2025-12-09
Minor clean-up of mlk_rej_uniform_native contract

mkannwischer:
Fixes single instance of -1 that should be MLK_NATIVE_FUNC_FALLBACK.

### DIFF
--- a/mlkem/src/native/api.h
+++ b/mlkem/src/native/api.h
@@ -413,8 +413,10 @@ __contract__(
   requires(memory_no_alias(r, sizeof(int16_t) * len))
   requires(memory_no_alias(buf, buflen))
   assigns(memory_slice(r, sizeof(int16_t) * len))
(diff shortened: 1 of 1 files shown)


mlkem/src/native/api.h
```diff
@@ -414,7 +414,7 @@ __contract__(
   requires(memory_no_alias(buf, buflen))
   assigns(memory_slice(r, sizeof(int16_t) * len))
   ensures(return_value == MLK_NATIVE_FUNC_FALLBACK || (0 <= return_value && return_value <= len))
```
hanno-becker: While at it, we might want to use `==>` here, too, and perhaps consolidate the two `ensures` clauses.

hanno-becker: I took the liberty to amend the PR

mkannwischer:
Thanks! LGTM.

Issue #1335, opened 2025-12-02
CBMC: Prove that x86_64 backend fulfills native api.h contracts


mkannwischer:
https://github.com/pq-code-package/mlkem-native/pull/1323 added the HOL-Light correctness proof of `polyvec_basemul_acc_montgomery_cached_asm_k2/3/4`.
Similar as for aarch64, the CBMC contract implied by this proof should be added to [arith_native_x86_64.h](https://github.com/pq-code-package/mlkem-native/blob/main/mlkem/src/native/x86_64/src/arith_native_x86_64.h) and then proven to match the contract in [api.h](https://github.com/pq-code-package/mlkem-native/blob/main/mlkem/src/native/api.h)

See the corresponding contract in [arith_native_aarch64.h](
https://github.com/pq-code-package/mlkem-native/blob/72b9197125bc8d0a5dc75da6327b4a75f8be41bd/mlkem/src/native/aarch64/src/arith_native_aarch64.h#L119) and [proof](https://github.com/pq-code-package/mlkem-native/tree/main/proofs/cbmc/polyvec_basemul_acc_montgomery_cached_k2_native_aarch64).

The following functions now have HOL-Light proofs and corresonding contracts and proofs should be added: 
 - polyvec_basemul_acc_montgomery_cached_asm_k2
 - polyvec_basemul_acc_montgomery_cached_asm_k3
 - polyvec_basemul_acc_montgomery_cached_asm_k4
 - ntt_avx2
 - intt_avx2
 - reduce_avx2

mkannwischer:
Once #1374 is merged, we should do the same with the NTT.

PR #1374, opened 2025-12-12
HOL-Light: Add proof for x86_64 forward NTT

mkannwischer:
This commit adds the HOL-Light correctness proof of the x86_64 forward NTT by @dkostic.

- Ported from https://github.com/awslabs/s2n-bignum/pull/314
- Resolves https://github.com/pq-code-package/mlkem-native/issues/1372

It also adds autogeneration of the embedded byte code and the constant twiddle factor array (qdata). 
CBMC proofs that the contract implied by the HOL-Light proofs also fulfills the api.h contracts is not yet added - we will do that as a part of #1335.


scripts/autogen
```diff
@@ -1203,6 +1203,51 @@ def gen_avx2_mulcache_twiddles():
...
+            1284,
+            770,
+            256,
+            3854,
+            3340,
+            2826,
+            2312,
+            1798,
+            1284,
+            770,
```
hanno-becker, line 1223: Deserves a comment, but should be a follow-up to avoid re-triggering proof CI

hanno-becker: Opened #1380

Issue #1373, opened 2025-12-12
Port HOL-Light correctness proof for AVX2 inverse NTT


hanno-becker:
https://github.com/awslabs/s2n-bignum/pull/315

hanno-becker:
@dkostic Wrote those proofs, so if there are any issues, he will be able to help.

PR #1378, opened 2025-12-14
HOL-Light: Add proof for x86_64 inverse NTT

mkannwischer:
This commit adds the HOL-Light correctness proof of the x86_64 inverse NTT by @dkostic.

- Ported from https://github.com/awslabs/s2n-bignum/pull/315
- Resolves https://github.com/pq-code-package/mlkem-native/issues/1373
- Depends on #1374 

It also adds autogeneration of the embedded byte code and the constant twiddle factor array (qdata).
CBMC proofs that the contract implied by the HOL-Light proofs also fulfills the api.h contracts is not yet added - we will do that as a part of https://github.com/pq-code-package/mlkem-native/issues/1335.

### DIFF
--- a/.github/workflows/hol_light.yml
+++ b/.github/workflows/hol_light.yml
@@ -189,6 +189,8 @@ jobs:
           # Dependencies on {name}.{S,ml} are implicit
           - name: mlkem_ntt
             needs: ["mlkem_specs.ml", "mlkem_zetas.ml"]
+          - name: mlkem_intt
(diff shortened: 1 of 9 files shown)


hanno-becker (approved):
LGTM, thanks @dkostic @mkannwischer.

Another thing we should do is remove magic size constants from the scripts, like we did with the AArch64 proofs. While we don't run SLOTHY, it'll still come handy if we ever make small changes to the code.

PR #1379, opened 2025-12-14
Bundle multiple allocations into workspaces

hanno-becker:
#1389 plus consolidation of multiple allocations into workspaces.

### DIFF
--- a/mlkem/src/kem.c
+++ b/mlkem/src/kem.c
@@ -43,29 +43,34 @@ MLK_EXTERNAL_API
 MLK_MUST_CHECK_RETURN_VALUE
 int crypto_kem_check_pk(const uint8_t pk[MLKEM_INDCCA_PUBLICKEYBYTES])
 {
-  int ret = 0;
-  MLK_ALLOC(p, mlk_polyvec, 1);
-  MLK_ALLOC(p_reencoded, uint8_t, MLKEM_POLYVECBYTES);
+  typedef struct
+  {
+    mlk_polyvec p;
+    MLK_ALIGN uint8_t p_reencoded[MLKEM_POLYVECBYTES];
+  } workspace;
 
-  if (p == NULL || p_reencoded == NULL)
+  int ret = 0;
+  MLK_ALLOC(ws, workspace, 1);
+
+  if (ws == NULL)
   {
     ret = MLK_ERR_OUT_OF_MEMORY;
     goto cleanup;
   }
 
-  mlk_polyvec_frombytes(p, pk);
-  mlk_polyvec_reduce(p);
-  mlk_polyvec_tobytes(p_reencoded, p);
+  mlk_polyvec_frombytes(&ws->p, pk);
+  mlk_polyvec_reduce(&ws->p);
+  mlk_polyvec_tobytes(ws->p_reencoded, &ws->p);
 
   /* We use a constant-time memcmp here to avoid having to
    * declassify the PK before the PCT has succeeded. */
-  ret = mlk_ct_memcmp(pk, p_reencoded, MLKEM_POLYVECBYTES) ? MLK_ERR_FAIL : 0;
+  ret =
+      mlk_ct_memcmp(pk, ws->p_reencoded, MLKEM_POLYVECBYTES) ? MLK_ERR_FAIL : 0;
 
 cleanup:
   /* Specification: Partially implements
    * @[FIPS203, Section 3.3, Destruction of intermediate values] */
-  MLK_FREE(p_reencoded, uint8_t, MLKEM_POLYVECBYTES);
-  MLK_FREE(p, mlk_polyvec, 1);
+  MLK_FREE(ws, workspace, 1);
   return ret;
 }
 
@@ -133,24 +138,29 @@ MLK_MUST_CHECK_RETURN_VALUE
 static int mlk_check_pct(uint8_t const pk[MLKEM_INDCCA_PUBLICKEYBYTES],
                          uint8_t const sk[MLKEM_INDCCA_SECRETKEYBYTES])
 {
-  int ret = 0;
-  MLK_ALLOC(ct, uint8_t, MLKEM_INDCCA_CIPHERTEXTBYTES);
-  MLK_ALLOC(ss_enc, uint8_t, MLKEM_SSBYTES);
-  MLK_ALLOC(ss_dec, uint8_t, MLKEM_SSBYTES);
+  typedef struct
+  {
+    MLK_ALIGN uint8_t ct[MLKEM_INDCCA_CIPHERTEXTBYTES];
+    MLK_ALIGN uint8_t ss_enc[MLKEM_SSBYTES];
+    MLK_ALIGN uint8_t ss_dec[MLKEM_SSBYTES];
+  } workspace;
 
-  if (ct == NULL || ss_enc == NULL || ss_dec == NULL)
+  int ret = 0;
+  MLK_ALLOC(ws, workspace, 1);
+
+  if (ws == NULL)
   {
     ret = MLK_ERR_OUT_OF_MEMORY;
     goto cleanup;
   }
 
-  ret = crypto_kem_enc(ct, ss_enc, pk);
+  ret = crypto_kem_enc(ws->ct, ws->ss_enc, pk);
   if (ret != 0)
   {
     goto cleanup;
   }
 
-  ret = crypto_kem_dec(ss_dec, ct, sk);
+  ret = crypto_kem_dec(ws->ss_dec, ws->ct, sk);
   if (ret != 0)
   {
     goto cleanup;
@@ -160,11 +170,12 @@ static int mlk_check_pct(uint8_t const pk[MLKEM_INDCCA_PUBLICKEYBYTES],
   /* Deliberately break PCT for testing purposes */
   if (mlk_break_pct())
   {
-    ss_enc[0] = ~ss_enc[0];
+    ws->ss_enc[0] = ~ws->ss_enc[0];
   }
 #endif /* MLK_CONFIG_KEYGEN_PCT_BREAKAGE_TEST */
 
-  ret = mlk_ct_memcmp(ss_enc, ss_dec, MLKEM_SSBYTES);
+  ret = mlk_ct_memcmp(ws->ss_enc, ws->ss_dec, MLKEM_SSBYTES);
+
   /* The result of the PCT is public. */
   MLK_CT_TESTING_DECLASSIFY(&ret, sizeof(ret));
 
@@ -177,9 +188,7 @@ cleanup:
 
   /* Specification: Partially implements
    * @[FIPS203, Section 3.3, Destruction of intermediate values] */
-  MLK_FREE(ss_dec, uint8_t, MLKEM_SSBYTES);
-  MLK_FREE(ss_enc, uint8_t, MLKEM_SSBYTES);
-  MLK_FREE(ct, uint8_t, MLKEM_INDCCA_CIPHERTEXTBYTES);
+  MLK_FREE(ws, workspace, 1);
   return ret;
 }
 #else  /* MLK_CONFIG_KEYGEN_PCT */
@@ -270,11 +279,16 @@ int crypto_kem_enc_derand(uint8_t ct[MLKEM_INDCCA_CIPHERTEXTBYTES],
                           const uint8_t pk[MLKEM_INDCCA_PUBLICKEYBYTES],
                           const uint8_t coins[MLKEM_SYMBYTES])
 {
-  int ret = 0;
-  MLK_ALLOC(buf, uint8_t, 2 * MLKEM_SYMBYTES);
-  MLK_ALLOC(kr, uint8_t, 2 * MLKEM_SYMBYTES);
+  typedef struct
+  {
+    MLK_ALIGN uint8_t buf[2 * MLKEM_SYMBYTES];
+    MLK_ALIGN uint8_t kr[2 * MLKEM_SYMBYTES];
+  } workspace;
 
-  if (buf == NULL || kr == NULL)
+  int ret = 0;
+  MLK_ALLOC(ws, workspace, 1);
+
+  if (ws == NULL)
   {
     ret = MLK_ERR_OUT_OF_MEMORY;
     goto cleanup;
@@ -287,26 +301,25 @@ int crypto_kem_enc_derand(uint8_t ct[MLKEM_INDCCA_CIPHERTEXTBYTES],
     goto cleanup;
   }
 
-  mlk_memcpy(buf, coins, MLKEM_SYMBYTES);
+  mlk_memcpy(ws->buf, coins, MLKEM_SYMBYTES);
 
   /* Multitarget countermeasure for coins + contributory KEM */
-  mlk_hash_h(buf + MLKEM_SYMBYTES, pk, MLKEM_INDCCA_PUBLICKEYBYTES);
-  mlk_hash_g(kr, buf, 2 * MLKEM_SYMBYTES);
+  mlk_hash_h(ws->buf + MLKEM_SYMBYTES, pk, MLKEM_INDCCA_PUBLICKEYBYTES);
+  mlk_hash_g(ws->kr, ws->buf, 2 * MLKEM_SYMBYTES);
 
   /* coins are in kr+MLKEM_SYMBYTES */
-  ret = mlk_indcpa_enc(ct, buf, pk, kr + MLKEM_SYMBYTES);
+  ret = mlk_indcpa_enc(ct, ws->buf, pk, ws->kr + MLKEM_SYMBYTES);
   if (ret != 0)
   {
     goto cleanup;
   }
 
-  mlk_memcpy(ss, kr, MLKEM_SYMBYTES);
+  mlk_memcpy(ss, ws->kr, MLKEM_SYMBYTES);
 
 cleanup:
   /* Specification: Partially implements
    * @[FIPS203, Section 3.3, Destruction of intermediate values] */
-  MLK_FREE(kr, uint8_t, 2 * MLKEM_SYMBYTES);
-  MLK_FREE(buf, uint8_t, 2 * MLKEM_SYMBYTES);
+  MLK_FREE(ws, workspace, 1);
   return ret;
 }
 
@@ -348,14 +361,19 @@ int crypto_kem_dec(uint8_t ss[MLKEM_SSBYTES],
                    const uint8_t ct[MLKEM_INDCCA_CIPHERTEXTBYTES],
                    const uint8_t sk[MLKEM_INDCCA_SECRETKEYBYTES])
 {
+  typedef struct
+  {
+    MLK_ALIGN uint8_t buf[2 * MLKEM_SYMBYTES];
+    MLK_ALIGN uint8_t kr[2 * MLKEM_SYMBYTES];
+    MLK_ALIGN uint8_t tmp[MLKEM_SYMBYTES + MLKEM_INDCCA_CIPHERTEXTBYTES];
+  } workspace;
+
   int ret = 0;
   uint8_t fail;
   const uint8_t *pk = sk + MLKEM_INDCPA_SECRETKEYBYTES;
-  MLK_ALLOC(buf, uint8_t, 2 * MLKEM_SYMBYTES);
-  MLK_ALLOC(kr, uint8_t, 2 * MLKEM_SYMBYTES);
-  MLK_ALLOC(tmp, uint8_t, MLKEM_SYMBYTES + MLKEM_INDCCA_CIPHERTEXTBYTES);
+  MLK_ALLOC(ws, workspace, 1);
 
-  if (buf == NULL || kr == NULL || tmp == NULL)
+  if (ws == NULL)
   {
     ret = MLK_ERR_OUT_OF_MEMORY;
     goto cleanup;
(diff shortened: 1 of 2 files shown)


hanno-becker:
@davidchisnall Seeing this will be relevant to your use case, you may want to have a look. It's understood that you'll need a context parameter as well, but this can be a follow-up PR.

hanno-becker:
~Still needs work since CBMC can't yet handle the consolidation into workspaces.~ Done in #1388 

Strictly speaking, the introduction of workspace structs could be deferred, since for now we test the custom heap-based allocation with C11 only and it's not part of the main source. Trialed this in #1389

davidchisnall:
I don't like the `goto cleanup` pattern, because it isn't something I can use for other kinds of error handling.  That might be something that would be possible to fix in a later commit but, for example, we'd like running out of stack space to do a `longjmp` and free all heap allocations that were created on the way (we have some SEH-like error handling that handles stack unwinding with some on-stack jump buffers, which is triggered automatically on stack overflow).  We'd also like to be able to use the same mechanism to report that getting entropy failed (currently it's assumed to succeed, which means that the system fails open).

There is also a tradeoff in consolidating the allocations.  Fewer calls to the allocator will be better for performance, but larger allocations are more likely to fail.  I put the performance / stack usage numbers in [this post](https://cheriot.org/pqc/2025/12/12/pqc-on-cheriot.html) for the Verilator simulator of the slowest CHERIoT core.  The signing and verification steps are 4-5M cycles, so a few hundred extra cycles for heap allocation is in the noise, but failing heap allocation is a problem.

hanno-becker:
> I don't like the goto cleanup pattern, because it isn't something I can use for other kinds of error handling. That might be something that would be possible to fix in a later commit but, for example, we'd like running out of stack space to do a longjmp and free all heap allocations that were created on the way (we have some SEH-like error handling that handles stack unwinding with some on-stack jump buffers, which is triggered automatically on stack overflow). We'd also like to be able to use the same mechanism to report that getting entropy failed (currently it's assumed to succeed, which means that the system fails open).

Why would the same mechanism not work for entropy failure? I was hoping that we could similarly set an error code (which is currently an unrefined `-1` but which I hope to split up in v2) and jump to `cleanup` upon RNG failure.

What would your suggestion be?

davidchisnall:
> Why would the same mechanism not work for entropy failure? I was hoping that we could similarly set an error code (which is currently an unrefined -1 but which I hope to split up in v2) and jump to cleanup upon RNG failure.

Yes, that would work.  Currently the get-entropy function has a `void` return and assumes success.  If allocation is in the same function as an entropy call, then it could also `goto` the same cleanup point.

hanno-becker:
> we'd like running out of stack space to do a longjmp and free all heap allocations that were created on the way (we have some SEH-like error handling that handles stack unwinding with some on-stack jump buffers, which is triggered automatically on stack overflow)

Does this need any co-operation from the application/library?

If not, it seems the current `goto cleanup` approach might be OK, if we also use it for other errors such as RNG failure.

hanno-becker:
> There is also a tradeoff in consolidating the allocations. Fewer calls to the allocator will be better for performance, but larger allocations are more likely to fail. I put the performance / stack usage numbers in [this post](https://cheriot.org/pqc/2025/12/12/pqc-on-cheriot.html) for the Verilator simulator of the slowest CHERIoT core. The signing and verification steps are 4-5M cycles, so a few hundred extra cycles for heap allocation is in the noise, but failing heap allocation is a problem.

This is an interesting point. I agree that the # of allocations is effectively irrelevant for performance here.

davidchisnall:
> Does this need any co-operation from the application/library?

That's a bit complicated.  If you take a synchronous fault, the switcher invokes some code in your compartment.  The default implementation finds the head of a linked list at the top (address, bottom logically) of your stack and pops a jump buffer off that.  This allows us to gracefully recover from memory-safety bugs (including stack overflow) and other kinds of errors.

> If not, it seems the current goto cleanup approach might be OK, if we also use it for other errors such as RNG failure.

It depends on whether it's local or non-local.  Stack overflow is typically not going to be in the function that contains the target of the `goto` (and, if it is, we can't rely on it because the current stack frame is in an undefined state when it happens).  RNG failure may be, or may be somewhere else.  In the case of any of these failures, the preferred behaviour is to bail to a top-level function, release any memory, and return failure.

davidchisnall:
> This is an interesting point. I agree that the # of allocations is effectively irrelevant for performance here.

Note that I said the opposite before (sorry!) but that was before I did any benchmarking.

mlkem/src/indcpa.h
mkannwischer, line 67: Document the return value here?

mlkem/src/kem.c
mkannwischer, line 1: Documentation on the return values in kem.h and mlkem_native.h needs to be adjusted.

mlkem/mlkem_native_config.h
```diff
@@ -443,6 +443,65 @@
...
+       if ((v) == NULL)                                           \
+       {                                                          \
+         ret = -1;                                                \
+         goto cleanup;                                            \
+       }                                                          \
+     } while(0)
+   #define MLK_CUSTOM_FREE(v, T, N)                               \
+     if ((v) != NULL)                                             \
+     {                                                            \
+       mlk_zeroize((v), sizeof(T) * (N));                         \
```
mkannwischer: zeroization was already performed in MLK_FREE - so like this you would be zeroizing it twice for no reason. This should be documented above as well.

mlkem/mlkem_native_config.h
```diff
@@ -443,6 +443,65 @@
...
+ *
+ *              Parameters to MLK_CUSTOM_ALLOC:
+ *              - T* v: Target pointer to modify. Can be assumed to be NULL.
+ *              - T: Type of structure to be allocated
+ *              - N: Number of elements to be allocated.
+ *
+ *              Parameters to MLK_CUSTOM_FREE:
+ *              - T* v: Target pointer to free. May be NULL.
+ *              - T: Type of structure to be freed.
+ *              - N: Number of elements to be freed.
```
mkannwischer, line 465: Out of interest: When would you ever need the T and N?

mlkem/mlkem_native_config.h
```diff
@@ -443,6 +443,65 @@
...
+ *              If this option is set, most of them will be allocated via
+ *              MLK_CUSTOM_ALLOC.
+ *
+ *              Parameters to MLK_CUSTOM_ALLOC:
+ *              - T* v: Target pointer to modify. Can be assumed to be NULL.
+ *              - T: Type of structure to be allocated
+ *              - N: Number of elements to be allocated.
+ *
+ *              Parameters to MLK_CUSTOM_FREE:
+ *              - T* v: Target pointer to free. May be NULL.
```
mkannwischer, line 463: Why can this be NULL?

test/configs.yml
```diff
@@ -388,6 +388,27 @@ configs:
...
+      MLK_CONFIG_CUSTOM_ALLOC_FREE:
+        content: |
+          #define MLK_CONFIG_CUSTOM_ALLOC_FREE
+          #if !defined(__ASSEMBLER__)
+          #include <stdlib.h>
+          #define MLK_CUSTOM_ALLOC(v, T, N)                              \
+            do                                                           \
+            {                                                            \
+              (v) = (T *)aligned_alloc(MLK_DEFAULT_ALIGN,                \
+                                       MLK_ALIGN_UP(sizeof(T) * (N)));   \
```
mkannwischer: I'm a little surprised here that the config has access to `MLK_DEFAULT_ALIGN` and `MLK_ALIGN_UP` without including common.h

mlkem/mlkem_native_config.h
hanno-becker: That's still a remnant for when it wasn't part of `MLK_FREE`, good spot.

mlkem/mlkem_native_config.h
hanno-becker: For `uint8_t` buffers for example.

mlkem/mlkem_native_config.h
hanno-becker: Because we may have multiple allocations and if the first one fails, the FREE for the second one will find the pointer NULL.

test/configs.yml
hanno-becker: Very good spot. Yes, this is deliberate, if a bit unsatisfactory, and a consequence of #1387. It's no issue here because every file includes `config.h` _through_ `common.h`.

mlkem/src/kem.c
hanno-becker: Done

mlkem/src/indcpa.h
hanno-becker: Done

mlkem/mlkem_native_config.h
mkannwischer: I can't follow. This question is about free, not alloc.

mlkem/mlkem_native_config.h
mkannwischer: I thought we are limited to exactly one allocation?

mlkem/mlkem_native_config.h
hanno-becker: Ah. I can't name an example off-hand, but it seems cleaner to provide the same information as for alloc. I initially had it to be able to include the zeroization.

hanno-becker:
I changed the approach to use an explicit error check in the code rather than expecting `MLK_CUSTOM_ALLOC` to do this. This makes it easier to switch between stack and heap allocation without running into C90 errors because of mixed code and declarations. It also gives us control over the error code, while the user-provided `MLK_CUSTOM_ALLOC` can follow the standard convention that failure is reported through a `NULL` return value.

Also, this is now based on #1389 which doesn't introduce workspaces.

mlkem/src/kem.c
```diff
@@ -339,14 +351,19 @@ int crypto_kem_dec(uint8_t ss[MLKEM_SSBYTES],
                    const uint8_t ct[MLKEM_INDCCA_CIPHERTEXTBYTES],
                    const uint8_t sk[MLKEM_INDCCA_SECRETKEYBYTES])
 {
+  typedef struct
+  {
+    MLK_ALIGN uint8_t buf[2 * MLKEM_SYMBYTES];
+    MLK_ALIGN uint8_t kr[2 * MLKEM_SYMBYTES];
+    uint8_t tmp[MLKEM_SYMBYTES + MLKEM_INDCCA_CIPHERTEXTBYTES];
```
mkannwischer: Is this buffer intentionally unaligned? Probably we want to align this too.

mkannwischer (changes requested):
Thanks @hanno-becker, one small nit on the code, otherwise this looks good to me.

Two points remain to be discussed:
1) The previous approach would allow a user to allocate different buffers in different memories, e.g., one could allocate the large matrix A on the heap while all other buffers stay on the stack. This is no longer possible with the workspace approach. This may be particularly relevant for embedded systems.
2) As @davidchisnall noted [here](https://github.com/pq-code-package/mlkem-native/pull/1379#issuecomment-3655396072) allocating everything at once is more likely to fail which is undesirable.

Any thoughts @davidchisnall, @gilles-peskine-arm, @hanno-becker?

mlkem/src/kem.c
hanno-becker: Done

hanno-becker:
Unless @davidchisnall @gilles-peskine-arm have concrete use cases in mind where we know that the monolithic allocation will cause problems (?), I would suggest to add workspaces for now and adjust things on demand. This is internal and can be changed at any time. Esp. for mldsa-native I am hoping that this will also help us to reduce CBMC overhead from dynamic allocation which is currently blocking proofs.

gilles-peskine-arm:
I very much prefer the monolithic allocation approach, unless there is an identified use case for fragmented allocation or monolithic allocation is too hard to implement.

Let's consider the three typical types of memory that might be used.

* Stack: if everything is on the stack, then monolithic or fragmented doesn't matter if the monolithic workspace is implemented optimally, meaning that it doesn't use more space than the stack would. E.g. a monolithic workspace can waste memory if it ends up being `struct {a; b; c;}` and you know that only two of `a`, `b` and `c` can ever be active at the same time. I would guess that this doesn't happen with crypto that is constant-time, because that tends to happen when you have code paths that depend on the data. But I'm not familiar enough with MLKEM and MLDSA to know if it's a problem.
* Heap: a monolithic allocation saves a bit of RAM overhead, but can be problematic if your heap is already fragmented. I think that if fragmenting the allocations is the only way your application can run, you're already too tight on memory and even a small change in the order of operations, or just letting your application run a little longer, would cause heap exhaustion even with smaller fragments.
* Global: if I have to reserve some global area for a costly operation, I'll just designate one area (or maybe one area per CPU core). I can't see any advantage for fragmentation.

There may be a reason to use different types of memory for some data. If a system has a small amount of memory that's better protected against snooping, that memory should be prioritized for key-equivalent secret data over message-specific secret data during signing. But I think if you want that level of control, you should be prepared to maintain your own platform-specific cryptography code, or at least your own patches. It's not a level of control that we intend to provide in TF-PSA-Crypto (whereas we would like to give a choice between stack and a user-provided buffer).

In addition, I would prefer to avoid having low-level functions that can fail: ideally some high-level function would reserve the necessary amount of memory (and generate the random data), and then all low-level functions can't fail. One benefit of that is that it reduces the risk of doing error handling wrong, although having a CBMC proof mostly mitigates that risk. Another benefit is the code size: the cost of all `if (failed) goto cleanup` adds up.

mlkem/src/kem.c
```diff
@@ -348,14 +361,19 @@ int crypto_kem_dec(uint8_t ss[MLKEM_SSBYTES],
                    const uint8_t ct[MLKEM_INDCCA_CIPHERTEXTBYTES],
                    const uint8_t sk[MLKEM_INDCCA_SECRETKEYBYTES])
 {
+  typedef struct
+  {
+    MLK_ALIGN uint8_t buf[2 * MLKEM_SYMBYTES];
+    MLK_ALIGN uint8_t kr[2 * MLKEM_SYMBYTES];
+    MLK_ALIGN uint8_t tmp[MLKEM_SYMBYTES + MLKEM_INDCCA_CIPHERTEXTBYTES];
+  } workspace;
```
gilles-peskine-arm, line 369: Could that struct be made public? In the short term, I think TF-PSA-Crypto will just put the workspace on the heap, or maybe give a choice between heap and stack. But in the longer term, we'd like the option of making it a global buffer, reserved at static link time.

hanno-becker: Could that not be achieved with a custom `MLK_ALLOC` which uses a `static` declaration?

gilles-peskine-arm: Good point, that does work for a global buffer that's reserved for MLKEM, and I think it also allows protecting that buffer with a mutex (alloc = lock and free = unlock).

But I don't think it allows what I was actually thinking of, which is a buffer that can be used for different things, such as using the same buffer for MLKEM and MLDSA (but only one at a time).

gilles-peskine-arm: (But to be clear, I'd prefer having the option to use a heap buffer quickly, and think about fancier stuff in version 2, than to spend a lot of time now to try to cover more use cases.)

mkannwischer:
> I very much prefer the monolithic allocation approach, unless there is an identified use case for fragmented allocation or monolithic allocation is too hard to implement.
> 
> Let's consider the three typical types of memory that might be used.
> 
> * Stack: if everything is on the stack, then monolithic or fragmented doesn't matter if the monolithic workspace is implemented optimally, meaning that it doesn't use more space than the stack would. E.g. a monolithic workspace can waste memory if it ends up being `struct {a; b; c;}` and you know that only two of `a`, `b` and `c` can ever be active at the same time. I would guess that this doesn't happen with crypto that is constant-time, because that tends to happen when you have code paths that depend on the data. But I'm not familiar enough with MLKEM and MLDSA to know if it's a problem.
> * Heap: a monolithic allocation saves a bit of RAM overhead, but can be problematic if your heap is already fragmented. I think that if fragmenting the allocations is the only way your application can run, you're already too tight on memory and even a small change in the order of operations, or just letting your application run a little longer, would cause heap exhaustion even with smaller fragments.
> * Global: if I have to reserve some global area for a costly operation, I'll just designate one area (or maybe one area per CPU core). I can't see any advantage for fragmentation.
> 
> There may be a reason to use different types of memory for some data. If a system has a small amount of memory that's better protected against snooping, that memory should be prioritized for key-equivalent secret data over message-specific secret data during signing. But I think if you want that level of control, you should be prepared to maintain your own platform-specific cryptography code, or at least your own patches. It's not a level of control that we intend to provide in TF-PSA-Crypto (whereas we would like to give a choice between stack and a user-provided buffer).
> 
> In addition, I would prefer to avoid having low-level functions that can fail: ideally some high-level function would reserve the necessary amount of memory (and generate the random data), and then all low-level functions can't fail. One benefit of that is that it reduces the risk of doing error handling wrong, although having a CBMC proof mostly mitigates that risk. Another benefit is the code size: the cost of all `if (failed) goto cleanup` adds up.

Thanks @gilles-peskine-arm for this extensive comment. Makes sense to me.
In that case we go ahead an merge the workspaces in mlkem-native now and also try to get that to work in mldsa-native.

mlkem/src/kem.c
hanno-becker: > But I don't think it allows what I was actually thinking of, which is a buffer that can be used for different things, such as using the same buffer for MLKEM and MLDSA (but only one at a time).

Got it. It feels wrong through that you would need to know the (internal) workspace structure for that. Would a global bump allocator not be enough for that purpose? This is what's already being tested [without workspaces] in [`test/test_alloc.c`](https://github.com/pq-code-package/mlkem-native/blob/main/test/src/test_alloc.c).

gilles-peskine-arm: If I reserve a separate memory area, I need to know how large it is. Obviously I can experiment, but it's easier if it's some constant in a header. Especially when the value gets reported through several layers (mldsa-native + mlkem-native → TF-PSA-Crypto → ACME Semiconductors BSP → configurable purpose-specific firmware → device-specific build).

This doesn't require exposing the structure, just its size. I'd guess the size is unlikely to change so a `#define WORKSPACE_SIZE` and put `static_assert(sizeof(struct) <= WORKSPACE_SIZE)` in `MLKEM_ALLOC` would work out.

mkannwischer: Yes, we can certainly do that. I'll open an issue for that.

hanno-becker: Note that, at present, there are still multiple layers of functions where workspaces are used -- so one needs to accumulate those sizes. It would be more work -- but perhaps the right thing to do -- to switch to a single workspace per top-level API.

mkannwischer: Something like this https://github.com/pq-code-package/mlkem-native/issues/1442? 

I think independenly we need to move all allocs to the top level, i.e., make sure that there is only a single invocation of `MLK_ALLOC`, right?

gilles-peskine-arm: Having multiple workspaces isn't a problem if I just want one global workspace, as long as the function call structure is known. In the simplest case, if all the function calls are nested (and there's no recursion, obviously), I can just add their sizes for each of the entry points that I use, at the integration level just above mlkem-native. For that aspect, I only want help from mlkem-native if there are parts that never overlap, e.g. if `entry_point()` calls `foo()` and `bar()` successively and each three have their own workspace then I'd like mlkem-native to help me figure out that the workspace size for `entry_point()` is `sizeof(entry_point_struct) + max(sizeof(foo_struct), sizeof(bar_struct))`. and not the sum of all three.

hanno-becker: Having a single top-level allocation would make it easy to define an upper bound for the required memory. But we can also simply measure the total memory usage in the bump allocation test (we already do that, actually)  and set the global constants accordingly and then size the bump allocator according to those constants, so we notice automatically when they get out of date. That may, for now, be a more pragmatic solution.

hanno-becker: @gilles-peskine-arm Do you have concerns with a 'simple-minded' approach of simply measuring the bump allocator's memory usage and providing defined constants for that?

mlkem/mlkem_native_config.h
hanno-becker: At the moment, we aren't; with this PR, we could in principle re-introduce this limitation, but I'd still prefer alloc/free to be general.

mlkem/src/kem.c
gilles-peskine-arm: Seems fine to me.

davidchisnall:
> Heap: a monolithic allocation saves a bit of RAM overhead, but can be problematic if your heap is already fragmented. I think that if fragmenting the allocations is the only way your application can run, you're already too tight on memory and even a small change in the order of operations, or just letting your application run a little longer, would cause heap exhaustion even with smaller fragments.

This really isn't the case for embedded targets.  Heap sizes of well under 1 MiB are almost universal and under 128 KiB is common.  There's a *massive* difference in the likelihood of being able to satisfy a single 60 KiB allocation and two 30 KiB ones in a long-running application.  A single 60 KiB allocation is likely to be close to half of the total heap size, which means that you can't allocate it even with a small amount of existing fragmentation.

hanno-becker:
Thank you @davidchisnall for chiming back in (and happy new year)!

I am open to reconsidering this. On mldsa-native, the hope was that fewer allocations would reduce CBMC proof complexity, but so far, this hope hasn't materialized (and we have not yet introduced workspaces, there). It might be more effective to have a wrapper doing the allocation and passing a myriad of pointers to a function doing the actual work. That's harder to read, but it should contain the proof complexity (the visual clutter added to the CBMC specs might be contained with something like https://github.com/pq-code-package/mldsa-native/pull/817).

Also, @gilles-peskine-arm's request of having a public constant indicating the memory usage can be achieved without workspaces (cf https://github.com/pq-code-package/mlkem-native/pull/1379#discussion_r2656165552).

hanno-becker:
@mkannwischer @gilles-peskine-arm Please chime in.

gilles-peskine-arm:
> This really isn't the case for embedded targets. Heap sizes of well under 1 MiB are almost universal and under 128 KiB is common. There's a massive difference in the likelihood of being able to satisfy a single 60 KiB allocation and two 30 KiB ones in a long-running application. A single 60 KiB allocation is likely to be close to half of the total heap size, which means that you can't allocate it even with a small amount of existing fragmentation.

As a former embedded OS architect, the systems I worked on reserved a static (or startup-time) amount of memory for critical systems such as cryptography, so that we knew in advance that crypto wouldn't fail midway due to lack of resources. distinct from heap which was used in less critical components where failing midway was acceptable.

Now, as a crypto library maintainer (TF-PSA-Crypto), I have known users who want to reserve a static area for ML-DSA verification, because the system's stack limit is too small, and the verifier has no heap. I also have known users who have a heap, including resource-constrained platforms where a few kilobytes can make a difference, but many of these users use the heap implementation that comes with our crypto library because they don't use a heap for anything else, and they would prefer if our crypto library didn't require a heap at all. (Not an option for our RSA code or most of our ECC code, but a long-term goal for us.) I also have known users who have a large heap and are concerned about how many thousands of concurrent connections they can establish, but this kind of high-end application is less of a priority for us.

So, in my experience, the case of a heap that's just big enough if you fragment the allocation is not very important, it's not something that TF-PSA-Crypto intends to cater for. However, it's a vast world and if there are environments where this is desirable, I certainly won't argue that mldsa-native shoult not provide this capability. That's not my call to make.

Issue #1377, opened 2025-12-14
Port HOL-Light correctness proof for AVX2 poly_reduce


mkannwischer:
https://github.com/awslabs/s2n-bignum/pull/297

PR #1381, opened 2025-12-14
HOL-Light: Add proof for x86_64 poly_reduce

mkannwischer:
This commit adds the HOL-Light correctness proof of
the x86_64 polynomial reduction by @dkostic.

- Ports https://github.com/awslabs/s2n-bignum/pull/297
- Resolves https://github.com/pq-code-package/mlkem-native/issues/1377

### DIFF
--- a/.github/workflows/hol_light.yml
+++ b/.github/workflows/hol_light.yml
@@ -197,6 +197,8 @@ jobs:
             needs: ["mlkem_specs.ml"]
           - name: mlkem_poly_basemul_acc_montgomery_cached_k4
             needs: ["mlkem_specs.ml"]
+          - name: mlkem_reduce
+            needs: ["mlkem_specs.ml"]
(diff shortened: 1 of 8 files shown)


mkannwischer:
The toolchain in mlkem-native produces slightly diffent bytecode than the one that is embedded in the s2n-bignum proof.
It appears that `vmovd %eax, %xmm2` is encoded differently.
See below the diff (red is s2n-bignum - green is mlkem-native). 
@dkostic, @hanno-becker, any idea what is going on there?

image

hanno-becker:
@mkannwischer We have observed in the past that assemblers don't exhibit uniform behavior, but I don't recall seeing this example. What we observed was the Mac assembler reordering `ADD` arguments in favor of shorter encodings -- that's why the NTT code in mldsa-native forces canonical ordering. @dkostic Have you noticed @mkannwischer's concrete issue?

Issue #1376, opened 2025-12-13
Introduce configurable alloc/free macros for local structures


hanno-becker:
We currently allocate all local structures on the stack, which is unacceptable for some environments. Even if we successfully reduce RAM usage, we are unlikely to get it down to a level that would be universally acceptable.

Task:
- [ ] Introduce macros `MLK_ALLOC(v, T, N)` and `MLK_FREE(v, T, N)` which assume a pointer variable `T* v` has already been declared, allocates a `T[N]`, and points v to it. This allocation can come from the heap or from the stack.
- [ ] For every allocation of a large buffer or large structure (esp. `mlk_poly`, `mlk_polyvec`, `mlk_polymat`), allocate switch to the pattern `T0* ptr0 = NULL; T1* ptr1 = NULL; ... MLK_ALLOC(ptr0, T0, N0); MLK_ALLOC(ptr1, T1, N1); ...` and `MLK_FREE(ptr0, T0, N0); MLK_FREE(ptr1, T1, N1); ...` at the end.
- [ ] For individual structs such as `mlk_poly`, pass N=1.
- [ ] For structured cleanup, every function involving `MLK_ALLOC` and `MLK_FREE` needs a `cleanup: ...` containing all `MLK_FREE(...)` invocations. This is needed for heap-based implementations of `MLK_ALLOC` which can fail.
- [ ] In the default stack implementation of `MLK_ALLOC(v, T, N)`, declare a helper `T[N]` on the stack and point `v` to it. For `MLK_FREE`, call mlk_zeroize.
- [ ] Add configuration options `MLK_CONFIG_CUSTOM_ALLOC` and `MLK_CONFIG_CUSTOM_FREE` which can be used to overwrite `MLK_ALLOC` and `MLK_FREE`.
- [ ] Add a custom config with a heap-based implementation of `MLK_ALLOC(v, T, N)`, setting `v` by calling malloc.

PR #1389, opened 2025-12-15
Make allocation of large structures/buffers configurable (without workspaces)

hanno-becker:
* Resolves https://github.com/pq-code-package/mlkem-native/issues/1376

This commit introduces the configuration option

  MLK_CONFIG_CUSTOM_ALLOC_FREE

which allows users to provide custom macros

  MLK_CUSTOM_ALLOC(v, T, N)
  MLK_CUSTOM_FREE(v, T, N)

that should be used in place of the default stack-allocation
to allocate/free large internal structures. Those macros are
then wrapped into MLK_ALLOC/MLK_FREE and used in the source.
Importantly, MLK_FREE adds zeroization ahead of calling
MLK_CUSTOM_FREE, so the latter need not take care of this.

The macros are put to use in kem.c and indcpa.c, but not yet
further down the call-stack.

The option is documented as experimental and (hence) unstable
so we have freedom to adjust this ahead of v2.

A custom configuration is added which implements the macros
based on C11's `aligned_alloc`. We enable the leak sanitizer
in the tests for this configuration to catch any missing calls
to MLK_FREE.

Since allocation can fail, additional error paths are introduced
to the functions using MLK_ALLOC. This uniformly follows the pattern
of all pointers being initialized to NULL before allocation, and a
cleanup section calling MLK_FREE on them before returning the result.

To distinguish between out-of-memory failure and other kinds of
functional failures, we introduce named error codes MLK_ERR_FAIL
and MLK_ERR_OUT_OF_MEMORY. This will likely need further refinement,
but this can only be done in v2.

Signed-off-by: Hanno Becker <beckphan@amazon.co.uk>


mlkem/src/indcpa.h
```diff
@@ -68,9 +68,9 @@ __contract__(
  *
  **************************************************/
 MLK_INTERNAL_API
-void mlk_indcpa_keypair_derand(uint8_t pk[MLKEM_INDCPA_PUBLICKEYBYTES],
-                               uint8_t sk[MLKEM_INDCPA_SECRETKEYBYTES],
-                               const uint8_t coins[MLKEM_SYMBYTES])
+int mlk_indcpa_keypair_derand(uint8_t pk[MLKEM_INDCPA_PUBLICKEYBYTES],
```
mkannwischer, line 72: CAn you add a `MLK_MUST_CHECK_RETURN_VALUE` to all functions that you have added an error code to?

mlkem/src/kem.c
```diff
@@ -223,14 +259,21 @@ int crypto_kem_enc_derand(uint8_t ct[MLKEM_INDCCA_CIPHERTEXTBYTES],
...
+    ret = MLK_ERR_OUT_OF_MEMORY;
+    goto cleanup;
+  }
 
   /* Specification: Implements @[FIPS203, Section 7.2, Modulus check] */
   if (crypto_kem_check_pk(pk))
   {
-    return -1;
+    ret = MLK_ERR_FAIL;
+    goto cleanup;
```
mkannwischer: This is wrong. check_pk could be failing due to not being able to allocate. That error should be forwarded.

mlkem/src/kem.c
```diff
@@ -283,21 +337,31 @@ int crypto_kem_dec(uint8_t ss[MLKEM_SSBYTES],
...
+    ret = MLK_ERR_OUT_OF_MEMORY;
+    goto cleanup;
+  }
 
   /* Specification: Implements @[FIPS203, Section 7.3, Hash check] */
   if (crypto_kem_check_sk(sk))
   {
-    return -1;
+    ret = MLK_ERR_FAIL;
+    goto cleanup;
```
mkannwischer: same here.

mlkem/src/indcpa.h
hanno-becker: Good idea. Done

mlkem/src/kem.c
hanno-becker: Oh man ... thank you for spotting that.

mlkem/src/indcpa.h
hanno-becker: We should apply this to all internal functions with return values; but no in this PR. #1399

mlkem/src/kem.c
hanno-becker: Done

mlkem/src/kem.c
hanno-becker: Done

PR #1390, opened 2025-12-15
CBMC: Remove uses of object_whole and same_object

### LINKED ISSUE #1388: CBMC: Replace object_whole with memory_slice in non-top-level contracts
* Analogous to https://github.com/pq-code-package/mldsa-native/issues/716 for mldsa-native, which was fixed in https://github.com/pq-code-package/mldsa-native/pull/730.

#1376 proposes top-level objects being bundled into a struct or union to allow for simple customization of the means of allocations.

This will break top-level CBMC proofs because many functions called by them use `object_whole` to specify the memory effect, but with objects bundled into a conglomerate object, that entire parent object would be tainted (it seems that this is a deficiency in the contract mechanism in CBMC, but this is another question).

Task:
- [ ] Identify all uses of object_whole applied to function arguments and replace them by appropriate uses of `memory_slice`.
- [ ] Check that the CBMC proofs still go through and adjust as needed.

hanno-becker:
* Fixes #1388 

In preparation for the introduction of workspace structures
at the higher-level APIs, this commit removes all uses of
the CBMC footprint predicate `object_whole` and instead uses
the more fine-grained `memory_slice`.

For `mlk_poly_getnoise_eta1122_4x`, we also remove the sole
use of `same_object` in the code base, and remove its definition
from cbmc.h.

### DIFF
--- a/dev/aarch64_clean/src/arith_native_aarch64.h
+++ b/dev/aarch64_clean/src/arith_native_aarch64.h
@@ -94,7 +94,7 @@ __contract__(
   requires(memory_no_alias(mlk_poly, sizeof(int16_t) * MLKEM_N))
   requires(zetas == mlk_aarch64_zetas_mulcache_native)
   requires(zetas_twisted == mlk_aarch64_zetas_mulcache_twisted_native)
-  assigns(object_whole(cache))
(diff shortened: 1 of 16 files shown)


hanno-becker:
@mkannwischer If you're happy with it, let's merge it directly as part of #1379

PR #1391, opened 2025-12-15
Serial FIPS202: Do not use x4 CBD sampling 

mkannwischer:
Before, MLK_CONFIG_SERIAL_FIPS202_ONLY would not affect the CBD sampling, but
only the parallel sampling of the matrix A into serial sampling.
This has two undesirable effecs:
1) Consumers settings MLK_CONFIG_SERIAL_FIPS202_ONLY will still have to
provide a fips202x4.h implementing shake256x4 which is confusing and
inconvienent.
2) Both in key generation and encryption, we would use 4-way batched CBD
sampling which in some places generates irrelevant polynomials (as 4-way
batched is commonly faster than 3x serial on large CPUs).

This commit fixes that. When MLK_CONFIG_SERIAL_FIPS202_ONLY is set, no
x4 functions for CBD sampling are used and, hence, fips202x4.h does no
longer have to be implemented.
This results in a speed-up for key generation, encapsulation, and
decapsulation.
To implement this, we require a poly_getnoise_eta1 (which was previously
removed).

A CBMC proof of poly_getnoise_eta1 and the proofs of the serial variants
of the indcpa_enc and indcpa_keypair_derand are added.

- Resolves https://github.com/pq-code-package/mlkem-native/issues/1274


mkannwischer:
We will merge #1392 instead.

PR #1392, opened 2025-12-15
Serial FIPS202: Do not use x4 Keccak for CBD sampling

### LINKED ISSUE #1274: Do not require fips202x4.h when using `MLK_CONFIG_SERIAL_FIPS202_ONLY`

mkannwischer:
Before, MLK_CONFIG_SERIAL_FIPS202_ONLY would not affect the CBD sampling, but
only the parallel sampling of the matrix A into serial sampling.
This has the undesirable effect that consumers settings
MLK_CONFIG_SERIAL_FIPS202_ONLY will still have to
provide a fips202x4.h implementing shake256x4 which is confusing and
inconvienent.

This commit fixes that. When MLK_CONFIG_SERIAL_FIPS202_ONLY is set,
poly_getnoise_eta1_4x and poly_getnoise_eta1122_4x do not use a 4-way
batched PRF (Keccak).
The fips202x4.h is no longer needed in that case.

CBMC proofs for the serial variants are added.
- Resolves #1274 
- Alternative to #1391


mlkem/src/poly_k.c
```diff
@@ -440,14 +455,16 @@ void mlk_poly_getnoise_eta1122_4x(mlk_poly *r0, mlk_poly *r1, mlk_poly *r2,
   /* On systems with fast batched Keccak, we use 4-fold batched PRF,
    * even though that means generating more random data in buf[2] and buf[3]
    * than necessary. */
-#if !defined(FIPS202_X4_DEFAULT_IMPLEMENTATION)
+#if !defined(FIPS202_X4_DEFAULT_IMPLEMENTATION) && \
```
hanno-becker, line 460: Do you want to remove `FIPS202_X4_DEFAULT_IMPLEMENTATION` at this occasion?

mkannwischer: That would come at a performance penalty for the C implementation. I'd prefer to evaluate the separately.

proofs/cbmc/poly_getnoise_eta1122_4x_serial/Makefile
```diff
@@ -0,0 +1,61 @@
...
+include ../Makefile_params.common
+
+HARNESS_ENTRY = harness
+HARNESS_FILE = poly_getnoise_eta1122_4x_harness
+
+# This should be a unique identifier for this proof, and will appear on the
+# Litani dashboard. It can be human-readable and contain spaces if you wish.
+PROOF_UID = mlk_poly_getnoise_eta1122_4x_serial
+
+DEFINES +=
```
hanno-becker: Is this missing `-DMLK_CONFIG_SERIAL_FIPS202_ONLY`?

I took the liberty to add that.

hanno-becker: On second thought, I think the new proof is redundant because the default proof covers the serial case, and we already have `poly_getnoise_eta1122_4x_native` which covers the batched case.

mlkem/src/poly_k.c
```diff
@@ -333,16 +333,31 @@ void mlk_poly_getnoise_eta1_4x(mlk_poly *r0, mlk_poly *r1, mlk_poly *r2,
   extkey[1][MLKEM_SYMBYTES] = nonce1;
   extkey[2][MLKEM_SYMBYTES] = nonce2;
   extkey[3][MLKEM_SYMBYTES] = nonce3;
+
+#if !defined(MLK_CONFIG_SERIAL_FIPS202_ONLY)
```
hanno-becker: I think this should follow the same pattern as `poly_getnoise_eta1122_4x` below. `MLK_CONFIG_SERIAL_FIPS202_ONLY` is effectively redundant here.

hanno-becker: I added a separate commit. Please check if you agree @mkannwischer The default CBMC proof turns into the one covering the serial case, the the `_native` one covers the batched case.

mkannwischer: Yes, good idea. LGTM.

hanno-becker (approved):
LGTM

PR #1393, opened 2025-12-15
CBMC: Add and prove x86_64 backend contracts

willieyz:
- Resolves: #1335 
- Rebased on #1432, tomont isn't actually use mlk_qdata.
- This PR also added proof corresponding these two hol-light: https://github.com/pq-code-package/mlkem-native/pull/1418
- This commit introduces CBMC verification for several x86-64 assembly functions. The contracts for these functions are derived from their corresponding HOL-Light correctness proofs.

- The following functions now have CBMC proofs based on their
  HOL-Light specifications:

  - polyvec_basemul_acc_montgomery_cached_asm_k2/k3/k4:
    - Pre-conditions:
      `<= &2 pow 12`(HOL_LIGHT) v.s. `MLKEM_UINT12_LIMIT + 1`(CBMC)
    - Adds a CBMC pre-condition:
      `requires(array_abs_bound(a, 0, 4 * MLKEM_N, MLKEM_UINT12_LIMIT +`
                                                                   `1))`
  - ntt_avx2:
    - Pre-condition:
      `<= &8191`(HOL_LIGHT) v.s. `8191 + 1`(CBMC)
    - Post-condition:
      `<= &23594`(HOL_LIGHT) v.s. `23594 + 1`(CBMC)
    - Adds a CBMC pre-condition:
      `requires(array_abs_bound(r, 0, MLKEM_N, 8192))`
    - Adds a CBMC post-condition:
      `ensures(array_abs_bound(r, 0, MLKEM_N, 23595))`

  - intt_avx2:
    - Post-condition:
      `<= &26631` v.s. `26631 + 1`(CBMC)
    - Adds a CBMC post-condition:
      `ensures(array_abs_bound(r, 0, MLKEM_N, 26632))`

   - reduce_avx2:
     - Post-condition:
       `< &3329`(HOL_LIGHT) v.s. `3329 == MLKEM_Q`(CBMC)
     - Adds a CBMC post-condition:
       `ensures(array_bound(r, 0, MLKEM_N, 0, MLKEM_Q))`

   - tobytes_avx2:
     - Pre-condition:
       `(!i. i < LENGTH l ==> val(EL i l) < 3329)`(HOL_LIGHT)
       v.s.
       `MLKEM_UINT12_LIMIT`(CBMC)
     - Adds a CBMC pre-condition:
       `requires(array_bound(a, 0, MLKEM_N, 0, MLKEM_Q))`

   - frombytes_avx2:
     - Post-conditions:
[...]

PR #1394, opened 2025-12-16
HOL-Light: x86 INTT output bound <=26632 -> <=26631

mkannwischer:
During https://github.com/pq-code-package/mlkem-native/pull/1393, we noticed that the HOL-Light proof for the x86 INTT proves the bound <=26632, while our contract requires <= 26631.
This commit corrects the bound in the HOL-Light specification.


mkannwischer:
Merged as a part of #1383

PR #1193, opened 2025-09-20
Add PPC64LE backend

hanno-becker:
This PR is a continuation of #1184, taking care of integration and testing.

- [ ] (Required) @mkannwischer or @hanno-becker review assembly and confirm validity of arithmetic bounds assumed by the frontend.
   - [x] **Issue:** InvNTT needs to move Montgomery scaling to the front to avoid overflow. @dannytsen 
- [ ] (Optional) Generate twiddles using `autogen` as done for the other backends
- [ ] (Required) Understand exact architecture requirement: Is the code supported on Power8 or not?
- [x] (Required) Rebase atop #1187 so we get further confidence in correctness through random tests

PR #1406, opened 2025-12-18
CI: Add native ppc64le (POWER10) CI 

mkannwischer:
Courtesy of @bhess /IBM, we now have a self-hosted POWER10 runner.
This commit adds functional tests to CI.
I had attempted to add it to the existign "kat_tests", but the sanitizer tests
don't work properly on that platform.
Note that opt tests are already enabled so that
https://github.com/pq-code-package/mlkem-native/pull/1193 can simply be
rebased on top of this.

### DIFF
--- a/.github/actions/bench/action.yml
+++ b/.github/actions/bench/action.yml
@@ -91,7 +91,7 @@ runs:
     - name: Check namespace
       shell: ${{ env.SHELL }}
       run: |
-        check-namespace
+        ./scripts/check-namespace
     - name: Store benchmark result
       uses: benchmark-action/github-action-benchmark@d48d326b4ca9ba73ca0cd0d59f108f9e02a381c7 # v1.20.4
       with:
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -69,6 +69,14 @@ jobs:
           ldflags: "-flto"
           bench_extra_args: "-r"
           nix_shell: ci-bench
+        - system: pqcp-ppc64
+          name:  ppc64le (POWER10) benchmarks
+          bench_pmu: PERF
+          archflags: "-mcpu=native"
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -173,6 +173,21 @@ jobs:
...
           cflags: "-DMLKEM_DEBUG"
           opt: 'opt'
+  ppc64le_tests:
+    if: github.repository_owner == 'pq-code-package' && !github.event.pull_request.head.repo.fork
+    name: Functional tests (ppc64le/POWER10, native)
+    runs-on: pqcp-ppc64
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - name: build + test
+        uses: ./.github/actions/multi-functest
```
hanno-becker, line 183: Could we set `MLK_FORCE_PPC64LE` to ensure that this is really ending up on a PPC64LE system that's detected as such?

hanno-becker (changes requested):
We should use `MLK_FORCE_PPC64LE` to double-check that we're on a PPC64LE system.

Also, can we add this to the benchmarking CI? We'll need to see if it's stable enough, but until we know better, we may as well try?

.github/workflows/ci.yml
mkannwischer: Added

mkannwischer:
> We should use `MLK_FORCE_PPC64LE` to double-check that we're on a PPC64LE system.
> 
> Also, can we add this to the benchmarking CI? We'll need to see if it's stable enough, but until we know better, we may as well try?

Great idea. I asked @bhess to help with this and we got it to work. I have no idea how meaningful and stable the benchmarks are, but we can find that out over time.

PR #1424, opened 2025-12-21
Add test for failing dynamic allocation

### LINKED ISSUE #1404: Add test for failing dynamic allocation
We now alloc custom alloc/free implementations via MLK_CONFIG_CUSTOM_ALLOC_FREE, but our testing does not cover the case of failing dynamic allocation.

**Task:** Write a custom test which uses a custom alloc/free implementation to trigger an allocation failure at an arbitrary point in the program.

Idea: Add a new test_alloc.c. The test uses a custom alloc/free where alloc maintains global state int counter and int fail_on_counter. counter gets increased with every allocation, and alloc fails precisely if counter == fail_on_counter (but still increases counter to not fail the next time). counter starts with 0, so if fail_on_counter == -1, no failure will happen, and instead one can check how many invocations to alloc have been made. Each test should first set fail_on_counter = -1 and run the functionality under test (keygen/enc/dec) to find out the number of allocations. It should then re-run the test counter times, resetting counter = 0 each time and setting fail_on_counter = i in the i-th iteration (i=0..(counter-1)). This way, each allocation failure path should be exercised.
For enc and dec, run keygen and keygen+enc first (without failure), so we're always in a successful context.

hanno-becker:
* Based on #1405 
* Resolves #1404 

```
Tests the top-level API correctly handles allocation failures by:
- Returning MLK_ERR_OUT_OF_MEMORY on allocation failure
- Cleaning up all allocated memory (no leaks)
- Freeing in LIFO order (enabling bump allocator use)

Uses custom bump allocator with LIFO tracking to systematically inject
failures at each allocation point and verify error handling.
```

### DIFF
--- a/test/test_alloc.c
+++ b/test/test_alloc.c
@@ -1,0 +1,402 @@
+/*
+ * Copyright (c) The mlkem-native project authors
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+ */
+#include <stddef.h>
+#include <stdio.h>
+#include <string.h>
+
+/* Expose declaration of allocator (normally internal) */
+#define MLK_BUILD_INTERNAL
+#include "../mlkem/mlkem_native.h"
+#include "notrandombytes/notrandombytes.h"
+
+/*
+ * This test checks that
+ * - we handle allocator failures correctly, propagating MLK_ERR_OUT_OF_MEMORY
+ *   and cleaning up all memory, and
+ * - we leak no memory, and
+ * - we always de-allocate in the reverse order of allocation, thereby
+ *   allowing the use of a bump allocator.
+ *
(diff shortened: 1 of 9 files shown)


test/test_alloc.c
```diff
@@ -0,0 +1,401 @@
...
+  {
+    fprintf(
+        stderr,
+        "ERROR: Attempting to free %s at %s:%d but allocation stack is empty\n",
+        var, file, line);
+    exit(1);
+  }
+
+  top = &alloc_stack[alloc_stack_top - 1];
+  if (top->addr != addr)
```
mkannwischer: Can you extend this to also error if the wrong size is passed to MLK_FREE? That is currently not detected.

hanno-becker: Good idea! Done

Issue #1413, opened 2025-12-18
Port HOL-Light performance improvements for NTT/INTT


mkannwischer:
- https://github.com/awslabs/s2n-bignum/pull/325

PR #1425, opened 2025-12-21
HOL-Light: Speed up NTT/INTT proofs

mkannwischer:
This commit ports @jargh's improvements to the NTT/INTT proofs in s2n-bignum which should result in vast proof performance improvements (1.6x- 4.3x according to the PR).

- Resolves https://github.com/pq-code-package/mlkem-native/issues/1413
- Ports https://github.com/awslabs/s2n-bignum/pull/325

### DIFF
--- a/proofs/hol_light/arm/proofs/mlkem_intt.ml
+++ b/proofs/hol_light/arm/proofs/mlkem_intt.ml
@@ -655,7 +655,7 @@ let MLKEM_INTT_CORRECT = prove
   (*** Simulate all the way to the end, in effect unrolling loops ***)
 
   MAP_UNTIL_TARGET_PC (fun n -> ARM_STEPS_TAC MLKEM_INTT_EXEC [n] THEN
-            (SIMD_SIMPLIFY_TAC [barred; barmul]))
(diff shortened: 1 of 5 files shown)


mkannwischer:
x86_64 NTT: 282m -> 34m 
x86_64 iNTT: 281m -> 44m 
AArch64 NTT: 54m -> 75m
AArch64 iNTT: 92m -> 63m

Did I do something wrong for the AArcch64 NTT? I can't find the mistake though.

jargh:
No, I also found that it didn't help and was even regressive for ARM. But since it substantially improves the x86 proofs and those are the more troubling bottleneck, I was satisfied for now. I don't exactly know why the difference is so pronounced, but I'll eventually revisit this and try and do everything better.

hanno-becker (approved):
This does not touch the spec and drastically improves the proof performance. I don't particularly like the approach taken (I think it would be more natural to keep terms as they are and instead not use unbounded tactics, e.g. bounded conversions or guarding the simplifier using congruence rules) but one can always switch to other alternative optimization approaches in the future.

Issue #1428, opened 2025-12-22
x86_64: Autogenerate entire constant table `qdata`


mkannwischer:
Currently in `consts.c` only the twiddle factors in the `qdata` table are autogenerated, while the other constants are hardcoded. 
We should port https://github.com/pq-code-package/mldsa-native/pull/812 to avoid the risk of the HOL-Light and the effective constant arrays diverging.

Issue #1380, opened 2025-12-14
Add commentary to autogeneration of twiddles in HOL-Light x86 NTT proof file


hanno-becker:
Leftover from #1374

mkannwischer:
This is a broader issue than just HOL-Light - there is a TODO for it in the code: 
https://github.com/pq-code-package/mlkem-native/blob/1c0c8e1ac4d99d77bb84803b9b19f248cb19558d/mlkem/src/native/x86_64/src/consts.c#L29-L38

We could consider autogenerating it.

PR #1430, opened 2025-12-23
x86_64: Autogenerate the entire constant array `qdata` for NTT/INTT

mkannwischer:
The constant array required for the x86_64 consists of multiple parts: first some index constants, then the twiddle factors for the NTT/INTT, and then the twiddle factors for the mulcache computation. Thus far, the autogeneration script would only generate the twiddle factors into two separate .i files that are then included into consts.c, while the index constants are hardcoded in the consts.c

However, for HOL-Light we require to generate the entire constant array. Having two diffent functions generating the constants is a risk that they may eventually diverge.
This commit fixes this by autogenerating the entire qdata in consts.c (the .i files is no longer needed). This way we can use the same generation function for both consts.c and the HOL-Light proofs.

While at it we also improve documentation of the index constants.

- Resolves https://github.com/pq-code-package/mlkem-native/issues/1428
- Resolves https://github.com/pq-code-package/mlkem-native/issues/1380


hanno-becker (changes requested):
Thank you @mkannwischer, this is great. Could you also take the opportunity to autogenerate `consts.h`? This way, we can emit the offsets into and length of the array from the same code that generates the array itself.

scripts/autogen
```diff
@@ -1223,36 +1223,39 @@ def gen_avx2_mulcache_twiddles():
...
+    ntt_zetas = list(gen_avx2_fwd_ntt_zetas())
+    mulcache_twiddles = list(gen_avx2_mulcache_twiddles())
 
+    revidxb_idx = 0
+    revidxd_idx = revidxb_idx + len(revidxb)
+    ntt_zetas_idx = revidxd_idx + len(revidxd)
+    mulcache_twiddles_idx = ntt_zetas_idx + len(ntt_zetas)
+
+    constants = revidxb + revidxd + ntt_zetas + mulcache_twiddles
+    offsets = [revidxb_idx, revidxd_idx, ntt_zetas_idx, mulcache_twiddles_idx]
```
hanno-becker: Nit: Use a dictionary here

PR #1434, opened 2025-12-26
test_alloc: Align bump buffer

hanno-becker:
test/test_alloc.c uses a custom bump allocator. While all allocation sizes are aligned to multiples of 32 bytes, the allocation buffer itself can be unaligned. This can crash the test on systems with alignment requirement, such as x86+AVX2.

This is not an issue in the actual implementation, which uses appropriate alignment constraints: See the default implementation of MLK_ALLOC in src/common.h.


hanno-becker:
We also need to add the alloc test to `scripts/tests` and the `functest` and `multi-functest` action.

PR #1441, opened 2025-12-31
nix: Add HOL-Light cross shells

hanno-becker:
As a first step towards developing and checking HOL-Light proofs for architectures different from the host architecture, this commit adds new nix shells for HOL-Light + cross toolchains.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -41,6 +41,10 @@
                 '')
             ];
           };
+          holLightShellHook = ''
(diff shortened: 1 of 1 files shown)


mkannwischer (approved):
Thanks! I tested that I can enter the `hol_light-cross-aarch64 ` shell on x86 Linux and run `hol.sh` there without any issues.
Have not tried to run it on MacOS / AArch64.

PR #1444, opened 2026-01-01
HOL-Light: Allow cross-generation of byte code

hanno-becker:
This commit adjusts the HOL-Light makefiles to support cross-compilation of the source under proof.

autogen is adjusted to cross-generate object code when requested via the existing option --force-cross.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -730,7 +730,7 @@ jobs:
       matrix:
         target:
           - system: macos-latest
-            nix_cache: 'false'
+            nix_cache: 'true'
             nix_shell: 'hol_light-cross-x86_64'
             extra_args: '--force-cross'
           # TODO: autogen does not yet work on macos15-intel (#1304)
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -69,10 +69,11 @@ jobs:
           devShell: ci
           gh_token: ${{ secrets.GITHUB_TOKEN }}
           script: |
-            # NOTE: we're not running cross compilation tests on macOS currently
-            #       and building cross compilation toolchain on macOS runner took
-            #       extremely long time
-            if [[ ${{ runner.os }} != 'macOS' ]]; then
+            # We only run cross-compilation checks for x86 on macos-latest,
+            # so restrict caching to the corresponding cross shell.
+            if [[ ${{ runner.os }} == 'macOS' ]]; then
+              nix develop .#ci-cross-x86_64 --profile tmp-cross
+            else
               nix develop .#ci-cross --profile tmp-cross
               # GH ubuntu-24.04 image tend to run outof space
               if [[ ${{ matrix.runner }} == 'ubuntu-24.04' ]]; then


scripts/autogen
```diff
@@ -3293,7 +3293,30 @@ def update_bytecode_in_proof_script(filepath, bytecode):
...
+def update_hol_light_bytecode_for_arch(arm_or_x86, force_cross=False):
+    status_update(
+        f"HOL Light bytecode ({arm_or_x86})",
+        f"Running make dump_bytecode ({arm_or_x86})... (this may take a few minutes)",
+    )
+
+    if arm_or_x86 == "arm":
+        source_arch = "aarch64"
+    else:
+        source_arch = "x86_64"
```
mkannwischer, line 3305: We should clean this up so that we are using the same identifier for the architectures everywhere. That requires quite a lot of changes though. What do you think?

.github/workflows/ci.yml
```diff
@@ -698,18 +698,19 @@ jobs:
       matrix:
         target:
           - system: macos-latest
-            nix_cache: 'false'
-            nix_shell: 'ci'
+            nix_shell: 'ci-cross-x86_64'
+            nix_cache: 'true'
```
mkannwischer, line 702: Are you sure that MacOS nix cache includes an x86 cross toolchain? I don't think so: 
```
            if [[ ${{ runner.os }} != 'macOS' ]]; then
              nix develop .#ci-cross --profile tmp-cross
              # GH ubuntu-24.04 image tend to run outof space
              if [[ ${{ matrix.runner }} == 'ubuntu-24.04' ]]; then
                nix-collect-garbage
              fi
            fi
```

scripts/autogen
```diff
@@ -3293,7 +3293,30 @@ def update_bytecode_in_proof_script(filepath, bytecode):
     update_file(filepath, updated_content)
 
 
-def update_hol_light_bytecode_for_arch(arm_or_x86):
+def update_hol_light_bytecode_for_arch(arm_or_x86, force_cross=False):
```
mkannwischer, line 3296: Do you also want to combine the `hol_light_bytecode_x86_64` and `hol_light_bytecode` jobs into one now?

.github/workflows/ci.yml
mkannwischer: We could include it in the cache if the job is to slow. Otherwise, `nix_cache: false` is probably not hurting you much here.

mkannwischer (approved):
Thanks @hanno-becker! Great improvement!
I tested that I can run
```
autogen --force-cross --update-hol-light-bytecode
```
on both AArch64 MacOS and x86_64 Linux and the byte code for the corresponding other arch gets correctly build. 

A couple of comments below, but they are not blocking.

.github/workflows/ci.yml
hanno-becker: > Otherwise, nix_cache: false is probably not hurting you much here.

Hm, I think it does? It takes >1h.

scripts/autogen
hanno-becker: Yes, let's do this separately.

scripts/autogen
hanno-becker: Yes, we can consider this now; it might be more cost-effective to use Arm runners for the x86 proofs. Let's do this separately, though?

scripts/autogen
hanno-becker: Opened https://github.com/pq-code-package/mlkem-native/issues/1448

.github/workflows/ci.yml
hanno-becker: I added `ci-cross-x86_64` to the nix cache for `macos-latest`.

scripts/autogen
hanno-becker: Opened https://github.com/pq-code-package/mlkem-native/issues/1449

PR #1445, opened 2026-01-01
CBMC: Use polyvec_permute contract in polymat_permute proof

mkannwischer:
The mlk_polymat_permute_bitrev_to_custom was split up into two parts to work around CBMC performance issues when proving it on top of the native backend.
However, it was forgotten to call mlk_polyvec_permute_bitrev_to_custom (which is proven for both the C and native backend) in the mlk_polymat_permute_bitrev_to_custom proof.
As a consequence, mlk_polymat_permute_bitrev_to_custom wasn't actually proven for the native backend, as the C version was inlined. This commit closes that gap by calling it by contract.


hanno-becker:
Very good observation, @mkannwischer. I wonder how we can automatically detect such situations?

hanno-becker (approved):
Astute observation, @mkannwischer. I opened https://github.com/pq-code-package/mlkem-native/issues/1446 to track how we could detect such issues in the future.

PR #1456, opened 2026-01-03
Use consistent architecture identifier

### LINKED ISSUE #1448: autogen: Use architecture identifiers consistently
`autogen` sometimes uses both `"aarch64"` and `"arm"` to represent 64-bit Arm platforms, and both `"x86_64"` and `"x86"` to represent 64-bit Intel/AMD platforms. This should consistent.

hanno-becker:
* Resolves #1448 

Use `aarch64` and `x86_64` consistently, removing uses of `arm` and `x86`,
most notably the HOL-Light proof directories.

While testing, noticed a missing bytecode dump for the native `tomont`
implementation on x86_64 and added it to x86_64/proofs/dump_bytecode.ml.

Also, noticed that ASM simplification from mlkem -> proofs/hol_light
was so far only done for the native architecture in scripts/autogen.
This is adjusted to run whenever a cross-compiler is available, matching
the behavior of autogen for the simplification from dev -> mlkem.


proofs/hol_light/aarch64/Makefile
```diff
@@ -141,4 +141,4 @@ FORCE: ;
 # https://stackoverflow.com/a/23569003/1488216
 NUM_CORES_FOR_SEMATEST = $(shell getconf _NPROCESSORS_ONLN)
 sematest: FORCE $(OBJ) $(SRC_ARM)/proofs/simulator_iclasses.ml $(SRC_ARM)/proofs/simulator.native
-	$(SRC)/tools/run-sematest.sh arm $(NUM_CORES_FOR_SEMATEST)
+	$(SRC)/tools/run-sematest.sh aarch64 $(NUM_CORES_FOR_SEMATEST)
```
mkannwischer, line 144: This seems to be an unintentional change? Should we remove the sematests?

hanno-becker: Good catch! You're right, but as you say this isn't  (was never?) functional anyway so does little harm.

PR #1458, opened 2026-01-05
CI: Add linting of commit messages

mkannwischer:
This commit adds linting of commit messages based on https://github.com/mristin/opinionated-commit-message. It enforces:
 - Separate subject from body with a blank line
 - Limit the subject line to 72 characters
 - Capitalize the subject line
 - Do not end the subject line with a period
 - Use the imperative mood in the subject line


mkannwischer:
This doesn't seem to work well unless we want to follow a much stricter commit message style like https://www.conventionalcommits.org/. I can't find another action that gives us the configurablility I would want.
Closing this again for now.

PR #1463, opened 2026-01-08
CT: Strengthen post-condition of `ct_memcmp`

mkannwischer:
To align with mldsa-native, this ports the strengthening of the post-condition: The return value is now guaranteed to be 0x00 (equal) or 0xFF (not-equal), rather than 0x00 (equal), not-0x00 (not-equal).


mkannwischer:
Should we rename `verify.{c,h}` to `ct.{c,h}` to align with mldsa-native?
It does not have much in common with the reference implementation's `verify` anymore.

PR #1465, opened 2026-01-09
Allow the public header to work in C++ mode.

davidchisnall:
Without this, includes of the header must be wrapped in `extern "C"`, which then causes problems if the config header includes headers that expect to be included in C or C++ mode.

### DIFF
--- a/mlkem/mlkem_native.h
+++ b/mlkem/mlkem_native.h
@@ -183,6 +183,11 @@
 
 #include <stdint.h>
 
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
 /*************************************************
  * Name:        crypto_kem_keypair_derand
  *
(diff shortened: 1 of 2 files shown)


mlkem/mlkem_native.h
```diff
@@ -183,6 +183,10 @@
 
 #include <stdint.h>
 
+#ifdef __cplusplus
+extern "C" {
```
hanno-becker:
```suggestion
extern "C" 
{
```
and we need to add `IndentExternBlock: NoIndent` to `.clang-format`, otherwise it wants to indent the whole file.

hanno-becker (changes requested):
LGTM, just a minor thing to appease `clang-format`.

mlkem/mlkem_native.h
davidchisnall: Done (both)

PR #1467, opened 2026-01-09
Add context parameter

hanno-becker:
Follow-up from #1339 by @davidchisnall:
* Creating PR from main repository to allow full CI to be run
* Minor fixes
* Running `format` and `autogen`
* Using the new option in the alloc failure test, which now passes the test context (including the bump allocator) explicitly rather than relying on a static global

### DIFF
--- a/test/configs/configs.yml
+++ b/test/configs/configs.yml
@@ -435,15 +435,18 @@ configs:
           #if !defined(MLK_CONFIG_KEYGEN_PCT)
           #define MLK_CONFIG_KEYGEN_PCT
           #endif
+      MLK_CONFIG_CONTEXT_PARAMETER: true
+      MLK_CONFIG_CONTEXT_PARAMETER_TYPE: struct test_ctx_t *
       MLK_CONFIG_CUSTOM_ALLOC_FREE:
         content: |
           #define MLK_CONFIG_CUSTOM_ALLOC_FREE
           #if !defined(__ASSEMBLER__)
           #include <stdlib.h>
-          void * custom_alloc(size_t sz, const char *file, int line, const char *var, const char *type);
-          void custom_free(void *p, size_t sz, const char *file, int line, const char *var, const char *type);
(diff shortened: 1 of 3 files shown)


hanno-becker:
We still need to work through some fundamental issues with the tests. Closing for now.

mlkem/src/kem.c
mkannwischer, line 206: This will result in an unused argument warning in case the context is used? 
Maybe add an
```c
#if defined(MLK_CONFIG_CONTEXT_PARAMETER)
((void)context);
#endif
```
?

mlkem/mlkem_native.h
```diff
@@ -215,7 +215,13 @@ MLK_API_MUST_CHECK_RETURN_VALUE
 int MLK_API_NAMESPACE(keypair_derand)(
     uint8_t pk[MLKEM_PUBLICKEYBYTES(MLK_CONFIG_API_PARAMETER_SET)],
     uint8_t sk[MLKEM_SECRETKEYBYTES(MLK_CONFIG_API_PARAMETER_SET)],
-    const uint8_t coins[2 * MLKEM_SYMBYTES]);
+    const uint8_t coins[2 * MLKEM_SYMBYTES]
+#ifdef MLK_CONFIG_CONTEXT_PARAMETER
+    ,
+    MLK_CONFIG_CONTEXT_PARAMETER_TYPE context
```
mkannwischer, line 221: The formatting here is rather ugly. but I guess there is no way around this?

mkannwischer:
I've ported this to mldsa-native to see if any problems surface: https://github.com/pq-code-package/mldsa-native/pull/874
No serious problems found.

mlkem/src/kem.c
mkannwischer: I pushed this fix.

mlkem/mlkem_native.h
mkannwischer: Nothing we can do about it, I think. Let's keep as is.

Issue #1442, opened 2025-12-31
Provide macro indicidating upper bound of memory allocations via MLK_ALLOC


mkannwischer:
Allowing customization of `MLK_ALLOC` enables consumers to implement their own memory allocation routine. Depending on how this is implemented, it may be important to know at compile time the maximum amount of memory that will be allocated (independent of this, it may be important that all memory is allocated in a single `MLK_ALLOC` - but that is a separate issue). 

We should provide a macro to consumers providing the maximum amount of memory any routine for any parameter set will require. The name is tbd, but it could be `MLK_MAX_WORKSPACE_SIZE` or `MLK_MAX_ALLOC`.

This should be verified in the MLK_ALLOC macro via `static_assert(sizeof(struct) <= MLK_MAX_WORKSPACE_SIZE)` (there we can only check a single allocation) and additionally in the alloc_test (here we can check the accumulated allocations).

Issue #869, opened 2025-03-12
Add clang 20 tests


mkannwischer:
`clang 20.1.0 ` just got released: https://releases.llvm.org/20.1.0/tools/clang/docs/ReleaseNotes.html

We should add that to CI (for compiler tests and constant-time tests) once it lands in nix-unstable (right now it is still [`clang 20.1.0-rc3`](https://search.nixos.org/packages?channel=unstable&show=clang_20&from=0&size=50&sort=relevance&type=packages&query=clang_20).)
I had a previous PR to test `clang 20.1.0-rc2` (#797) - so this should be straightforward.

PR #1468, opened 2026-01-10
mlkem_native.h: Introduce `MLK_TOTAL_ALLOC` constants

mkannwischer:
For consumers implementing custom allocations (in cases where the stack is not large enough), it may be important to know how much space has to be reserved for allocations of large structures.

This commit adds per-operation allocation constants:
 - MLK_TOTAL_ALLOC_{512,768,1024}_{KEYPAIR,ENCAPS,DECAPS} for each parameter set and operation, indicating peak memory consumption.
 - MLK_MAX_TOTAL_ALLOC_{KEYPAIR,ENCAPS,DECAPS} for maximum across parameter sets per operation.
 - MLK_TOTAL_ALLOC_{512,768,1024} for maximum across operations per parameter set.
 - MLK_MAX_TOTAL_ALLOC for maximum across all parameter sets and operations.

The alloc_test is extended to check that the constants indeed match the peak memory consumption of the corresponding functions.

I included the extension from #869 that differentiates between keygen w/ and w/o PCT. The commits are kept separate though to follow the same structure as in mldsa-native. 

 - Ported from https://github.com/pq-code-package/mldsa-native/pull/850
 - Ported from https://github.com/pq-code-package/mldsa-native/pull/869
 - Resolves https://github.com/pq-code-package/mlkem-native/issues/1442


mkannwischer:
~Note that the keypair numbers include the pair-wise consistency test (PCT) - keypair without PCT will be considerably lower. 
We probably want to provide the numbers both w/ and w/o PCT, but that needs more changes (also in mldsa-native).
I prefer to do that as a follow-up.~

I have included the extension as a separate commit.

mlkem/mlkem_native.h
```diff
@@ -428,4 +428,81 @@ int MLK_API_NAMESPACE(check_sk)(
...
+#define MLK_TOTAL_ALLOC_1024_KEYPAIR MLK_TOTAL_ALLOC_1024_KEYPAIR_NO_PCT
+#endif
+
+/*
+ * `MLK_MAX_TOTAL_ALLOC_{KEYPAIR,ENCAPS,DECAPS}` is the maximum across all
+ * parameter sets for each operation.
+ */
+#define MLK_MAX_TOTAL_ALLOC_KEYPAIR MLK_TOTAL_ALLOC_1024_KEYPAIR
+#define MLK_MAX_TOTAL_ALLOC_ENCAPS MLK_TOTAL_ALLOC_1024_ENCAPS
+#define MLK_MAX_TOTAL_ALLOC_DECAPS MLK_TOTAL_ALLOC_1024_DECAPS
```
hanno-becker: In a single-level build of level != 1024, this is a bit confusing. I am inclined to stick with constants which make the level specific.

hanno-becker (changes requested):
I think it is slightly confusing to expose `MLK_MAX_TOTAL_ALLOC` as the maximum across all security levels, even in a single-level build. I'd prefer to only export constants explicitly labeled by a security level.

mkannwischer:
> I think it is slightly confusing to expose `MLK_MAX_TOTAL_ALLOC` as the maximum across all security levels, even in a single-level build. I'd prefer to only export constants explicitly labeled by a security level.

Okay, yes, makes sense. I have removed the macros. Also removed it in mldsa-native in https://github.com/pq-code-package/mldsa-native/pull/869.
I kept it as a separate commit, so that each change has a corresponding one in mldsa-native.

hanno-becker (approved):

PR #1473, opened 2026-01-11
autogen: fix print for narrow terminals

L-series:
This commit replaces the hardcoded width of 160 in `clear_status_line`
with the width of ones terminal window. This resolves the issue in which
we see autogen generate large amounts of new lines for narrow terminals
(width <160). In case we are running on a system with a non-interactive
shell (e.g. some of the CI runners) we fall back to the default.

Signed-off-by: Andreas Hatziiliou <andreas.hatziiliou@savoirfairelinux.com>

### DIFF
--- a/scripts/autogen
+++ b/scripts/autogen
@@ -202,10 +202,16 @@ BLUE = "\033[94m"
 BOLD = "\033[1m"
 NORMAL = "\033[0m"
 
+# Set terminal width, fallback to default on systems where none is present.
+try:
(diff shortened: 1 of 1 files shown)


hanno-becker:
Thank you for noting this, @L-series. The fix looks good to me.

I imagine there are Python packages one can use for the sort of status update done in `autogen`. Mid-term, we may want to switch to those rather than baking our own.

PR #1476, opened 2026-01-12
CI: Bump OpenTitan version

mkannwischer:
This commit bumps to the most recent commit from
https://github.com/zerorisc/expo.

The integration of mlkem-native was changed in
https://github.com/zerorisc/expo/pull/186 to use workbuffer arguments
instead of using the default stack allocations to allow running with small
stack configurations.
The patch we use to only run the derandomized tests in CI is updated
accordingly.

- Resolves https://github.com/pq-code-package/mlkem-native/issues/1528


mkannwischer:
Some problems with the Python dependencies. Reported upstream: https://github.com/zerorisc/expo/issues/183

hanno-becker (approved):

PR #1478, opened 2026-01-13
CI: Comment on PRs if CBMC proofs fail

hanno-becker:
This commit extends the CI to comment on PRs for which CBMC proofs fail. In this case, the table of failing proofs and their runtime is printed.


hanno-becker:
Closing in favor of #1486

PR #1480, opened 2026-01-15
Update x4 Keccak with verified assembly for perfromance data

manastasova:
Testing PR to collect performance data.


mkannwischer:
You'll need to open this PR from a local branch to use the EC2 benchmarking.

PR #1482, opened 2026-01-16
sys.h: Fix __attribute__ usage for non-GCC compilers

mkannwischer:
Restructure `MLD_INLINE` and `MLD_ALWAYS_INLINE` into separate blocks, each independently guarded:

`MLD_INLINE`:
- MSVC: `__inline`
- C99 or `inline` macro defined: `inline`
- GCC/Clang C90: `__attribute__((unused))` to silence warnings
- Other C90: empty

`MLD_ALWAYS_INLINE`:
- MSVC: `__forceinline`
- GCC/Clang C99+: `MLD_INLINE __attribute__((always_inline))`
- Other: `MLD_INLINE` (no forced inlining)

This fixes two issues:
1. If `inline` was defined as a macro before including sys.h, the MSVC check was bypassed and `__attribute__` was used, causing MSVC failures.
2. For non-MSVC/GCC/Clang compilers like IAR, `__attribute__` was used unconditionally, causing build failures.

Also fix typo: `defined(clang)` -> `defined(__clang__)`

- Ported from https://github.com/pq-code-package/mldsa-native/pull/884

### DIFF
--- a/mlkem/src/indcpa.c
+++ b/mlkem/src/indcpa.c
@@ -380,36 +380,32 @@ int mlk_indcpa_keypair_derand(uint8_t pk[MLKEM_INDCPA_PUBLICKEYBYTES],
                               const uint8_t coins[MLKEM_SYMBYTES],
                               MLK_CONFIG_CONTEXT_PARAMETER_TYPE context)
 {
-  typedef struct
-  {
-    MLK_ALIGN uint8_t buf[2 * MLKEM_SYMBYTES];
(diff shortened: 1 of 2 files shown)


mkannwischer:
The commit message still says MLD_*... that needs to be changed to MLK_* everywhere.

mkannwischer:
Merged as a part of https://github.com/pq-code-package/mlkem-native/pull/1483

PR #1486, opened 2026-01-16
CI/CBMC: Report CBMC runtimes and warn about regressions

### LINKED ISSUE #1322: CBMC: Warn about increase of proof times
Unless we specifically look for the runtime of the CBMC CI jobs, we don't currently notice if a CBMC proof's runtime increases significantly.

**Task:**
- [ ] Using the same approach used for benchmarking data, persist CBMC performance data in a branch on every merge. Ideally, we'd have one data point per combination of function and parameter set, and also store relevant metadata such as the commit hash and versions of CBMC, bitwuzla, z3.
- [ ] Once CBMC completes successfully in an open PR, warn about any function whose CBMC runtime is either a) more than 50% slower (made up number -- TBD) than previously, b) slower than 10min (made up number -- TBD).

hanno-becker:
This commit adds to the CBMC CI action another step which, for PRs,
- emits one comment per security level showing the CBMC runtimes
  and the relative slowdown/speedup compares to the last run on `main`
- warns about severe proof regressions (>50% for now) and failures.

Comments are updated when the CI is run multiple times on the same PR.

* Resolves #1322

### DIFF
--- a/.github/actions/cbmc/action.yml
+++ b/.github/actions/cbmc/action.yml
@@ -51,7 +51,17 @@ runs:
             EOF
       - name: Run CBMC proofs (MLKEM_K=${{ inputs.mlkem_k }})
         shell: ${{ env.SHELL }}
+        env:
+          GH_TOKEN: ${{ inputs.gh_token }}
         run: |
           echo "::group::cbmc_${{ inputs.mlkem_k }}"
-          tests cbmc --mlkem-parameter-set ${{ inputs.mlkem_k }} --per-proof-timeout 1800;
+          tests cbmc --mlkem-parameter-set ${{ inputs.mlkem_k }} --per-proof-timeout 1800 --output-result-json ${{ github.workspace }}/cbmc_result_k${{ inputs.mlkem_k }}.json || cbmc_failed=true
           echo "::endgroup::"
+          python3 ${{ github.action_path }}/report.py \
+            --results-json "${{ github.workspace }}/cbmc_result_k${{ inputs.mlkem_k }}.json" \
+            --mlkem-k "${{ inputs.mlkem_k }}" \
+            --min-runtime 20 \
+            --regression-threshold 1.5 \
+            --gh-pages-branch gh-pages \
+            --data-dir dev/cbmc
+          if [ "$cbmc_failed" = "true" ]; then exit 1; fi
--- a/.github/actions/cbmc/report.py
+++ b/.github/actions/cbmc/report.py
@@ -1,0 +1,265 @@
+#!/usr/bin/env python3
+# Copyright (c) The mlkem-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+"""CBMC proof results reporter for GitHub Actions."""
+
+import argparse
+import base64
+import json
+import os
+import subprocess
+import tempfile
+from collections import namedtuple
+
+# GitHub context
+GITHUB_REPOSITORY = os.environ.get("GITHUB_REPOSITORY", "")
+GITHUB_REF = os.environ.get("GITHUB_REF", "")
+GITHUB_SHA = os.environ.get("GITHUB_SHA", "")[:7]
+
+COMMENT_TAG = "cbmc-report"
+
+# Status symbols
+OK = "✅"
+WARN = "⚠️"
+FAIL = "❌"
+
+ProofResult = namedtuple(
+    "ProofResult", ["name", "status", "current", "previous", "change"]
+)
+
+
+def get_args():
+    parser = argparse.ArgumentParser(description="CBMC proof results reporter")
+    parser.add_argument("--results-json", required=True)
+    parser.add_argument("--mlkem-k", required=True, choices=["2", "3", "4"])
+    parser.add_argument("--min-runtime", type=int, default=20)
+    parser.add_argument("--regression-threshold", type=float, default=1.5)
+    parser.add_argument("--gh-pages-branch", default="gh-pages")
+    parser.add_argument("--data-dir", default="dev/cbmc")
+    return parser.parse_args()
+
+
+def get_pr_number():
+    """Get PR number from GitHub event."""
+    event_path = os.environ.get("GITHUB_EVENT_PATH")
+    if not event_path or not os.path.exists(event_path):
+        return None
+    with open(event_path) as f:
+        event = json.load(f)
+    return event.get("pull_request", {}).get("number")
+
+
+def fetch_baseline(cfg):
+    """Fetch baseline data from gh-pages branch."""
+    baseline_file = f"{cfg.data_dir}/mlkem-{cfg.param_set}.json"
+    subprocess.run(
+        ["git", "fetch", "origin", cfg.gh_pages_branch, "--depth=1"],
+        capture_output=True,
+    )
+    result = subprocess.run(
+        ["git", "show", f"origin/{cfg.gh_pages_branch}:{baseline_file}"],
+        capture_output=True,
+        text=True,
+    )
+    if result.returncode == 0 and result.stdout.strip():
+        return json.loads(result.stdout)
+    return None
+
+
+def render_table(rows):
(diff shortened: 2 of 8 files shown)


.github/actions/cbmc/action.yml
```diff
@@ -50,8 +50,34 @@ runs:
               - $(bash --version | grep -m1 "")
             EOF
       - name: Run CBMC proofs (MLKEM_K=${{ inputs.mlkem_k }})
+        id: cbmc_run
```
mkannwischer: Out of interest: what are these ids for?

.github/actions/cbmc/action.yml
```diff
@@ -50,8 +50,34 @@ runs:
...
+        run: |
+          case "${{ inputs.mlkem_k }}" in
+            2) echo "name=512" >> $GITHUB_OUTPUT ;;
+            3) echo "name=768" >> $GITHUB_OUTPUT ;;
+            4) echo "name=1024" >> $GITHUB_OUTPUT ;;
+          esac
+      - name: Setup GitHub CLI
+        uses: ./.github/actions/setup-os
+        with:
+          packages: gh
```
mkannwischer: can we set this up with nix please?

.github/actions/cbmc/action.yml
```diff
@@ -50,8 +50,34 @@ runs:
...
+          tests cbmc --mlkem-parameter-set ${{ inputs.mlkem_k }} --per-proof-timeout 1800 || echo "cbmc_failed=true" >> $GITHUB_OUTPUT
           echo "::endgroup::"
+      - name: Map MLKEM_K to variant name
+        id: process_results
+        shell: bash
+        run: |
+          case "${{ inputs.mlkem_k }}" in
+            2) echo "name=512" >> $GITHUB_OUTPUT ;;
+            3) echo "name=768" >> $GITHUB_OUTPUT ;;
+            4) echo "name=1024" >> $GITHUB_OUTPUT ;;
```
mkannwischer: why do we need the name in this action? Just pass mlkem_k to cbmc-report?

.github/workflows/all.yml
mkannwischer, line 1: revert the changes in this file?

proofs/cbmc/run-cbmc-proofs.py
```diff
@@ -261,6 +261,18 @@ def run_build(litani, jobs, fail_on_proof_failure, summarize):
         sys.exit(1)
 
     if summarize:
+        # Export result JSON if CBMC_RESULT_JSON env var is set
+        result_json = os.getenv("CBMC_RESULT_JSON")
```
mkannwischer: nit: maybe don't use an env variable for this, but just add it as an parameter to the test script that forwards it to this script?

proofs/cbmc/lib/summarize.py
```diff
@@ -20,7 +20,12 @@ def get_args():
             "flags": ["--run-file"],
             "help": "path to the Litani run.json file",
             "required": True,
-        }
+        },
+        {
+            "flags": ["--output-result-json"],
+            "help": "path to export unified result JSON",
```
mkannwischer: what does unified stand for here?

mkannwischer (changes requested):

.github/actions/cbmc/action.yml
hanno-becker: Why? There's no reason (IMO) to run the post-processing and commenting in a nix shell

.github/actions/cbmc/action.yml
hanno-becker: So you can refer to outputs

.github/workflows/all.yml
hanno-becker: Done

.github/actions/cbmc/action.yml
hanno-becker: Okay

proofs/cbmc/run-cbmc-proofs.py
```diff
@@ -261,6 +261,18 @@ def run_build(litani, jobs, fail_on_proof_failure, summarize):
...
+        if result_json:
+            cmd_summarize = [
+                sys.executable,
+                str(pathlib.Path(__file__).parent / "lib" / "summarize.py"),
```
mkannwischer: How did we call the summarize.py script before? I can't see it in this PR.

proofs/cbmc/run-cbmc-proofs.py
hanno-becker: Ack

proofs/cbmc/lib/summarize.py
hanno-becker: Remnant from the time when there were two JSONs for failed vs. all proofs. Removing

.github/actions/cbmc/report.py
```diff
@@ -0,0 +1,262 @@
...
+        "PUT",
+        "-F",
+        f"message=Update CBMC results for ML-KEM-{PARAMETER_SET} ({GITHUB_SHA})",
+        "-F",
+        f"content={content_b64}",
+        "-F",
+        f"branch={GH_PAGES_BRANCH}",
+    ]
+    if sha:
+        cmd.extend(["-F", f"sha={sha}"])
```
mkannwischer, line 241: I'm not sure what is happening here. What's this sha argument?

.github/actions/cbmc/report.py
mkannwischer, line 1: maybe reconsider using https://pypi.org/project/github/ here? The direct invocations of the `gh` api are a little hard to understand

.github/actions/cbmc/report.py
hanno-becker: If you update a file you need to pass the hash of the previous version

.github/actions/cbmc/report.py
hanno-becker: I don't see value in switching to that; if we find ourselves doing much more manual manipulation of comments, maybe, but for now, I think the present approach via `gh` is good enough.

proofs/cbmc/run-cbmc-proofs.py
hanno-becker: Good spot. We didn't call it via subprocess. I'll try to switch it back.

.github/actions/cbmc/action.yml
mkannwischer: In no other workflow are we calling `setup-os` half way through. Adding `gh` to the ci-cbmc shell should be straightforward. 
Ultimately, so that we have control over when the tool gets updated, not Github or Ubuntu.

.github/actions/cbmc/report.py
mkannwischer: okay

.github/actions/cbmc/action.yml
hanno-becker: Fair enough -- I'll change.

proofs/cbmc/run-cbmc-proofs.py
hanno-becker: Done

proofs/cbmc/lib/summarize.py
hanno-becker: Removed

proofs/cbmc/run-cbmc-proofs.py
hanno-becker: Moved this and other environment variables into parameters

.github/actions/cbmc/action.yml
hanno-becker: Done

.github/actions/cbmc/action.yml
hanno-becker: Done

mkannwischer (approved):
Thank you for the clean-up and for all the work @hanno-becker!
This is a great improvement that will make our life much easier.

mkannwischer:
Tested it once more in https://github.com/pq-code-package/mlkem-native/pull/1492. Nice :)

PR #1490, opened 2026-01-16
Examples: Fix feature detection via auto.mk

mkannwischer:
The examples are using auto.mk to detect features such as AVX2. However, auto.mk depends on variables defined in compiler.mk which was not included previously disabling the entire feature detection and causing the native examples running without the native code. This commit includes compiler.mk which fixes the problem.

- Port of https://github.com/pq-code-package/mldsa-native/pull/890


hanno-becker (approved):
LGTM. Tested locally on an Arm and x86 machine that `monolithic_build_multilevel_native` does exercise the expected backends.

Issue #1304, opened 2025-11-13
autogen does not work on macos-15-intel


hanno-becker:
It was noticed in https://github.com/pq-code-package/mlkem-native/pull/1303 that `autogen` does not appear to succeed on `macos-15-intel` runners.

f15hr:
Adding this here but in #1498 I figured this:

The main issue is that simpasm conflicts with the way MacOS Intel Mach-O files are generated. Essentially, control flow statements like jumps, use relative offsets such as <label+offset> which can not be trivial changed to just <resolved label>. I attempted both replacing the relative offset with a label (by resolving the labels with the linker and just swapping them) and also keeping the relative offset in place. While both version compiled, the resulting binaries differed greatly from the original; In fact, they did so in the same way, turning into `monoliths' replacing the control flow with one block. Attached are the diffs showing this. I think more time spent here may not be worth it, so I will move on to other more important issues for now.
[DIFFS.txt](https://github.com/user-attachments/files/24991024/DIFFS.txt)

mkannwischer:
> Adding this here but in [#1498](https://github.com/pq-code-package/mlkem-native/pull/1498) I figured this:
> 
> The main issue is that simpasm conflicts with the way MacOS Intel Mach-O files are generated. Essentially, control flow statements like jumps, use relative offsets such as <label+offset> which can not be trivial changed to just . I attempted both replacing the relative offset with a label (by resolving the labels with the linker and just swapping them) and also keeping the relative offset in place. While both version compiled, the resulting binaries differed greatly from the original; In fact, they did so in the same way, turning into `monoliths' replacing the control flow with one block. Attached are the diffs showing this. I think more time spent here may not be worth it, so I will move on to other more important issues for now. [DIFFS.txt](https://github.com/user-attachments/files/24991024/DIFFS.txt)

Thanks @f15hr for looking into this! I agree that it is probably not worth to spend more time on.
After all Apple themselves will stop supporting their Intel machines starting with MacOS 27 (which will be released later this year).

I think we can close this, @hanno-becker? Or are we expecting any developers with Intel-based Macs?

PR #1498, opened 2026-01-19
Fix autogen for macos-15-intel

f15hr:
"Re"introducing MacOS 15 intel runners, which are currently not working according to #1304 because of `autogen`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,714 +9,716 @@ on:
   workflow_dispatch:
 
 jobs:
-  build_kat:
-    strategy:
-      fail-fast: false
-      matrix:
-        external:
-         - ${{ github.repository_owner != 'pq-code-package' }}
-        target:
-         - runner: macos-latest
-           name: 'MacOS (aarch64)'
-           arch: mac
-           mode: native
-           nix_shell: ci
-         - runner: macos-15-intel
-           name: 'MacOS (x86_64)'
-           arch: mac
-           mode: native
-           nix_shell: ci
-         - runner: ubuntu-24.04-arm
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -734,9 +735,10 @@ jobs:
             nix_shell: 'hol_light-cross-x86_64'
             extra_args: '--force-cross'
           # TODO: autogen does not yet work on macos15-intel (#1304)
-          # - system: macos-15-intel
-          #   nix_cache: 'false'
-          #   nix_shell: 'ci'
+          - system: macos-15-intel
+            nix_cache: 'true'
```
mkannwischer: We are not building a nix cache for Intel MacOS - so this is a no-op.

.github/workflows/ci.yml
mkannwischer, line 1: Please trim down the CI while experiemnting, so that only the relevant jobs are run.

f15hr:
I'm going to close this PR for now. The main issue is that `simpasm` conflicts with the way MacOS Intel Mach-O files are generated. Essentially, control flow statements like jumps, use relative offsets such as `<label+offset>` which can not be trivial changed to just `<resolved label>`. I attempted both replacing the relative offset with a label (by resolving the labels with the linker and just swapping them) and also keeping the relative offset in place. While both version compiled, the resulting binaries differed greatly from the original; In fact, they did so in the same way, turning into `monoliths' replacing the control flow with one block. Attached are the diffs showing this. I think more time spent here may not be worth it, so I will move on to other more important issues for now.
[DIFFS.txt](https://github.com/user-attachments/files/24991024/DIFFS.txt)

Issue #1502, opened 2026-01-20
Armv8.1-M: Add PMU benchmarking code


mkannwischer:
Add benchmarking code using the PMU cycle counters.
- Must be functional on MPS3
- Ideally should be tested by the qemu CI tests.

PR #1514, opened 2026-01-21
Add PMU cycle counting for Armv8.1-M 

mkannwischer:
Add PMU-based cycle counting support for Armv8.1-M Cortex-M processors.
This uses the CMSIS PMU APIs for portable cycle counter access.

 - Resolves https://github.com/pq-code-package/mlkem-native/issues/1502

- Hoisted out from https://github.com/pq-code-package/mldsa-native/pull/511

### DIFF
--- a/.github/workflows/baremetal.yml
+++ b/.github/workflows/baremetal.yml
@@ -24,6 +24,7 @@ jobs:
            kat: true
            acvp: true
            alloc: true
+           bench: true
          - runner: ubuntu-latest
            name: 'AVR ATmega128RFR2 (modified for 32K RAM)'
            makefile: test/baremetal/platform/avr/platform.mk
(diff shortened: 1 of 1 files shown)


.github/workflows/baremetal.yml
```diff
@@ -49,3 +51,15 @@ jobs:
...
+        if: ${{ matrix.target.bench }}
+        uses: ./.github/actions/bench
+        env:
+          EXTRA_MAKEFILE: ${{ matrix.target.makefile }}
+        with:
+          name: ${{ matrix.target.name }}
+          nix-shell: ${{ matrix.target.nix-shell }}
+          gh_token: ${{ secrets.GITHUB_TOKEN }}
+          perf: PMU
+          opt: false
```
hanno-becker, line 64: We'll soon want `true` here?

hanno-becker (approved):
I cannot test that the PMU code is correct on HW, but the CI and a local check confirm that it at least builds and runs in QEMU, which is good enough for now.

PR #1515, opened 2026-01-21
x86_64: Remove native poly_decompress_d4/d10 implementations

mkannwischer:
Benchmarking on my laptop shows these AVX2 intrinsics implementations are actually slower than the C fallback code that gcc generates.

This commit removes the two functions.


mkannwischer:
Appears to be highly platform-dependent :(

PR #1521, opened 2026-01-22
Optimized AVX2 batched Keccak_x4

manastasova:
This commit improves the performance of the AVX2 Keccak-F1600x4 implementation by:
 * Replace gather-based loading (vpgatherqq) with explicit loads and transpose operations using unpack and permute instructions.
 * Replace individual scatter stores with batched 4-lane transpose and store operations (MLK_SCATTER_STORE256_4X).
 * Merged prepareTheta into thetaRhoPiChiIota at the start of each round.
 * Eliminated separate E state variables by using temporaries (Tba, etc.) and copying back to A at end of each round, removing A/E alternation.
 * Changed ROUNDS24 from 24 unrolled macro calls to a loop-based MLK_ROUNDS_x2 processing 2 rounds per iteration.

### DIFF
--- a/mlkem/src/fips202/native/x86_64/src/KeccakP_1600_times4_SIMD256.c
+++ b/mlkem/src/fips202/native/x86_64/src/KeccakP_1600_times4_SIMD256.c
@@ -31,7 +31,7 @@ http://creativecommons.org/publicdomain/zero/1.0/
  * - Eliminated separate E state variables by using temporaries (Tba, etc.)
  *   and copying back to A at end of each round, removing A/E alternation.
  * - Changed ROUNDS24 from 24 unrolled macro calls to a loop-based
- *   MLK_ROUNDS_x2 processing 2 rounds per iteration.
+ *   MLK_ROUNDS_X2 processing 2 rounds per iteration.
  */
 
 #include "../../../../common.h"
@@ -347,7 +347,7 @@ static MLK_ALIGN const uint64_t mlk_keccakf1600RoundConstants[24] = {
   } while (0)
 
 /* clang-format off */
-#define MLK_ROUNDS_x2                                               \
+#define MLK_ROUNDS_X2                                               \
       do {                                                          \
         int i;                                                      \
         for (i = 0; i < 12; i++) {                                  \
@@ -362,7 +362,7 @@ void mlk_keccakf1600x4_permute24(void *states)
   __m256i *statesAsLanes = (__m256i *)states;
   MLK_DECLARE_ABCDE;
   MLK_COPY_FROM_STATE(A, statesAsLanes);
-  MLK_ROUNDS_x2;
+  MLK_ROUNDS_X2;
   MLK_COPY_TO_STATE(statesAsLanes, A);
 }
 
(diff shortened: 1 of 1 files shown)


mkannwischer:
Thanks @manastasova - these are very nice performance gains!
I briefly looked at the code and so far it looks good to me - will do a more thorough review later.

Could you elaborate what the further plan is with this code?
I thought the goal was to replace this implementation with assembly so that is can be verified?
What's stopping us from doing that already now with the new tricks you have found?
Apologies if my question was answered in one of the other PRs - I did not follow closely so far.

manastasova:
Thanks for having a look @mkannwischer!
The next steps are to verify the assembly code using HOL Light and replace the intrinsics with inline assembly once the proof is complete. In the meantime, the intrinsics-based implementation can be merged to realize the performance gains.

mlkem/src/fips202/native/x86_64/src/KeccakP_1600_times4_SIMD256.c
```diff
@@ -370,94 +334,36 @@ static MLK_ALIGN const uint64_t mlk_keccakf1600RoundConstants[24] = {
...
 /* clang-format on */
 
 void mlk_keccakf1600x4_permute24(void *states)
 {
   __m256i *statesAsLanes = (__m256i *)states;
-  MLK_DECLARE_ABCDE MLK_COPY_FROM_STATE(A, statesAsLanes)
-      MLK_ROUNDS24 MLK_COPY_TO_STATE(statesAsLanes, A)
+  MLK_DECLARE_ABCDE;
+  MLK_COPY_FROM_STATE(A, statesAsLanes);
+  MLK_ROUNDS_x2;
```
hanno-becker: Nit: The other macros take `A` as an explicit argument; one should keep it uniform.

In fact, it's not clear why this is a macro in the first place (pre-existing, of course).

hanno-becker: Nit: `s/MLK_ROUNDS_x2/MLK_ROUNDS_X2/`

manastasova: `A` is passed as an explicit argument to `MLK_COPY_FROM_STATE(A, statesAsLanes)` and `MLK_COPY_TO_STATE(statesAsLanes, A)` but not to `MLK_ROUNDS24`.

Indeed, it does not have to be a macro, but for consistency (both with pre-existing code and upstream https://github.com/KeccakTeam/KeccakCodePackage/blob/e7a08f7baa3d43d64f5c21e641cb18fe292f2b75/lib/low/KeccakP-1600-times4/AVX2/KeccakP-1600-times4-AVX2.c#L811C5-L811C13) and to facilitate potential future changes to the loop itself, I'd prefer to keep it as a macro for now.

However, I'm happy to inline it if you feel strongly about it.

manastasova: Thank you! Just updated.

hanno-becker:
@manastasova I assume this is no longer relevant? Please re-open if you disagree.

PR #1524, opened 2026-01-23
Armv8.1-M: Add native Keccak x4 XORBytes and ExtractBytes 

### LINKED ISSUE #1507: Armv8.1-M: Add native bitinterleaving x4

mkannwischer:
See individual commits

- Resolves #1507

### DIFF
--- a/dev/fips202/armv81m/src/state_extract_bytes_x4_mve.S
+++ b/dev/fips202/armv81m/src/state_extract_bytes_x4_mve.S
@@ -1,0 +1,333 @@
+/*
+ * Copyright (c) The mlkem-native project authors
+ * Copyright (c) The mldsa-native project authors
+ * Copyright (c) 2026 Arm Limited
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+ */
+
+// ---------------------------------------------------------------------------
+// Overview
+// ---------------------------------------------------------------------------
+// MVE/Helium implementation of KeccakF1600x4_StateExtractBytes
+// (inverse of state_xor_bytes_x4_mve.S).
+//
+// void KeccakF1600x4_StateExtractBytes(state, d0, d1, d2, d3, offset, length)
+//
+// Reads 'length' bytes from the bit-interleaved Keccak state starting at
+// byte 'offset', recombines the even and odd halves of each lane back
+// into plain bytes, and writes them to four output buffers (d0..d3).
+//
+// ---------------------------------------------------------------------------
+// Bit-interleaving background
+// ---------------------------------------------------------------------------
+// Each 64-bit Keccak lane is stored as two 32-bit words:
+//   even half -- bits 0, 2, 4, ..., 62 of the lane
+//   odd half  -- bits 1, 3, 5, ..., 63 of the lane
+// This representation allows 64-bit lane rotations (used in the Keccak
+// round function) to be implemented as pairs of 32-bit rotations.
+//
+// Batched (x4) processing:
+//   Four Keccak instances are processed as a batch.  Their states are
+//   stored interleaved in a single 800-byte buffer: first the even
+//   halves of all 25 lanes (400 bytes), then the odd halves (400 bytes).
+//   Within each 16-byte row, the four u32 words correspond to
+//   instances 0..3 of the same lane, enabling SIMD-parallel operations
+//   across all four instances.
+//
+// State memory layout (25 lanes x 4 instances x 2 halves):
+//   S[i][l]_even/odd = even/odd half of lane l, instance i  (u32)
+//   Each row is 16 bytes (one Q-register).
+//   Offset  Contents
+//     0     S[0][ 0]_even, S[1][ 0]_even, S[2][ 0]_even, S[3][ 0]_even
+//    16     S[0][ 1]_even, S[1][ 1]_even, S[2][ 1]_even, S[3][ 1]_even
+//    ...
+//   384     S[0][24]_even, S[1][24]_even, S[2][24]_even, S[3][24]_even
+//   400     S[0][ 0]_odd,  S[1][ 0]_odd,  S[2][ 0]_odd,  S[3][ 0]_odd
+//   416     S[0][ 1]_odd,  S[1][ 1]_odd,  S[2][ 1]_odd,  S[3][ 1]_odd
+//    ...
+//   784     S[0][24]_odd,  S[1][24]_odd,  S[2][24]_odd,  S[3][24]_odd
+//
+// ---------------------------------------------------------------------------
+// Three-phase structure
+// ---------------------------------------------------------------------------
+//   Prologue -- if offset is not 8-byte aligned, extract
+//               min(length, 8-(offset%8)) bytes via predicated byte stores.
+//   Main     -- process full 8-byte groups: load even/odd lane pair,
+//               de-interleave, scatter-store to output buffers.
+//   Tail     -- extract remaining <8 bytes via predicated byte stores.
+
+#include "../../../../common.h"
+#if defined(MLK_FIPS202_ARMV81M_NEED_X4) && \
+    !defined(MLK_CONFIG_MULTILEVEL_NO_SHARED)
+/* simpasm: header-end */
+
+.thumb
+.syntax unified
+.text
+
+// ---------------------------------------------------------------------------
+// deinterleave_even: inverse of the even-bit compaction.  Inflates the even
+// half stored in \e back into byte positions.
+// Inputs:  \e (compacted even bits)
+// Outputs: \e (even bits expanded to byte positions, odd bits are garbage)
+// Clobbers: \tmp
+// ---------------------------------------------------------------------------
+.macro deinterleave_even e, tmp
+    // Inflate e                              +------+-----------+-----------+-----------+
+    // Single-element annotation follows.  e: |   X  |     X     | l[0],l[1] | l[2],l[3] |
+    vsli.u32    \e, \e, #8              // e: |   X  | l[0],l[1] | l[2],l[3] | l[2],l[3] |
+    vsli.u16    \e, \e, #4              // e: | l[0] | l[1],l[1] | l[3],l[2] | l[3],l[3] |
+    //                                        +------+-----------+-----------+-----------+
+    // Now, expand the lower bits  e: | XXXX3210 | tmp: | XXXXXXXX |
+    vsli.u8     \e, \e, #1    //   e: | XXX32100 | tmp: | XXXXXXXX |  (start re-packing low bits)
+    vshr.u8     \tmp, \e, #3  //   e: | XXX32100 | tmp: | XXXXXX32 |
+    vsli.u8     \e, \tmp, #4  //   e: | XX322100 | tmp: | XXXXXX32 |  (assemble nibbles)
+    vshr.u8     \tmp, \e, #5  //   e: | XX322100 | tmp: | XXXXXXX3 |
+    vsli.u8     \e, \tmp, #6  //   e: | X3322100 | tmp: | XXXXXXX3 |  (finalize byte compaction)
+.endm
+
+// ---------------------------------------------------------------------------
+// from_bit_interleaving_4x: reconstruct byte vectors from bit-interleaved
+// even/odd halves.
+// Inputs:  \qe = even half, \qo = odd half
+//          (per 32b lane: lo16 = bytes 0..3, hi16 = bytes 4..7)
+// Outputs: \qe = [d0l, d1l, d2l, d3l] (low  4 bytes per instance)
+//          \qo = [d0h, d1h, d2h, d3h] (high 4 bytes per instance)
+// Clobbers: \rtmp, \qt0, \qt1, \qt2
+// ---------------------------------------------------------------------------
+.macro from_bit_interleaving_4x qe, qo, qt0, qt1, qt2, rtmp
+    //     +------+------+------+------+------+------+------+------+
+    // qe: | E0l  | E0u  | E1l  | E1u  | E2l  | E2u  | E3l  | E3u  |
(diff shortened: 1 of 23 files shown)


mlkem/src/fips202/native/api.h
```diff
@@ -55,4 +55,49 @@ __contract__(
...
+ * XORing input data into the state and extracting output data from the state.
+ * This is particularly useful for backends that use a different internal state
+ * representation (e.g., bit-interleaved), as conversion can happen during
+ * XOR/extract rather than before/after each permutation.
+ *
+ * MLK_USE_FIPS202_X4_XOR_BYTES_NATIVE: Backend provides native XOR bytes
+ * MLK_USE_FIPS202_X4_EXTRACT_BYTES_NATIVE: Backend provides native extract
+ * bytes
+ *
+ * If MLK_FIPS202_X4_NATIVE_CUSTOM_STATE_REPRESENTATION is set, both XOR and
```
hanno-becker: My understanding is that this function is implicitly assumed to preserve the 0 state, and is therefore not used in the FIPS202 implementation at all. This should be documented.

It is only used in the unit test for greater coverage.

We should consider one of the following:
* Explicitly mention that this is only relevant for testing, and that it may hence be implemented in a non-performant way. In particular, we don't have to implement it in assembly either.
* Remove it altogether and change the unit test for Keccak-F1600 by randomizing the input through randomized XOR bytes.

mkannwischer: As discussed offline, the test was changed to test xor_bytes + permute + extract_bytes. This works around the custom state representation not allowing to test the native function.
The `MLK_FIPS202_X4_NATIVE_CUSTOM_STATE_REPRESENTATION` is no longer needed.

mlkem/src/fips202/native/api.h
```diff
@@ -55,4 +55,35 @@ __contract__(
...
 #endif /* MLK_USE_FIPS202_X4_NATIVE */
 
+/*
+ * Native x4 XOR bytes and extract bytes interface.
+ *
+ * These functions allow backends to provide optimized implementations for
+ * XORing input data into the state and extracting output data from the state.
+ * This is particularly useful for backends that use a different internal state
+ * representation (e.g., bit-interleaved), as conversion can happen during
+ * XOR/extract rather than before/after each permutation.
```
hanno-becker, line 64: Nit: It should still be mentioned that we assume that the custom presentation of the 0 state is the 0 state.

dev/fips202/armv81m/src/state_xor_bytes_x4_mve.S
```diff
@@ -0,0 +1,189 @@
...
+    @ decrement length
+    le  lr, keccak_f1600_x4_state_xor_bytes_aligned_loop_start
+keccak_f1600_x4_state_xor_bytes_aligned_loop_end:
+    vadd.u32 q0, q4, r3
+    vpop {d8 - d15}
+    pop {r4, pc}
+
+@ ----------------------------------------------------------------------------
+@ Local subroutine: to_bit_interleaving_4x
+@ TODO: This duplication should be eliminated
```
hanno-becker: What's going to happen with this?

dev/fips202/armv81m/src/state_xor_bytes_x4_mve.S
```diff
@@ -0,0 +1,189 @@
...
+        vsri.u8 q7, q3, #1          // ..........*....................................................................................
+        vshr.u8 q1, q2, #2          // ............*..................................................................................
+        vsli.u8 q2, q1, #1          // ..............*................................................................................
+        vshl.u8 q5, q6, #2          // ................*..............................................................................
+        vsri.u8 q6, q5, #1          // ..................*............................................................................
+        vshr.u8 q4, q0, #2          // ....................*..........................................................................
+        vsli.u8 q0, q4, #1          // ......................*........................................................................
+        vshl.u8 q5, q7, #3          // ........................*......................................................................
```
hanno-becker: Note that you can trade some shifts for MULs for (much) better pipelining.

hanno-becker: (A right shift can be expressed as a high mul)

dev/fips202/armv81m/src/state_xor_bytes_x4_mve.S
mkannwischer: Before we merge I would like KeccakF1600x4_StateXORBytes and KeccakF1600x4_StateExtract_bytes fully in assembly instead of split between C and multiple asm functions. 
Once that is done the duplication will be gone.

mlkem/src/fips202/native/auto.h
```diff
@@ -19,10 +19,9 @@
...
 
-/* We do not yet include the FIPS202 backend for Armv8.1-M+MVE by default
- * as it is still experimental and undergoing review. */
-/* #if defined(MLK_SYS_ARMV81M_MVE) */
-/* #include "armv81m/mve.h" */
-/* #endif */
+/* Enable the Armv8.1-M + MVE FIPS202 backend when available. */
+#if defined(MLK_SYS_ARMV81M_MVE)
+#include "armv81m/mve.h"
+#endif
```
mkannwischer: This should not be changed.

mlkem/src/fips202/native/armv81m/src/keccak_f1600_x4_mve.c
```diff
@@ -9,117 +10,24 @@
 #if defined(MLK_FIPS202_ARMV81M_NEED_X4) && \
     !defined(MLK_CONFIG_MULTILEVEL_NO_SHARED)
 
+#include <arm_mve.h>
+#include <stddef.h>
```
mkannwischer: Should not be required.

dev/fips202/armv81m/src/state_xor_bytes_x4_mve.S
mkannwischer: Eliminated now.

dev/fips202/armv81m/src/state_xor_bytes_x4_mve.S
mkannwischer: @bremoran, I'll leave it to you if you want to look into this now, or if we open an issue for this to investigate it later.

mlkem/src/fips202/native/api.h
mkannwischer: I added a comment about that.

mlkem/src/fips202/native/auto.h
mkannwischer: reverted to the original state.

mlkem/src/fips202/native/armv81m/src/keccak_f1600_x4_mve.c
mkannwischer: removed.

dev/fips202/armv81m/src/state_xor_bytes_x4_mve.S
mkannwischer: I opened https://github.com/pq-code-package/mlkem-native/issues/1532 so we don't forget about this.

mkannwischer:
I have opened https://github.com/pq-code-package/mlkem-native/issues/1531 to remind us that we still have to superoptimize this code.

mlkem/src/fips202/keccakf1600.c
```diff
@@ -82,6 +82,10 @@ void mlk_keccakf1600x4_extract_bytes(uint64_t *state, unsigned char *data0,
                                      unsigned char *data3, unsigned offset,
                                      unsigned length)
 {
+#if defined(MLK_USE_FIPS202_X4_EXTRACT_BYTES_NATIVE)
+  mlk_keccakf1600_extract_bytes_x4_native(state, data0, data1, data2, data3,
+                                          offset, length);
+#else
```
hanno-becker: Can we follow the existing pattern here? 
* Add `foo_c` with the C version
* Have `foo` dispatch between `foo_native` and `foo_c`
* Add error code to `foo_c`

mkannwischer: Good point. Added.
Also added the corresponding _native CBMC proofs.

hanno-becker (changes requested):
Following other backend functions, could we have a separate CBMC proof for `foo_c`, and then the `foo` and `foo_native` call `foo_c` by contract? AFAIU, we currently still inline `foo_c` into the proofs of `foo` and `foo_native`.

mlkem/src/fips202/keccakf1600.c
hanno-becker: Thank you! I think we're still missing a separate proof for `_c` and for the `extract/xor_bytes` and `exctract/xor_bytes_native` proofs to call `_c` by contract.

mkannwischer:
> Following other backend functions, could we have a separate CBMC proof for `foo_c`, and then the `foo` and `foo_native` call `foo_c` by contract? AFAIU, we currently still inline `foo_c` into the proofs of `foo` and `foo_native`.

I don't see a separate proof for `mlk_keccakf1600_permute_c`. I followed the same approach.

hanno-becker:
> I don't see a separate proof for mlk_keccakf1600_permute_c. I followed the same approach.

Ok, we can do this as a follow-up. This seems to be inconsistent between arithmetic and FIPS backend.

mlkem/src/fips202/keccakf1600.c
hanno-becker: Agreed offline to do this in a follow-up.

dev/fips202/armv81m/src/state_extract_bytes_x4_mve.S
```diff
@@ -0,0 +1,333 @@
...
+// ---------------------------------------------------------------------------
+// from_bit_interleaving_4x: reconstruct byte vectors from bit-interleaved
+// even/odd halves.
+// Inputs:  q0 = even half, q1 = odd half
+//          (per 32b lane: lo16 = bytes 0..3, hi16 = bytes 4..7)
+// Outputs: q0 = [d0l, d1l, d2l, d3l] (low  4 bytes per instance)
+//          q1 = [d0h, d1h, d2h, d3h] (high 4 bytes per instance)
+// Clobbers: r0, q2, q3, q4
+// ---------------------------------------------------------------------------
+.macro from_bit_interleaving_4x
```
hanno-becker: Similar to `deinterleave_even`, it would be more robust and readable if the input/output and temporaries in `from_bit_interleaving_4x` were named macro arguments.

mlkem/src/fips202/native/armv81m/src/state_extract_bytes_x4_mve.S
```diff
@@ -0,0 +1,290 @@
...
+//   384     S[0][24]_even, S[1][24]_even, S[2][24]_even, S[3][24]_even
+//   400     S[0][ 0]_odd,  S[1][ 0]_odd,  S[2][ 0]_odd,  S[3][ 0]_odd
+//   416     S[0][ 1]_odd,  S[1][ 1]_odd,  S[2][ 1]_odd,  S[3][ 1]_odd
+//    ...
+//   784     S[0][24]_odd,  S[1][24]_odd,  S[2][24]_odd,  S[3][24]_odd
+//
+// ---------------------------------------------------------------------------
+// Three-phase structure
+// ---------------------------------------------------------------------------
+//   Prologue -- if offset is not 8-byte aligned, extract
```
hanno-becker, line 53: Do we even need `offset`? I may be overlooking something, but it seems to be `0` at all call-sites.

hanno-becker: We're not using it in mlkem/mldsa-native, but an incremental squeeze would need it. So, for the sake of keeping doors open for using the code in other contexts, it makes sense to keep it.

dev/fips202/armv81m/src/state_extract_bytes_x4_mve.S
```diff
@@ -0,0 +1,333 @@
...
+keccak_f1600_x4_state_extract_bytes_asm_main_loop_start:
+    vldrw.u32 qd0, [rSE], #16
+    vldrw.u32 qd1, [rSO], #16
+
+    // De-interleave (clobbers r0, q2, q3, q4)
+    from_bit_interleaving_4x
+
+    // Scatter-store 8 bytes per instance (two u32 stores with post-increment)
+    vstrw.u32   qd0, [qP, #4]!
+    vstrw.u32   qd1, [qP, #4]!
```
hanno-becker, line 275: The loop invariant is slightly awkward here, seeing that `qP` is 4 bytes _below_ the current data. But it seems that unless we want to add an explicit `vadd` for `qP`, there is no other way with the (lack of) variants of scatter-`vstrw.u32` available.

dev/fips202/armv81m/src/state_extract_bytes_x4_mve.S
mkannwischer: Changed here: https://github.com/pq-code-package/mlkem-native/compare/f5de3c946af9d3419825a52bae0bc3e579fede5f..07baf7d75d5046afb1331f0db3fe92a89ac628e6

mkannwischer:
> Apologies for the long delay, @bremoran @mkannwischer. I wanted to look at this properly and couldn't find time until now.
> 
> I have reviewed the assembly and am very happy with it. I applaud you for the clarity and documentation, which is great and will simplify any further work in the future.
> 
> As mentioned, I think some uniformity regarding the presence/absence of macro arguments would be an improvement, but I leave it to you whether you want to do this as a follow-up or not.

Thanks @hanno-becker  - I agree that adding the macro arguments explicilty makes it easier to read. I added it in https://github.com/pq-code-package/mlkem-native/compare/f5de3c946af9d3419825a52bae0bc3e579fede5f..07baf7d75d5046afb1331f0db3fe92a89ac628e6.

PR #1525, opened 2026-01-27
Add rejection key comparision for negative tests

abhi-dev-engg:
This PR compares the key generated during " corrupted cipher-text" test matches the key as prescribed in the FIPS-203 Algorithm spec


mkannwischer:
> This PR compares the key generated during " corrupted cipher-text" test matches the key as prescribed in the FIPS-203 Algorithm spec

Thanks!
The ACVP tests already have test coverage for the implicit rejection as well. I'm not sure if we need to specifically add this to the functional tests.

abhi-dev-engg:
> > This PR compares the key generated during " corrupted cipher-text" test matches the key as prescribed in the FIPS-203 Algorithm spec
> 
> Thanks! The ACVP tests already have test coverage for the implicit rejection as well. I'm not sure if we need to specifically add this to the functional tests.

This was added because we were validating negative test cases by comparing the encapsulated secret with the decapsulated secret to ensure they were different; however, we were not verifying whether the decapsulated key in this scenario matched the expected value as per FIPS-203.

please feel free to close if you feel is not needed.

mkannwischer:
> > > This PR compares the key generated during " corrupted cipher-text" test matches the key as prescribed in the FIPS-203 Algorithm spec
> > 
> > 
> > Thanks! The ACVP tests already have test coverage for the implicit rejection as well. I'm not sure if we need to specifically add this to the functional tests.
> 
> This was added because we were validating negative test cases by comparing the encapsulated secret with the decapsulated secret to ensure they were different; however, we were not verifying whether the decapsulated key in this scenario matched the expected value as per FIPS-203.

You are right that the functional test `test_mlkem.c` did not validate that behaviour - it's specific to FIPS203, while `test_mlkem.c` was kept general and would apply to any KEM (it's up for discussion if we want to keep it that way or not). The ACVP tests, however, do validate exactly the FIPS203 behaviour for this case, i.e., making sure that in the case of rejection, the key matches the expectation.
My comment was just to clarify that this is not a test gap, it was just tested in another place. Whether additional test coverage would be useful or not is something we can discuss.

abhi-dev-engg:
> 
> You are right that the functional test `test_mlkem.c` did not validate that behaviour - it's specific to FIPS203, while `test_mlkem.c` was kept general and would apply to any KEM (it's up for discussion if we want to keep it that way or not). The ACVP tests, however, do validate exactly the FIPS203 behaviour for this case, i.e., making sure that in the case of rejection, the key matches the expectation. My comment was just to clarify that this is not a test gap, it was just tested in another place. Whether additional test coverage would be useful or not is something we can discuss.

Agree completely, again I leave the decision upto you regarding the PR :)

abhi-dev-engg:
closing this PR as it seems there is no hope of it getting merged

mkannwischer:
> closing this PR as it seems there is no hope of it getting merged

I would not say that there is no hope for this to get merged. You can keep this open if you want and we'll get to it later when we have time. We are just busy right now with other things higher on the priority list. Apologies.

abhi-dev-engg:
> > closing this PR as it seems there is no hope of it getting merged
> 
> I would not say that there is no hope for this to get merged. You can keep this open if you want and we'll get to it later when we have time. We are just busy right now with other things higher on the priority list. Apologies.

no worries at all. I have reopened the PR

PR #1527, opened 2026-01-28
Armv8.1-M: Add clean x4 Keccak code and SLOTHY Makefile

### LINKED ISSUE #1510: Armv8.1-M: Add Keccak x4 assembly and SLOTHY Makefile

mkannwischer:
Add the clean (non-optimized) Keccak x4 assembly and Makefile to enable regeneration of the SLOTHY-optimized assembly.

Extend autogen --slothy to support new function, and also add --slothy armv81m and --slothy aarch64 group options. Update CI to only run aarch64 SLOTHY targets.

Optimization of Armv8.1-M assembly is not currently tested in CI, but has been tested locally.

Update SLOTHY to version 0.1.10 as the code requires disabling the address offset fixup which is not supported for Armv8.1-M in the older versionf of SLOTHY.

 - Resolves #1510

### DIFF
--- a/dev/fips202/armv81m_symbolic/keccak_f1600_x4_mve_clean.S
+++ b/dev/fips202/armv81m_symbolic/keccak_f1600_x4_mve_clean.S
@@ -1,0 +1,481 @@
+/*
+ * Copyright (c) The mlkem-native project authors
+ * Copyright (c) 2025 Arm Limited
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+ */
+
+/*yaml
+  Name: keccak_f1600_x4_mve_asm
+  Description: Armv8.1-M MVE implementation of 4-way parallel Keccak-f[1600] permutation using bit-interleaved state
+  Signature: void mlk_keccak_f1600_x4_mve_asm(void *state, void *tmpstate, const uint32_t *rc)
+  ABI:
+    r0:
+      type: buffer
+      size_bytes: 800
+      permissions: read/write
+      c_parameter: void *state
+      description: Four bit-interleaved Keccak states (low halves followed by high halves)
+    r1:
+      type: buffer
+      size_bytes: 800
+      permissions: read/write
+      c_parameter: void *tmpstate
+      description: Temporary storage for intermediate state
+    r2:
+      type: buffer
+      size_bytes: 192
+      permissions: read
+      c_parameter: const uint32_t *rc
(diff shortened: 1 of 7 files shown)


dev/fips202/armv81m_symbolic/keccak_f1600_x4_mve_clean.S
```diff
@@ -0,0 +1,481 @@
...
+    vstrw.32 qA00_h, [\state_h, #A__00] // @slothy:writes=A\state_h\()__00
+.endm
+
+.text
+.balign 8
+.type MLK_ASM_NAMESPACE(keccak_f1600_x4_mve_asm), %function
+.global MLK_ASM_NAMESPACE(keccak_f1600_x4_mve_asm)
+MLK_ASM_FN_SYMBOL(keccak_f1600_x4_mve_asm)
+
+    push {r3,r4,r5,r6,r7,r8,r9,r10,r11,r12,lr}
```
mkannwischer: This actually got changed in https://github.com/pq-code-package/mlkem-native/pull/1558 - we should align that once it gets merged.

nix/slothy/default.nix
```diff
@@ -17,12 +17,12 @@ let
 in
 stdenvNoCC.mkDerivation rec {
   pname = "slothy-cli";
-  version = "915c224166207ce07b31152194305c3b6687d09b";
+  version = "0.1.10";
```
hanno-becker, line 20: Is this still the right version to update to?

scripts/autogen
```diff
@@ -4175,6 +4177,11 @@ def _main():
         "keccak_f1600_x4_v8a_scalar_hybrid_asm",
         "keccak_f1600_x4_v8a_v84a_scalar_hybrid_asm",
     ]
+    slothy_armv81m_targets = [
+        "keccak_f1600_x4_mve",
+    ]
+    slothy_all_targets = slothy_aarch64_targets + slothy_armv81m_targets
+    slothy_choices = slothy_all_targets + ["aarch64", "armv81m"]
```
hanno-becker, line 4184: Can we explicitly prefix the targets with the architecture? E.g. `--slothy aarch64/* mve/*`?

nix/slothy/default.nix
mkannwischer: No. we should take 0.2.2

hanno-becker (changes requested):
We need a CI job which re-runs SLOTHY on the MVE kernel and _then_ runs the baremetal tests.

PR #1535, opened 2026-01-31
CI: Bump AWS-LC to v1.67.0

mkannwischer:
This commit also
 - Stops testing against the main branch of AWS-LC as that so far has not revealed any interesting problems and risks that breaking changes in AWS-LC break our CI; We should instead update AWS-LC more frequently
 - Fixes a typo in integration-awslc.yml matrix.flags ->  matrix.test.flags
 - Removes a temporary workaround in the setup action that is no longer needed

### DIFF
--- a/.github/workflows/integration-awslc-main.yml
+++ b/.github/workflows/integration-awslc-main.yml
@@ -1,0 +1,22 @@
+# Copyright (c) The mlkem-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+name: AWS-LC (HEAD)
+permissions:
+  contents: read
+on:
+  schedule:
+    # Run daily at 04:00 UTC
+    - cron: '0 4 * * *'
+  workflow_dispatch:
+
+jobs:
+  awslc_head:
+    name: AWS-LC (HEAD)
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    uses: ./.github/workflows/integration-awslc.yml
+    with:
+      commit: main
+    secrets: inherit


.github/workflows/all.yml
```diff
@@ -67,25 +67,15 @@ jobs:
...
-    secrets: inherit
-  awslc_integration_head:
-    name: AWS-LC (HEAD)
-    permissions:
-      contents: 'read'
-      id-token: 'write'
-    needs: [ base ]
-    uses: ./.github/workflows/integration-awslc.yml
-    with:
-      commit: main
```
hanno-becker, line 88: I am a bit wary that this delays us noticing breaking changes in AWS-LC. Just because it hasn't happened doesn't mean it won't.

At the same time, I think we need to be pragmatic about where we spend our CI resources.

What about keeping the `main` check but only running the `aws_lc_integration_fips` job?

mkannwischer: I think running tests against AWS-LC `main` in every PR is the wrong approach.
In the long term, we don't want AWS-LC breakages to block important other PRs into the main branch.

If you want to detect breakage due to recent changes in AWS-LC, I think the correct approach would be a scheduled job (e.g., daily) that tests mlkem-native main with AWS-LC main. We just need to make sure that we get notfications when it fails.

Would that work? Should be straightforward to set that up with our current CI.

mkannwischer: I added a commit implementing that.

hanno-becker: Thanks!

hanno-becker (approved):
Thanks @mkannwischer !

Issue #1523, opened 2026-01-23
Consolidate backend `MLK_USE_XXX` flags indicating function support


mkannwischer:
In the arithmetic backend, we signal support for certain functions using macros of the form
`MLK_USE_NATIVE_NTT`, while in the FIPS202 backend, we use `MLK_USE_FIPS202_X1_NATIVE`

We should consolidate that to `MLK_USE_NATIVE_XXX`

PR #1541, opened 2026-02-02
Consolidate MLK_USE_NATIVE_XXX macros

mkannwischer:
In the arithmetic backend, we signal support for certain functions using macros of the form MLK_USE_NATIVE_XXX (e.g., MLK_USE_NATIVE_NTT), while in the FIPS202 backend, we use MLK_USE_XXX_NATIVE (e.g., MLK_USE_FIPS202_X1_NATIVE). This commit consolidates it to MLK_USE_NATIVE_XXX.

- Resolves https://github.com/pq-code-package/mlkem-native/issues/1523


hanno-becker (changes requested):
I support this, but it is an API break, so we need a fall-back path, ideally with a warning being emitted when the now-deprecated defines are used.

mkannwischer:
Closing this as I don't care enough about it.

PR #1542, opened 2026-02-02
WIP: HOL-Light: Prove correctness of AVX2 `poly_decompress_d10`

mkannwischer:


mkannwischer:
> Please can we try using the same structure as for the other compression routines, i.e. a uniform macro-based loop?

Done.

PR #1544, opened 2026-02-04
AArch64: Use alignment-safe Neon loads/stores and test in CI

mkannwischer:
This PR fixes alignment issues in AArch64 Neon assembly that cause faults on bare-metal systems without an MMU or when alignment checking is enabled.

On AArch64, certain Neon load/store instructions like ldp/stp with Q-registers require 16-byte alignment. While Linux and most operating systems handle unaligned accesses transparently via the MMU, bare-metal environments running on Device memory fault immediately on misaligned accesses.

The batched Keccak implementations process two sequential 200-byte states. The first state is 16-byte aligned, but the second starts at offset 200 bytes, which is only 8-byte aligned. The ldp/stp instructions for Q-registers on this second pointer caused alignment faults. Similar issues existed in other assembly routines using unaligned addresses.

The fix replaces alignment-requiring instructions with their alignment-safe equivalents (ld1/st1), which have no alignment requirement and work correctly on both normal and Device memory.

This PR also adds a bare-metal AArch64-virt QEMU target that runs tests without an MMU, catching any remaining unaligned Neon accesses.

Thanks to @flynd who has reported this issue in https://github.com/pq-code-package/mldsa-native/pull/934.


proofs/hol_light/aarch64/proofs/keccak_f1600_x2_v84a.ml
```diff
@@ -255,7 +256,7 @@ let KECCAK_F1600_X2_V84A_CORRECT = prove
   ENSURES_WHILE_UP_TAC `24` `pc + 0xb8` `pc + 0x1c8`
    `\i s.
       wordlist_from_memory(rc,24) s = round_constants /\
-      read X0 s = a /\
+      read X0 s = word_add a (word 192) /\
```
hanno-becker, line 259: Does this even matter?

.github/workflows/baremetal.yml
```diff
@@ -66,3 +66,34 @@ jobs:
...
+        uses: ./.github/actions/functest
+        env:
+          EXTRA_MAKEFILE: test/baremetal/platform/aarch64-virt/platform.mk
+        with:
+          nix-shell: cross-aarch64-embedded
+          gh_token: ${{ secrets.GITHUB_TOKEN }}
+          opt: opt
+          func: true
+          kat: false
+          acvp: false
```
hanno-becker: Why are we not running those tests?

proofs/hol_light/aarch64/proofs/keccak_f1600_x2_v84a.ml
mkannwischer: This is definitely needed. I tried removing the invariant, but it breaks.
I do not understand `ENSURES_WHILE_UP_TAC` enough to know if there is an easier way to express that this loop does not change `x0`.

mkannwischer:
> Thank you @mkannwischer, this looks great!
> 
> I noticed that the KAT tests don't work because the exec wrapper swallows the QEMU semihosting output, which appears to go to `stderr` by default. I think the exec wrapper needs to forward stderr or consolidate stdout + stderr.

Very good point. I fixed that by switching to `stderr`.
`ACVP` tests were also not working because the input arguments were not passed correctly. I also fixed that.
You can see the diff before rebase on top of main [here](https://github.com/pq-code-package/mlkem-native/compare/49f425a71f8f59949ad67ff2849e33257b3d0c7f..5a20d36e7095b75514ccc54579cc134257752fc1).

.github/workflows/baremetal.yml
mkannwischer: Fixed.

PR #1547, opened 2026-02-05
Create Nucleo N657x0-Q Platform

bremoran:
This change introduces a new bare-metal test platform for the Nucleo N657x0-Q board and integrates it into the project’s Nix devshell and test build system.

## Summary
- Adds a dedicated Nix devshell target for `nucleo-n657x0-q`.
- Introduces a full bare-metal platform under `test/baremetal/platform/nucleo-n657x0-q/`, including build rules, execution wrapper, argument packing utility, and platform sources.
- Updates HAL and test make components to integrate the new platform.

## Changes
- `flake.nix`: Add `.#nucleo-n657x0-q` devshell.
- `nix/nucleo-n657x0-q/default.nix`: Define board devshell and tooling.
- `nix/util.nix`: Utility update for new platform.
- `test/baremetal/platform/nucleo-n657x0-q/README.md`: Platform documentation.
- `test/baremetal/platform/nucleo-n657x0-q/platform.mk`: Makefile integration for the platform.
- `test/baremetal/platform/nucleo-n657x0-q/exec_wrapper.py`: Execution wrapper for running tests on target.
- `test/baremetal/platform/nucleo-n657x0-q/make_argv_bin.py`: Helper to pack argv for target.
- `test/baremetal/platform/nucleo-n657x0-q/src/*.c`, `*.h`: Platform sources (cmdline, semihosting syscall, helpers).
- `test/hal/hal.c`, `test/hal/pmu_armv8.h`: Minor HAL updates for compatibility.
- `test/mk/components.mk`: Include the new platform components.

## Why
- Enables running and validating ML-KEM tests on the Nucleo N657x0-Q board.
- Establishes a consistent devshell and build workflow for this platform.

## Usage
- Devshell: `nix develop .#nucleo-n657x0-q`
- Build tests: `make test EXTRA_MAKEFILE=test/baremetal/platform/nucleo-n657x0-q/platform.mk`

## Notes
- No breaking changes expected to other platforms.
- Documentation included under the platform’s `README.md`.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -36,6 +36,8 @@ jobs:
           ldflags: "-flto"
           bench_extra_args: ""
           nix_shell: bench
+          extra_makefile: ""
+          zephyr_target: ""
         - system: rpi5
           name: Arm Cortex-A76 (Raspberry Pi 5) benchmarks
           bench_pmu: PERF
@@ -45,6 +47,8 @@ jobs:
           bench_extra_args: ""
           nix_shell: bench
           cross_prefix: ""
+          extra_makefile: ""
+          zephyr_target: ""
         - system: a55
           name: Arm Cortex-A55 (Snapdragon 888) benchmarks
           bench_pmu: PERF
@@ -53,6 +57,8 @@ jobs:
           ldflags: "-flto -static"
           bench_extra_args: -w exec-on-a55
           nix_shell: bench
+          extra_makefile: ""
+          zephyr_target: ""
         - system: bpi
           name: SpacemiT K1 8 (Banana Pi F3) benchmarks
           bench_pmu: PERF
@@ -62,6 +68,8 @@ jobs:
           bench_extra_args: -w exec-on-bpi
           cross_prefix: riscv64-unknown-linux-gnu-
           nix_shell: cross-riscv64
+          extra_makefile: ""
+          zephyr_target: ""
         - system: m1-mac-mini
           name: Mac Mini (M1, 2020) benchmarks
           bench_pmu: MAC
@@ -70,6 +78,8 @@ jobs:
           ldflags: "-flto"
           bench_extra_args: "-r"
           nix_shell: bench
+          extra_makefile: ""
+          zephyr_target: ""
         - system: pqcp-ppc64
           name:  ppc64le (POWER10) benchmarks
           bench_pmu: PERF
@@ -79,11 +89,27 @@ jobs:
           bench_extra_args: "-r"
           nix_shell: ''
           cross_prefix: ""
+          extra_makefile: ""
+          zephyr_target: ""
+        - system: nucleo-n657x0
+          name: Arm Cortex-M55 (NUCLEO-N657X0-Q) benchmarks
+          bench_pmu: NO
+          archflags: ""
+          cflags: ""
+          ldflags: ""
+          bench_extra_args: ""
+          nix_shell: zephyr
+          cross_prefix: ""
+          extra_makefile: test/zephyr/platform.mk
+          zephyr_target: nucleo-n657x0-q
     if: github.repository_owner == 'pq-code-package' && !github.event.pull_request.head.repo.fork && (github.event.label.name == 'benchmark' || github.ref == 'refs/heads/main')
     runs-on: self-hosted-${{ matrix.target.system }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - uses: ./.github/actions/bench
+        env:
+          EXTRA_MAKEFILE: ${{ matrix.target.extra_makefile }}
+          ZEPHYR_TARGET: ${{ matrix.target.zephyr_target }}
         with:
           name: ${{ matrix.target.name }}
           cflags: ${{ matrix.target.cflags }}
--- a/scripts/tests
+++ b/scripts/tests
@@ -762,8 +762,11 @@ class Tests:
                     test_type, self.do_opt(), suppress_output=False
                 )
 
-        if resultss is None:
+        # Nothing below should write benchmark output unless the run produced
+        # results and all compile/run steps succeeded.
+        if resultss is None or len(self.failed) > 0:
             self.check_fail()
+            return
 
         # NOTE: There will only be one items in resultss, as we haven't yet decided how to write both opt/no-opt benchmark results
         for k, results in resultss.items():
--- a/test/hal/hal.c
+++ b/test/hal/hal.c
@@ -124,8 +124,8 @@ uint64_t get_cyclecounter(void) { return DWT->CYCCNT; }
 #elif defined(ARMCM55)
 /* Cortex-M55: Use dedicated PMU */
 #include <ARMCM55.h>
+#include <pmu_armv8.h>
 #include <system_ARMCM55.h>
-#include "pmu_armv8.h"
 
 void enable_cyclecounter(void)
 {
--- a/test/mk/components.mk
+++ b/test/mk/components.mk
@@ -155,6 +155,9 @@ $(call MAKE_OBJS, $(MLKEM1024_DIR), $(EXTRA_SOURCES)): CFLAGS += $(EXTRA_SOURCES
 endif
 else
 $(ALL_BINS): $$(TEST_SRCS) $(LIB_SRCS)
+# Extra per-binary prerequisites a custom-build platform needs (e.g. Zephyr's
+# app inputs and active-target marker, set in test/zephyr/platform.mk).
+$(ALL_BINS): $(CUSTOM_BUILD_DEPS)
 endif
 
 # ABI checker
--- a/test/zephyr/app/shim_nucleo_n657x0_q.c
+++ b/test/zephyr/app/shim_nucleo_n657x0_q.c
@@ -1,0 +1,240 @@
+/*
+ * Copyright (c) The mlkem-native project authors
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+ */
+
+/*
+ * Zephyr entrypoint shim for RAM-loaded NUCLEO-N657X0-Q tests.
+ *
+ * The OpenOCD/GDB wrapper breaks at get_bootargs() after Zephyr has cleared
+ * BSS, restores a Zephyr dynamic bootargs command line into
+ * mlk_bootargs_block, then continues. Zephyr parses the command line and calls
+ * main(argc, argv). Target stdout/stderr goes through ITM stimulus port 0 and
+ * is captured by the host over SWO. The test return code is passed to
+ * nucleo_test_done(), where the host wrapper reads r0.
+ */
+
+#include <cmsis_core.h>
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stm32n6xx_ll_bus.h>
+#include <stm32n6xx_ll_gpio.h>
+#include <stm32n6xx_ll_system.h>
+#include <string.h>
+
+extern int mlk_test_main(int argc, char **argv);
+
+void nucleo_test_breakpoint(int rc);
+int __wrap_printf(const char *format, ...);
+int __wrap_fprintf(FILE *stream, const char *format, ...);
+int __wrap_puts(const char *s);
+int __wrap_putchar(int c);
+int __wrap_fflush(FILE *stream);
+__attribute__((noreturn)) void nucleo_test_done(int rc);
+__attribute__((noreturn)) void __wrap_exit(int status);
+const char *get_bootargs(void);
+
+#define BOOTARGS_BLOCK_SIZE (64U * 1024U)
+#define NUCLEO_ITM_LAR (*(volatile uint32_t *)0xE0000FB0UL)
+
+__asm__(
+    ".global mlk_bootargs_block\n"
+    ".set mlk_bootargs_block, 0x340b0000\n");
+
+extern char mlk_bootargs_block[BOOTARGS_BLOCK_SIZE];
+
+static char nucleo_stdout_buf[8192];
+static size_t nucleo_stdout_len;
+static bool nucleo_swo_enabled;
+
+static void nucleo_swo_pin_setup(void)
+{
+  LL_AHB4_GRP1_EnableClock(LL_AHB4_GRP1_PERIPH_GPIOB);
+  LL_GPIO_SetPinMode(GPIOB, LL_GPIO_PIN_5, LL_GPIO_MODE_ALTERNATE);
+  LL_GPIO_SetPinSpeed(GPIOB, LL_GPIO_PIN_5, LL_GPIO_SPEED_FREQ_VERY_HIGH);
+  LL_GPIO_SetPinPull(GPIOB, LL_GPIO_PIN_5, LL_GPIO_PULL_NO);
+  LL_GPIO_SetAFPin_0_7(GPIOB, LL_GPIO_PIN_5, LL_GPIO_AF_0);
+}
+
+static void nucleo_swo_enable(void)
+{
+  if (nucleo_swo_enabled)
+  {
+    return;
+  }
+
+  nucleo_swo_pin_setup();
+  LL_BUS_EnableClock(LL_APB3);
+  LL_DBGMCU_EnableDBGClock();
+  LL_DBGMCU_EnableTPIUExportClock();
+  CoreDebug->DEMCR |= CoreDebug_DEMCR_TRCENA_Msk;
+
+  NUCLEO_ITM_LAR = 0xC5ACCE55UL;
+  ITM->TER = 0UL;
+  ITM->TCR &= ~ITM_TCR_ITMENA_Msk;
+  while ((ITM->TCR & ITM_TCR_BUSY_Msk) != 0UL)
+  {
+  }
+  ITM->TPR = 0UL;
+  ITM->TCR =
+      ITM_TCR_ITMENA_Msk | ITM_TCR_DWTENA_Msk | (1UL << ITM_TCR_TRACEBUSID_Pos);
+  ITM->TER = 1UL;
+  nucleo_swo_enabled = true;
+}
+
+static void nucleo_swo_write_byte(char ch)
+{
+  nucleo_swo_enable();
+
+  if (((ITM->TCR & ITM_TCR_ITMENA_Msk) == 0UL) || ((ITM->TER & 1UL) == 0UL))
+  {
+    return;
+  }
+
+  while (ITM->PORT[0U].u32 == 0UL)
+  {
+  }
+  ITM->PORT[0U].u8 = (uint8_t)ch;
+}
+
+static void nucleo_swo_write_raw(const char *src, size_t len)
+{
+  for (size_t offset = 0; offset < len; offset++)
+  {
+    nucleo_swo_write_byte(src[offset]);
+  }
+}
+
+static void nucleo_swo_wait_idle(void)
+{
+  for (uint32_t timeout = 10000000UL; timeout > 0U; timeout--)
+  {
+    if ((ITM->TCR & ITM_TCR_BUSY_Msk) == 0UL)
+    {
+      break;
+    }
+  }
+  __DSB();
+}
+
+__attribute__((noinline, used)) void nucleo_test_breakpoint(int rc)
+{
+  (void)rc;
+  __asm__ volatile("bkpt 0" ::: "memory");
+}
+
+static void nucleo_stdout_flush(void)
+{
+  if (nucleo_stdout_len == 0U)
+  {
+    return;
+  }
+
+  nucleo_swo_write_raw(nucleo_stdout_buf, nucleo_stdout_len);
+  nucleo_stdout_len = 0U;
+}
+
+static void nucleo_stdout_write(const char *src, size_t len)
+{
+  for (size_t i = 0; i < len; i++)
+  {
+    nucleo_stdout_buf[nucleo_stdout_len++] = src[i];
+    if (src[i] == '\n' || nucleo_stdout_len == sizeof(nucleo_stdout_buf))
+    {
+      nucleo_stdout_flush();
+    }
+  }
+}
+
+static int nucleo_vprintf(const char *format, va_list ap)
+{
+  char buf[8192];
+  int rc = vsnprintf(buf, sizeof(buf), format, ap);
+
+  if (rc <= 0)
+  {
+    return rc;
+  }
+
+  nucleo_stdout_write(
+      buf, (size_t)((rc < (int)sizeof(buf)) ? rc : (int)sizeof(buf) - 1));
+  return rc;
+}
+
+int __wrap_printf(const char *format, ...)
+{
+  va_list ap;
+  int rc;
+
+  va_start(ap, format);
+  rc = nucleo_vprintf(format, ap);
+  va_end(ap);
+  return rc;
+}
+
+int __wrap_fprintf(FILE *stream, const char *format, ...)
+{
+  va_list ap;
+  int rc;
+
+  (void)stream;
+  va_start(ap, format);
+  rc = nucleo_vprintf(format, ap);
+  va_end(ap);
+  return rc;
+}
+
+int __wrap_puts(const char *s)
+{
+  size_t len = strlen(s);
+
+  nucleo_stdout_write(s, len);
+  nucleo_stdout_write("\n", 1);
+  return (int)len + 1;
+}
+
+int __wrap_putchar(int c)
+{
+  char ch = (char)c;
+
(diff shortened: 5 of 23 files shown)


hanno-becker:
@bremoran What's your plan for this?

bremoran:
I am hoping to finalise this but I'm not able to spend time on it right now.

mkannwischer:
> I am hoping to finalise this but I'm not able to spend time on it right now.

Thanks for the update. Marking it as draft for now. Please mark it as ready when you have finished it.

bremoran:
> Thanks for the update. Marking it as draft for now. Please mark it as ready when you have finished it.

Done!

.github/workflows/base.yml
```diff
@@ -23,6 +23,20 @@ jobs:
           nix-shell: linter
           gh_token: ${{ secrets.GITHUB_TOKEN }}
           cross-prefix: "aarch64-unknown-linux-gnu-"
+  ubuntu_arm:
```
hanno-becker: Is this not subsumed by the `quickcheck` job?

nix/nucleo-n657x0-q/default.nix
```diff
@@ -0,0 +1,247 @@
...
+    tpl="Projects/NUCLEO-N657X0-Q/Templates/Template"
+
+    if [ ! -d "$tpl" ]; then
+      echo "ERROR: expected Cube template at $tpl"
+      exit 1
+    fi
+
+    # Copy CMSIS headers needed by system/startup from the Cube tree
+    mkdir -p "$outp/Drivers/CMSIS/Core"
+    mkdir -p "$outp/Drivers/CMSIS/Device"
```
hanno-becker: The install phase seems pretty long. Could you provide some overview in the beginning as to what is happening here?

bremoran: Yes, I have added some comments that will hopefully make things clearer. I have also simplified the nix file.

.github/workflows/base.yml
bremoran: Yes, I suspect it is. I think this is because of a bad rebase; I'll prune it.

.github/workflows/bench.yml
```diff
@@ -79,11 +79,22 @@ jobs:
...
+          bench_extra_args: ""
+          nix_shell: nucleo-n657x0-q
+          cross_prefix: ""
+    if: ${{ github.repository_owner == 'pq-code-package' && (github.event.label.name == 'benchmark' || github.ref == 'refs/heads/main') }}
     runs-on: self-hosted-${{ matrix.target.system }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - uses: ./.github/actions/bench
+        env:
+          EXTRA_MAKEFILE: ${{ matrix.target.system == 'nucleo-n657x0' && 'test/baremetal/platform/nucleo-n657x0-q/platform.mk' || '' }}
```
hanno-becker: Can this be part of the matrix?

.github/workflows/ci.yml
```diff
@@ -191,6 +191,29 @@ jobs:
           nix-shell: ''
           gh_token: ${{ secrets.GITHUB_TOKEN }}
           cflags: "-DMLKEM_DEBUG -DMLK_FORCE_PPC64LE"
+  nucleo_n657x0_q_tests:
```
hanno-becker: Why is this done here and not alongside the existing baremetal tests?

test/baremetal/platform/nucleo-n657x0-q/README.md
```diff
@@ -0,0 +1,258 @@
+<!--
+Copyright (c) The mldsa-native project authors
+Copyright (c) The mlkem-native project authors
+Copyright (c) Arm Ltd.
+SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+-->
+
+# NUCLEO-N657X0-Q Baremetal Platform
+
+This platform runs ML-KEM tests on the ST NUCLEO-N657X0-Q board using the
```
hanno-becker: Does QEMU support this board, so one could, alternatively, emulate the final binary?

test/baremetal/platform/nucleo-n657x0-q/flexmem_configure.py
```diff
@@ -0,0 +1,179 @@
+#!/usr/bin/env python3
```
hanno-becker: Shallow comment: Can we keep the top-level of the platform clean and ideally confined to `exec_wrapper.py`, `platform.mk`, `README.md`?

test/mk/components.mk
```diff
@@ -103,6 +103,11 @@ $(MLKEM512_DIR)/bin/bench_components_mlkem512: $(MLKEM512_DIR)/test/hal/hal.c.o
 $(MLKEM768_DIR)/bin/bench_components_mlkem768: $(MLKEM768_DIR)/test/hal/hal.c.o
 $(MLKEM1024_DIR)/bin/bench_components_mlkem1024: $(MLKEM1024_DIR)/test/hal/hal.c.o
 
+# Build hal.c without pedantic diagnostics (overrides global -pedantic/-Wpedantic)
+$(MLKEM512_DIR)/test/hal/hal.c.o: CFLAGS := $(filter-out -pedantic -Wpedantic -Werror -Wconversion -Wsign-conversion,$(CFLAGS)) -Wno-pedantic -Wno-conversion -Wno-sign-conversion -Wno-error
+$(MLKEM768_DIR)/test/hal/hal.c.o: CFLAGS := $(filter-out -pedantic -Wpedantic -Werror -Wconversion -Wsign-conversion,$(CFLAGS)) -Wno-pedantic -Wno-conversion -Wno-sign-conversion -Wno-error
+$(MLKEM1024_DIR)/test/hal/hal.c.o: CFLAGS := $(filter-out -pedantic -Wpedantic -Werror -Wconversion -Wsign-conversion,$(CFLAGS)) -Wno-pedantic -Wno-conversion -Wno-sign-conversion -Wno-error
```
hanno-becker: hal.c seems small enough that one should be able to adjust the code instead of silencing the warnings?

scripts/tests
```diff
@@ -759,6 +759,10 @@ class Tests:
 
         if resultss is None:
             self.check_fail()
+            return
+
+        if len(self.failed) > 0:
+            self.check_fail()
```
hanno-becker: What's this?

.github/workflows/bench.yml
bremoran: Yes, agreed. I’ll add an `extra_makefile` field to the benchmark matrix and pass that through as `EXTRA_MAKEFILE`, with an empty value for the existing non-baremetal targets. That removes the target-specific expression from the step.

.github/workflows/ci.yml
bremoran: Good point. This should live with the existing baremetal workflow. I’ll move the `NUCLEO-N657X0-Q` job into `.github/workflows/baremetal.yml` as a matrix target, while keeping the self-hosted/fork guard because this one needs attached hardware rather than QEMU.

test/baremetal/platform/nucleo-n657x0-q/README.md
bremoran: I don’t think QEMU currently supports the `STM32N657/NUCLEO-N657X0-Q` machine model. QEMU’s Arm docs list only a small set of STM32 boards, and this board is not among them; they also note that firmware images generally need an exact matching board model to boot. I’ll add a short note to the README saying this platform currently requires hardware. ([qemu.org](https://www.qemu.org/docs/master/system/target-arm.html))

That said, I have just stumbled across https://renode.io which may be relevant to simulating hardware targets outside the baseline supported by QEMU.

test/baremetal/platform/nucleo-n657x0-q/flexmem_configure.py
bremoran: Yes. I’ll move host-side helpers such as flexmem_configure.py, run_test_after_flexmem.py, make_argv_bin.py, and the host tests under the existing nucleo_host/ area, leaving only exec_wrapper.py, platform.mk, and README.md at the platform top level.

test/mk/components.mk
bremoran: Agreed. I’ll remove this per-target warning suppression and fix the warnings in `test/hal/hal.c` directly. The warning relaxation should stay limited to vendor/Cube sources in the platform makefile.

test/baremetal/platform/nucleo-n657x0-q/flexmem_configure.py
bremoran: I have a fix staged, but I don't have my nucleo with me today so I cannot test it until later in the day.

nix/nucleo-n657x0-q/default.nix
```diff
@@ -0,0 +1,195 @@
...
+
+{ stdenvNoCC
+, fetchFromGitHub
+, writeText
+}:
+# Construct a minimal STM32CubeN6-based platform environment for building
+# nucleo-n657x0-q benchmarks. The package copies only the required Cube headers
+# and sources, then patches selected files where needed to get the intended
+# RAM-loaded benchmark behavior.
+stdenvNoCC.mkDerivation {
```
mkannwischer: Would it be an option to just move all the patched files to pqmx and consume them from there?

test/baremetal/platform/nucleo-n657x0-q/src/libfns.c
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) The mldsa-native project authors
```
hanno-becker: Is this file needed? It doesn't seem to be listed in the extra sources.

test/baremetal/platform/nucleo-n657x0-q/test_nucleo_host.py
```diff
@@ -0,0 +1,635 @@
+# Copyright (c) The mlkem-native project authors
```
hanno-becker: This file is not exercised, it seems. Can it be removed?

test/baremetal/platform/nucleo-n657x0-q/README.md
```diff
@@ -0,0 +1,258 @@
...
+- data/stack: default 128 KiB DTCM
+- no flash memory regions or flash LMAs
+
+`linker/ram_secure.ld` is used by tests after FLEXMEM reset has applied:
+
+- vector table and executable sections: expanded 256 KiB ITCM at `0x00000000`
+- `.data`, `.bss`: expanded 256 KiB DTCM at `0x30000000`
+- argv block: expanded 256 KiB ITCM at `0x00000000`
+- stdout capture: AXI SRAM at `0x34080000`
+- stack: top 192 KiB of DTCM, with `_estack = 0x30040000` and
```
hanno-becker: This seems out of date with the linker script which uses 240 KiB. Can you sync the README with the current state of the PR?

test/baremetal/platform/nucleo-n657x0-q/run_test_after_flexmem.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
```
hanno-becker: (Where) is this exercised? Drop it or add to CI?

nix/st-openocd/default.nix
```diff
@@ -0,0 +1,69 @@
...
+, autoconf
+, automake
+, libtool
+, which
+}:
+
+openocd.overrideAttrs (old: rec {
+  pname = "st-openocd";
+  version = "unstable-2026-05-01";
+
```
mkannwischer: I have moved this directly into nixpkgs and I'll try to upstream it so we do not have to compile thsi from source: https://github.com/pq-code-package/mlkem-native/commit/47c5987718bdf13521fc847b5aa9f5e6735c05c8

mkannwischer: opened a PR here: https://github.com/NixOS/nixpkgs/pull/531668

nix/nucleo-n657x0-q/default.nix
bremoran: The issue with this is the license here. If this is acceptable in pqmx, then we can move the files.

https://github.com/STMicroelectronics/STM32CubeN6/blob/main/Projects/NUCLEO-N657X0-Q/Templates/LICENSE.md

.github/workflows/ci.yml
hanno-becker: This still seems open

scripts/tests
hanno-becker: @bremoran ping

bremoran: This block exists because failed benchmark runs return booleans instead of output strings, and the code was bailing out before parsing those booleans. I’ll make that explicit in control flow.

scripts/format
```diff
@@ -69,7 +69,7 @@ info "Expanding tabs"
 expand-tabs()
 {
   git ls-files -- ":/" ":/!:Makefile" ":/!:**/Makefile" ":/!:**/Makefile.*" ":/!:Makefile.*" ":/!:*.mk" ":/!:*.patch" ":/!:*.S" ":/!:*.inc" ":/!:nix/valgrind/*.txt" | xargs -P "$nproc" -I {} sh -c '
-    if [ ! -L {} ] && grep -Pq '"'"'\t'"'"' "{}"; then
+    if [ ! -L {} ] && grep -q "$(printf '"'"'\t'"'"')" "{}"; then
```
hanno-becker: Why was that change made? It seems unrelated

scripts/lint
```diff
@@ -268,9 +268,9 @@ gh_group_end
 
 check-tabs()
 {
-  for file in $(git ls-files -- ":/" ":/!:Makefile" ":/!:**/Makefile" ":/!:**/Makefile.*" ":/!:Makefile.*" ":/!:**/Kconfig" ":/!:*.mk" ":/!:*.patch" ":/!:*.S" ":/!:*.inc" ":/!:nix/valgrind/*.txt"); do
-    if [[ ! -L $file ]] && grep -Pq '\t' "$file"; then
-      l=$(grep -Pn '\t' "$file" | head -1 | cut -d: -f1)
```
hanno-becker, line 273: Why was this change made?

test/zephyr/app/shim_nucleo_n657x0_q.c
```diff
@@ -0,0 +1,235 @@
...
+static size_t nucleo_stdout_len;
+static bool nucleo_swo_enabled;
+
+static void nucleo_swo_pin_setup(void)
+{
+  LL_AHB4_GRP1_EnableClock(LL_AHB4_GRP1_PERIPH_GPIOB);
+  LL_GPIO_SetPinMode(GPIOB, LL_GPIO_PIN_5, LL_GPIO_MODE_ALTERNATE);
+  LL_GPIO_SetPinSpeed(GPIOB, LL_GPIO_PIN_5, LL_GPIO_SPEED_FREQ_VERY_HIGH);
+  LL_GPIO_SetPinPull(GPIOB, LL_GPIO_PIN_5, LL_GPIO_PULL_NO);
+  LL_GPIO_SetAFPin_0_7(GPIOB, LL_GPIO_PIN_5, LL_GPIO_AF_0);
```
hanno-becker, line 59: Can you remind me why we need stuff like this? I thought the point of using Zephyr was that we can piggy back on their platform code.

hanno-becker (commented):
Some questions about seemingly unrelated changes in `scripts/*`, and the bench CI seems to be failing.

Other than that: Apologies it fell out of my context window, but I am confused why on the one hand we seem to reuse (some of) the Zephyr platform, but on the other hand provide a lot of custom code in `shim_nucleo_n657x0_q.c` which I would have expected to come from Zephyr. Can you explain, if only for the record?

test/zephyr/app/shim_nucleo_n657x0_q.c
bremoran: When I did my original port to Nucleo, I quickly identified that semihosting I/O was poor performance due to using single-character-writes. This becomes extremely painful in KAT runs.

In my first attempt, I replaced semihosted I/O with an in-memory buffer for output, which was data-dumped after the run. This was a very unusual approach, so I replaced it with SWO output. I was hoping that this would allow me to get a very high output data rate since SWO is supposed to be able to be clocked up to 16MHz, however in practice this only landed at 1MHz.

I see several options:
1. We accept low speed by default and stick with semihosting.
2. We leave this as is, as a shim.
3. We convert to using UART and use the builtin Zephyr configuration for that
4. We convert to using logging and use the builtin Zephyr logging configuration

Given that I was only able to get 1Mbps out of the SWO configuration I'm not sure that it is getting the intended benefits. I'm happy to give UART a try and see if it gets us further.

bremoran:
> Some questions about seemingly unrelated changes in `scripts/*`, and the bench CI seems to be failing.
> 
> Other than that: Apologies it fell out of my context window, but I am confused why on the one hand we seem to reuse (some of) the Zephyr platform, but on the other hand provide a lot of custom code in `shim_nucleo_n657x0_q.c` which I would have expected to come from Zephyr. Can you explain, if only for the record?

I put a more complete explanation here: https://github.com/pq-code-package/mlkem-native/pull/1547#discussion_r3494759177

But the point was essentially that the default we were using (semihosting) was not effective on real hardware and we needed something faster. I'll do some experiments and see if converting to UART is favourable in terms of complexity.

scripts/format
bremoran: Agreed, this was an unrelated portability cleanup for avoiding `grep -P` which is not available on all platforms. I’ve reverted it from this PR to keep the NUCLEO platform change focused.

scripts/lint
bremoran: This was an unrelated portability cleanup for avoiding `grep -P`. I’ve reverted it from this PR to keep the NUCLEO platform change focused.

bremoran:
> But the point was essentially that the default we were using (semihosting) was not effective on real hardware and we needed something faster. I'll do some experiments and see if converting to UART is favourable in terms of complexity.

UART is simpler on the target (zephyr handles more of it) but more complex on the host due to the need to identify which Virtual Com Port belongs to which debugger. With SWO, this is automatic. I am inclined to keep SWO on this basis.

test/hal/hal.c
```diff
@@ -51,6 +51,35 @@ void enable_cyclecounter(void) {}
 void disable_cyclecounter(void) {}
 uint64_t get_cyclecounter(void) { return k_cycle_get_32(); }
 
+#elif defined(CYCCNT_CYCLES)
+
+#if defined(__ARM_ARCH_8M_MAIN__) || defined(__ARM_ARCH_8_1M_MAIN__)
+
+#if defined(ARMCM55)
```
hanno-becker: (Where) is this used? It seems this is overriden by `__ZEPHYR__` above

test/zephyr/nucleo_n657x0_q/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+Use the Zephyr platform makefile and select the NUCLEO target:
+
+```sh
+nix develop .#zephyr
+make run_func_512 EXTRA_MAKEFILE=test/zephyr/platform.mk ZEPHYR_TARGET=nucleo-n657x0-q
+```
+
+The execution wrapper configures FLEXMEM before each ELF run. The supported
+Zephyr hardware run targets are:
```
hanno-becker: (Why) Are the other targets not supported? We need to clearly distinguish a) not working, b) makes no sense c) works in principle, but takes too long, d) actually it works and we should test it. For example, the unit tests would seem to fall under d)?

test/zephyr/nucleo_n657x0_q/README.md
```diff
@@ -0,0 +1,86 @@
...
+export SWO_PIN_FREQ=1000000
+```
+
+`OPENOCD`, `OPENOCD_INTERFACE`, `OPENOCD_TARGET`, `OPENOCD_TRANSPORT`, `GDB`,
+`NM`, and `READELF` can override the default tools and OpenOCD scripts.
+
+## FLEXMEM Sequence
+
+STM32N657X0 starts with a smaller reset-time TCM layout. The NUCLEO runner
+expands both ITCM and DTCM before each RAM-loaded Zephyr test:
```
hanno-becker: How big is the reset-time TCM? Do we need to dynamically expand it?

test/zephyr/platform.mk
```diff
@@ -50,6 +59,29 @@ ZEPHYR_FIPS202_BACKEND := $(if $(filter 1,$(OPT)),$(strip $(ZEPHYR_FIPS202_BACKE
 
 ZEPHYR_APP := $(PLATFORM_PATH)/app
 ZEPHYR_BUILD_DIR := $(BUILD_DIR)/zephyr/$(ZEPHYR_TARGET)
+ZEPHYR_ACTIVE_TARGET := $(BUILD_DIR)/zephyr/.active-target
```
hanno-becker, line 62: What's this for?

test/zephyr/platform.mk
```diff
@@ -50,6 +59,29 @@ ZEPHYR_FIPS202_BACKEND := $(if $(filter 1,$(OPT)),$(strip $(ZEPHYR_FIPS202_BACKE
...
+	-DZEPHYR_NUCLEO_N657X0_Q=ON \
+	-DEXTRA_CONF_FILE=$(ZEPHYR_NUCLEO_CONF) \
+	-DDTC_OVERLAY_FILE=$(ZEPHYR_NUCLEO_OVERLAY))
+
+.PHONY: zephyr_target_marker_force
+$(ZEPHYR_ACTIVE_TARGET): zephyr_target_marker_force
+	$(Q)[ -d $(@D) ] || mkdir -p $(@D)
+	$(Q)if [ ! -f $@ ] || [ "$$(cat $@)" != "$(ZEPHYR_TARGET)" ]; then \
+		echo "$(ZEPHYR_TARGET)" > $@; \
+	fi
```
hanno-becker, line 89: What do we need this for? Why is this not added into the `CUSTOM_BUILD` recipe?

hanno-becker (changes requested):
A few question remain, see comments

test/hal/hal.c
bremoran: I've removed this section.

test/zephyr/nucleo_n657x0_q/README.md
bremoran: Reset-time DTCM: 128KiB/ITCM: 64KiB.

For anything other than benchmark code, this is irrelevant and this target has 4.2MiB of AXI RAM with 32KiB caches.

For benchmark code, to eliminate cache effects and memory access effects, we need to run from ITCM and to use only DTCM for data. It is possible that, if we carefully partitioned the memory, we could assign everything that is not in the benchmark path to normal AXISRAM while keeping the stack and data in DTCM and, similarly, keeping only the benchmarked code in ITCM, while leaving all other code in AXISRAM.

I don't know if that would be enough DTCM for the full 4-wide accelerated mlkem-native. I know it is not enough for mldsa-87.

test/zephyr/platform.mk
bremoran: This is because I do not have a sim target for nucleo n657x0, so I'm simulating with mps3-an547. That means that I need rebuilds when I test for correctness on mps3, then benchmark on nucleo n657x0. This is probably specific to my workflow, however.

test/zephyr/platform.mk
bremoran: This is part of the same machinery for target switching as described above.

test/zephyr/platform.mk
hanno-becker: Ah. So this is to automatically trigger a rebuild if you change the board. That makes sense and is generally useful, I think. Can you just add a comment to this effect?

test/zephyr/nucleo_n657x0_q/README.md
bremoran: The latest update makes it clear that there are some known not working targets, but the rest should run.

scripts/tests
```diff
@@ -1391,8 +1394,8 @@ def cli():
     bench_parser.add_argument(
         "-c",
         "--cycles",
-        help="Method for counting clock cycles. PMU requires (user-space) access to the Arm Performance Monitor Unit (PMU). PERF requires a kernel with perf support. MAC works on some Apple platforms, at least Apple M1.",
-        choices=["NO", "PMU", "PERF", "MAC"],
+        help="Method for counting clock cycles. CYCCNT uses the Cortex-M DWT cycle counter. PMU requires (user-space) access to the Arm Performance Monitor Unit (PMU). PERF requires a kernel with perf support. MAC works on some Apple platforms, at least Apple M1.",
```
hanno-becker: Remnant of the `CYCCNT` mode that has been removed

bremoran:
FWIW the 1 failed test appears unrelated.

hanno-becker (approved):
Thank you for the work and patience, @bremoran! :pray:

bremoran:
> Thank you for the work and patience, @bremoran! 🙏

Thank you for the many reviews and the guidance @hanno-becker! 🙏

PR #1559, opened 2026-02-10
Keypair: zeroize pk and sk in case of PCT failure

mkannwischer:
If the pair-wise consistency test fails (e.g., due to OOM), we currently do still return the pk and sk.
This commit changes that to zeroize the buffers instead which lowers the risk of API mis-use.

### DIFF
--- a/mlkem/src/kem.c
+++ b/mlkem/src/kem.c
@@ -217,7 +217,7 @@ int mlk_kem_keypair_derand(uint8_t pk[MLKEM_INDCCA_PUBLICKEYBYTES],
   ret = mlk_indcpa_keypair_derand(pk, sk, coins, context);
   if (ret != 0)
   {
-    return ret;
+    goto cleanup;
   }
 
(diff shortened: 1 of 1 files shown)


mlkem/src/kem.c
```diff
@@ -234,6 +234,8 @@ int mlk_kem_keypair_derand(uint8_t pk[MLKEM_INDCCA_PUBLICKEYBYTES],
   ret = mlk_check_pct(pk, sk, context);
   if (ret != 0)
   {
+    mlk_zeroize(pk, MLKEM_INDCCA_PUBLICKEYBYTES);
+    mlk_zeroize(sk, MLKEM_INDCCA_SECRETKEYBYTES);
```
hanno-becker, line 244: I'd prefer a generic cleanup section which clears `pk` and `sk` if the return value is not zero. Functionally, this is equivalent _at the moment_, but it would no longer be if an error condition arose in `indcpa_keypair_derand` which saw `pk/sk` written to before an error happens.

mkannwischer: Agreed. Changed.

Issue #1261, opened 2025-10-30
HOL-Light: Import constant-time proofs from s2n-bignum


hanno-becker:
Thanks to @aqjune-aws, s2n-bignum now includes infrastructure for reasoning about constant-time properties of code. This infrastructure has in particular been used to prove constant-time'ness of the AArch64 backend of mlkem-native.

**Task:** 
* Import new specifications and proofs into mlkem-native and its CI
* Document and explain what has been proved

mkannwischer:
@jammychiou1, are you interested in looking at this issue?

jammychiou1:
Thank you @mkannwischer for the opportunity! I'm certainly interested in learning more about HOL-Light and s2n-bignum's verification framework, but unfortunately I have to pass on that for now...

I did take a quick look at it: Importing the proofs doesn't seem too difficult, but I'm not sure if I can understand them enough to correctly "Document and explain what has been proved" in reasonable time.

I wish I could dive deeper into this, but I might have to prioritize other tasks for the next few weeks.

mkannwischer:
Tracking progress: 

  ### AArch64                                                                                                                                                               
   
  - [x] `mlkem_poly_tobytes`                                                                                                                                                
  - [x] `mlkem_intt`         
  - [x] `mlkem_ntt`
  - [x] `mlkem_poly_basemul_acc_montgomery_cached_k2`
  - [x] `mlkem_poly_basemul_acc_montgomery_cached_k3`
  - [x] `mlkem_poly_basemul_acc_montgomery_cached_k4`
  - [x] `mlkem_poly_mulcache_compute`
  - [x] `mlkem_poly_reduce`
  - [x] `mlkem_poly_tomont`
  - [ ] `mlkem_rej_uniform`
  - [x] `keccak_f1600_x1_scalar`
  - [x] `keccak_f1600_x1_v84a`
  - [x] `keccak_f1600_x2_v84a`
  - [x] `keccak_f1600_x4_v8a_scalar`
  - [x] `keccak_f1600_x4_v8a_v84a_scalar`

  ### x86_64

  - [x] `mlkem_intt`
  - [x] `mlkem_ntt`
  - [x] `mlkem_poly_basemul_acc_montgomery_cached_k2`
  - [x] `mlkem_poly_basemul_acc_montgomery_cached_k3`
  - [x] `mlkem_poly_basemul_acc_montgomery_cached_k4`
  - [x] `mlkem_mulcache_compute`
[...]

PR #1570, opened 2026-02-16
HOL-Light: Add constant-time proofs for various x86_64 functions

mkannwischer:
- Ports the `mlkem_reduce` CT+memory safety proof from s2n-bignum
- Also adds similar proofs for `mlkem_tomont`, `mlkem_unpack`, `mlkem_tobytes`, `mlkem_frombytes`, and `mlkem_mulcache_compute` - they are not present in s2n-bignum but closely follow `mlkem_reduce`
- Towards #1261


proofs/hol_light/x86_64/proofs/mlkem_frombytes.ml
```diff
@@ -472,3 +472,45 @@ let MLKEM_FROMBYTES_SUBROUTINE_CORRECT = prove(
...
+
+(* ------------------------------------------------------------------------- *)
+(* Constant-time and memory safety proof.                                    *)
+(* ------------------------------------------------------------------------- *)
+
+needs "x86_64/proofs/mlkem_utils.ml";;
+needs "x86_64/proofs/subroutine_signatures.ml";;
+
+let full_spec,public_vars = mk_safety_spec
+    ~keep_maychanges:true
```
hanno-becker: Can we keep this consistent with AArch64? I don't see the value in setting this option.

mkannwischer: Should be doable. I'll give it a try.

mkannwischer: Fixed - see https://github.com/pq-code-package/mlkem-native/compare/3d2e47a..5a22128daf29e105017904f71361203dc22a2001.

Issue #1520, opened 2026-01-22
format: Don't allow tabs


hanno-becker:
Our `format` script does so far not complain about tabs in source files. We should find a way to force whitespaces instead of tabs.

mkannwischer:
What also bother's me is that we have/had tabs mixed with spaces in Makefiles in places we could also just just spaces.

hanno-becker:
`clang-format` doesn't seem to support this (there's only `UseTab` but that concerns tabs introduced by clang-format, I think), but `expand/unexpand` seem standard.

PR #1545, opened 2026-02-04
HOL Light: Prove correctness of AVX2 `poly_compress_d{4,5,10,11}`

mkannwischer:

PR #1543, opened 2026-02-04
HOL Light: Prove correctness of AVX2 `poly_decompress_d{4,5,10,11}`

mkannwischer:

PR #1573, opened 2026-02-20
Format: Don't allow tabs

mkannwischer:
- Resolves https://github.com/pq-code-package/mlkem-native/issues/1520


mkannwischer:
Let's hold this back until we have merged #1543 and #1545.

hanno-becker:
> Let's hold this back until we have merged https://github.com/pq-code-package/mlkem-native/pull/1543 and https://github.com/pq-code-package/mlkem-native/pull/1545.

Those are merged now. Removing DO-NOT-MERGE.

mkannwischer:
> Should we also set `UseTab: Never` in `.clang-format` to prevent formatting-introduced tabs?

Isn't that already the case when you do `BasedOnStyle: Google`?

hanno-becker:
@mkannwischer Good point, you are right.

PR #1575, opened 2026-02-24
HOL-Light: Add missing x86_64 CT proofs and lift to subroutine level

mkannwischer:


proofs/hol_light/x86_64/proofs/mlkem_reduce.ml
```diff
@@ -459,30 +478,84 @@ needs "x86_64/proofs/subroutine_signatures.ml";;
...
+                    read RSP s = stackpointer /\
+                    read (memory :> bytes64 stackpointer) s = returnaddress /\
+                    C_ARGUMENTS [a] s /\
+                    read events s = e)
+               (\s. read RIP s = returnaddress /\
+                    read RSP s = word_add stackpointer (word 8) /\
+                    (exists e2.
+                         read events s = APPEND e2 e /\
+                         e2 = f_events a pc stackpointer returnaddress /\
+                         memaccess_inbounds e2 [a,512; a,512; stackpointer,8]
```
mkannwischer: The duplicated buffers should still be eliminated via `MEMACCESS_INBOUNDS_DEDUP_CONV` - making sure this works without first.

mkannwischer: Fixed.

proofs/hol_light/x86_64/proofs/mlkem_intt.ml
```diff
@@ -1246,3 +1246,44 @@ let MLKEM_INTT_SUBROUTINE_CORRECT  = prove
...
+                    read RIP s = word pc /\
+                    C_ARGUMENTS [a; zetas] s /\
+                    read events s = e)
+               (\s.
+                    read RIP s = word (pc + 3341) /\
+                    (exists e2.
+                         read events s = APPEND e2 e /\
+                         e2 = f_events zetas a pc /\
+                         memaccess_inbounds e2
+                           [a,512; zetas,1248] [a,512]))
```
hanno-becker: Sorry, I mentioned this before but forgot the outcome. Can we use a `LENGTH` value here instead of hardcoding `1248`?

mkannwischer: Yes, we can. I just forgot this one instance :(

hanno-becker: Hold on, I didn't yet look to the next commits.

mkannwischer: Oh yes, this is an outdated diff. This hopefully is resolved by the later commit. We should clean up the commit history once you are happy.

proofs/hol_light/x86_64/proofs/mlkem_tomont.ml
```diff
@@ -350,29 +369,83 @@ needs "x86_64/proofs/subroutine_signatures.ml";;
...
+                    read events s = e)
+               (\s. read RIP s = returnaddress /\
+                    read RSP s = word_add stackpointer (word 8) /\
+                    (exists e2.
+                         read events s = APPEND e2 e /\
+                         e2 = f_events a pc stackpointer returnaddress /\
+                         memaccess_inbounds e2 [a,512; stackpointer,8]
+                                               [a,512; stackpointer,8]))
+               (\s s'. true)`,
+  MATCH_ACCEPT_TAC(ADD_IBT_RULE MLKEM_TOMONT_NOIBT_SUBROUTINE_SAFE));;
```
hanno-becker, line 454: General comment: We should keep the `ASSERT_CONCL_TAC` in the final subroutine spec, connecting the spec that's being proven to the spec emitted by `mk_safety_spec`.

proofs/hol_light/x86_64/proofs/mlkem_reduce.ml
```diff
@@ -459,30 +478,84 @@ needs "x86_64/proofs/subroutine_signatures.ml";;
 needs "common/consttime_utils.ml";;
 
 let full_spec,public_vars = mk_safety_spec
-    ~keep_maychanges:false
+    ~keep_maychanges:true
```
hanno-becker, line 481: What relevance does this change actually have, seeing that you don't use `full_spec` anywhere, and I assume `public_vars` does not depend on the settingof `keep_maychanges`?

Do we want to keep `keep_maychanges:false` here and connect  `full_spec` to the final subroutine spec via `ASSERT_CONCL_TAC`?

mkannwischer: Very good catch - I followed what is in s2n-bignum, but that should really be added. 

Unfortunately, `mk_safety_spec` is incompatible with out `*_SUBROUTINE_CORRECT` theorems - I, however, managed to use the full_spec in the core `_SAFE` proof. Is that sufficient, or could you take a look how `mk_safety_spec` has to be modified to make it functional with the subroutine thoerems?

mkannwischer: There is another problem though: ASSERT_CONCL_TAC checks that the hand-written goal matches full_spec syntactically. The hand-written goal uses LENGTH qdata_full * 2 everywhere, andLENGTH_SIMPLIFY_CONV normalizes it all to 1248. However, full_spec mixes representations: mk_safety_spec computes 624*2=1248 numerically from the subroutine signature for memaccess_inbounds, but copies LENGTH qdata_full * 2 verbatim from the correctness theorem's nonoverlapping preconditions. So after LENGTH_SIMPLIFY_CONV on the goal, nonoverlapping doesn't match (1248 in goal vs LENGTH qdata_full * 2 in full_spec). We can work around this by also applying LENGTH_SIMPLIFY_CONV to full_spec.

hanno-becker: Understood and agreed.

PR #1582, opened 2026-02-26
Add SOUNDNESS.md

hanno-becker:
Formal verification is never absolute. SOUNDNESS.md aims to map out what is proved about mlkem-native, what is assumed, and where the gaps and risks lie.

This is difficult to get right. All feedback is welcome. Even once we merge an initial version of this document, I expect it to be a "living document" -- people will (hopefully, anyway) point out gaps that we missed, suggest new mitigations, etc. Here, we merely try to make a start so we can share it and kick off the process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,9 @@ All AArch64 and x86_64 assembly is proved functionally correct at the object-cod
 interactive theorem prover and the [s2n-bignum](https://github.com/awslabs/s2n-bignum/) verification infrastructure (which includes models of the
 relevant parts of the Arm and x86 architectures). See [proofs/hol_light](proofs/hol_light) for details.
 
+**NOTE:** Formal Verification is never absolute. See [SOUNDNESS.md](SOUNDNESS.md) for a detailed analysis of the scope, assumptions and risks of the formal verification
+efforts around mlkem-native.
+
 ## Security
 
 All assembly in mlkem-native is constant-time in the sense that it is free of secret-dependent control flow, memory access,
--- a/SOUNDNESS.md
+++ b/SOUNDNESS.md
@@ -1,0 +1,539 @@
+[//]: # (SPDX-License-Identifier: CC-BY-4.0)
+
+# Formal Verification in mlkem-native: Scope, Assumptions, Risks
+
+This document describes the scope, assumptions and risks of the formal verification
+efforts around mlkem-native.
+
+The parts of the analysis pertaining to HOL Light and s2n-bignum are largely shared with
+the corresponding [^s2n-bignum-soundness].
+
+## Overview
+
+Formal verification is never absolute. Every verification effort links formal objects --
+specifications and models -- to informal, real-world requirements and systems. This document
+maps out what is proved about mlkem-native, what is assumed, and where the gaps and risks lie.
+
+Our goal is to provide evidence rooted in formal reasoning for the statement that
+"mlkem-native implements the FIPS203 standard[^FIPS203] and does not leak secrets through timing variations".
+
+The narrative for the argument is the following, which is common in formal verification:
+- [Informal] Here is the real physical system, in all its complexity.
+- [Formal] Here is a formal model of the real system which we believe approximates its behavior.
+- [Formal] Here is how we can formally specify the behavior of this formal model.
+- [Formal] Here is how we formally prove the formal specification.
+- [Informal] Here is why we think this approximates the desired properties of the real system.
+
+Diagrammatically, this can be depicted as follows:
+
+```
+
+                                                  "Here is how we can formally specify
+                                                   the behavior of mlkem-native w.r.t.
+                                                   the machine's formal model."
+
+         Informal                                              Formal
+  ┌─────────────────────┐                             ┌─────────────────────────┐
+  │                     │            Gap A            │                         │
+  │  Desired behavior   │<· · · · · · · · · · · · · ·>│   Formal specification  │
+  │  of actual system   │   Is this what we wanted?   │                         │
+  │                     │                             │                         │
+  └─────────────────────┘                             └────────────▲────────────┘
+           ^                                                       │
+           :                                    Is the argument    │
+           :                                    sound?             │  "Here is how we formally
+           : GOAL                                                  │   prove mlkem-native's
+           :                                         Gap C: prover │   formal specification."
+           :                                         trust (TCB)   │
+           v                                                       │
+  ┌─────────────────────┐                             ┌────────────┴────────────┐
+  │                     │            Gap B            │                         │
+  │  Actual system      │<· · · · · · · · · · · · · ·>│     Formal model of     │
+  │                     │   Is this how the system    │     system and code     │
+  │  Code running on    │   behaves?                  │  Code representation +  │
+  │  real hardware      │                             │  execution semantics    │
+  └─────────────────────┘                             └─────────────────────────┘
+
+  "Here's the real system."                           "Here's how we think it can
+                                                       be formally approximated."
+
+```
+
+This methodology introduces three fundamental soundness gaps:
+- Gap A: Does the formal specification capture what we want to say about the real system?
+- Gap B: Is the formal model a faithful reflection of the real system?
+- Gap C: Is the formal argument trustworthy for why the formal model satisfies its specification?
+
+For mlkem-native, this structure is instantiated twice -- once for each verification
+stack.
+
+- **CBMC[^CBMC]** for the C code: memory safety, type safety, and absence of undefined behavior.
+- **HOL-Light[^HOL-Light] + s2n-bignum[^s2n-bignum]** for the assembly backends: functional correctness, memory safety, and secret-independent
+  execution.
+
+```
+         Informal                            Formal
+  ┌─────────────────────┐          ┌──────────────────────────┐
+  │                     │          │                          │
+  │  ML-KEM spec        │          │  CBMC          HOL Light │
+  │  (FIPS 203),        │          │  contracts     specs     │
+  │  constant-time      │   Gap A  │  (C funcs)      (ASM)    │
+  │  requirements,      │<· · · · >│     :             :      │
+  │  ABI, ...           │          │     :             :      │
+  │                     │          │     :             :      │
+  │                     │          │     :             :      │
+  └─────────────────────┘          └─────▲─────────────▲──────┘
+                                         │             │
+                                   Gap C │        Gap C│
+                                         │             │
+                                   CBMC +│      HOL Light kernel
+                             SMT solvers │             │
+                                         │             │
+  ┌─────────────────────┐          ┌─────┴─────────────┴─────┐
+  │                     │          │                         │
+  │  Compiled binary    │          │                         │
+  │  on real hardware   │   Gap B  │  CBMC's C      ISA      │
+  │                     │<· · · · >│  semantics    model     │
+  │  (+ CPU errata,     │          │  model      (arm.ml/    │
+  │   physical faults,  │          │             x86.ml)     │
+  │   reassembly risk)  │          │                         │
+  └─────────────────────┘          └─────────────────────────┘
+```
+
+---
+
+## Gap A: Do the formal specifications match actual requirements?
+
+This is the gap between what we *specify* and what users and applications actually *need*.
+It has multiple facets, including: Whether all system components are covered by specification (breadth),
+whether the specifications are strong enough (depth), whether they faithfully capture the informal requirements
+(faithfulness), and whether they compose to a claim about the whole system (consistency).
+
+### A1. Coverage: Breadth
+
+Which components of the system are captured by formal specification? The primary risk is
+that our coverage claims are wrong -- that a C or assembly function slips through without
+a specification, or that an unspecified component is incorrect.
+
+**What is covered.**
+
+- All C code in the core library has CBMC contracts.
+- All AArch64 assembly routines have HOL Light specifications.
+- All x86_64 assembly routines have HOL Light specifications.
+
+**What is NOT covered.**
+
+- **Backends for platforms other than AArch64 and x86_64** (e.g. RISC-V RVV, Armv8.1-M MVE, PowerPC):
+  Where present, these are not yet covered by specification.
+
+The full test suite (functional tests, KAT, ACVP, unit tests) validates functional
+correctness empirically across all platforms and configurations, but there is currently
+no automatic coverage check for CBMC or HOL Light. A function or configuration could slip
+through without being covered by specification.
+
+**Potential improvements.**
+- Add automatic proof coverage check to CI. ([#1423](https://github.com/pq-code-package/mlkem-native/issues/1423), [#1594](https://github.com/pq-code-package/mlkem-native/issues/1594))
+- Add verification coverage for all backends, using existing or new methodologies ([#1595](https://github.com/pq-code-package/mlkem-native/issues/1595)).
+
+### A2. Coverage: Depth
+
+Do the specifications have the desired depth? The risk is that undesired behavior occurs
+which is outside the scope of specification.
+
+We aim for mlkem-native to be **functionally correct** (the code computes the
+right answer as per FIPS 203), **memory-safe** (it only accesses memory within
+the bounds of what is provided or allocated) and **constant-time** (no
+secret-dependent timing variation).
+
+**Assembly (HOL Light).** With one exception, the ASM specifications capture functional
+correctness, memory safety, and secret-independent execution. The one exception remaining is
+rejection sampling: the native implementations for this function only have functional
+correctness specifications, but no specifications of memory safety (the functions are safely
+variable-time, so no constant-time proofs are needed).
+
+**C code (CBMC).** Our use of CBMC focuses on **memory safety and type safety** -- absence
+of undefined behavior including out-of-bounds access, integer overflow, null pointer
+dereference, division by zero, undefined shifts, and lossy type conversions. The proofs
+also capture limited aspects of functional behavior in simple cases -- for example,
+coefficient bounds after arithmetic operations, or the functional specification of
+`mlk_ct_memcmp`. See [`proofs/cbmc/Makefile.common`](proofs/cbmc/Makefile.common) for the
+full list of checked classes of undefined behavior.
+
+The CBMC proofs do **not** currently cover:
+
+- **Functional correctness.** There is no machine-checked proof that the C code computes
+  the right answer per FIPS 203.
+- **Constant-time execution.** Side-channel resistance at the C level is not formally
+  proved.
+
+These gaps are mitigated in complementary ways. The full test suite (functional tests, KAT,
+ACVP, unit tests) validates functional correctness empirically across all platforms and
+configurations. For arithmetic correctness specifically, the most subtle bugs in ML-KEM
+implementations are rare overflows in the optimized polynomial arithmetic -- precisely the
+kind of bug that the type-safety and integer-overflow proofs are designed to catch: the CBMC
+contracts track coefficient bounds through the arithmetic pipeline, and overflow in
+intermediate computations would be flagged as undefined behavior. Constant-time properties
+are tested empirically using valgrind across many compilers and optimization levels (see
+the corresponding [CI job](.github/workflows/ct-tests.yml) for
+the full list), and the C code uses value barriers to prevent harmful compiler optimizations.
+
+**Potential improvements.**
+- Add memory-safety proof for rejection sampling routines. ([#1596](https://github.com/pq-code-package/mlkem-native/issues/1596))
+- Add automatic extraction of compiler coverage documentation from CI. ([#1608](https://github.com/pq-code-package/mlkem-native/issues/1608))
+- Introduce additional verification tooling that allows us to express functional correctness
+  and constant-time properties for the C code. ([#1597](https://github.com/pq-code-package/mlkem-native/issues/1597), [#1598](https://github.com/pq-code-package/mlkem-native/issues/1598))
+
+### A3. Specification faithfulness
+
+For the specifications that do exist, does their formal meaning capture their intent? The
+risk is that a specification does not express what we informally intend it to express.
+
+**Correctness.** The informal intent is to express that the top-level
+API behaves as specified by NIST's ML-KEM standard FIPS 203. As we do not yet address
+functional correctness at the C level, there is no formal specification of ML-KEM as a
+whole in mlkem-native. In our current setting -- capturing functional behavior only at the
+assembly level -- faithfulness requires that the HOL Light correctness specification style
+faithfully captures program behavior, and that the abstract HOL functions used in the
+specifications (e.g., the Number Theoretic Transform) are faithful representations of the
+mathematical functions in the standard.
+
+**Constant-time.** The informal intent is that mlkem-native does not
+leak secrets through timing side channels. Since we only capture constant-time properties
+at the assembly level, faithfulness requires that the HOL Light constant-time specification
+style faithfully captures constant-time execution.
+
+The s2n-bignum formal model approximates constant-time'ness as follows: It introduces the
+notion of *microarchitectural events* that flags selected instructions that are known to exhibit variable timing or
+influence timing otherwise (e.g., through caches): Examples are branches, load/store instructions, or variable-time
+instructions such as divisions. The constant-time specifications then posit that the trace of microarchitectural events
+emitted by a program can be expressed as a function of public variables such as input/output pointers or pointers
+to constant tables. Conversely, and somewhat implicitly, the absence of the (typically secret)
+_data_ behind those pointers as a parameter to the event-generating function implies that there
+is are no secret-dependent branches, load/stores, etc. It is the responsibility of the proof-writer
+to identify what is meant to be public vs. secret, and parametrize the event-generating function
+accordingly.
+
+The formal notion of constant-time'ness used in s2n-bignum does not and cannot guarantee
+that the hardware executes those instructions in constant time: Unmodeled microarchitectural
+effects such as speculative execution or Hertzbleed-style frequency scaling can still leak information.
+Moreover, some hardware provides opt-in guarantees for a listed set of instructions -- Arm's DIT (Data Independent
+Timing, Armv8.4-A onwards) and Intel's DOIT (Data Operand Independent Timing, Ice Lake onwards). We do not yet claim
+full alignment between the set of instructions marked as event-generating in s2n-bignum and the set of instructions
+outside of the scope of DIT/DOIT. Finally, our notion of microarchitectural event is grounded in timing as the
+observable -- power and frequency-based side channels are out of scope.
+
+**Safety.** For the bulk of the CBMC specifications which do not capture
+functional correctness, memory- and type-safety are implicit in the CBMC configuration
+and not explicitly stated in the CBMC contracts. Additional clauses in the CBMC contracts
+(such as memory assumptions and footprint) need not be human-validated except for the top-level
+API, since their adequacy is judged by the machine-checked compositionality against their
+call-sites. The fidelity of CBMC specifications therefore reduces to whether the top-level
+CBMC specifications capture the footprint and assumptions of the top-level FIPS 203 API,
+and whether CBMC is correctly configured to include memory- and type-safety implicitly in
+each function contract.
+
+All specifications are written in a high-level mathematical style (HOL Light) or a simple
+declarative style using auditable macros (CBMC), both far simpler than the implementations.
+The approach to Hoare-style correctness specifications in HOL Light has been carefully audited
+and is extensively used for numerous other assembly kernels in s2n-bignum, without infidelities
+being detected. For CBMC, modular proof means that an insufficient specification on a callee will
+cause the caller's proof to fail. Top-level CBMC specifications are straightforward as their inputs
+and outputs are merely byte buffers of standardized lengths; all these buffer length values are
+defined in [params.h](mlkem/src/params.h) and easily auditable to align with FIPS 203.
+
+Residual risks remain: The constant-time specification style in HOL Light/s2n-bignum could
+fail to detect a practically relevant class of variable-time execution. A supposedly constant-time
+specification could erroneously include secret data as a parameter to the event-generating function.
+A misconfiguration of CBMC could cause the implicit expectation of memory- and type-safety being
+included in the CBMC specifications not to hold. Or, a CBMC proof could incorrectly disable a safety
+check.
+
+**Potential improvements.**
+- Derive the CBMC configuration from a machine- and human-readable source documenting
+  the desired configuration options and their
+  meaning. ([#1599](https://github.com/pq-code-package/mlkem-native/issues/1599))
+- Align the notion of event-generating instruction in s2n-bignum with the set of instructions
+  outside of the scope of DIT/DOIT ([s2n-bignum/#361](https://github.com/awslabs/s2n-bignum/issues/361)).
+- Highlight the notion of public vs. secret data more explicitly in the s2n-bignum constant-time
+  specifications, rather than implicitly in the list of parameters to the event-generating function ([s2n-bignum/#362](https://github.com/awslabs/s2n-bignum/issues/362)).
+- Provide a document which explains the specification style for correctness and constant-time
+  properties in HOL Light. ([#1600](https://github.com/pq-code-package/mlkem-native/issues/1600))
+
+### A4. Specification consistency
+
+Do the individual specifications compose to a coherent claim about the whole system? The
+risks are that a contract assumed during proof differs from the contract that is proved,
+that a contract assumed during proof is never proved, or that the bridge between the
+CBMC and HOL Light verification stacks introduces an inconsistency.
+
+#### CBMC compositionality
+
+When CBMC proves a function F using the contract of a callee G (via `USE_FUNCTION_CONTRACTS`),
+it assumes G's contract as an axiom. The contract it assumes must be the same contract that
+is proved for G in G's own proof. In mlkem-native, this is enforced structurally: each
+function has a single contract written at its declaration site, and CBMC uses that same
+contract both when proving the function and when assuming it as a callee. There is no
+mechanism by which the "assumed" and "proved" versions can diverge.
+
+This structural guarantee is reinforced by the
+[check-contracts](scripts/check-contracts) script, which looks for CBMC functions with a
+contract but no proof. Additionally, bounds assertions in CBMC specifications are mirrored
+by runtime debug assertions at the beginning and end of the respective function, so that if a
+function is not covered by CBMC -- erroneously or deliberately -- tests in debug mode can
+still catch gross mismatches between what the caller provides/assumes and what the callee
+assumes/provides.
+
+Residual risks remain: a function's proof could be present but not run in CI (e.g., due to
+a CI configuration error), or a function's proof could be absent but
+[check-contracts](scripts/check-contracts) could fail to detect it due to a bug or
+misconfiguration.
+
+#### The bridge between CBMC and HOL Light
+
+Where C code calls into assembly, the assembly function has both a HOL Light specification
+(proved against the object code) and a CBMC contract (assumed by CBMC when proving the C
+caller). These two specifications are written in different languages and are currently
+kept in sync by hand. Failure to do so may invalidate the safety and correctness claims
+for mlkem-native.
+
+For each assembly function, the CBMC specification is typically a subset of the HOL Light
+specification obtained by removing aspects of functional correctness -- which, as discussed
+above, are not yet covered in CBMC. What remains are statements about memory footprint,
+arithmetic bounds, and constant tables.
+
+**Example.** For the AArch64 NTT:
+
+- HOL Light proves: If the input coefficients satisfy `abs(ival(x i)) <= &8191`, then the output
+  satisfies `abs(ival zi) <= &23594` and `(ival zi == forward_ntt (ival o x) i) (mod &3329)`.
+  In other words, we provide a description of the underlying modular arithmetic function (here, the NTT),
+  plus a bound on the concrete being computed.
+- The CBMC contract on `mlk_ntt_asm` simplifies this to the mere bounds assertions
+  `requires(array_abs_bound(p, 0, MLKEM_N, 8192))`
+  and `ensures(array_abs_bound(p, 0, MLKEM_N, 23595))`, omitting the description of the
+  functional behavior.
+- HOL Light assumes that auxiliary arguments point to pre-computed constant tables for
+  the NTT: `C_ARGUMENTS [a; z_12345; z_67] s /\ ntt_constants z_12345 z_67 s`. CBMC similarly
+  assumes `requires(twiddles12345 == mlk_aarch64_ntt_zetas_layer12345)` and `requires(twiddles56 ==
+  mlk_aarch64_ntt_zetas_layer67)` for the C arguments used for the constant tables.
+
+**Mitigations.**
+Both the HOL Light proofs and the CBMC contracts contain comments of the form
+`/* This must be kept in sync with the HOL Light specification in ... */`
+(or vice versa) to flag the manual dependency and provide a review checkpoint.
+Moreover, constant tables used in the HOL Light and C code are auto-generated from the same code
+in [`scripts/autogen`](scripts/autogen). This code is run as part of CI and thereby establishes that
+constant tables remain in sync.
+The full test suite exercises the native backend code paths, catching
+gross mismatches. Furthermore, CBMC proves the C wrapper correct *assuming* the assembly
+contract, so an inconsistency between the wrapper's precondition and the assembly's
+precondition would typically cause the wrapper proof to fail.
+
+Despite these safeguards, residual risks remain. A transcription error could cause the CBMC
+contract on the assembly function to fail to faithfully reflect the HOL Light specification
+-- e.g., an off-by-one in a bound, a missing aliasing constraint, or a wrong constant. An
+ABI mismatch could cause the C calling convention assumed by the CBMC contract to differ
+from the actual register/stack layout used by the assembly. And a semantic gap between the
+CBMC contract language and HOL Light's logic -- for example, differing signed vs. unsigned
+interpretation of bounds -- could cause the bridge to be unsound.
+
+**Potential improvements.**
+- Establish a machine-checked link between the HOL Light specifications and the CBMC
+  contracts. ([#1601](https://github.com/pq-code-package/mlkem-native/issues/1601))
+
+---
+
+## Gap B: Do the formal models match the actual systems?
+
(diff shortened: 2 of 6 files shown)


SOUNDNESS.md
```diff
@@ -0,0 +1,484 @@
...
+
+**Residual risks.** The CBMC TCB is orders of magnitude larger than HOL Light's. There is
+no independent proof-checking mechanism analogous to Candle or HOLTrace. The soundness
+guarantee for the C proofs is therefore fundamentally weaker than for the assembly proofs.
+
+## Summary of risks
+
+| # | Risk | Primary mitigation |
+|---|------|--------------------|
+| A0 | Not all code has formal specs | Intrinsics and non-x86/AArch64 backends: testing only; *planned replacement with assembly* |
```
mkannwischer: We will have replaced x86_64 intrinsics shortly - needs updating.

SOUNDNESS.md
```diff
@@ -0,0 +1,484 @@
...
+  harmful compiler optimizations.
+
+**Future work.**
+- Add memory-safety proof for rejection sampling routines. (TODO: Ref tracking issue)
+- Introduce additional verification tooling that allows us to express functional correctness
+  and constant-time properties for the C code. (TODO: Ref tracking issue)
+
+### A2. Specification faithfulness
+
+For the specifications that do exist, does their formal meaning capture their intent?
```
hanno-becker: One can argue this is rather part of Gap B..

SOUNDNESS.md
hanno-becker: x86 Keccak is unknown because of the AVX512 issue.

SOUNDNESS.md
```diff
@@ -0,0 +1,484 @@
...
+## Gap C: Is the proof infrastructure sound?
+
+### C1. HOL Light kernel and OCaml runtime
+
+HOL Light has a small trusted kernel (~400 lines of OCaml) implementing 10 primitive
+inference rules and 3 axioms. All theorems must be constructed through this kernel (the
+LCF architecture). No soundness bugs have been found since 2003.
+
+Independent reassurance is provided by Candle (a formally verified prover for the same
+logic) and HOLTrace (independent proof trace checking).
```
mkannwischer: This could use some links and possibly attribution?

tautschnig:
A first comment, independent of actual contents of the doc: I think this needs to get linked from `README.md` to make it discoverable.

SOUNDNESS.md
```diff
@@ -0,0 +1,546 @@
+[//]: # (SPDX-License-Identifier: CC-BY-4.0)
+
+# Formal Verification in mlkem-native: Scope, Assumptions, Risks
+
+Formal verification is never absolute. Every verification effort links formal objects --
```
tautschnig: This is a neat introduction, but I'd suggest to first have an executive summary that states what this document is about.

SOUNDNESS.md
```diff
@@ -0,0 +1,546 @@
...
+maps out what is proved about mlkem-native, what is assumed, and where the gaps and risks lie.
+
+The parts of the analysis pertaining to HOL Light and s2n-bignum are largely shared with
+the corresponding [HOL Light / s2n-bignum Soundness
+Document](https://github.com/awslabs/s2n-bignum/blob/main/doc/s2n_bignum_soundness.md).
+
+## Overview
+
+Our goal is to provide evidence rooted in formal reasoning for the statement that
+"mlkem-native implements the FIPS 203 standard and does not leak secrets through timing variations".
```
tautschnig: There should be a link to FIPS 203.

SOUNDNESS.md
```diff
@@ -0,0 +1,546 @@
...
+
+The narrative for such argument is typically the following:
```
tautschnig: Apart from s/such argument/such an argument/ I am struggling to understand whether "typically" tells me that "this is how everyone else does it" or "we do the following, and this is in line with how everyone else does it."

SOUNDNESS.md
```diff
@@ -0,0 +1,546 @@
...
+
+  "Here's the real system."                           "Here's how we think it can
+                                                       be formally approximated."
+
+```
+
+This methodology introduces three fundamental soundness gaps:
+- Gap A: Does the formal specification capture what we want to say about the real system?
+- Gap B: Is the formal model a faithful reflection of the real system?
+- Gap C: Is the formal argument trustworthy for why the formal model satisfies its specification?
```
tautschnig, line 65: I'd like more precise language here. What does the "real system" even refer to? Is that the implementation as C code (with its semantics rooted in the C standard), or is it the actual compiled and deployed software operating in some place (in our universe). In other words: does the "real system" suffer from compiler bugs, broken hardware, cosmic radiation?

SOUNDNESS.md
```diff
@@ -0,0 +1,546 @@
...
+contracts track coefficient bounds through the arithmetic pipeline, and overflow in
+intermediate computations would be flagged as undefined behavior. Constant-time properties
+are tested empirically using valgrind with a variable-latency patch across many compilers
+(GCC 4.8–15, Clang 14–21) and optimization levels (-O0 through -Ofast), and the C code
+uses value barriers to prevent harmful compiler optimizations.
+
+**Future work.**
+- Add memory-safety proof for rejection sampling routines. (TODO: Ref tracking issue)
+- Introduce additional verification tooling that allows us to express functional correctness
+  and constant-time properties for the C code. (TODO: Ref tracking issue)
```
tautschnig: Acknowledging that this is a tangent: do these aspects really require additional tooling? Functional correctness should mainly be a problem of extending existing contracts with more properties. Constant-time properties require relational verification, and Vladimir Klebanov has published a bunch of papers using CBMC for this purpose.

SOUNDNESS.md
```diff
@@ -0,0 +1,546 @@
...
+
+See [s2n-bignum soundness, B2](https://github.com/awslabs/s2n-bignum/blob/main/doc/s2n_bignum_soundness.md#b2-elf-object-code-loader)
+for further details on the ELF loader.
+
+**Reassembly risk.** When mlkem-native's `.S` files are assembled on a different system --
+whether by mlkem-native's own build, by a downstream consumer such as AWS-LC, or by a
+cross-compilation toolchain -- there is currently no systematic check that the resulting
+object code matches the bytes the proofs were verified against.
+Assembler bugs, version differences, or different assembler dialects can and
+do produce different object code. This has been observed in practice.
```
tautschnig: Is there anything you can reference with regard to "has been observed in practice" -- like GitHub issues or other public documents?

SOUNDNESS.md
```diff
@@ -0,0 +1,546 @@
...
+high-assurance deployments in physically hostile environments, additional countermeasures
+at the hardware or protocol level would be needed.
+
+### B5. C semantics model fidelity (CBMC)
+
+CBMC translates C source code into an internal representation and then into SMT formulas.
+This translation must faithfully model the semantics of C as approximated by the collective
+interpretation of C by common compilers. Bugs in CBMC's C-to-SMT translation could cause it
+to miss real undefined behavior and to accept incorrect code.
+
```
tautschnig, line 459: It could actually go both ways: we could also be wrongly be reporting issues.

SOUNDNESS.md
```diff
@@ -0,0 +1,546 @@
...
+**Mitigations.**
+- CBMC is a mature, widely-used tool with an active community and extensive
+  test suite.
+- mlkem-native's CBMC proofs are run on every CI commit, providing continuous regression
+  testing.
```
tautschnig: While that is true, it would just mean running the same (possibly buggy) CBMC version every time. You might also add that CBMC is being kept up to date.

SOUNDNESS.md
```diff
@@ -0,0 +1,546 @@
...
+### B5. C semantics model fidelity (CBMC)
+
+CBMC translates C source code into an internal representation and then into SMT formulas.
+This translation must faithfully model the semantics of C as approximated by the collective
+interpretation of C by common compilers. Bugs in CBMC's C-to-SMT translation could cause it
+to miss real undefined behavior and to accept incorrect code.
+
+**Mitigations.**
+- CBMC is a mature, widely-used tool with an active community and extensive
+  test suite.
```
tautschnig: The same should be said about the underlying SMT solvers being used.

SOUNDNESS.md
```diff
@@ -0,0 +1,546 @@
...
+    cannot verify that the compiled code actually performs the zeroization.
+  - **Value barriers.** mlkem-native's C code uses value barriers (`mlk_value_barrier_u32`
+    etc.) to prevent the compiler from introducing secret-dependent branches or
+    variable-time instructions. These barriers rely on inline assembly or volatile
+    semantics to defeat compiler optimization. CBMC sees the source-level semantics
+    (where the barrier is an identity function) and cannot verify that the compiled code
+    preserves the constant-time property.
+- **Memory model.** CBMC's memory model uses a flat, object-based representation that may
+  not match the target platform's memory layout. In particular, CBMC does not model
+  platform-specific alignment requirements, padding, or struct layout beyond what the C
```
tautschnig: CBMC actually does model padding and struct layout the same way the compiler will apply it (but then those are not specified in the standard, and CBMC will only do verification in line the platform's default compiler, not for every possible compiler).

SOUNDNESS.md
```diff
@@ -0,0 +1,546 @@
...
+verification: the proofs reason about an idealized machine, not the physical device. For
+high-assurance deployments in physically hostile environments, additional countermeasures
+at the hardware or protocol level would be needed.
+
+### B5. C semantics model fidelity (CBMC)
+
+CBMC translates C source code into an internal representation and then into SMT formulas.
+This translation must faithfully model the semantics of C as approximated by the collective
+interpretation of C by common compilers. Bugs in CBMC's C-to-SMT translation could cause it
+to miss real undefined behavior and to accept incorrect code.
```
tautschnig: In keeping with what was said earlier about other verification work: it may be just fair to say that such CBMC bugs have actually been seen.

SOUNDNESS.md
```diff
@@ -0,0 +1,546 @@
...
+  not match the target platform's memory layout. In particular, CBMC does not model
+  platform-specific alignment requirements, padding, or struct layout beyond what the C
+  standard requires. For mlkem-native this is low-risk since the code avoids
+  platform-dependent memory tricks, but it is a gap in principle.
+
+### B6. Execution-environment assumptions
+
+Both the HOL Light ISA model and CBMC's C semantics model make assumptions about
+the execution environment. If these assumptions are violated at runtime, the proofs'
+guarantees do not hold.
```
tautschnig: I'd say "need not hold" -- taking the 64-bit vs 32-bit aspect that you mention right away: I don't think there is evidence that any proofs break on 32-bit platforms, they just haven't been attempted.

SOUNDNESS.md
```diff
@@ -0,0 +1,546 @@
...
+- **SMT solvers** (Z3, Bitwuzla): These are complex software systems. A solver bug
+  directly compromises soundness. Unlike HOL Light's LCF architecture, where bugs in
+  proof automation cannot compromise soundness, a solver bug in CBMC's backend directly
+  invalidates the proof.
+- **The C compiler used by CBMC**: CBMC uses a C preprocessor and parser; bugs in these
+  components could affect the analysis.
+
+**Mitigations.**
+
+- CBMC has been in development since 2001 and is widely used in industry (including
```
tautschnig: I'm not entirely sure about 2001, checking with @kroening.

SOUNDNESS.md
hanno-becker: Changed to: "The narrative for the argument the following, which is common in formal verification:"

SOUNDNESS.md
hanno-becker: I mean the latter

hanno-becker: Do you have suggestions? Above I say "real physical system", but spelling this out everywhere does not read well.

SOUNDNESS.md
hanno-becker: I don't have experience with this, but my impression is that it would be difficult/impossible to express something as complex as the entire FIPS203 spec in CBMC?

SOUNDNESS.md
hanno-becker: Yes, we have observed the Mac assembler swapping around addition arguments to reduce code size on x86. I will see if I can dig it up.

SOUNDNESS.md
hanno-becker: Yes, but that's not a soundness risk?

SOUNDNESS.md
hanno-becker: Good point!

SOUNDNESS.md
hanno-becker: True, fixed.

hanno-becker: On second thought, this rather belongs to Gap C.

SOUNDNESS.md
hanno-becker: Very important point. I added a paragraph.

SOUNDNESS.md
hanno-becker: True. I added "...; this has happened in the past."

SOUNDNESS.md
hanno-becker: Yes, I agree -- changed.

SOUNDNESS.md
hanno-becker: Changed to >20 years for now.

SOUNDNESS.md
hanno-becker: Done

README.md
```diff
@@ -61,10 +61,12 @@ All C code in [mlkem/src/*](mlkem) and [mlkem/src/fips202/*](mlkem/src/fips202)
...
 
-All AArch64 assembly and all x86_64 ML-KEM arithmetic assembly is proved functionally correct at the object-code level. This uses the [HOL-Light](https://github.com/jrh13/hol-light)
-interactive theorem prover and the [s2n-bignum](https://github.com/awslabs/s2n-bignum/) verification infrastructure (which includes models of the
-relevant parts of the Arm and x86 architectures). See [proofs/hol_light](proofs/hol_light) for details.
-The x86_64 FIPS-202 (Keccak) backend is still in C intrinsics and not yet covered by proof.
+All AArch64 assembly is proved functionally correct at the object-code level. This uses the [HOL-Light](https://github.com/jrh13/hol-light)
+interactive theorem prover and the [s2n-bignum](https://github.com/awslabs/s2n-bignum/) verification infrastructure (which includes a model of the
+relevant parts of the Arm architecture). See [proofs/hol_light/aarch64](proofs/hol_light/aarch64) for details.
+
+See [SOUNDNESS.md](SOUNDNESS.md) for a detailed analysis of the scope, assumptions and risks of the formal verification
```
mkannwischer: Shall we add a `Formal verification is never absolute.` before that? That captures it well, I think.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+**What is covered.**
+
+- All C code in the core library has CBMC contracts.
+- All AArch64 assembly routines have HOL Light specifications.
+- All x86_64 assembly routines have HOL Light specifications.
+
+**What is NOT covered.**
+
+- **x86_64 intrinsics-based C code.** The native Keccak-F1600 implementations for
+  x86_64 are written in intrinsics and not covered by any specification.
```
mkannwischer: We should not forget to remove that; I expect we can merge Keccak asm shortly.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+- **Other backends** (RISC-V RVV, experimental Armv8.1-M MVE, PowerPC):
```
mkannwischer: PowerPC is not merged yet. Maybe we keep RVV and MVE - but phrase them as examples here so that we are not out of date so quickly.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+  these are not covered by specification.
+
+The full test suite (functional tests, KAT, ACVP, unit tests) validates functional
+correctness empirically across all platforms and configurations, but there is currently
+no automatic check that every C function has a CBMC contract or that every assembly
```
mkannwischer: All C function having a contract is not the important part here.
C functions without contracts are still fine to be inlined into the proof of the caller.
If I understand correctly, the risk is that the C function is called by contract, but there actually is no contract because it defaults to an empty contraact.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+routine has a HOL Light proof. A function could slip through without specification.
+
+**Potential improvements.**
+- Add automatic proof coverage check to CI. (TODO: Ref tracking issue)
+- Replace all intrinsics-based code with pure assembly, bringing it into the scope of
+  HOL Light verification. (TODO: Ref tracking issue)
```
mkannwischer: Should not forget to remove that should we indeed merge Keccak today.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+- Extend assembly verification to all backends. (TODO: Ref tracking issue)
+
+### A2. Coverage: Depth
+
+Do the specifications have the desired depth? The risk is that undesired behavior occurs
+which is outside the scope of specification.
+
+Currently, we aim for mlkem-native to be **functionally correct** (the code computes the
```
mkannwischer: Remove "Currently".
Should we mention something about memory safety here?

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+  proved.
+
+These gaps are mitigated in complementary ways. The full test suite (functional tests, KAT,
+ACVP, unit tests) validates functional correctness empirically across all platforms and
+configurations. For arithmetic correctness specifically, the most subtle bugs in ML-KEM
+implementations are rare overflows in the optimized polynomial arithmetic -- precisely the
+kind of bug that the type-safety and integer-overflow proofs are designed to catch: the CBMC
+contracts track coefficient bounds through the arithmetic pipeline, and overflow in
+intermediate computations would be flagged as undefined behavior. Constant-time properties
+are tested empirically using valgrind with a variable-latency patch across many compilers
```
mkannwischer:
```suggestion
are tested empirically using valgrind across many compilers
```

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+(GCC 4.8–15, Clang 14–21) and optimization levels (-O0 through -Ofast), and the C code
```
mkannwischer: Replace list of compilers with a reference to https://github.com/pq-code-package/mlkem-native/blob/main/.github/workflows/ct-tests.yml to avoid being outdated immediately.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+**Constant-time.** The informal intent is that mlkem-native does not
+leak secrets through timing side channels. Since we only capture constant-time properties
+at the assembly level, faithfulness requires that the HOL Light constant-time specification
+style faithfully captures constant-time execution.
+
+The s2n-bignum formal model approximates constant-time'ness as follows: It introduces the
+notion of *microarchitectural events* that flag uses of instructions with potentially
+data-dependent timing (e.g., integer division). The constant-time proofs then establish
+that (a) the sequence of instructions executed and all memory addresses accessed are independent
+of secret inputs, and that (b) no variable-timing microarchitectural events occur.
```
mkannwischer: How is defined what inputs are secret? I have not seen any secret/public typing yet.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+
+The formal notion of constant-time'ness used in s2n-bignum does not and cannot guarantee
+that the hardware executes those instructions in constant time: microarchitectural effects
+(variable-latency instructions, Hertzbleed-style frequency scaling) could still leak
+information. Some hardwaren provides opt-in guarantees for a listed set of instructions -- ARM's DIT (Data Independent
+Timing, Armv8.4-A onwards) and Intel's DOITM (Data Operand Independent Timing Mode, Ice Lake onwards) -- but even with
```
mkannwischer:
```suggestion
information. Some hardwaren provides opt-in guarantees for a listed set of instructions -- Arm's DIT (Data Independent
Timing, Armv8.4-A onwards) and Intel's DOIT (Data Operand Independent Timing, Ice Lake onwards) -- but even with
```

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+
+The s2n-bignum formal model approximates constant-time'ness as follows: It introduces the
+notion of *microarchitectural events* that flag uses of instructions with potentially
+data-dependent timing (e.g., integer division). The constant-time proofs then establish
+that (a) the sequence of instructions executed and all memory addresses accessed are independent
+of secret inputs, and that (b) no variable-timing microarchitectural events occur.
+
+The formal notion of constant-time'ness used in s2n-bignum does not and cannot guarantee
+that the hardware executes those instructions in constant time: microarchitectural effects
+(variable-latency instructions, Hertzbleed-style frequency scaling) could still leak
```
mkannwischer: speculative execution should be named here

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+these enabled, coverage is limited to specific instruction sets, and power- or frequency-based side channels remain
+outside scope. See [^s2n-bignum-soundness] for full details.
+
+**Safety.** For the bulk of the CBMC specifications which do not capture
+functional correctness, memory- and type-safety are implicit in the CBMC configuration
+and not explicitly stated in the CBMC contracts. Additional clauses in the CBMC contracts
+(such as memory assumptions and footprint) need not be human-validated except for the top-level
+API, since their adequacy is judged by the machine-checked compositionality against their
+call-sites. The fidelity of CBMC specifications therefore reduces to whether the top-level
+CBMC specifications capture the footprint and assumptions of the top-level FIPS 203 API,
```
mkannwischer, line 231: Maybe be worth to mention here that the top-level specs are straightforward as inputs for ML-KEM are arbitrary fixed-length byte arrays.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+and whether CBMC is correctly configured to include memory- and type-safety implicitly in
+each function contract.
+
+All specifications are written in a high-level mathematical style (HOL Light) or a simple
+declarative style using auditable macros (CBMC), both far simpler than the implementations.
+The Hoare-style correctness specifications in HOL Light have been carefully audited and are
```
mkannwischer: What specifications are you talking about here? 
I thought these are the ML-KEM specs, but then `used for numerous other assembly kernels` does not make sense. Should be clarified.

SOUNDNESS.md
mkannwischer: a not here - there is a paragraph lower where it fits better.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+extensively used for numerous other assembly kernels in s2n-bignum, without infidelities
+being detected. For CBMC, modular proof means that a wrong specification on a callee will
+typically cause the caller's proof to fail. Top-level CBMC specifications don't use magic
```
mkannwischer: `typically` is imprecise here. In what cases would it not fail? I can only think of contracts being too strong.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+constants; all buffer length values are defined in [params.h](mlkem/src/params.h) and
```
mkannwischer: What do you mean by magic constants in this case? I'd rather mention what I suggested above here:  hat the top-level specs are straightforward as inputs for ML-KEM are arbitrary fixed-length byte arrays.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+it assumes G's contract as an axiom. The contract it assumes must be the same contract that
+is proved for G in G's own proof. In mlkem-native, this is enforced structurally: each
+function has a single contract written at its declaration site, and CBMC uses that same
+contract both when proving the function and when assuming it as a callee. There is no
+mechanism by which the "assumed" and "proved" versions can diverge.
+
+This structural guarantee is reinforced by the
+[check-contracts](scripts/check-contracts) script, which looks for CBMC functions with a
+contract but no proof. Additionally, bounds assertions in CBMC specifications are mirrored
+by executable checks at the beginning and end of the respective function, so that if a
```
mkannwischer: executable checks -> runtime debug assertions?

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+  and memory footprint).
+
+Both the HOL Light proofs and the CBMC contracts contain comments of the form
+`/* This must be kept in sync with the HOL Light specification in ... */`
+(or vice versa) to flag the manual dependency.
+
+**Example.** For the AArch64 NTT:
+
+- HOL Light proves: if input coefficients satisfy `abs(ival(x i)) <= 8191`, then output
+  satisfies `abs(ival zi) <= 23594` and `ival zi ≡ forward_ntt(ival ∘ x)(i) (mod 3329)`.
```
mkannwischer: Can we stick to actual HOL-Light syntax here:
`(ival zi == avx2_forward_ntt (ival o x) i) (mod &3329) `

mkannwischer: the other snippets also need slight changes

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+caller). These two specifications are written in different languages and maintained
+independently. Their consistency is checked manually, as follows, requiring that:
+
+- The CBMC contract on the assembly function reflects aspects of the HOL Light
+  specification relevant to memory- and type-safety (typically bounds assumptions
+  and memory footprint).
+
+Both the HOL Light proofs and the CBMC contracts contain comments of the form
+`/* This must be kept in sync with the HOL Light specification in ... */`
+(or vice versa) to flag the manual dependency.
```
mkannwischer: It seems we have forgotten this comment in some places - I will open a PR shortly to fix that.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+hand-written from the ARM and Intel architecture reference manuals. Errors in those
+references, misunderstandings, or transcription mistakes could silently invalidate proofs.
+
+**Mitigation: co-simulation testing.** s2n-bignum's CI includes a co-simulation test
+(`simulator.ml` + `simulator.c`) that repeatedly picks random instruction encodings and
+random register/flag states, decodes them, executes them both symbolically through the
+formal model and natively on real hardware, and compares results. This exercises both the
+ISA semantics and the decoder. It in CI and covers all register-to-register instruction
+forms with randomized operands, as well as memory-accessing instructions via dedicated
+harnesses for various addressing modes.
```
mkannwischer: "It in CI" is not a proper sentence.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+
+### B2. Object code verification and reassembly (assembly)
+
+The HOL Light proofs work at the level of object-code byte sequences, not the assembly source.
+This takes the assembler out of the TCB of the proof. However, two risks remain.
+
+**ELF loader.** An OCaml ELF loader extracts the `.text` section (and, where applicable,
+the `.rodata` section) from each object file for verification. If it extracts the wrong
+bytes, the proof applies to different code than what runs in production. This risk is
+mitigated by the fact that the proof engineer must know the exact byte sequence to write
```
mkannwischer: maybe rephrase this.

The byte sequence is embedded in the proof script including human-readable decoding .... the proof engineer should know the decoding, but expecting them to know the exact same byte sequence is maybe a little strong.

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+
+See [^s2n-bignum-soundness] for further details on the ELF loader.
+
+**Reassembly risk.** When mlkem-native's `.S` files are assembled on a different system --
+whether by mlkem-native's own build, by a downstream consumer such as AWS-LC, or by a
+cross-compilation toolchain -- there is currently no systematic check that the resulting
+object code matches the bytes the proofs were verified against.
+Assembler bugs, version differences, or different assembler dialects can and
+do produce different object code (for example, it has been observed that some x86
+compilers swap operands of AVX2 `VPADD` instruction to reduce code size).
```
mkannwischer:
```suggestion
assemblers swap operands of AVX2 `VPADD` instruction to reduce code size).
```

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+at the hardware or protocol level would be needed.
+
+### B5. C semantics model fidelity
+
+CBMC gives meaning to C by translating C source into an internal representation and then into SMT formulas. Bugs in
+CBMC's C-to-SMT translation could cause it to miss undefined behavior and to accept incorrect code; this has happened in
+the past. However, CBMC is a mature, widely-used tool with an active community and extensive test suite, and
+mlkem-native uses the latest CBMC version.
+
+Also, the C language has many conformant implementations. For example, the width of pointer types could be
```
mkannwischer, line 460: Should this be explicility structured as a list of mitigations like in other sections?

SOUNDNESS.md
```diff
@@ -0,0 +1,517 @@
...
+- Run CBMC proofs against 16-bit and 32-bit C compilers.
+
+---
+
+## Gap C: Is the proof infrastructure sound?
+
+There is risk that the formal statements made about the formal model are proved,
+yet not true, because of an unsoundness in the underlying proof infrastructure.
+
+### C1. HOL Light kernel and OCaml runtime
```
mkannwischer, line 489: paragraph does not state anything about the OCaml runtime

SOUNDNESS.md
mkannwischer: Fixed: https://github.com/pq-code-package/mlkem-native/pull/1593

README.md
hanno-becker: Yes, good idea.

SOUNDNESS.md
hanno-becker: Agree

SOUNDNESS.md
hanno-becker: Ack

SOUNDNESS.md
hanno-becker: I agree.

SOUNDNESS.md
hanno-becker: Yes

SOUNDNESS.md
hanno-becker: > We aim for mlkem-native to be **functionally correct** (the code computes the
right answer as per FIPS 203), **memory-safe** (it only accesses memory within
the bounds of what is provided or allocated) and **constant-time** (no
secret-dependent timing variation).

?

SOUNDNESS.md
hanno-becker: I mean the way HOL Light assembly specs are phrased

hanno-becker: > The apprach to Hoare-style correctness specifications in HOL Light has been carefully audited
and is extensively used for numerous other assembly kernels in s2n-bignum

Clearer?

SOUNDNESS.md
hanno-becker: Fair point. "Wrong" is imprecise here; the adequacy of a contract _is_ whether it works for the caller.

> For CBMC, modular proof means that an insufficient specification on a callee will
cause the caller's proof to fail. 

Ok?

SOUNDNESS.md
hanno-becker: > Top-level CBMC specifications are straightforward as their inputs
and outputs are merely byte buffers of standardized lengths; all these buffer lengths values are
defined in [params.h](mlkem/src/params.h) and easily auditable to align with FIPS 203.

Better?

SOUNDNESS.md
hanno-becker: Thanks!

SOUNDNESS.md
hanno-becker: "provide" instead of "know"?

SOUNDNESS.md
hanno-becker: At present, the text isn't uniform in this regard... I meandered between bullets and prose. At the moment I am inclined to accept a bit of variance, but I expect us to iterate towards a more uniform format in future versions.

mkannwischer: okay

SOUNDNESS.md
mkannwischer: okay

SOUNDNESS.md
hanno-becker: This is part of the `subroutine_signatures`

mkannwischer: Really? I have not seen any function that declares a public input. Do you have an example how it looks like?

hanno-becker: You are right, I mixed things up. There is no definition of what is secret or public. The crux is what appears as a parameter in the event-generating function -- this is typically a bunch of pointers, but no data variable. Explaining this falls under https://github.com/pq-code-package/mlkem-native/issues/1600

hanno-becker: I tried to clarify things in https://github.com/pq-code-package/mlkem-native/pull/1582/commits/630cd8dcde93c41abd4a0792ed856f3cb61d3182

SOUNDNESS.md
```diff
@@ -284,30 +296,37 @@ misconfiguration.
...
+  assumes `requires(twiddles12345 == mlk_aarch64_ntt_zetas_layer12345)` and `requires(twiddles56 ==
+  mlk_aarch64_ntt_zetas_layer67)` for the C arguments used for the constant tables.
 
 **Mitigations.**
-The "keep in sync" comments in both the CBMC and the HOL Light specifications provide a
-review checkpoint. The full test suite exercises the native backend code paths, catching
+Both the HOL Light proofs and the CBMC contracts contain comments of the form
+`/* This must be kept in sync with the HOL Light specification in ... */`
+(or vice versa) to flag the manual dependency and provide a review checkpoint.
+Moreover, constant tables used in the HOL Light and C code are generated from the same code in
```
mkannwischer: Mention that this is automatically checked in CI

SOUNDNESS.md
```diff
@@ -0,0 +1,543 @@
...
+This methodology introduces three fundamental soundness gaps:
+- Gap A: Does the formal specification capture what we want to say about the real system?
+- Gap B: Is the formal model a faithful reflection of the real system?
+- Gap C: Is the formal argument trustworthy for why the formal model satisfies its specification?
+
+For mlkem-native, this structure is instantiated twice -- once for each verification
+stack.
+
+- **[^CBMC]** for the C code: memory safety, type safety, and absence of undefined behavior.
+- **[^HOL-Light] + [^s2n-bignum]** for the assembly backends: functional correctness, memory safety, and secret-independent
```
mkannwischer: This does not get rendered well. 
```suggestion
- **CBMC[^CBMC]** for the C code: memory safety, type safety, and absence of undefined behavior.
- **HOL-Light[^HOL-Light] + s2n-bignum[^s2n-bignum]** for the assembly backends: functional correctness, memory safety, and secret-independent
```

SOUNDNESS.md
```diff
@@ -0,0 +1,543 @@
...
+the corresponding [^s2n-bignum-soundness].
+
+## Overview
+
+Formal verification is never absolute. Every verification effort links formal objects --
+specifications and models -- to informal, real-world requirements and systems. This document
+maps out what is proved about mlkem-native, what is assumed, and where the gaps and risks lie.
+
+Our goal is to provide evidence rooted in formal reasoning for the statement that
+"mlkem-native implements the [^FIPS203] standard and does not leak secrets through timing variations".
```
mkannwischer:
```suggestion
"mlkem-native implements the FIPS203 standard[^FIPS203] and does not leak secrets through timing variations".
```

SOUNDNESS.md
```diff
@@ -0,0 +1,543 @@
...
+kind of bug that the type-safety and integer-overflow proofs are designed to catch: the CBMC
+contracts track coefficient bounds through the arithmetic pipeline, and overflow in
+intermediate computations would be flagged as undefined behavior. Constant-time properties
+are tested empirically using valgrind across many compilers and optimization levels (see
+the corresponding [CI job](.github/workflows/ct-tests.yml) for
+the full list), and the C code uses value barriers to prevent harmful compiler optimizations.
+
+**Potential improvements.**
+- Add memory-safety proof for rejection sampling routines. ([#1596](https://github.com/pq-code-package/mlkem-native/issues/1596))
+- Add automatic extraction of compiler coverage documentation from CI. (TODO: Ref tracking issue)
```
mkannwischer: leftover TODO

SOUNDNESS.md
```diff
@@ -0,0 +1,543 @@
...
+  Where present, these are not yet covered by specification.
+
+The full test suite (functional tests, KAT, ACVP, unit tests) validates functional
+correctness empirically across all platforms and configurations, but there is currently
+no automatic coverage check for CBMC or HOL Light. A function or configuration could slip
+through without being covered by specification.
+
+**Potential improvements.**
+- Add automatic proof coverage check to CI. ([#1423](https://github.com/pq-code-package/mlkem-native/issues/1423), [#1594](https://github.com/pq-code-package/mlkem-native/issues/1594))
+- Replace all intrinsics-based code with pure assembly, bringing it into the scope of
```
mkannwischer: Should this be removed?
I am not sure if RV64 intrinsics shoul be replaced too. I remember Markku saying that intrinsics are default for RVV development.

SOUNDNESS.md
```diff
@@ -0,0 +1,543 @@
...
+check.
+
+**Potential improvements.**
+- Derive the CBMC configuration from a machine- and human-readable source documenting
+  the desired configuration options and their
+  meaning. ([#1599](https://github.com/pq-code-package/mlkem-native/issues/1599))
+- Align the notion of event-generating instruction in s2n-bignum with the set of instructions
+  outside of the scope of DIT/DOIT.
+- Highlight the notion of public vs. secret data more explicitly in the s2n-bignum constant-time
+  specifications, rather than implicitly in the list of parameters to the event-generating function.
```
mkannwischer: Should these two also get a tracking issue?

mkannwischer (approved):
Thanks @hanno-becker. This is a great start and it was long overdue that we write such a document. 

I left a few last nits that you may still want to address before merging. Please also squash everything into one commit.

SOUNDNESS.md
```diff
@@ -0,0 +1,543 @@
...
+**Residual risks.** The CBMC TCB is orders of magnitude larger than HOL Light's. There is
+no independent proof-checking mechanism analogous to Candle or HOLTrace. The soundness
+guarantee for the C proofs is therefore fundamentally weaker than for the assembly proofs.
+
+**Potential improvements**:
+- Systematically introduce redundancy by employing more than one SMT solver backend
+  for CBMC functions. ([#1603](https://github.com/pq-code-package/mlkem-native/issues/1603))
+
+<!--- bibliography --->
+[^CBMC]: Amazon Web Services Diffblue: C Bounded Model Checker, [https://github.com/diffblue/cbmc](https://github.com/diffblue/cbmc)
```
rod-chapman: This does not make sense. What is "Amazon Web Services Diffblue"??? Should there be an "and" in there somewhere? Or just remove these words.

SOUNDNESS.md
```diff
@@ -0,0 +1,543 @@
...
+  proof automation cannot compromise soundness, a solver bug in CBMC's backend directly
+  invalidates the proof.
+- **The C compiler used by CBMC**: CBMC uses a C preprocessor and parser; bugs in these
+  components could affect the analysis.
+
+**Mitigations.**
+
+- CBMC has been in development for more than 20 years and is widely used in industry (including
+  at Amazon for AWS-LC and other projects).
+- The SMT solvers are independently developed and extensively tested.
```
rod-chapman, line 521: Could we or should we say anything about how AWS curates and validates solver versions?

SOUNDNESS.md
```diff
@@ -0,0 +1,543 @@
...
+- mlkem-native's proofs are run continuously in CI, providing regression testing.
+
+**Residual risks.** The CBMC TCB is orders of magnitude larger than HOL Light's. There is
+no independent proof-checking mechanism analogous to Candle or HOLTrace. The soundness
+guarantee for the C proofs is therefore fundamentally weaker than for the assembly proofs.
+
+**Potential improvements**:
+- Systematically introduce redundancy by employing more than one SMT solver backend
```
rod-chapman, line 529: We could also switch to using an SMT solver that produces independently checkable results - e.g. CVC5.

rod-chapman (commented):
See comments inline.

SOUNDNESS.md
hanno-becker: Done

SOUNDNESS.md
hanno-becker: I agree. I weakened this to covering all backends through verification, without specifying what the methodology ought to be.

hanno-becker: Renamed #1595 accordingly.

SOUNDNESS.md
hanno-becker: Added #1608

SOUNDNESS.md
hanno-becker: I leave this to you; we can potentially add it later.

SOUNDNESS.md
hanno-becker: Interesting. I am generalizing the comment.

SOUNDNESS.md
hanno-becker: Thank you for spotting this, this is auto-generated from `bibliography.yml` which was missing an explicit listing of Author's rather than a comma separated list.

SOUNDNESS.md
hanno-becker: Yes, done

SOUNDNESS.md
rod-chapman: Accepted

SOUNDNESS.md
rod-chapman: Leave for later agreed.

SOUNDNESS.md
rod-chapman: Accepted

rod-chapman (approved):
Approved.

SOUNDNESS.md
hanno-becker: Let's follow up offline on this.

SOUNDNESS.md
hanno-becker: Done

SOUNDNESS.md
hanno-becker: Kept it for now.

SOUNDNESS.md
tautschnig: Agreed.

SOUNDNESS.md
tautschnig: Ok, but actually I got confirmation that 2001 is correct.

SOUNDNESS.md
hanno-becker: I left a comment

PR #1619, opened 2026-03-12
RFC: Add incremental encaps API to support ML-KEM Braid

mkannwischer:
Split ML-KEM encapsulation into two phases (mlk_kem_enc_derand_u / mlk_kem_enc_v) to support protocols like Braid that need to interleave encapsulation with other operations between computing the u- and v-components of the ciphertext. The first phase only requires the public seed and H(pk), not the full public key vector. Internally, K-PKE.Encrypt is refactored into mlk_indcpa_enc_u + mlk_indcpa_enc_v. The non-incremental KEM path calls mlk_indcpa_enc directly to avoid serialization overhead. The intermediate noise polynomial epp is serialized as 4-bit nibbles (128 bytes) - this is primarily done to not require a pre-condition on the allowed values.


hanno-becker (changes requested):
What's the purpose of 0a01cc44941fff5b7925fea60c7569a60a7a5627? Tests also serve as documentation, and using internal constants rather than public ones sets a wrong example.

If this is needed, can it be done in a preparatory PR? It seems unrelated to this PR.

mkannwischer:
> What's the purpose of [0a01cc4](https://github.com/pq-code-package/mlkem-native/commit/0a01cc44941fff5b7925fea60c7569a60a7a5627)? Tests also serve as documentation, and using internal constants rather than public ones sets a wrong example.
> 
> If this is needed, can it be done in a preparatory PR? It seems unrelated to this PR.

The main question here is if we want to add the new API in mlkem_native.h or not. If we don't, we can't test the API in the standard test_mlkem.c, but we could add it in a separate test that includes kem.h, but not mlkem_native.h.
The purpose of https://github.com/pq-code-package/mlkem-native/commit/0a01cc44941fff5b7925fea60c7569a60a7a5627 was to get something to work first, so we can discuss how we want to proceed.

I agree with you that we don't want to keep it as is right now.

hanno-becker:
Seeing that you also observed a slowdown on x86, I wonder if we should treat the incremental API as internal by default and only expose it in the public API if some new option `MLK_CONFIG_ENABLE_MLKEM_BRAID` it set?

mkannwischer:
> Seeing that you also observed a slowdown on x86, I wonder if we should treat the incremental API as internal by default and only expose it in the public API if some new option `MLK_CONFIG_ENABLE_MLKEM_BRAID` it set?

So we would expose it in mlkem_native.h, but only if `MLK_CONFIG_ENABLE_MLKEM_BRAID` is set? Works for me, but it needs a bunch of code in mlkem_native.h including some length constants for the serialized state, etc.

Let's first figure out if marking the incremental API as static resolves our performance issues. If that is the case, I'm all for it.

hanno-becker:
> Let's first figure out if marking the incremental API as static resolves our performance issues. If that is the case, I'm all for it.

Agree.

Issue #1591, opened 2026-02-28
HOL-Light: Check in CI if expected theorems are proven


mkannwischer:
All of our HOL-Light are expected to have the same set of proven theorems.
For example, for the NTT on x86_64, we expect to have the following theorems: 
- MLKEM_NTT_CORRECT
- MLKEM_NTT_NOIBT_SUBROUTINE_CORRECT
- MLKEM_NTT_SUBROUTINE_CORRECT
- MLKEM_NTT_SAFE
- MLKEM_NTT_NOIBT_SUBROUTINE_SAFE
- MLKEM_NTT_SUBROUTINE_SAFE

If we also implement proofs for the Windows ABI (see #1587), we would have even more theorems.
It's imaginable that one forgets to add some of those (e.g., the safety proofs) and we currently rely on reviews to catch such mistakes. 

We should instead check this in CI, i.e., validate that at the end of the proof scripts the expected sets of theorems is available.

mkannwischer:
Resolved by https://github.com/pq-code-package/mlkem-native/pull/1634

Issue #1585, opened 2026-02-27
HOL-Light: Extend `tests` script to allow cross-platform proof checking


mkannwischer:
Currently, the `tests` script only supports checking the proofs for the native platform.
If one wants to check a proof for a different platform, one has to enter the cross shell (e.g., `nix develop .#hol_light-cross-x86_64`), then navigate to the x86_64 directory, and call the make `mlkem/XXX.correct` target directly.

We should extend the `tests` script to allow cross-platform proof checking

mkannwischer:
Resolved by https://github.com/pq-code-package/mlkem-native/pull/1634

PR #1634, opened 2026-03-20
HOL-Light: improve tooling and CI

L-series:
This commit aims to resolve the two following issues:
* https://github.com/pq-code-package/mlkem-native/issues/1585
* https://github.com/pq-code-package/mlkem-native/issues/1591

Note that the `check-theorems` script will currently fail the HOL-Light CI because for AArch, the naming convention for the memory safety proof (usually _SAFE) is not respected in `mlkem_rej_uniform` (it uses _MEMSAFE). @hanno-becker please advise if I should modify this.


L-series:
Modified the two remaining occurences of _MEMSAFE to _SAFE in order to fit the expected format for aarch proofs.

scripts/tests
```diff
@@ -1481,6 +1490,21 @@ def cli():
...
+        choices=["aarch64", "x86_64"],
+        default=None,
+    )
+
+    hol_light_parser.add_argument(
+        "--use-nix",
+        help="Use nix cross environment for the specified architecture.",
+        action="store_true",
+        default=False,
+    )
```
mkannwischer: We don't have this anywhere else in the test scripts, we assume you are already running in the correct nix shell.

scripts/tests
```diff
@@ -1050,12 +1052,19 @@ class Tests:
                     os.remove(os.path.join(proof_dir, proof_target))
                 except FileNotFoundError:
                     pass
+
+                make_cmd = [
+                    "make",
+                    f"mlkem/{func}.correct",
+                ] + self.make_j()
```
mkannwischer: I don't think the `-j` works here, does it?

scripts/check-theorems
mkannwischer, line 1: can't we include this in the regular lint script and just always go through all proofs?

.github/workflows/hol_light.yml
```diff
@@ -39,12 +39,37 @@ concurrency:
...
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/setup-shell
+        with:
+          gh_token: ${{ secrets.GITHUB_TOKEN }}
+          nix-shell: 'hol_light'
+          script: |
+            scripts/check-theorems ${{ matrix.arch }}
```
mkannwischer: I would prefer if we do not have to add a separate script for this.

scripts/check-theorems
```diff
@@ -0,0 +1,48 @@
...
+  # mlkem_rej_uniform on x86_64 only has a functional correctness proof
+  [[ $ARCH == "x86_64" && $routine == "mlkem_rej_uniform" ]] && suffixes=("CORRECT")
+
+  for sfx in "${suffixes[@]}"; do
+    if ! grep -q "_${sfx}" "$PROOFS_DIR/${routine}.ml"; then
+      echo "FAIL: ${routine}_${sfx} not found in ${routine}.ml" >&2
+      ((ERRORS++)) || true
+    fi
+  done
+done
```
mkannwischer: While at it, could you grep for CHEAT in the script?
`CHEAT_TAC` is the only way in HOL-Light to pass `prove` without actually proving it.

scripts/check-theorems
```diff
@@ -0,0 +1,48 @@
...
+
+for routine in $PROOFS; do
+  suffixes=("CORRECT" "SAFE" "SUBROUTINE_SAFE")
+  [[ $ARCH == "x86_64" ]] && suffixes+=("SUBROUTINE_CORRECT" "NOIBT_SUBROUTINE_CORRECT" "NOIBT_SUBROUTINE_SAFE")
+
+  # mlkem_rej_uniform on x86_64 only has a functional correctness proof
+  [[ $ARCH == "x86_64" && $routine == "mlkem_rej_uniform" ]] && suffixes=("CORRECT")
+
+  for sfx in "${suffixes[@]}"; do
+    if ! grep -q "_${sfx}" "$PROOFS_DIR/${routine}.ml"; then
```
mkannwischer: can you actually grep for {suffix} = prove or {suffix} = time prove?

mkannwischer (changes requested):
Thanks @L-series - I left a couple of comments.

This PR actually takes another approach that I had in mind: I thought that after running each HOL-Light proof we check for the expected theorems to be present.
The approach implemented here is more like a linter. But I support the addition!
The `-a` flag is highly appreciated - thank you!

scripts/tests
L-series: Noted, removed this argument.

scripts/tests
L-series: That is correct, I didn't realize this at first. Removed.

scripts/check-theorems
L-series: Moved the check into the linting script

.github/workflows/hol_light.yml
L-series: Done, moved the theorem check into `lint`

scripts/check-theorems
L-series: Added a grep for this which will fail the lint if it is present.

scripts/check-theorems
L-series: Done, however please note that I modified some of the proofs as they included double whitespaces before the `= prove`. I opted for this as it seems cleaner than having a more complicated regexp. As an aside, perhaps there is some tool to apply standard formatting onto the *.ml files?

L-series:
> Thanks @L-series - I left a couple of comments.
> 
> This PR actually takes another approach that I had in mind: I thought that after running each HOL-Light proof we check for the expected theorems to be present. The approach implemented here is more like a linter. But I support the addition! The `-a` flag is highly appreciated - thank you!

Thank you for the review! Indeed, the idea I had in mind was that one should avoid using compute on proofs which do not contain the required theorems. However, I see now that there is still value running the HOL CI even if the routine is only partially proven.

scripts/lint
```diff
@@ -295,6 +295,51 @@ gh_group_start "Check contracts"
...
+    local proofs_dir="$ROOT/proofs/hol_light/$arch/proofs"
+    local proofs
+    proofs=$("$ROOT/proofs/hol_light/$arch/list_proofs.sh")
+
+    for routine in $proofs; do
+      local suffixes=("CORRECT" "SAFE" "SUBROUTINE_SAFE")
+      if [[ $arch == "x86_64" && ${routine} == "mlkem_rej_uniform" ]]; then
+        suffixes=("CORRECT")
+      elif [[ $arch == "x86_64" ]]; then
+        suffixes+=("SUBROUTINE_CORRECT" "NOIBT_SUBROUTINE_CORRECT" "NOIBT_SUBROUTINE_SAFE")
```
mkannwischer: I tried removing the `SUBROUTINE_SAFE` proof from the mlkem_poly_compress_d4.ml proof and the script did not catch it. Is that because `NOIBT_SUBROUTINE_SAFE` matches the `SUBROUTINE_SAFE` suffix? That should be fixed.

We can actually check the full theorem name, not just the suffix. It should be possible to derive the theorem names from the file names - we should enforce that and adjust theorem names where needed.

scripts/lint
```diff
@@ -295,6 +295,51 @@ gh_group_start "Check contracts"
...
+  local success=true
+  for arch in aarch64 x86_64; do
+    local proofs_dir="$ROOT/proofs/hol_light/$arch/proofs"
+    local proofs
+    proofs=$("$ROOT/proofs/hol_light/$arch/list_proofs.sh")
+
+    for routine in $proofs; do
+      local suffixes=("CORRECT" "SAFE" "SUBROUTINE_SAFE")
+      if [[ $arch == "x86_64" && ${routine} == "mlkem_rej_uniform" ]]; then
+        suffixes=("CORRECT")
```
mkannwischer: Why is there this exceptions? There are memory safety proofs for this routine?

mkannwischer (changes requested):
Thanks @L-series for the changes! I'm happy with the overall approach now.

Two comments still need to be addressed. Thank you!

proofs/hol_light/aarch64/proofs/mlkem_rej_uniform.ml
```diff
@@ -1755,7 +1755,7 @@ let STRIP_EXISTS_ASSUM_TAC =
 (* The main memory safety theorem.                                           *)
 (* ========================================================================= *)
 
-let MLKEM_REJ_UNIFORM_MEMSAFE = prove
+let MLKEM_REJ_UNIFORM_SAFE = prove
```
mkannwischer: This was intentionally named this way and I would prefer to keep it like that because we do not have any constant time proofs at this time.
Can you implement it as: if mlkem_rej_uniform then MEMSAFE else SAFE? That would be better, I believe.

mkannwischer:
@L-series, gentle ping. Could you please adjust this PR so we can get this merged?

mkannwischer:
@L-series, are you still working on this?

mkannwischer:
@L-series, do you expect to finish this PR soon?

scripts/lint
L-series: I misunderstood as I was reading `proofs/hol_light/README.md` which states 
>The one exception is rejection sampling (`rej_uniform_aarch64_asm`), which only has a functional correctness specification.

scripts/lint
L-series: Fixed, I now added a `theorem-prefix()` function which derives the generic proof name from the file name, however there are quite a few proof names which now need to be changed.

L-series:
> @L-series, do you expect to finish this PR soon?

Apologies for the long delay, I was away traveling this last month. I just pushed some changes and I've reworked `check-theorems` to derive the expected theorem prefix directly
  from the filename (strip `_aarch64_asm` / `_avx2_asm`, uppercase, prepend `MLKEM_` for
  non-keccak) and grep for the fully-qualified theorem name. The only conditional is `MEMSAFE`
  instead of `SAFE` for `rej_uniform_*`, per your earlier comment.
  
However, I just wanted to make sure that this is okay with you as it would require renaming around 15 proofs.

mkannwischer:
> > @L-series, do you expect to finish this PR soon?
> 
> Apologies for the long delay, I was away traveling this last month. I just pushed some changes and I've reworked `check-theorems` to derive the expected theorem prefix directly from the filename (strip `_aarch64_asm` / `_avx2_asm`, uppercase, prepend `MLKEM_` for non-keccak) and grep for the fully-qualified theorem name. The only conditional is `MEMSAFE` instead of `SAFE` for `rej_uniform_*`, per your earlier comment.
> 
> However, I just wanted to make sure that this is okay with you as it would require renaming around 15 proofs.

Thanks for picking this upa gain @L-series! Welcome back!
Renaming theorems to be uniform is no problem.

mkannwischer:
@L-series, is this ready for review?  Thanks for the work!

L-series:
@mkannwischer Yes it is! Please let me know if the renaming scheme is appropriate. Given that the HOL proofing CI is passing I assume I didn't break any of the proofs :sweat_smile:

scripts/lint
```diff
@@ -353,6 +353,80 @@ gh_group_start "Linting Doxygen comments"
...
+          "${prefix}_${safe}"
+          "${prefix}_NOIBT_SUBROUTINE_CORRECT"
+          "${prefix}_NOIBT_SUBROUTINE_${safe}"
+        )
+      fi
+
+      for theorem in "${expected[@]}"; do
+        if ! grep -qE "^[[:space:]]*let ${theorem}[[:space:]]+=[[:space:]]+(time[[:space:]]+)?prove" "$file"; then
+          gh_error "${routine}" "" "Missing theorem" "${file}: ${theorem} not found"
+          success=false
```
hanno-becker: Suggestion: Can we complement `list_proofs.sh` by `list_thms.sh` which runs this grep and emits all (so-declared) theorems? I'd like to have a little logic as possible in `lint` itself.

scripts/lint
```diff
@@ -353,6 +353,80 @@ gh_group_start "Linting Doxygen comments"
...
+        "${prefix}_SUBROUTINE_CORRECT"
+        "${prefix}_SUBROUTINE_${safe}"
+      )
+      if [[ $arch == "x86_64" ]]; then
+        expected+=(
+          "${prefix}_${safe}"
+          "${prefix}_NOIBT_SUBROUTINE_CORRECT"
+          "${prefix}_NOIBT_SUBROUTINE_${safe}"
+        )
+      fi
```
hanno-becker, line 460: It would be nice to make this logic and exception more prominent... I wonder if one could define this in some TOML/YAML in proofs/{arch}... not sure if it's worth it. Ideas?

scripts/lint
L-series: Will do!

scripts/lint
L-series: Hmm.. I think the advantages of such a move are twofold:
* It would make the maintaining the naming conventions and exceptions such as `rej_uniform` simpler and more visible as they are in the same directory as the proofs. 
* It could be useful if there are any ideas to reuse or refer to the naming conventions elsewhere in the project.

However, if the sole use is within the linting script for CI, I think that it would be preferable to keep it as is as we would not have to add any additional files or TOML/YAML parsing logic.

L-series (commented):
I just realized that the comments I had written many weeks ago as well as last week which were labelled "pending", were in a  "pending to be submitted" not a "pending to be reviewed" state. I wasn't aware that one had to submit the comments through the files changed page 😓

scripts/lint
mkannwischer: I don't think this is done yet, is it?

L-series: Done! Sorry for the delay.

scripts/lint
```diff
@@ -411,6 +411,82 @@ gh_group_start "Linting Doxygen comments"
...
+
+      for theorem in "${expected[@]}"; do
+        if ! grep -qx "$theorem" <<<"$theorems"; then
+          gh_error "${routine}" "" "Missing theorem" "${file}: ${theorem} not found"
```
mkannwischer: I think the first argument of gh_error should be the file path so that the error shows up in the right place.

mkannwischer (changes requested):
Thanks @L-series. One nit - rest looks good to me.

scripts/lint
L-series: Ah, good catch! Changed it.

L-series:
Hi @mkannwischer, it seems the AWS-LC integration test is now failing due to renamed assembly routines. Were there any changes made on that end recently?

mkannwischer:
> Hi @mkannwischer, it seems the AWS-LC integration test is now failing due to renamed assembly routines. Were there any changes made on that end recently?

That's really odd. I rebased on top of main to check if it's still there. Hopefully some glitch.

mkannwischer (approved):
Thanks @L-series for the updates. LGTM. Could you port this to mldsa-native please?


@hanno-becker, could you also take a quick look, please?

scripts/tests
```diff
@@ -1525,6 +1527,14 @@ def cli():
...
     )
 
+    hol_light_parser.add_argument(
+        "-a",
+        "--arch",
+        help="Architecture for which to list/run HOL_LIGHT proofs (aarch64 or x86_64).",
+        choices=["aarch64", "x86_64"],
+        default=None,
+    )
+
```
hanno-becker, line 1537: We don't seem to test this cross-build option yet -- we probably should (in a small example)?

mkannwischer: Lets add it in a follow-up?

hanno-becker: ok

hanno-becker (approved):
Thanks @L-series!

We should check the `--arch` flag in CI, but it can be a follow-up.

scripts/lint
hanno-becker: Ok

scripts/tests
mkannwischer: Open an issue to track it: https://github.com/pq-code-package/mlkem-native/issues/1792

L-series:
> Thanks @L-series for the updates. LGTM. Could you port this to mldsa-native please?
> 
Will do! Should get around to it later today or perhaps tomorrow.

hanno-becker:
@L-series Gentle ping. What's your plan for the mldsa port?

PR #1636, opened 2026-03-23
Wycheproof: Fix incorrect assertion for invalid encaps/decaps tests

mkannwischer:
The encaps and combined test functions asserted that K should NOT match the expected value for invalid test cases where the binary succeeded. This is wrong: for invalid inputs, the binary should reject the input via input validation, not succeed and produce a different K.

Fix this by restructuring all test functions to branch on tc["result"] first, then verify the binary's behavior:
- "invalid": assert the binary rejected the input (_error or decode_error)
- "valid": assert outputs match expected values
- anything else: fail with unexpected result

Additional changes:
- Remove "acceptable" as a result type (not used in any test vectors)
- In keygen_seed tests, only "valid" results exist; simplify accordingly
- In semi_expanded_decaps tests, assert that invalid tests are rejected rather than silently ignoring errors on valid tests
- In combined tests, distinguish keygen decode errors from runtime errors, and always verify ek when keygen succeeds

### DIFF
--- a/test/wycheproof/wycheproof_client.py
+++ b/test/wycheproof/wycheproof_client.py
@@ -103,17 +103,17 @@ def run_keygen_seed_test(data_file):
         for tc in tg["tests"]:
             info(f"  tcId={tc['tcId']} ... ", end="")
             out = run_binary([binary, "keygen_seed", f"seed={tc['seed']}"])
-            if "_error" in out or "decode_error" in out:
-                assert (
-                    tc["result"] == "invalid"
-                ), f"binary error on non-invalid tcId={tc['tcId']}"
-            elif tc["result"] in ("valid", "acceptable"):
+            if tc["result"] == "valid":
                 assert (
                     out["ek"].upper() == tc["ek"].upper()
                 ), f"ek mismatch tcId={tc['tcId']}"
                 assert (
                     out["dk"].upper() == tc["dk"].upper()
                 ), f"dk mismatch tcId={tc['tcId']}"
+            else:
+                assert (
+                    False
+                ), f"Unexpected result '{tc['result']}' for tcId={tc['tcId']}"
             info("ok")
             count += 1
     info(f"  {count} keygen_seed tests passed")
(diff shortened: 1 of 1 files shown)


test/wycheproof/wycheproof_client.py
```diff
@@ -132,21 +132,21 @@ def run_encaps_test(data_file):
         for tc in tg["tests"]:
             info(f"  tcId={tc['tcId']} ... ", end="")
             out = run_binary([binary, "encaps", f"ek={tc['ek']}", f"m={tc['m']}"])
-            if "_error" in out or "decode_error" in out:
+            if tc["result"] == "invalid":
                 assert (
-                    tc["result"] == "invalid"
-                ), f"binary error on non-invalid tcId={tc['tcId']}"
-            elif tc["result"] in ("valid", "acceptable"):
+                    "_error" in out or "decode_error" in out
```
hanno-becker, line 138: What is the `"_error"` branch supposed to catch in addition to `decode_error`?

test/wycheproof/wycheproof_client.py
```diff
@@ -132,21 +132,21 @@ def run_encaps_test(data_file):
...
                 assert (
                     out["K"].upper() == tc["K"].upper()
                 ), f"K mismatch tcId={tc['tcId']}"
-            elif tc["result"] == "invalid":
+            else:
                 assert (
-                    out["K"].upper() != tc["K"].upper()
-                ), f"K should not match for invalid tcId={tc['tcId']}"
+                    False
+                ), f"Unexpected result '{tc['result']}' for tcId={tc['tcId']}"
```
hanno-becker: This is slightly misleading. "Unexpected result" to me suggests that something was wrong with the client's output -- but instead, we simply don't understand the test case?

hanno-becker:
```suggestion
                ), f"Unsupported test result '{tc['result']}' for tcId={tc['tcId']}"
```

hanno-becker (changes requested):
One nit, otherwise it looks good. Thank you very much for catching this!

test/wycheproof/wycheproof_client.py
hanno-becker: I took the liberty to fix this.

hanno-becker: I took the liberty to fix this.

test/wycheproof/wycheproof_client.py
hanno-becker: I added a comment

hanno-becker (approved):
@mkannwischer LGTM. Please check my edit and feel free to merge then.

PR #1640, opened 2026-03-24
Refactor keccak_squeezeblock_x4() to improve proof time.

rod-chapman:
Removes pointer arithmetic in favour of an integer current_offset value. Simplifies loop invariant.

Proof time on M1/macOS falls from 127s to 25s for all values of MLKEM_K.

With gcc 14.2.0 on macOS, generated code at -O3 is identical to the original, so no performance impact expected.


mkannwischer (approved):
Thanks @rod-chapman. LGTM and CI confirms that this is a massive proof performance gain.

Could you kindly also make that change in mldsa-native?

PR #1650, opened 2026-04-14
Add Rust bindings

Rixxc:
This PR proposes a Rust binding for mlkem-native.

Some notes about the implementation:

- The bindings implement the KEM trait from the RustCrypto kem crate, which makes it a easy drop in replacement for the RustCrypto ml-kem crate
- Rust compiles the mlkem-native with avx2 support if the target architecture supports it
- acvp_mlkem.rs implements the acvp test binary using the Rust API, which allows the binding to be tested against the acvp test vectors
- The implementation performs a key check when deserializing the encapsulation and decapsulation keys. This is a bit redundant as the checks are also performed on every encap and decap operation by the mlkem-native library. I chose to do this as the trait definition of the kem crate do not allow for those operations to fail, which means the only option to handle those errors is to panic. But the keys are checked upon construction this case should never be triggered with a safe program.

Please let me know what you think about this approach.


Cargo.lock
```diff
@@ -0,0 +1,579 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
+version = 4
+
+[[package]]
```
hanno-becker: @Rixxc Can you comment on what all these packages are needed for?

bindings/rust/mlkem-native/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "mlkem-native"
+version = "1.1.0-rc.1"
+edition = "2021"
```
hanno-becker: Wrong year?

Rixxc: The edition field is used by the compiler. It specifies a set of features that the Rust compiler has to support to be able to compile the project.

Cargo.lock
Rixxc: There are only a few dependencies in that that are actually used at build- and runtime. The dependencies are:

# Runtime Dependencies

- kem: which contains interface definition that the mlkem-native crate implements. This is the same interface that the ml-kem crate from RustCrypto uses. This allows for the mlkem-native crate to be used in generic code that expects objects that implement this interface and can be used as an easy drop-in replacement for the already existing and quit popular ml-kem crate.
- crypto-common: Is a dependency from kem
- hybrid-array: Is needed to implement the kem interface
- typenum: pulled in by hybrid-array
- rand_core: This crate contains interface definitions for RNGs, similar to the kem crate

# Build Dependencies

- cc: Is needed to compile the C code of mlkem-native during build time of this crate
- find-msvc-tools: pulled in by cc, needed to compile the C code under Windows
- shlex: pulled in by cc, used to escape strings for a POSIX shell

# Other Dependencies

The rest of the dependencies are just dev dependencies. They are pulled in due to the getrandom feature of of the kem create. They are only used in the examples in the mlkem-native create.

bindings/rust/mlkem-native/Cargo.toml
Rixxc: I've updated the edition to the latest one

Rixxc: I've updated the edition to the latest one

Rixxc:
I had to move the package definition to the root of the project as otherwise we most likely will encounter issues when trying to publish the crate

Rixxc:
Shall I squash the commits before this PR gets merged? Also is the crate name of mlkem-native fine or should it rather be mlkem-native-rs?

bindings/rust/mlkem-native/build.rs
```diff
@@ -0,0 +1,59 @@
...
+    // CPU feature is present at compile time (e.g. via
+    // `RUSTFLAGS="-C target-cpu=native"` or
+    // `RUSTFLAGS="-C target-feature=+avx2"`).
+    let target_features = env::var("CARGO_CFG_TARGET_FEATURE").unwrap_or_default();
+    let use_avx2 = env::var("CARGO_CFG_TARGET_ARCH").unwrap_or_default() == "x86_64"
+        && target_features.split(',').any(|f| f == "avx2");
+
+    if use_avx2 {
+        println!("cargo:rustc-cfg=mlkem_native_avx2");
+    }
```
hanno-becker, line 26: We'd need support for at least AArch64 as well, ideally also RV64.

bindings/rust/mlkem-native/src/lib.rs
```diff
@@ -0,0 +1,431 @@
...
+                    assert_eq!(ret, 0, "enc_derand failed — key may have been corrupted");
+                    (ct, ss)
+                }
+            }
+
+            /// The decapsulation (secret) key.
+            ///
+            /// The encapsulation key is stored alongside to support
+            /// [`kem::Decapsulator::encapsulation_key`].
+            pub struct DecapsulationKey {
```
hanno-becker, line 134: Please consider overwriting `Drop` to implement zeroization. How do other RustCrypto KEMs handle zeroization of stack material?

bindings/rust/mlkem-native/src/lib.rs
```diff
@@ -0,0 +1,431 @@
...
+                    // for invalid ciphertexts per FIPS 203 §8.3); non-zero only if the
+                    // hash check on the secret key fails, which is guarded against on
+                    // construction.
+                    let ret = unsafe {
+                        $dec(
+                            ss.as_mut_slice().as_mut_ptr(),
+                            ct.as_slice().as_ptr(),
+                            self.sk.as_ptr(),
+                        )
+                    };
```
hanno-becker, line 172: Every `unsafe` block should have a `# Safety ...` comment explaining why the code is safe.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,49 @@
...
+      - uses: ./.github/actions/setup-shell
+        name: ACVP test
+        with:
+          nix-shell: rust
+          gh_token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            cargo build --release --bin acvp_mlkem
+            for version in v1.1.0.39 v1.1.0.40 v1.1.0.41; do
+              python3 test/acvp/acvp_client.py \
+                --binary ./target/release/acvp_mlkem \
```
hanno-becker, line 47: This doesn't seem to work locally. How can I run the ACVP tests?

.github/workflows/rust.yml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: ./.github/actions/setup-shell
+        with:
+          nix-shell: rust
+          gh_token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            cargo build --workspace
+            RUSTFLAGS="-C target-feature=+avx2" cargo build --workspace
+            cargo test --workspace
```
hanno-becker, line 36: We should not unconditionally pass `+avx2` if this is also running on AArch64 runners?

hanno-becker (changes requested):
Thank you @Rixxc for picking this up again. I left a first bunch of comments.

We need to get this through CI. I suggest we add you as a contributor (cf https://github.com/pq-code-package/tsc/pull/209/) so you can re-open the PR from the main repository and run the full CI.

hanno-becker:
@Rixxc Could you re-open this PR from a branch on the main repo? You now her permission to push to non-`main` branches.

Rixxc:
Yes, I'll do that. Closing the PR now and reopening it

PR #1652, opened 2026-04-17
Make --slice-formula default for all CBMC proofs

rod-chapman:
CBMC team have recommended that, with CBMC 6.9.0 and above, the `--slice-formula` option for CBMC
should always be used, and that --no-array-field-sensitivity should never be used.

This PR implements the first of those changes.

1. Add --slice-formula to Makefile.common
2. Remove --slice-formula from per-function Makefiles
3. SMT only Z3 tactic for indcpa_enc() stabilizes proof time for this function for all values of K and on macOS/Linux. 

Proof times on r8g instance, all proofs, total CPU "user time" for MLKEM_K=2,3,4
```
main branch: 33m35s, 33m21s, 33m13s (note much improved stability for values of K)
this branch: 33m47s, 33m3s, 32m57s
```
so no significant change.

Running all proofs on r8g with -j64 completes all proofs in 3 minutes real time.

### DIFF
--- a/proofs/cbmc/Makefile.common
+++ b/proofs/cbmc/Makefile.common
@@ -281,10 +281,6 @@ PROVER_NAME=cvc5
 SMT_FORMAT := --cvc5
 endif
 
-# CBMC flags used for property checking and coverage checking
-
-CBMCFLAGS += $(CBMC_FLAG_FLUSH)
-
 # CBMC 6.0.0 enables all standard checks by default, which can make coverage analysis
 # very slow. See https://github.com/diffblue/cbmc/issues/8389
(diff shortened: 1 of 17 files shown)


mkannwischer:
What is the reason for not removing --no-array-field-sensitivity yet?

rod-chapman:
Just trying to do it and review it one step at a time...

rod-chapman:
It's also possible that adding --slice-formula will be OK and effective, but removing --no-array-field-sensitivity might cause unexpected problems, so doing it in two putts seems sensible.

hanno-becker (approved):

PR #1653, opened 2026-04-19
Lint: Replace `black` with `ruff`

mkannwischer:
This PR replaced `black` with `ruff` which is much faster and also offers linting.

A few real issues ruff flagged:         
  - run_cbmc() in scripts/tests references `log` before it is defined, and the message is a plain string where `{self.args.start_with}`                      
  - Two bare `except:` clauses in scripts/cfify which would swallow a CTRL+C
  - A handful of unused variables, unused argparse subparsers, and stray `f""` prefixes.

Being able to catch those issues in the future appears worthwhile to me.

I also fixed the other stylistic warnings (that I do not feel strongly about).
For now I have ignored `E501` complaining about too long lines - that would require a lot of changes.

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,0 +1,25 @@
+# Copyright (c) The mlkem-native project authors
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+# In addition to ruff's default (*.py, *.pyi), format and lint the
+# extension-less python scripts under scripts/.
+extend-include = [
+    "scripts/autogen",
+    "scripts/cfify",
+    "scripts/check-namespace",
+    "scripts/simpasm",
+    "scripts/tests",
+]
+
+[lint]
+select = ["E", "F", "W"]
+ignore = [
+    "E501", # line too long
+    "E713", # `not in` rather than `not X in Y`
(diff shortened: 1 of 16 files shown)


ruff.toml
```diff
@@ -0,0 +1,25 @@
...
+# SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+
+# In addition to ruff's default (*.py, *.pyi), format and lint the
+# extension-less python scripts under scripts/.
+extend-include = [
+    "scripts/autogen",
+    "scripts/cfify",
+    "scripts/check-namespace",
+    "scripts/simpasm",
+    "scripts/tests",
```
hanno-becker, line 14: There are more python scripts in `scripts`: `stack`, `check-contracts`, `check-magic`

hanno-becker (approved):
@mkannwischer LGTM except that we don't seem to list all scripts from `scripts` in the TOML. Thank you also for the clean commit structure.

ruff.toml
mkannwischer: Good catch. Those were missing in the `lint` before this PR. Fixed.

PR #1657, opened 2026-04-20
Correct declaration and comments re size of rej-uniform lookup table

rod-chapman:
Review of the AArch64 back-end shows that comments pertaining to the size of the rej_uniform lookup were are incorrect in a number of places. The table is 256 * 16 bytes, so 4096 bytes, NOT 2048 bytes as appears in the comments.


hanno-becker (approved):
Thanks @rod-chapman !

PR #1658, opened 2026-04-21
Port: Allow tables to be declared static

mkannwischer:
When building single-CU builds and MLK_CONFIG_INTERNAL_API_QUALIFIER is set to make internal API static, also make the data tables static. This makes the data symbols local instead of global, hiding them from the scope of the rest of the application.

Add MLK_INTERNAL_DATA_DECLARATION (defaulting to extern) and MLK_INTERNAL_DATA_DEFINITION (defaulting to empty) macros. When MLK_CONFIG_INTERNAL_API_QUALIFIER is set, both expand to its value (e.g. static).

Also add explicit lengths to all constant array declarations and refactor scripts/autogen to use a new emit_c_array() helper for all generated C array definitions, so sizes are always explicit (e.g., mlk_rej_uniform_table[4096] instead of mlk_rej_uniform_table[]).

- Ports https://github.com/pq-code-package/mldsa-native/pull/1043


hanno-becker (approved):
As far as I can see, this is a faithful port. Thank you @mkannwischer !

PR #1661, opened 2026-04-22
check-contracts: Detect contracts separated from prototype by a comment

mkannwischer:
The previous regex skipped contracts with a block comment between the prototype's `)` and `__contract__`, silently missing six contracts (polyvec_basemul_acc_montgomery_cached_asm_k{2,3,4} on both aarch64 and x86_64). Strip comments before matching, and extend the exception list to cover the `_asm_k{2,3,4}` suffixes alongside `_native`, `_asm`, and `_avx2`. Also fix a typo in the `_avx2` suffix.

Also align DECREASES_EXCEPTIONS handling with mldsa-native.

- Ported from pq-code-package/mldsa-native#1051.

### DIFF
--- a/scripts/check-contracts
+++ b/scripts/check-contracts
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 # Copyright (c) The mlkem-native project authors
+# Copyright (c) The mldsa-native project authors
 # SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
 
 #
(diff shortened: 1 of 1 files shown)


scripts/check-contracts
```diff
@@ -55,7 +68,8 @@ def is_exception(funcname):
     if (
         funcname.endswith("_native")
         or funcname.endswith("_asm")
-        or funcname.endswith("avx2")
+        or funcname.endswith(("_asm_k2", "_asm_k3", "_asm_k4"))
+        or funcname.endswith("_avx2")
```
mkannwischer, line 72: I wonder if we should unify the asm symbols. It's somewhat inconsistent right now.

Also both x86_64 and aarch64 backend use the `mlk_polyvec_basemul_acc_montgomery_cached_asm_k{2,3,4}` symbols which looks unintentional.

hanno-becker: I think that's a good idea, yes.

hanno-becker (approved):

PR #1664, opened 2026-04-23
Remove 2 dead instruction from AArch64 back-end poly_tomont()

### LINKED ISSUE #1656: Dead code in AArch64 poly_tomont_asm.S
In the "clean" version of poly_tomont() in the AArch64 backend (`dev/aarch64_clean/src/poly_tomont_asm.S`) the `modulus_twisted` constant is not used, so the instructions
```
        mov wtmp, #20159 // Barrett twist of 1 wrt 2^27
        dup modulus_twisted.8h, wtmp
```
are dead. They could be removed. The named register `tmp0` (currently defined as `v6`) could be re-defined to be `v5`, freeing a register for potential optimization with SLOTHY.

rod-chapman:
Fixes #1656 
1. Remove 2 dead instructions and re-assign registers
2. Re-apply SLOTHY
3. Re-autogen all files
4. Update HOL-Light bytecode for poly_tomont()

The HOL-Light proof scripts itself did not require modifiction

### DIFF
--- a/dev/aarch64_opt/src/poly_tomont_asm.S
+++ b/dev/aarch64_opt/src/poly_tomont_asm.S
@@ -42,7 +42,6 @@
         factor            .req v2
         factor_t          .req v3
         modulus           .req v4
-        modulus_twisted   .req v5
 
         tmp0              .req v6
 
@@ -53,9 +52,6 @@ MLK_ASM_FN_SYMBOL(poly_tomont_asm)
 
         mov wtmp, #3329 // ML-KEM modulus
         dup modulus.8h, wtmp
-
-        mov wtmp, #20159 // Barrett twist of 1 wrt 2^27
-        dup modulus_twisted.8h, wtmp
 
         mov wtmp, #-1044 // 2^16 % 3329
         dup factor.8h, wtmp
@@ -217,7 +213,6 @@ poly_tomont_loop:
         .unreq factor
         .unreq factor_t
         .unreq modulus
-        .unreq modulus_twisted
 
         .unreq tmp0
 
--- a/dev/aarch64_clean/src/poly_tomont_asm.S
+++ b/dev/aarch64_clean/src/poly_tomont_asm.S
@@ -43,7 +43,6 @@
         factor            .req v2
         factor_t          .req v3
         modulus           .req v4
-        modulus_twisted   .req v5
 
         tmp0              .req v6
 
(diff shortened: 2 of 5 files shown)


rod-chapman:
On a Graviton4 (r8g) instance, the `tests bench --components` benchmark says exec time for poly_tomont() reduces from 58 to 56 cycles.

dev/aarch64_opt/src/poly_tomont_asm.S
mkannwischer, line 1: The changes are not affecting the code optimized by SLOTHY, right? To avoid unnecessary churn, I'd prefer if we do not reoptimize the loop.

mkannwischer (changes requested):
Thanks @rod-chapman for fixing. I would prefer if we do not run the SLOTHY re-optimization as this makes the diff larger than it needs to be. Or am I missing something here?

dev/aarch64_opt/src/poly_tomont_asm.S
rod-chapman: Well... removing the 2 dead instructions meant that one more register could be made available to SLOTHY, so there was the possibility of it finding a better schedule, so I wanted to give it the opportunity to try.

The HOL-Light specification and proof script have not changed at all, so I am happy that the new code is equivalent to the old.

mkannwischer: No? The code that is input to SLOTHY never used modulus_twisted, so all that SLOTHY sees is that you changed tmp0 from v6 to v5, but that does not matter to SLOTHY as that all becomes symbolic internally.

Of course it's great to know that the HOL-Light proofs are scheduling-agnostic as intended, but I prefer to have a smaller diff to the code that's actually deployed in AWS-LC/s2n-bignum/liboqs unless there is a good reason to change (e.g., better performance).

hanno-becker: I agree with @mkannwischer

rod-chapman: So you suggest revert to the previous version of aarch64_opt/src/poly_tomont_asm.S and manually remove the two offending instructions only?

rod-chapman: OK. I will make it so.

rod-chapman: Done. I squashed before push to minimize the apparent diff.

mkannwischer (approved):
LGTM. Many thanks to @rod-chapman for spotting and fixing this one.

PR #1668, opened 2026-04-27
Add ACVP test Windows support (C backend)

willieyz:
This PR adds ACVP test support for Windows using the C backend by adding the compilation and link rules in `Makefile.Microsoft_nmake`,  also integrate to CI testing flow.

- `mlkem-native` currently has no ACVP test coverage on Windows. This PR establishes the foundation for Windows support by enabling ACVP tests against the portable C backend, which is a start before solving the more complex work of enabling x86_64 and AArch64 assembly backends on Windows.

### DIFF
--- a/Makefile.Microsoft_nmake
+++ b/Makefile.Microsoft_nmake
@@ -65,6 +65,21 @@ OPT = 0
     @if NOT EXIST $(MLKEM1024_BUILD_DIR)\test mkdir $(MLKEM1024_BUILD_DIR)\test
     $(CC) $(CFLAGS) /D MLK_CONFIG_PARAMETER_SET=1024 /c /Fo$(MLKEM1024_BUILD_DIR)\test\ $<
 
+# compilation of acvp test for mlkem512
+{test\acvp}.c{$(MLKEM512_BUILD_DIR)\test\acvp}.obj::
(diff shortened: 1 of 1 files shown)


.github/workflows/base.yml
```diff
@@ -187,6 +187,25 @@ jobs:
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756 # v1.13.0
+      - name: Build ACVP test
+        shell: powershell
+        run: |
+          nmake /f ./Makefile.Microsoft_nmake acvp
+      - name: Run ACVP test
+        shell: powershell
+        run: |
+          python test/acvp/acvp_client.py --version ${{ matrix.acvp-version }}
```
mkannwischer: Can this be integrated into 
```
nmake /f ./Makefile.Microsoft_nmake quickcheck
```

Would be good if we could slowly align that one to have the same behavior in the Makefile.

We don't need to run multiple ACVP versions on Windows.

mkannwischer (changes requested):
Thanks @willieyz for looking into this. Seems to do the right thing according to the CI logs.

One suggestion. Please also combine it into a single commit.

mkannwischer (approved):
Thanks @willieyz. LGTM and runs fine in CI.

PR #1676, opened 2026-05-05
HOL-Light: Prefix imports to separate mlkem-native from s2n-bignum

hanno-becker:
Ports: https://github.com/pq-code-package/mldsa-native/pull/1088/

Issue: every `needs` directive in mlkem-native proof files resolved ambiguously against both $PROOF_DIR and $S2N_BIGNUM_DIR, so a file with the same relative path in either tree could silently shadow the other.

Proposed resolution: Every import now declares its source at the call site with a namespace prefix.

  needs "mlkem_native/common/mlkem_specs.ml";;   (* local *)
  needs "s2n_bignum/arm/proofs/base.ml";;        (* external *)
  needs "Library/words.ml";;                     (* HOL-Light stdlib *)

The nix shellHook creates a temporary .imports directory with symlinks `mlkem_native` and `s2n_bignum` pointing to the local resp. s2n-bignum directories, and HOLLIGHT_LOAD_PATH points at that import directory. The `mlkem_native/` and `s2n_bignum/` namespaces are disjoint by prefix, so path collisions become structurally impossible.

A new scripts/check-hol-light-imports enforces the rule across proofs/hol_light/**/*.ml and is wired into scripts/lint and both build-proof.sh scripts. s2n-bignum imports must additionally appear on proofs/hol_light/s2n_bignum_allowlist.txt, so new cross-repo dependencies surface in review.

A side-benefit of exposing proofs/hol_light/.imports is that it is easier to navigate to the nix-imported s2n-bignum sources for inspection.

### DIFF
--- a/.github/workflows/hol_light.yml
+++ b/.github/workflows/hol_light.yml
@@ -71,7 +71,7 @@ jobs:
           nix-shell: 'hol_light'
           script: |
             make -C proofs/hol_light/aarch64 mlkem/poly_tobytes_aarch64_asm.o
-            echo 'needs "aarch64/proofs/poly_tobytes_aarch64_asm.ml";;' | hol.sh
+            echo 'needs "mlkem_native/aarch64/proofs/poly_tobytes_aarch64_asm.ml";;' | hol.sh
   hol_light_proofs:
     needs: [ hol_light_bytecode ]
     strategy:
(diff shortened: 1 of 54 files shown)


mkannwischer (approved):
Thanks for the port @hanno-becker. LGTM.

PR #1691, opened 2026-05-12
Avoid mismatched comparison [full CI]

mkannwischer:
Running full CI on #1690.


nmouha:
Thank you, @mkannwischer!

I've fixed the CI failure and replaced `requires(len <= UINT16_MAX)` by `requires(len <= MLK_MAX_BUFFER_SIZE)`.

Could you pull in my update to #1690 and see if CI passes on your end as well? Thanks!

PR #1692, opened 2026-05-12
common: Move backend includes into new backends.h shim

### LINKED ISSUE #1182: Remove backend header import from common.h
Currently, common.h includes the backend headers with the `static inline` wrappers around backend functionality. For some olders compilers (gcc-4.8) in -O0 this leads to linker failures if the backend itself is not linked. This makes it hard to test files in isolation -- specifically, in #1135 I'm struggling to create a build which includes specific backend assembly files without including the entire backends (including twiddle tables).

Consider removing the backend header import from `common.h`, and instead import it manually where needed.

hanno-becker:
* Resolves #1182

common.h previously pulled in MLK_CONFIG_ARITH_BACKEND_FILE and MLK_CONFIG_FIPS202_BACKEND_FILE for every translation unit, which caused gcc-4.8 / -O0 link failures when a TU was compiled in isolation without the backend (issue #1182).

Introduce mlkem/src/backends.h as the single shim that materialises backend metadata and the MLK_CHECK_APIS sanity includes. Consumer C files (compress.c, indcpa.c, poly.c, poly_k.c, sampling.c, fips202/keccakf1600.c) and backend-internal asm/C sources include backends.h instead of either the old common.h block or a direct ../meta.h / ../auto.h reference.


.github/workflows/integration-opentitan.yml
```diff
@@ -75,9 +75,21 @@ jobs:
...
+          # Inject the backends.h header into the BUILD file's hdrs list.
+          # Upstream expo's BUILD.mlkem_native.bazel enumerates every header
+          # explicitly, and mlkem-native split the backend includes out of
+          # common.h into a new backends.h shim that the expo pin doesn't
+          # know about. Drop this patch once expo lists backends.h itself.
+          BUILD=third_party/mlkem_native/BUILD.mlkem_native.bazel
+          if ! grep -q '"mlkem/src/backends.h"' "$BUILD"; then
+            sed -i 's|"mlkem/src/common.h",|"mlkem/src/common.h",\n        "mlkem/src/backends.h",|' "$BUILD"
+          fi
+
```
mkannwischer: Do you mind doing this as a patch? That will make it easier to actually update alter.

integration/aws-lc/pre_import.patch
```diff
@@ -1,10 +1,25 @@
...
+   s2n_header=$(if [[ "$file" == *"aarch64"* ]]; then echo "_internal_s2n_bignum_arm.h"; else echo "_internal_s2n_bignum_x86_att.h"; fi)
+   sed "${SED_I[@]}" "s/#include \"\.\.\/\.\.\/\.\.\/common\.h\"/#include \"$s2n_header\"/" "$file"
+
++  # Drop the backends.h include. In mlkem-native it surfaces the
++  # MLK_ARITH_BACKEND_* gate macro to the asm sources, but unifdef above
++  # has already statically resolved that gate, so the include has nothing
++  # left to provide here. Leaving it in would also break the build, since
++  # the relative path (../../../backends.h) does not resolve in the
++  # AWS-LC vendored layout.
++  sed "${SED_I[@]}" '/^#include "\.\.\/\.\.\/\.\.\/backends\.h"$/d' "$file"
```
mkannwischer, line 14: Would it be cleaner for AWS-LC to gate the include via the `MLK_CONFIG_USE_NATIVE_BACKEND_ARITH` macro? Probably I am missing soemthing here.

mkannwischer (changes requested):

integration/aws-lc/pre_import.patch
hanno-becker: The importer patches the `common.h` include to an s2n-bignum include and manually removes the backend guards so the code is unconditionally included. We hence don't need `backends.h` anymore here.

hanno-becker: An alternative would be to import _just_ backends.h (because that also includes common.h) in the ASM files, and then just patch that. That's a smaller change but deviates from the convention that all files should first include common.h.

.github/workflows/integration-opentitan.yml
hanno-becker: Will do

hanno-becker: Done

hanno-becker:
Perhaps we can wait merging this until it is confirmed that the ABI checker is unblocked by it.

hanno-becker:
I found a better way to implement the ABI checker that does not require this change.

PR #1698, opened 2026-05-18
Add kat test Windows support (C backend only)

willieyz:
- Related: #1668

This PR adds Windows support for the C backend KAT tests.

Currently, Unix KAT tests rely on shell pipelines (sha256sum, cut, xargs) not available on Windows.                                                   
Introduce `test/src/kat_client.py`, a Python script that runs gen_KAT binaries, SHA-256-hashes their stdout, and checks against META.yml, replacing the shell one-liner for each security level:                  
                                                                        
```                                                                     
 set -o pipefail; $W $MLKEM512_DIR/bin/gen_KAT512 | $SHA256SUM | \ cut -d " " -f 1 | xargs ./META.sh ML-KEM-512 kat-sha256
```                                                                     
Also, refactor gen_KAT{512,768,1024}, gen_KAT, run_kat in Makefile.Microsoft_nmake. now quickcheck align with Unix Makefile.


test/src/kat_client.py
mkannwischer, line 1: I'm not a fan of introducing a second way of running KAT tests.

@hanno-becker, would it be an option to also migrate the other KAT tests to use a Python wrapper?

mkannwischer (changes requested):
@willieyz, I have discussed this with @hanno-becker: We want to keep KAT checking uniform across platforms. I think the Python approach you have picked is likely the most portable. Could you also switch the Makefile to use that script? Please also clean up all unnecessary files that won't be needed anymore then.

willieyz:
> @willieyz, I have discussed this with @hanno-becker: We want to keep KAT checking uniform across platforms. I think the Python approach you have picked is likely the most portable. Could you also switch the Makefile to use that script? Please also clean up all unnecessary files that won't be needed anymore then.

Hello @mkannwischer, thank you for the review and suggestion!
I have addressed your feedback, both the Unix `Makefile` and `Makefile.Microsoft_nmake` now use `test/src/kat_client.py` to run the KAT tests, so KAT checking is uniform across platforms. 
`META.sh` and the `SHA256SUM` detection block in the Makefile have been removed as they are no longer needed.
Please take another look when you have time, thank you and Hanno again for the suggesions and help!

test/src/kat_client.py
```diff
@@ -0,0 +1,103 @@
...
+            err(f"META.yml: no kat-sha256 entry for {scheme_name}")
+            failed = True
+            continue
+        if not run_kat_single(scheme_name, level, ref_hashes[scheme_name]):
+            failed = True
+
+    if failed:
+        sys.exit(1)
+
+    info("ALL GOOD!")
```
mkannwischer: Let's simplify this to 
```py
  def run_kat_single(level, ref_hash):
      scheme_name = f"ML-KEM-{level}"
      binary = get_kat_binary(level)
      ...

  def main():
      parser = argparse.ArgumentParser()
      parser.add_argument("--scheme", choices=["512", "768", "1024"],
                          help="Run KAT for only one parameter set (default: all)")
      args = parser.parse_args()

      levels = [int(args.scheme)] if args.scheme else [512, 768, 1024]
      ref_hashes = read_meta_hashes()

      failed = False
      for level in levels:
          scheme_name = f"ML-KEM-{level}"
          ref = ref_hashes.get(scheme_name)
          if ref is None:
              err(f"META.yml: no kat-sha256 entry for {scheme_name}")
              failed = True
              continue
          if not run_kat_single(level, ref):
              failed = True

      if failed:
          sys.exit(1)
      info("ALL GOOD!")
```

mkannwischer (changes requested):
Thanks @willieyz. Looks good to me now. One small nit, then we can merge it.

test/src/kat_client.py
willieyz: Hello, @mkannwischer , thank you for your review!
I have updated `kat_client.py` according to your suggesion: remove the redundant tuple list and lookup,
please take another look when you have time, thank you!

After this PR merged, I will port this to mldsa-native, thank you again for your help!

mkannwischer (approved):
Thanks @willieyz. LGTM now. 


@hanno-becker, could you take a look if you agree with this change?

test/src/kat_client.py
hanno-becker: I don't object to a Python script, but I would prefer if we would stick to the approach of running the KAT binary directly and just piping its output into the KAT checker. cf https://github.com/pq-code-package/mlkem-native/commit/2fb7f1739d3003dc1c1762d3aa8d893883deb5fe WDYT

mkannwischer: yes agreed, that's cleaner. Can you push that?

hanno-becker (approved):

PR #1700, opened 2026-05-20
CI: Add more legacy compiler tests and move to scheduled workflow 

### LINKED ISSUE #1557: CI: Do not use `nixpkgs` for compiler tests and constant-time tests
For our compiler tests and constant-time tests we have been relying on many compiler versions being present in `nixpkgs`.

However, old compiler versions are dropped from `nixpkgs` forcing us to rely on old versions of `nixpkgs` to keep our tests functional.
For example, nixpkgs 25.11 dropped support for gcc 11, gcc 12, clang 14, clang 15, clang 16, clang 17, zig 0.12. Temporarily we ca source those from 24.05 which still has all of those, but eventually those are going to be dropped from the nix binary cache which will force us to compile all those from source.

I don't think this is sustainable in the long term.
We should consider alternative ways to test older versions of compilers in CI. It may be straightforward to achieve the same tests using `docker`.

mkannwischer:
Our compiler tests and constant-time tests have been steadily
growing over time which is not sustainable in the medium term.

This commit solves that by splitting the set up of compiler versions
into two
1) Regular compilers - tested as before on every PR. These are compilers
that are supported by the most recent nixpkgs release.

2) Legacy compilers - tested in a scheduled job or when
`legacy-compiler-tests` label is added to the PR. Currently the
job is scheduled to run daily, but in the long term we may want
to switch it to weekly.

Legacy compiler shells and overlay overrides live in nix/legacy/, with
their own nixpkgs revisions pinned via fetchTarball.
The main flake.lock therefore stays free of legacy inputs.

This allows us to extend comiler coverage by  gcc 6, 8, 9, 10, zig 0.10 +0.11, 
and clang 6..13 for both CT and compiler tests.

My expectation is that the legacy compilers only get changed
whenever nixpkgs drops a compiler version.

- Resolves https://github.com/pq-code-package/mlkem-native/issues/1557

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -214,7 +214,7 @@ jobs:
           check_namespace: 'false'
           extra_args: "--fips202-aarch64-backend ${{ matrix.backend }}"
   compiler_tests:
-    name: Compiler tests  (${{ matrix.compiler.name }}, ${{ matrix.target.name }}, ${{ matrix.cflags }})
+    name: Compiler tests  (${{ matrix.compiler.family }}-${{ matrix.compiler.version }}, ${{ matrix.target.name }}, ${{ matrix.cflags }})
     strategy:
       fail-fast: false
       matrix:
@@ -226,90 +226,55 @@ jobs:
            name: 'x86_64'
          - runner: macos-latest
            name: 'macos'
+        # Oldest available + 3 latest per family; intermediate versions run in legacy-compilers.
         compiler:
-         - name: gcc-4.8
+         - family: gcc
+           version: "4.8"
            shell: gcc48
            darwin: False
-           c17: False
-           c23: False
            opt: all
            examples: true
-         - name: gcc-4.9
-           shell: gcc49
-           darwin: False
-           c17: False
-           c23: False
-           opt: all
-           examples: true
-         - name: gcc-7
-           shell: gcc7
-           darwin: False
-           c17: False
-           c23: False
-           opt: all
-           examples: true
-         - name: gcc-11
-           shell: gcc11
-           darwin: True
-           c17: True
-           c23: False
-           opt: all
-           examples: true
-         - name: gcc-13
+         - family: gcc
+           version: "13"
            shell: gcc13
            # TODO: gcc13 broken on macOS in nixpkgs 25.11
            darwin: False
-           c17: True
-           c23: False
            opt: all
            examples: true
-         - name: gcc-14
+         - family: gcc
+           version: "14"
            shell: gcc14
            darwin: True
-           c17: True
-           c23: True
            opt: all
            examples: true
-         - name: gcc-15
+         - family: gcc
+           version: "15"
            shell: gcc15
            darwin: True
-           c17: True
-           c23: True
            opt: all
            examples: true
-         - name: clang-18
-           shell: clang18
-           darwin: True
-           c17: True
-           c23: True
+         - family: clang
+           version: "6"
+           shell: clang6
+           darwin: False
            opt: all
            examples: true
-         - name: clang-19
-           shell: clang19
-           darwin: True
-           c17: True
-           c23: True
-           opt: all
-           examples: true
-         - name: clang-20
+         - family: clang
+           version: "20"
            shell: clang20
            darwin: True
-           c17: True
-           c23: True
            opt: all
            examples: true
-         - name: clang-21
+         - family: clang
+           version: "21"
            shell: clang21
            darwin: True
-           c17: True
-           c23: True
            opt: all
            examples: true
-         - name: clang-22
+         - family: clang
+           version: "22"
            shell: clang22
            darwin: True
-           c17: True
-           c23: True
            opt: all
            examples: true
          # CPU flags are not correctly passed to the zig assembler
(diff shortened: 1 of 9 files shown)


mkannwischer:
For a long time I tried many other ways to install such old compilers.
For example, https://github.com/pq-code-package/mlkem-native/pull/1688 tries to user Docker for gcc, and https://github.com/KyleMayes/install-llvm-action for LLVM. It seems to be working, but it becomes somewhat hacky and it cannot give us the same coverage as nix can. For example, the gcc4.8/4.9 containers don't work with current docker - and they are not maintained.

The main reason for not using nix was the fear that cache entries in the nix binary cache will eventually disappear forcing our CI to build the compilers from source. So far we have not seen that, so I think we can defer solving this problem until it actually materializes. There is some on going discussion on [garbage collecting in the nix binary cache ](https://discourse.nixos.org/t/garbage-collecting-cache-nixos-org/74249) as their S3 bill is apparently quite high, so we may eventually be affected.

.github/workflows/ci.yml
```diff
@@ -227,34 +227,6 @@ jobs:
          - runner: macos-latest
            name: 'macos'
         compiler:
-         - name: gcc-4.8
```
hanno-becker, line 230: I am slightly concerned by dropping gcc-4.8 because it has in the past flagged issues not seen in any other compiler. For example, it seems to emit code for unused `static inline` functions and thereby enforce that all functions referenced in those inline functions are actually linked. This was (and is) an issue in #1135 where we imported the backend headers for the ABI checks but didn't want to link against the constant tables.

mkannwischer: we are not dropping 4.8 - it's just moved to the other workflow

hanno-becker: I understand that. But I am uncomfortable not running it in every PR, because gcc-4 has a history of demonstrating differing behavior I'd like to be aware of as early as possible, not after a merge.

mkannwischer: There is no technical reason to not run that on every PR, so I am happy to add that back if we keep the list of tested compilers short.
What would be a good list?

mkannwischer: Something like the latest 3 versions available + the oldest version (which is gcc 4.8 and clang-6?

hanno-becker: Yes, let's start with that.

.github/workflows/legacy-compilers.yml
```diff
@@ -0,0 +1,457 @@
...
+        target:
+         - runner: ubuntu-24.04-arm
+           name: 'aarch64'
+         - runner: ubuntu-latest
+           name: 'x86_64'
+        # Keep in sync with compiler_tests_macos below.
+        # ci.yml covers the oldest available + 3 latest per family on every PR;
+        # the intermediate versions are tested here.
+        compiler:
+         - name: gcc-4.9
```
hanno-becker: I'd rather err on the side of caution and test all versions here, including the ones already tested in the main CI. This way one does not have to keep them in sync.

.github/workflows/legacy-compilers.yml
```diff
@@ -0,0 +1,457 @@
...
+           c17: True
+           c23: True
+           opt: all
+           examples: true
+         - name: zig-0.11
+           shell: zig0_11
+           c17: True
+           c23: False
+           examples: False
+           opt: no_opt
```
hanno-becker, line 195: Might be worth a comment again why we don't test the opt version here.

.github/workflows/legacy-compilers.yml
```diff
@@ -0,0 +1,457 @@
...
+          func: true
+          kat: false
+          acvp: false
+          wycheproof: false
+          examples: ${{ matrix.compiler.examples }}
+          opt: ${{ matrix.compiler.opt }}
+          nix-shell: ${{ matrix.compiler.shell }}
+          cflags: "-std=c11 ${{ matrix.cflags }}"
+      - name: native build+functest (C17)
+        if: ${{ matrix.compiler.c17 }}
```
hanno-becker: While at it, one could make this a bit leaner by explicitly encoding for each compiler family the version that added support for Cxx -- rather than having a `c17/c23` flag in the matrix. Similarly the logic of when to run examples and when to run opt/no-opt could be captured once here rather than scattered in the matrix.

.github/workflows/legacy-compilers.yml
```diff
@@ -0,0 +1,457 @@
...
+      max-parallel: 10
+      matrix:
+        system: [ubuntu-latest, ubuntu-24.04-arm]
+        compiler:
+          - gcc49
+          - gcc6
+          - gcc7
+          - gcc8
+          - gcc9
+          - gcc10
```
hanno-becker: This looks so much nicer than the big matrix above.

.github/workflows/legacy-compilers.yml
mkannwischer: okay, added all versions here.

.github/workflows/legacy-compilers.yml
mkannwischer: added.

.github/workflows/legacy-compilers.yml
mkannwischer: I replaced the c17/c23 flags by an explicit version and then only run the step if the version is >= than the one that added support. I can't think of an easier way right now, but feel free to do a follow-up PR.

.github/workflows/legacy-compilers.yml
mkannwischer: Yes, but it also encodes much less information.

.github/workflows/ci.yml
```diff
@@ -226,90 +226,55 @@ jobs:
...
-           c23: True
+         - name: clang-6
+           shell: clang6
+           version: 6
+           darwin: False
            opt: all
            examples: true
          - name: clang-20
            shell: clang20
+           version: 20
```
hanno-becker: Nit: If we anyway have `version`, we may split it into `compiler` and `version` and drop `name` and `shell`, and also avoid the `startsWith` below.

mkannwischer: Done

hanno-becker (approved):

PR #1701, opened 2026-05-21
Make MLK_ASM_* macros overwritable

hanno-becker:


hanno-becker:
This is causing non-trivial complications in AWS-LC's FIPS build. Not pursuing further for now.

PR #1702, opened 2026-05-22
nix: Bump z3 to 4.16.0 from nixpkgs-unstable

mkannwischer:
It is somewhat inconvienent for developers having to compile z3 from source. While this a one-off step, I would still like to avoid it if possible.

This commit sources z3 from nixpkgs-unstable rather than the local override so we pick up the binary cache, and aligns the pinned version with what nixpkgs 26.05 will ship.


mkannwischer:
Let's hold this back until we have #1703 has been addressed.

mkannwischer:
superseded by https://github.com/pq-code-package/mlkem-native/pull/1762

PR #1704, opened 2026-05-22
x86_64: 32-byte align Keccak x4 AVX2 stack frame

hanno-becker:
For better performance, align stack to 32-byte in the AVX2 x4 Keccak backend implementation.

Updates HOL-Light proofs accordingly. Unfortunately, no existing tactic for the extension from the 'core' to the 'subroutine' proofs can handle the pattern we use, so we fall back to some ad-hoc tactics. As and when s2n-bignum adds support for stack alignment to its automation, this can hopefully be removed.

We also extend scripts/cfify to track the CFA register. A `mov rsp, %REG` re-anchors the CFA on REG, and subsequent modifications to the RSP do not require CFI directives. We handle this by conditioning the rules subq/addq->cfi_adjust_cfa_offset on the operand being the current CFA reg.

**NOTE:** Compared to the mldsa-native port of this change, scripts/cfify required an additional fix: the original re-anchor rule fired on every `movq %rsp, %REG`, including scratch base-pointer copies (e.g. the `rep movsb` source setup in rej_uniform_avx2_asm.S). That misclassified the scratch copy as an alignment anchor, dropping the legitimate `.cfi_adjust_cfa_offset` on the matching `addq $N, %rsp` and emitting a spurious `.cfi_def_cfa_register`. cfify now scans forward to the next ret and only re-anchors when a matching `movq %REG, %rsp` restore is found in the same function body. This change needs porting to mldsa-native as well.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -36,15 +36,6 @@ jobs:
           ldflags: "-flto"
           bench_extra_args: ""
           nix_shell: bench
-        - system: rpi5
-          name: Arm Cortex-A76 (Raspberry Pi 5) benchmarks
-          bench_pmu: PERF
-          archflags: "-mcpu=cortex-a76 -march=armv8.2-a"
-          cflags: "-flto -DMLK_FORCE_AARCH64"
-          ldflags: "-flto"
-          bench_extra_args: ""
-          nix_shell: bench
-          cross_prefix: ""
         - system: a55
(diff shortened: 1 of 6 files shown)


mkannwischer (approved):
Thanks @hanno-becker - this looks good to me.

The benchmarking CI was failing due to the A76 runner having a full disk. I can't access that machine anymore, so I have temporarily disabled the benchmarks in a commit stapled onto this PR.

PR #1708, opened 2026-05-26
CI: Move RISC-V runner jobs into a standalone workflow

mkannwischer:
This RISC-V runners are often having capacity problems. This PR moves them to a separate workflow, so the rest of the CI can run in parallel.

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -34,6 +34,13 @@ jobs:
       id-token: 'write'
     uses: ./.github/workflows/nix.yml
     secrets: inherit
+  riscv:
+    name: RISC-V
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    needs: [ base ]
+    uses: ./.github/workflows/riscv.yml
   ci:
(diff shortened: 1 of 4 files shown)


hanno-becker (commented):
Seems like a good idea to me

.github/workflows/all.yml
```diff
@@ -34,6 +34,12 @@ jobs:
       id-token: 'write'
     uses: ./.github/workflows/nix.yml
     secrets: inherit
+  riscv:
+    name: RISC-V
+    permissions:
```
hanno-becker, line 39: `needs: [base]`?

hanno-becker (changes requested):
Seeing that there aren't that many RISC-V runners available, we should not unnecessarily exercise them; let's only run the RV workflow once the base workflow passes.

.github/workflows/all.yml
mkannwischer: Added.

hanno-becker (approved):

PR #1709, opened 2026-05-27
CI: Add gcc16 tests

mkannwischer:


.github/workflows/ci.yml
```diff
@@ -253,6 +246,12 @@ jobs:
            darwin: True
            opt: all
            examples: true
+         - family: gcc
+           version: "16"
```
mkannwischer, line 250: reminder for myself: gcc16 also needs to be added to the ct tests.

mkannwischer: done

hanno-becker (approved):
Might be worth mentioning in the commit message that gcc13 is dropped in accordance with the new 'policy' of testing only the latest 3 compiler versions per-PR.

PR #1711, opened 2026-05-30
nix: bump to nixpkgs 26.05

mkannwischer:
Update the nixpkgs input from nixos-25.11 to nixos-26.05 and adjust the
flake to the new package set:

- Default gcc changed from 14.3.0 to 15.2.0

- slothy: use the new pkgs.slothy from nixpkgs instead of the custom
  build; nix/slothy/default.nix now just re-exports it with a commented
  override for pinning a specific upstream revision. Drops the
  python3-for-slothy (unstable ortools) workaround.

- hol_light: hol_light is now taken directly from nixpkgs, as I have
  upstreamed a recent version and made the necessary adjustments to
  make it directly usable. Similar as for slothy, we keep a separate
  package to make it easier to overwrite the version in the future.

- python: provide script dependencies (pyparsing, sympy, ...) via
  python3.withPackages instead of bare python3Packages entries in a
  symlinkJoin, which no longer reach sys.path. Fixes autogen's
  "No module named 'pyparsing'".

- autogen: use the snake_case pyparsing API (parse_string/parse_all) to
  silence the deprecation warning from the newer pyparsing.

- cbmc: bitwuzla version 0.8.2 -> 0.9.0.

- autogen: Reformatted with newer ruff version

- platform tests: Increase disk on x86 from 20 GiB to 30 GiB as nix derivation
  does not fit anymore otherwise.

- clang 18: Clang 18 from 26.05 no longer compiles on MacOS - take it from
  the 24.05 channel instead.
  
  
  
  - Resolves #1719

PR #1712, opened 2026-05-30
nix: Use HOL-Light from nixpkgs

mkannwischer:
- I'm trying to upstream the latest HOL-Light to make it useable for us here: https://github.com/NixOS/nixpkgs/pull/525966


mkannwischer:
My upstream changes have been merged and backported to 26.05, so we can include the changes in #1711.

PR #1714, opened 2026-06-01
CI: Split mingw-w64 tests into per-PR and daily sets

mkannwischer:
Previously every PR tested a fixed set of mingw-w64 versions.

Mirror the gcc/clang/zig setup: base.yml now tests only the oldest working version (5.4.0) plus the three most recent majors (13.2.0, 14.2.0, 15.2.0) on every PR, while legacy-compilers.yml tests each major version (5.4.0 through 15.2.0) daily.

- Ports pq-code-package/mldsa-native#1138.

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -191,8 +191,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


mkannwischer:
Will merge this as a part of #1730

PR #1558, opened 2026-02-09
Armv8.1-M: Add CFI directives for stack unwinding 

mkannwischer:
Extend scripts/cfify with Armv8.1-M architecture support, handling
push/pop, vpush/vpop, sub/add sp, and bx lr.
Enable cfify for armv81m in scripts/autogen.

- Resolves https://github.com/pq-code-package/mlkem-native/issues/1517

PR #1720, opened 2026-06-05
Aligned with #1558

bremoran:


mkannwischer:
Thanks @bremoran. Can you push that change directly to that branch and please also use a more descriptive commit message.

PR #1727, opened 2026-06-07
 Fix signed-shift UB on 16-bit-int targets in `mlk_rej_uniform_c` and `mlk_poly_frombytes_c`

### LINKED ISSUE #1726: Signed-shift UB in `mlk_rej_uniform_c` on 16-bit-int targets
https://github.com/pq-code-package/mlkem-native/blob/bd4c076dff29cb7570a172c799af26d878d4d1ef/mlkem/src/sampling.c#L59

On 16-bit-`int` targets (AVR, MSP430), `buf[pos+1]` promotes to signed `int`, so `<< 8` overflows `INT_MAX` for bytes ≥ 128.

This does not cause problems in the AVR tests today, but we should fix it (e.g. cast to `uint16_t`) and investigate why the AVR compiler does not flag it.

Acceptance criteria:
- [ ] If possible, extend AVR test in CI so it flags the current code as UB.
- [ ] Fix the UB, e.g. by promoting to `uint16_t` prior to applying the shift.

mkannwischer:
Fixes signed-shift UB and adds UBSan for AVR to catch similiar UB.
See individual commit messages.

- Resolves #1726

### DIFF
--- a/mlkem/src/compress.c
+++ b/mlkem/src/compress.c
@@ -641,7 +641,12 @@ __contract__(
     const uint8_t t0 = a[3 * i + 0];
     const uint8_t t1 = a[3 * i + 1];
     const uint8_t t2 = a[3 * i + 2];
-    r->coeffs[2 * i + 0] = (int16_t)(t0 | ((t1 << 8) & 0xFFF));
+    /* Safety:
+     * - The explicit cast to uint16_t ensures that << 8 does
+     *   not signed-overflow even on a 16-bit system.
+     * - The cast to int16_t is safe due to the explicit 0xFFF truncation.
+     */
+    r->coeffs[2 * i + 0] = (int16_t)(t0 | (((uint16_t)t1 << 8) & 0xFFF));
     r->coeffs[2 * i + 1] = (int16_t)((t1 >> 4) | (t2 << 4));
   }
 
--- a/test/baremetal/platform/avr/platform.mk
+++ b/test/baremetal/platform/avr/platform.mk
@@ -26,6 +26,13 @@ CFLAGS += \
 	-fno-fat-lto-objects \
 	-DNTESTS_FUNC=10
 
+# Trap on signed-shift UB; needs -fno-wrapv as avr-gcc defaults to -fwrapv.
+# On UB, abort() in avr_wrapper.c is called.
+CFLAGS += \
+	-fno-wrapv \
+	-fsanitize=shift \
+	-fsanitize-trap=shift
+
 CFLAGS += $(CFLAGS_EXTRA)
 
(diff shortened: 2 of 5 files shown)


test/baremetal/platform/avr/platform.mk
```diff
@@ -26,6 +26,13 @@ CFLAGS += \
 	-fno-fat-lto-objects \
 	-DNTESTS_FUNC=10
 
+# Trap on signed-shift UB; needs -fno-wrapv as avr-gcc defaults to -fwrapv.
+# On UB, abort() in avr_wrapper.c is called.
+CFLAGS += \
+	-fno-wrapv \
+	-fsanitize=shift \
+	-fsanitize-trap=shift
```
hanno-becker, line 34: Why don't we use `-fsanitize=undefined`?

hanno-becker (changes requested):
Thank you for fixing this!

I tested locally that, without the code change, the UB is caught. However, the tests don't fail (err code 0 still), which should be fixed.

```
ERROR: abort() called (e.g. UBSan trap)

All good!```

**Comment by mkannwischer** [comment]
> Thank you for fixing this!
> 
> I tested locally that, without the code change, the UB is caught. However, the tests don't fail (err code 0 still), which should be fixed.
> 
> ```
> ERROR: abort() called (e.g. UBSan trap)
> 
> All good!```
> ```

That's an independent problem, see https://github.com/pq-code-package/mlkem-native/issues/1728. I'm on it.

File: mlkem/src/compress.c [hunk 2]
```diff
@@ -641,7 +641,9 @@ __contract__(
     const uint8_t t0 = a[3 * i + 0];
     const uint8_t t1 = a[3 * i + 1];
     const uint8_t t2 = a[3 * i + 2];
-    r->coeffs[2 * i + 0] = (int16_t)(t0 | ((t1 << 8) & 0xFFF));
+    /* Safety: The cast prevents signed overflow on platforms with
+     * 16-bit int. */
+    r->coeffs[2 * i + 0] = (int16_t)(t0 | (((uint16_t)t1 << 8) & 0xFFF));
```
hanno-becker, line 649: Strictly speaking, there's another safety argument here, namely that the conversion to `int16_t` in the end is safe (because we truncated to 0xFFF -- otherwise it wouldn't be).

hanno-becker: Added

hanno-becker (approved):

test/baremetal/platform/avr/platform.mk
mkannwischer: This seems to cause some problems in the ACVP tests for ML-KEM-1024. Let's merge this bug fix first, and extend the scope in a follow-up

PR #1729, opened 2026-06-07
Fix minor inconsistencies

mkannwischer:
Spotted by Claude Opus 4.8:

- indcpa.c: Fix comment on the unused noise buffer in mlk_indcpa_enc (K=3): it is NULL and skipped, not "overwritten later".
- verify.h: Document the arithmetic right-shift assumption in the mask helpers, matching the existing note in poly.c.
- keccakf1600.c: Parenthesize the arguments of MLK_KECCAK_ROL.
- fips202x4: const-qualify the input pointers of mlk_shake256x4.
- notrandombytes.h: Fix truncated SPDX identifier (MI -> MIT).
- riscv.yml: Gate the RISE RISC-V runner jobs on non-fork PRs,
  consistent with the other jobs not running on Github runners.

### DIFF
--- a/mlkem/src/fips202/keccakf1600.c
+++ b/mlkem/src/fips202/keccakf1600.c
@@ -31,7 +31,7 @@
 #if !defined(MLK_CONFIG_MULTILEVEL_NO_SHARED)
 
 #define MLK_KECCAK_NROUNDS 24
-#define MLK_KECCAK_ROL(a, offset) ((a << offset) ^ (a >> (64 - offset)))
+#define MLK_KECCAK_ROL(a, offset) (((a) << (offset)) ^ ((a) >> (64 - (offset))))
 
 void mlk_keccakf1600_extract_bytes(uint64_t *state, unsigned char *data,
                                    unsigned offset, unsigned length)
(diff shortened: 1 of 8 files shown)


mlkem/src/fips202/keccakf1600.c
```diff
@@ -31,7 +31,7 @@
 #if !defined(MLK_CONFIG_MULTILEVEL_NO_SHARED)
 
 #define MLK_KECCAK_NROUNDS 24
-#define MLK_KECCAK_ROL(a, offset) ((a << offset) ^ (a >> (64 - offset)))
+#define MLK_KECCAK_ROL(a, offset) (((a) << (offset)) ^ ((a) >> (64 - (offset))))
```
hanno-becker, line 34: Is there not a flag for the preprocesor that would forbid non-hygenic arguments?

hanno-becker (approved):

mlkem/src/fips202/keccakf1600.c
hanno-becker: This is rather for a static analysis check. I'm a bit shocked the none of our existing checks catches this.

mkannwischer: Agreed, we should put tooling in place to catch this.

PR #1730, opened 2026-06-07
x86_64: Use SysV calling convention for assembly on Windows

mkannwischer:
- Depends on #1714


mkannwischer:
@hanno-becker, do you foresee any issues in AWS-LC with this change?

mlkem/src/sys.h
```diff
@@ -226,6 +226,18 @@
 #define MLK_MUST_CHECK_RETURN_VALUE
 #endif
 
+/* The x86_64 assembly backend uses the SysV calling convention. On Windows,
+ * where the Microsoft x64 calling convention is the default, it can still be
+ * used with compilers that allow choosing the calling convention per
```
hanno-becker, line 248: Hm, so how does a consumer indicate whether their compiler supports this ABI? Shouldn't we set `MLK_SYSV_ABI` to the below whenever not set, _and_ have a flag `MLK_SYSV_ABI_SUPPORTED` which indicates whether the tool chain supports it. And then the backend is only used when the ABI is supported?

mkannwischer: Okay, i see how this is confusing.
The current logic is 
```c
#if defined(__x86_64__)
#define MLK_SYS_X86_64
#if defined(__AVX2__)
#define MLK_SYS_X86_64_AVX2
#endif
#endif /* __x86_64__ */
```
`__x86_64__` is defined by clang/gcc/mingw, but not by MSVC, so this acts as a gate to disable the backends. We should probably refine that.

mkannwischer: I have pushed a proposal. WDYT?

hanno-becker (approved):
`__attribute__((sysv_abi))` seems very well supported, so it's OK to assume it based on `__GNUC__` or `__clang__` being set.

mlkem/src/sys.h
hanno-becker: I think that looks good

PR #1732, opened 2026-06-07
RV64: Minor bug fix and constant-time hardening

mkannwischer:


mlkem/src/native/riscv64/src/rv64v_poly.c
```diff
@@ -71,26 +71,37 @@ static inline vint16m1_t fq_barrett(vint16m1_t a, size_t vl)
...
+  m = __riscv_vand_vx_i16m1(m, MLKEM_Q, vl); /* m = (x < 0) ?  Q : 0 */
+  rx = __riscv_vadd_vv_i16m1(rx, m, vl);     /* x += m               */
   return rx;
 }
 
-/* Conditionally subtract Q (if Q or above) */
+/* Conditionally subtract Q (if Q or above).
+ * See constant-time note on fq_cadd above. */
 
 static inline vint16m1_t fq_csub(vint16m1_t rx, size_t vl)
```
hanno-becker: This function seems dead and could be removed at this occasion.

hanno-becker (changes requested):
The changes look good to me, but one function patched here is actually dead and can just be removed.

mlkem/src/native/riscv64/src/rv64v_poly.c
mkannwischer: oh. why does the monobuild not flag that?

mkannwischer: because it's inline. 
Side note: We should not be using inline directly here and instead use `MLK_INLINE` to be compatible with C90. 
~~If we had a C90 monobuild running on risc-v, this would have been caught.~~

hanno-becker: It would be useful to have a way to catch this, agree.

mkannwischer: I have removed `fq_csub`. I have also fixed the C90 issues and added CI for that.
But this **does not** catch dead code because `MLK_INLINE` maps to `__attribute__((unused))` under C90. I don't know a good way that would allow us catch dead inlined functions right now.

.github/actions/multi-functest/action.yml
```diff
@@ -27,7 +27,7 @@ inputs:
     description: LDFLAGS to pass to linking
     default: ""
   compile_mode:
-    description: all | native | cross-x86_64 | cross-aarch64 | cross-riscv64 | cross-riscv32
+    description: all | native | cross-x86_64 | cross-aarch64 | cross-aarch64_be | cross-riscv32
```
hanno-becker: Hm, removing RV64 support from `multi-functest` seems wrong. How about allowing the caller to specify exec-wrapper arguments instead?

hanno-becker (changes requested):
Functionally, I'm fine with the changes, but I think removing RV64 support from the multi-functest action is not the right approach.

.github/actions/multi-functest/action.yml
mkannwischer: Yes, you are right, your suggestion is better. Changed.

.github/workflows/ci.yml
```diff
@@ -36,11 +36,6 @@ jobs:
            arch: x86_64
            mode: cross-x86_64
            nix_shell: cross-x86_64
-         - runner: ubuntu-24.04-arm
-           name: 'ubuntu-latest (aarch64)'
-           arch: riscv64
-           mode: cross-riscv64
-           nix_shell: cross-riscv64
```
hanno-becker, line 43: Rather than moving this into a separate job, can we now just have N entries for RV64 in the existing matrix? That would keep the cross-arch tests together.

mkannwischer: Done.

hanno-becker (approved):
LGTM, thank you @mkannwischer for the adjustments!

PR #1734, opened 2026-06-07
AVR: Propagate test exit codes through simavr

### LINKED ISSUE #1728: AVR test harness does not propagate exit codes; functional test failures go unnoticed
The AVR baremetal tests run under simavr, which always exits with code 0 regardless of what the firmware does: `return 1` from `main()`, `cli(); sleep_cpu()`, or `abort()` all yield simavr exit code 0.

Since `scripts/tests` validates the functional test purely by exit code, the AVR functional test currently cannot fail in CI (short of hanging into the timeout). The KAT and ACVP tests are unaffected, as they validate the program output instead. This also limits the UBSan-trap detection added for #1726: a trap prints an error, but the functional test still reports success.

simavr gained an exit-code mechanism in buserror/simavr@c9354b3 (`SIMAVR_CMD_EXIT_CODE_*`), which is not in the v1.7 release we get from nixpkgs, but could be backported via a patch in `nix/avr/`.

mkannwischer:
simavr always exited with code 0 regardless of the firmware behaviour, so the AVR functional test could not fail in CI. simavr recently gained exit-code commands (buserror/simavr@c9354b3, not yet in a release): the firmware declares a command register via AVR_MCU_SIMAVR_COMMAND, and writing SIMAVR_CMD_EXIT_CODE_0/1 to it makes simavr exit 0/1. Backport this as a patch to our simavr build. Override exit() in the test wrapper to forward the return value of main(), and make abort() exit nonzero so that UBSan traps fail the test.

The .mmcu section holding the command register declaration must be placed outside the flash address range (following the linker flags used by simavr's own examples): the linker otherwise places it between the .text and .data load segments, and simavr then loads .data at the wrong address, breaking the firmware.

- Resolves #1728

### DIFF
--- a/nix/avr/default.nix
+++ b/nix/avr/default.nix
@@ -10,6 +10,11 @@ let
       ./simavr-32kb-ram.patch
       ./simavr-uart-output-fix.patch
       ./simavr-16k-eeprom.patch
+      # Exit-code commands (SIMAVR_CMD_EXIT_CODE_*), not yet in a release
+      (pkgs.fetchpatch {
+        url = "https://github.com/buserror/simavr/commit/c9354b32e057e409c2fbc9454e26db3b3103c26a.patch";
+        hash = "sha256-JKIjBmvlIrUe/0lJ2ngVgN3sCB0UJYWtv2dNp0wcPEY=";
+      })
     ];
   });
 in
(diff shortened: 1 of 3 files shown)


hanno-becker (approved):
LGTM, and I tested locally that it works.

PR #1735, opened 2026-06-07
autogen: Allow --force-cross to target specific architectures

hanno-becker:
Bare --force-cross forced every architecture, pulling the arm-embedded and ppc64le toolchains into the HOL-Light object-code check shells even though that job only regenerates aarch64/x86_64 bytecode.

--force-cross now accepts an architecture list (--force-cross=aarch64,x86_64); the CI job uses it and the now-unneeded gcc-arm-embedded and toolchain_ppc64le are dropped from the hol_light-cross-* shells.

### DIFF
--- a/scripts/autogen
+++ b/scripts/autogen
@@ -3130,13 +3130,37 @@ def normalize_asm_macro_syntax():
     run_parallel(files, normalize_asm_macro_syntax_for_file)
 
 
+# Architectures autogen knows how to (cross-)compile assembly for. Used to
+# expand a bare `--force-cross` (no value) into "force all architectures".
+FORCE_CROSS_ALL_ARCHES = {"aarch64", "x86_64", "armv81m", "ppc64le"}
+
+
+def resolve_force_cross(value):
+    """Normalize the --force-cross CLI value into a set of architectures.
+
+    Cross-compilation always happens for an architecture when its cross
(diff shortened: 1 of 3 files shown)


scripts/autogen
```diff
@@ -4525,7 +4549,17 @@ def _main():
...
+        "--force-cross",
+        nargs="?",
+        const="all",
+        default=None,
+        metavar="ARCHES",
+        help="Treat a missing cross toolchain as an error instead of silently "
+        "skipping. Without a value, requires all architectures "
+        f"({','.join(sorted(FORCE_CROSS_ALL_ARCHES))}); or pass a comma-separated "
+        "list to require only those, e.g. --force-cross=aarch64,x86_64.",
+    )
```
mkannwischer, line 4563: Hmm, this doesn't validate that what you pass to `--force-cross` is actually valid. Can we change this to be consistent to what we are doing for the SLOTHY optimiztation, i.e., use `choices`? I think we list you can pass is then space separated, not comma separated, but I have not tested it.

mkannwischer (changes requested):

scripts/autogen
hanno-becker: Yes!

mkannwischer (approved):
Thanks for the changes. LGTM now.

PR #1737, opened 2026-06-07
Add clang-tidy to lint

hanno-becker:
Add clang-tidy to lint with pedantic config

Run clang-tidy as part of scripts/lint. clang-tidy needs a C compiler to resolve system headers, so the lint CI job moves from the compiler-less 'linter' shell to 'ci' (linters + native toolchain); the now-unused 'linter' devShell is removed.

Pre-suppress style checks hostile to this codebase; fix or suppress the substantive findings, including breaking the common.h <-> verify.h include cycle and renaming the definition parameters of poly_ntt, poly_invntt_tomont and poly_tomsg to match their header declarations.

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -20,7 +20,6 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - uses: ./.github/actions/lint
         with:
-          nix-shell: linter
           gh_token: ${{ secrets.GITHUB_TOKEN }}
           cross-prefix: "aarch64-unknown-linux-gnu-"
   quickcheck:
--- a/scripts/lint
+++ b/scripts/lint
@@ -194,6 +194,64 @@ lint-c-files()
 checkerr "Lint C" "$(lint-c-files)"
 gh_group_end
 
+lint-clang-tidy()
+{
+  if ! command -v clang-tidy >/dev/null; then
+    gh_error_simple "clang-tidy missing" "clang-tidy is not installed"
+    error "Lint clang-tidy"
+    SUCCESS=false
+    gh_summary_failure "Lint clang-tidy"
+    return 0
+  fi
+
+  # Files to analyse, by language. Add directories here to widen coverage;
+  # native backends are deliberately excluded (they need per-arch flags).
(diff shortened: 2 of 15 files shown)


.github/workflows/base.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - uses: ./.github/actions/lint
         with:
-          nix-shell: linter
+          nix-shell: ci
```
mkannwischer: nit: Change the default `nix-shell` of the link action and remove it here.

scripts/lint
```diff
@@ -194,6 +194,43 @@ lint-c-files()
...
+    error "Lint clang-tidy"
+    SUCCESS=false
+    gh_summary_failure "Lint clang-tidy"
+    return 0
+  fi
+
+  local nproc out success=true
+  nproc=$(getconf _NPROCESSORS_ONLN 2>/dev/null || echo 1)
+  for params in 512 768 1024; do
+    if ! out=$(git ls-files -- ":/mlkem/src/*.c" | grep -v "/native/" |
```
mkannwischer: What's the reasoning for not including the native backends?

scripts/lint
```diff
@@ -194,6 +194,43 @@ lint-c-files()
...
+      xargs -P "$nproc" -I {} \
```
mkannwischer, line 225: I suppose the output of this is completely unreadable if more than one file has mistakes? But I guess without it the runtime is unmanagable?

mkannwischer (approved):
Thanks @hanno-becker! Looks good to me. Some comments/questions below that you can ignore if you want.

scripts/lint
hanno-becker: This needs further thought and work because `clang-tidy` needs a proper compilation command for the files. Let's start with the core files. I had missed `fips202/*.c` though, which is now added.

.github/workflows/base.yml
hanno-becker: Fixed

scripts/lint
hanno-becker: Leaving this as-is for now.

PR #1742, opened 2026-06-11
SLOTHY: Pin to upstream 4302933

mkannwischer:


mkannwischer:
Does not appear to matter much for performance here.

Issue #1744, opened 2026-06-14
Avoid z3 soundness issue #9550


mkannwischer:
A soundness issue was reported in https://github.com/Z3Prover/z3/issues/9550 which affects the version of Z3 we are using through CBMC.

At this point, it is unclear if the soundness issue is affecting our proofs, but we must either A) update Z3 to a fixed version, or B) update CBMC to a version avoiding the soundness bug, as soon as possible.

See https://github.com/pq-code-package/mldsa-native/issues/1126

mkannwischer:
Workaround has been merged in CBMC: https://github.com/diffblue/cbmc/pull/9011

PR #1745, opened 2026-06-14
nix: Bump CBMC to latest develop to work around z3 soundness issue

mkannwischer:
To work around the z3 soundness issue (https://github.com/pq-code-package/mlkem-native/issues/1744), this commit updates to the unstable CBMC version which has a workaround for the issue - seehttps://github.com/diffblue/cbmc/pull/9011.

### DIFF
--- a/nix/cbmc/default.nix
+++ b/nix/cbmc/default.nix
@@ -15,13 +15,19 @@ buildEnv {
   paths =
     builtins.attrValues {
       cbmc = cbmc.overrideAttrs (_: {
-        version = "6.9.0";
+        # We pin an unstable CBMC develop snapshot rather than the 6.9.0
+        # release to pick up the workaround for the z3 soundness issue
+        # (https://github.com/pq-code-package/mlkem-native/issues/1744,
+        # https://github.com/diffblue/cbmc/pull/9011).
+        # TODO: switch back to a tagged release once one ships with the fix.
+        version = "6.9.0-unstable-2026-06-13";
         src = fetchFromGitHub {
           owner = "diffblue";
           repo = "cbmc";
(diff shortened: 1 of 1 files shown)


nix/cbmc/default.nix
```diff
@@ -15,13 +15,14 @@ buildEnv {
   paths =
     builtins.attrValues {
       cbmc = cbmc.overrideAttrs (_: {
-        version = "6.9.0";
+        version = "6.9.0-unstable-2026-06-13";
```
hanno-becker, line 23: Add a comment why we don't use a stable version here?

nix/cbmc/default.nix
```diff
@@ -15,13 +15,14 @@ buildEnv {
...
         src = fetchFromGitHub {
           owner = "diffblue";
           repo = "cbmc";
-          hash = "sha256-SMJBnzoyTwcwJa9L2X1iX2W4Z/Mwoirf8EXfoyG0dRI=";
-          tag = "cbmc-6.9.0";
+          hash = "sha256-2RcLAttkORSv3DqwIb1pj73aeYYY6r1RGpytTIW5NAc=";
+          rev = "7750282a89dc57bb1a8e5a1ec544e8584be0cee4";
         };
+        doInstallCheck = false;
```
hanno-becker, line 30: For my benefit, what's this about?

mkannwischer: for some nix packages, the install step checks if `--version` contains the version string set by the package. In this case that would be `6.9.0-unstable-2026-06-13` so the install step fails because the version is actually reported as `6.9.0`. 
So you can either keep the version at `6.9.0` or you need to skip that step.

nix/cbmc/default.nix
mkannwischer: Yes, good point. Added.

hanno-becker (approved):

PR #1747, opened 2026-06-14
tests: Add --platform convenience flag for target selection

hanno-becker:
Cross-platform tests currently require callers to look up and repeat compiler prefixes, target flags, execution wrappers, or platform Makefiles.

Add --platform profiles that capture those compile and run settings. For example:

```
  ./scripts/tests func --platform linux/aarch64
  ./scripts/tests func --platform linux/riscv64-rvv256
  ./scripts/tests func --platform baremetal/avr --opt=no_opt
  ./scripts/tests func --platform zephyr
```

Explicit --cross-prefix, --exec-wrapper, and --extra-makefile overrides remain available. CI now invokes these same commands directly, and the test driver reports missing compilers or runners with the relevant nix develop shells.


hanno-becker:
@mkannwischer This has annoyed me for a while, but the present proposal is not yet ideal because it does not yet cover the baremetal platforms.

One could perhaps have `--platform cross/{ARCH}` and `--platform baremetal/{PLATFORM}`? WDYT

.github/workflows/baremetal.yml
mkannwischer: Drop Makefile.

scripts/tests
```diff
@@ -32,6 +34,220 @@ def dict2str(dict):
...
+        "ldflags": ldflags,
+        "exec_wrapper": exec_wrapper,
+        "nix_shells": [nix_shell, "cross"],
+    }
+
+
+PLATFORMS = {
+    "native": {
+        "nix_shells": ["default"],
+    },
```
mkannwischer: The BUILDING lists "ci" as the nix shell for this platform.
And I also think this entry is never used as check_platform_tools is going to terminate early.

scripts/tests
```diff
@@ -32,6 +34,220 @@ def dict2str(dict):
...
+def apply_platform_profile(args):
+    if args.platform is None:
+        if args.auto is None:
+            args.auto = True
+        return
+
+    profile = PLATFORMS[args.platform]
+
+    if args.cross_prefix == "":
+        args.cross_prefix = profile.get("cross_prefix", "")
```
mkannwischer: This results in CROSS_PREFIX being defined as "".
For zephyr that means that
```
CROSS_PREFIX ?= arm-none-eabi-
```
is ineffective.

.github/workflows/zephyr.yml
```diff
@@ -26,9 +26,9 @@ jobs:
       - name: zephyr build + test
         uses: ./.github/actions/functest
         env:
-          EXTRA_MAKEFILE: test/zephyr/platform.mk
           ZEPHYR_TARGET: ${{ matrix.target.board }}
         with:
+          platform: zephyr
```
mkannwischer, line 31: Should the zephyr platform also be used in the bench.yml?

mkannwischer (changes requested):
Thanks @hanno-becker. Looks like a very convenient improvement.

scripts/tests
hanno-becker: Yes, that's true today, though I think for future robustness it's natural to suggest the default shell for native builds. I'll leave `default` and adjust BUILDING.

scripts/tests
hanno-becker: I believe we treat unset and empty variables largely the same, and I'm not sure we should start deviating from that. I think the is rather with Zephyr's platform.mk which contrary to other platform .mk's uses `?=` instead of `=`. I'll change it to `CROSS_PREFIX = arm-none-eabi=` for uniformity, and then the above code path is fine.

scripts/tests
```diff
@@ -437,7 +660,7 @@ class Tests:
         return self.args.opt.lower() in ["all", "no_opt"]
 
     def compile_mode(self):
-        return "Cross" if self.args.cross_prefix != "" else "Native"
+        return "Cross" if self.args.cross_prefix else "Native"
```
hanno-becker: This seems wrong. The cross prefix can now also come from the platform. Also, it is unclear if we really care about the cross prefix or rather native vs _some_ platform.

.github/workflows/zephyr.yml
hanno-becker: Yes, done

scripts/tests
```diff
@@ -32,6 +34,227 @@ def dict2str(dict):
...
+            setattr(args, name, None)
+
+    if args.platform is None:
+        if args.auto is None:
+            args.auto = True
+        return
+
+    profile = PLATFORMS[args.platform]
+
+    if args.cross_prefix is None and "cross_prefix" in profile:
```
hanno-becker, line 153: @mkannwischer I uniformly moved to the behavior of only touching the environment if a non-empty value is provided on the command line or the profile for the target variable.

scripts/tests
hanno-becker: Fixed

.github/workflows/baremetal.yml
hanno-becker: Done

scripts/tests
hanno-becker: I changed my mind here and adapted `tests` so it touches the environment only for non-empty values of `args` or `profile`.

hanno-becker:
@mkannwischer I removed the "preflight" check trying to anticipate missing binaries. This would have been a nice-to-have but is too difficult to get working across the different ways in which the binaries and compilers can be configured.

mkannwischer (approved):
Thanks! I've checked that all my previous comments are addressed.

PR #1738, opened 2026-06-08
Add Windows MSVC test support: rng_fail, alloc, wycheproof

willieyz:
mlkem-native already runs tests on Windows via Mingw-w64, but the MSVC/nmake build only covers the functional tests. This PR closes the remaining gap by adding the following tests to `Makefile.Microsoft_nmake`, against the portable C backend:

- `rng_fail`: link rules added.
- `alloc`: added `test/src/test_alloc.c` and `test/configs/test_alloc_config.h`, referencing the Linux Makefile. Needs to rebuild the `.obj` files since it uses a different config from the default functional test. A new `ALLOC_CFLAGS` macro keeps the build rules short.
- `wycheproof`: test obj building rules and link rules added.

Unit tests are intentionally omitted from this PR and can be added later once native backend support becomes available on Windows.

PR #1749, opened 2026-06-15
Add alloc test Windows support (C backend only)

willieyz:
This PR adds alloc test support for Windows using the C backend.

mlkem-native currently has no alloc test coverage on Windows. Referencing the Linux Makefile, this PR adds the alloc test
(`test/src/test_alloc.c`, `test/configs/test_alloc_config.h`) to `Makefile.Microsoft_nmake`. Note that this alloc test needs to rebuild the `.obj` files, since it uses a different config from the default functional test. I also define a new macro `ALLOC_CFLAGS` to reduce the length of the build rules


willieyz:
Based on the review comment: https://github.com/pq-code-package/mlkem-native/pull/1738#pullrequestreview-4505171684, the remaining gaps should be addressed in a single PR. As a result, I will close this PR and cherry-pick the relevant commit into PR #1738.

PR #1750, opened 2026-06-16
Add Zephyr-based test platform for QEMU MPS boards

mkannwischer:


test/zephyr/platform.mk
```diff
@@ -0,0 +1,96 @@
...
+		$(if $(ZEPHYR_FIPS202_BACKEND),-DCONFIG_FIPS202_MVE_BACKEND=y) \
+		$(4) \
+		-DUSER_CACHE_DIR=$(abspath $(ZEPHYR_BUILD_DIR)/$(2)/.cache) \
+		>/dev/null
+	$$(Q)cmake --build $(ZEPHYR_BUILD_DIR)/$(2) >/dev/null
+	$$(Q)[ -d $$(@D) ] || mkdir -p $$(@D)
+	$$(Q)cp $(ZEPHYR_BUILD_DIR)/$(2)/zephyr/zephyr.elf $$@
+endef
+
+$(eval $(call ZEPHYR_BIN,512,test_mlkem512,test/src/test_mlkem.c))
```
hanno-becker: Why do we need those per-target recipes here? This wasn't needed in previous baremetal platforms, and doesn't scale well as new make targets are added.

hanno-becker (changes requested):
I very much like the consolidation of targets and am generally in favor of introducing Zephyr.

However, I'm not yet happy with how all make targets have to be manually adjusted in the new platform.mk. Can we do better and keep building the test sources in the same way as for other baremetal platforms, but merely add some configurable 'finalize-build' step which bundles the compiled .o files and the Zephyr sources into a Zephyr binary? So rather than having only an `EXTRA_SOURCES` for a baremetal platform, we generalize this to a more elaborate custom build finalization step -- while still keeping the way the test sources themselves are compiled? Just thinking out loud.

mkannwischer (commented):
Thanks @hanno-becker for taking over and cleaning up my changes. I agree with the updates.

hanno-becker (approved):

PR #1754, opened 2026-06-23
AArch64: Remove dead instructions from NTT, invNTT, rej_uniform

hanno-becker:
* Suggested by: @dominic-mulligan-aws

Static analysis surfaced a handful of instructions whose results are never observed. Remove them.

* NTT / invNTT: the loop-tail `subs count, count, #1` sets NZCV, but the loop is closed by `cbnz` (which tests the register) and neither file contains any flag-reading instruction. Replace with a plain `sub`.

* rej_uniform: drop three dead instructions on the loop-24 / return paths:
  - `sub buflen, buflen, #24` (buflen is never read again before ret),
  - `add output_tmp, output_tmp, ctr1, lsl #1` (overwritten by the following `mov output_tmp, output_tmp_base` with no intervening read),
  - `b rej_uniform_return` to the immediately following label.


mkannwischer (approved):
Good catch! Thanks @hanno-becker!

hanno-becker:
> Good catch! Thanks @hanno-becker!

The 10min of applause should be directed at @dominic-mulligan-aws

PR #1762, opened 2026-06-26
Upgrade to tagged release of CBMC 6.10.0

### LINKED ISSUE #1744: Avoid z3 soundness issue #9550

rod-chapman:
Fixes #1744 

Update NIX config to use the tagged release of CBMC 6.10.0

### DIFF
--- a/nix/cbmc/default.nix
+++ b/nix/cbmc/default.nix
@@ -15,19 +15,13 @@ buildEnv {
   paths =
     builtins.attrValues {
       cbmc = cbmc.overrideAttrs (_: {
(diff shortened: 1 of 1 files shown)


mkannwischer (approved):
Thanks @rod-chapman!

PR #1766, opened 2026-06-29
Fix Armv8.1-M x4 xor_bytes alignment fallback

### LINKED ISSUE #1759: Armv8.1-M: Native XOR bytes fails unit test
**Issue:** The native implementation of `xor_bytes` for Armv8.1-M fails its unit test because the load instructions https://github.com/pq-code-package/mlkem-native/blob/main/mlkem/src/fips202/native/armv81m/src/state_xor_bytes_x4_mve.S#L198-L200 require word alignment which is not guaranteed in the intended generality of the function (any offset, any input length).

**IMPORTANT:** This is **not** an issue for the top-level ML-KEM API: The ASM works correctly if either `offset % 4 == 0`, or if `inlen < 8`, which is given for all uses by `fips202*.c`.

**Secondary issue:** The unit test actually fails in CI, but the baremetal exec wrapper somehow does not propagate the error. This needs fixing, too.

bremoran:
Fixes #1759.

Corrects the xor-bytes unaligned access

### DIFF
--- a/dev/fips202/armv81m/src/state_xor_bytes_x4_mve.S
+++ b/dev/fips202/armv81m/src/state_xor_bytes_x4_mve.S
@@ -51,8 +51,9 @@
 // ---------------------------------------------------------------------------
 //   Prologue -- if offset is not 8-byte aligned, absorb
 //               min(length, 8-(offset%8)) bytes via predicated byte loads.
-//   Main     -- process full 8-byte groups via word-level gather loads,
-//               bit-interleave, then VEOR into even/odd state halves.
+//   Main     -- process full 8-byte groups via vldrw.u32 gather loads when
+//               32-bit aligned, otherwise byte loads; bit-interleave, then
+//               VEOR into even/odd state halves.
 //   Tail     -- absorb remaining <8 bytes via predicated byte loads.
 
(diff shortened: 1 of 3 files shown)


mlkem/src/fips202/native/armv81m/mve.h
```diff
@@ -49,6 +49,30 @@ static MLK_INLINE int mlk_keccakf1600_xor_bytes_x4_native(
...
+
+  main_length = length - byte_prefix;
+  if (main_length >= 8u &&
+      ((((uintptr_t)data0 + byte_prefix) |
+        ((uintptr_t)data1 + byte_prefix) |
+        ((uintptr_t)data2 + byte_prefix) |
+        ((uintptr_t)data3 + byte_prefix)) &
+       3u) != 0u)
+  {
+    return MLK_NATIVE_FUNC_FALLBACK;
```
hanno-becker: I don't think we can use fallback here, because the C version it not bit-interleaved?

hanno-becker: ping @bremoran

bremoran:
@hanno-becker how does the fix look to you?

PR #1780, opened 2026-07-03
Add gcc to the NIX shell on macOS to support CBMC.

rod-chapman:
Using GCC for pre-processing source on all platforms offers the hope of
better stability of CBMC proofs by having the same set of header
files on all platforms.

Adjustment to the CBMC Makefile.common to always use gcc for CBMC pre-processing step.

Leaves CC set as "clang" for compilation and testing on macOS.

### DIFF
--- a/nix/util.nix
+++ b/nix/util.nix
@@ -49,6 +49,15 @@ rec {
       # - On some machines, `native-gcc` needed to be evaluated lastly (placed as the last element of the toolchain list), or else would result in environment variables (CC, AR, ...) overriding issue.
     pkgs.lib.optionals cross [ pkgs.qemu pkgs.gcc-arm-embedded x86_64-gcc aarch64-gcc riscv64-gcc riscv32-gcc ppc64le-gcc ]
     ++ pkgs.lib.optionals (cross && pkgs.stdenv.isLinux && pkgs.stdenv.isx86_64) [ aarch64_be-gcc ]
+    # On Darwin, install GCC in addition to Clang. We reuse the exact same
+    # `pkgs.gcc` attribute that the Linux native toolchain wraps (`native-gcc`
+    # = `wrap-gcc pkgs`, whose compiler is `pkgs.gcc.cc`). Because both sides
+    # reference the same package from the flake-locked nixpkgs, the Darwin GCC
+    # version is always identical to the Linux one (do NOT pin a hardcoded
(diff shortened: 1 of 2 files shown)


mkannwischer (changes requested):
Thanks @rod-chapman. Can you please do this change together with the CBMC changes? Like this it's a no-op that doesn't add any value as far as I can tell.

rod-chapman:
OK..  if you're happy with the NIX config change, I will update the CBMC Makefile.common now to explicitly choose gcc for pre-processing.

hanno-becker (approved):
I ran a spot-check on my M1 and did not notice any meaningful performance difference; insofar I'm open to trying this.

nix/util.nix
```diff
@@ -49,7 +49,16 @@ rec {
...
     ++ pkgs.lib.optionals (cross && pkgs.stdenv.isLinux && pkgs.stdenv.isx86_64) [ aarch64_be-gcc ]
-    ++ pkgs.lib.optionals cross [ native-gcc ]
+    # On Darwin, install GCC in addition to Clang. We reuse the exact same
+    # `pkgs.gcc` attribute that the Linux native toolchain wraps (`native-gcc`
+    # = `wrap-gcc pkgs`, whose compiler is `pkgs.gcc.cc`). Because both sides
+    # reference the same package from the flake-locked nixpkgs, the Darwin GCC
+    # version is always identical to the Linux one (do NOT pin a hardcoded
+    # `gccNN` here, or the two platforms could drift apart on a nixpkgs bump).
+    # `pkgs.gcc` is listed *before* `native-gcc` (clang on Darwin) so that
+    # clang's cc-wrapper setup hook runs last and CC remains "clang".
```
hanno-becker: Actually, thanks to @mkannwischer's sharp eye, this isn't true. The ordering is irrelevant here for what you end up with as `$CC`.

hanno-becker: @rod-chapman Can you simplify the PR accordingly please?

hanno-becker (changes requested):
See comment

nix/util.nix
rod-chapman: Done

hanno-becker (approved):

hanno-becker:
@rod-chapman Can you clean up the history please

mkannwischer (approved):

PR #1787, opened 2026-07-06
Make tests and examples compatible with reduced-API configurations

hanno-becker:
The previous tests and examples assumed all APIs (keygen, encaps,
    decaps) were always available: encaps tests generated a keypair as
    setup, decaps tests generated a keypair and ciphertext as setup, and
    the alloc test required keygen for every test function. This made them
    incompatible with reduced-API configurations where some APIs are
    disabled via MLK_CONFIG_NO_KEYPAIR_API, MLK_CONFIG_NO_ENCAPS_API, or
    MLK_CONFIG_NO_DECAPS_API.

    This commit introduces auto-generated test vectors and refactors all
    tests and examples so that each operation (keygen, encaps, decaps) can
    be tested independently using pre-computed test vectors.

    scripts/notrandombytes:
    - Add Python implementation of the SURF-based deterministic test PRNG,
      matching the C version in test/notrandombytes/
    - Used by autogen --test-vectors to generate reproducible randomness
      for test vector generation

    Test vector generation (scripts/autogen --test-vectors):
    - Add --test-vectors flag to scripts/autogen that invokes the ACVP
      binaries with randomness from scripts/notrandombytes to generate
      test/test_vectors/expected_test_vectors.h and the multilevel variant
    - Generate pk, sk, ct, ss vectors for all three parameter sets
      (512, 768, 1024)
    - Include z, d, m byte-arrays in the header so consumers don't need
      to hardcode them
    - Array dimensions are explicit in the generated header

    Test changes (test_mlkem.c):
    - The existing tests (test_keys, test_invalid_pk, etc.) require all
      three APIs and are left unchanged, guarded by
      !MLK_CONFIG_NO_KEYPAIR_API && !MLK_CONFIG_NO_ENCAPS_API &&
      !MLK_CONFIG_NO_DECAPS_API
    - Add test_kem_expected() as a new minimal test that works in
      reduced-API configurations: each block (keygen, encaps, decaps) is
      independently guarded and uses test vectors directly, so e.g. the
      decaps block can run without keygen by using test_vector_sk

    test_alloc.c and test_rng_fail.c:
    - Both files exercise the same 7 API entry points (keypair_derand,
      keypair, enc_derand, enc, check_pk, dec, check_sk), each
      independently guarded by the minimal required API
    - Encaps tests use test_vector_pk directly (no keygen dependency)
    - Decaps tests use test_vector_sk/ct directly (no keygen or encaps
      dependency)
    - main() uses r |= pattern for error accumulation

    test_stack.c:
    - Guard each test function (keygen/encaps/decaps stack measurement)
      by the matching API config; unavailable modes print SKIPPED

    gen_KAT/wycheproof/acvp:
    - KAT and Wycheproof stub out to a SKIPPED main() when any API is
      disabled (they inherently need all three)
    - ACVP guards each mode (keyGen / encapDecap encapsulation /
      encapDecap decapsulation / encapsulationKeyCheck /
      decapsulationKeyCheck) by the matching API config, so it builds
      and works in any subset

    Example refactoring:
    - Hoist test logic into static example_xxx() functions with
      #if/#else/#endif guards and SKIPPED stubs for disabled APIs
    - main() is a flat sequence of r |= example_xxx() calls
    - example_end_to_end runs the full keygen->encaps->decaps flow only
      when all three APIs are enabled; example_expected_vectors runs
      independent keygen/encaps/decaps KAT checks for whichever APIs are
      enabled
    - Multilevel examples use per-level static functions (test_expected_512,
      test_expected_768, test_expected_1024) with the same per-API guards
    - basic_deterministic uses test_vector_z/d/m from the header instead
      of hardcoded byte arrays
    - Add expected_test_vectors.h symlink (single- or multi-level) in each
      example directory pointing to test/test_vectors/

    CI (config-variations):
    - Add keygen-only, encaps-only, decaps-only, keygen-encaps, keygen-decaps,
      encaps-decaps test configurations, covering all 6 combinations of
      1 or 2 enabled APIs from {keygen, encaps, decaps}


test/src/test_mlkem.c
```diff
@@ -159,6 +165,76 @@ static int test_invalid_ciphertext(void)
...
+}
+#endif /* !MLK_CONFIG_NO_KEYPAIR_API */
+
+#if !defined(MLK_CONFIG_NO_ENCAPS_API)
+static MLK_NOINLINE int test_kem_expected_encaps(void)
+{
+  printf("XEN\n");
+  return 0;
+}
+#endif /* !MLK_CONFIG_NO_ENCAPS_API */
```
mkannwischer: Is this intentional?

test/src/test_mlkem.c
```diff
@@ -159,6 +165,76 @@ static int test_invalid_ciphertext(void)
...
+ * frames -- this reduces peak stack usage on memory-constrained targets
+ * such as AVR.
+ *
+ * These are only compiled when at least one API is disabled -- when all
+ * three are enabled, the roundtrip loop in main() provides equivalent
+ * coverage, and keeping the pre-computed vectors out of `.data` matters
+ * on RAM-constrained targets (AVR .data would otherwise grow by ~6 KB).
+ */
+#if defined(MLK_CONFIG_NO_KEYPAIR_API) || defined(MLK_CONFIG_NO_ENCAPS_API) || \
+    defined(MLK_CONFIG_NO_DECAPS_API)
```
mkannwischer: Maybe we can always run those tests, not just in the reduced-API config?

mlkem/src/kem.c
```diff
@@ -207,6 +212,7 @@ static int mlk_check_pct(uint8_t const pk[MLKEM_INDCCA_PUBLICKEYBYTES],
  *            - We optionally include PCT which is not present in
  *              the reference code. */
 MLK_EXTERNAL_API
+MLK_MUST_CHECK_RETURN_VALUE
```
mkannwischer, line 215: We should really have linting for this.

hanno-becker: Hm, here's it's the definition, not the declaration -- I'm not sure it's actually needed. A linter is a good idea, though.

test/src/test_mlkem.c
hanno-becker: Yes, Claude tried to be too clever here.

test/src/test_mlkem.c
hanno-becker: No.

test/src/test_mlkem.c
```diff
@@ -167,19 +261,42 @@ int main(void);
...
   unsigned i;
+  int r;
+
+  /* Fixed test-vector smoke test: exercises whichever of keygen/encaps/decaps
+   * are enabled against pre-computed vectors. Mark public test vectors
+   * DECLASSIFIED and the secret key test vector SECRET so valgrind CT
+   * testing sees the same taint model as production usage. */
+  MLK_CT_TESTING_DECLASSIFY(test_vector_pk, CRYPTO_PUBLICKEYBYTES);
+  MLK_CT_TESTING_DECLASSIFY(test_vector_ct, CRYPTO_CIPHERTEXTBYTES);
+  MLK_CT_TESTING_SECRET(test_vector_sk, CRYPTO_SECRETKEYBYTES);
```
mkannwischer, line 273: Also mark  d,z,m as secret?

test/src/test_mlkem.c
```diff
@@ -27,11 +28,16 @@
...
 
 
-static int test_keys_core(uint8_t pk[CRYPTO_PUBLICKEYBYTES],
-                          uint8_t sk[CRYPTO_SECRETKEYBYTES],
-                          uint8_t ct[CRYPTO_CIPHERTEXTBYTES],
-                          uint8_t key_a[CRYPTO_BYTES],
-                          uint8_t key_b[CRYPTO_BYTES])
+#if !defined(MLK_CONFIG_NO_KEYPAIR_API) && \
+    !defined(MLK_CONFIG_NO_ENCAPS_API) && !defined(MLK_CONFIG_NO_DECAPS_API)
+/* Force out-of-line so the ~6.4KB of stack buffers below stay in short-lived
```
mkannwischer: I am a little confused that in this file we are not gating for NO_RANDOMIZED_API. Are we not testing that configuration? Maybe that should be added as a follow up.

mkannwischer (changes requested):
Thanks @hanno-becker for implementing this feature and aligning with mlkem-native. 
CI isn't quite happy yet.

test/src/test_mlkem.c
hanno-becker: That's true. Again it's the same as in mldsa-native. Let's have a follow-up.

hanno-becker: Note that we do still exercise the fixed vectors.

hanno-becker: Opened https://github.com/pq-code-package/mldsa-native/issues/1317

test/src/test_mlkem.c
hanno-becker: Done

hanno-becker:
@mkannwischer Thank you for your review, I believe I have addressed your feedback. Can you re-review?

mkannwischer (approved):
Thanks @hanno-becker. This looks great to me now.

PR #1790, opened 2026-07-07
zephyr: keep project CFLAGS off Zephyr internals

bremoran:
Forwarding mlkem-native CFLAGS into the Zephyr app build is intentional: it keeps the Zephyr test and benchmark sources on the same warning policy as the normal make build, including `-Werror`, `-Wpedantic`, `-Wconversion` and `-Wsign-conversion`. Keep that behavior. The mlkem-native app sources still receive the expanded flags explicitly through `ZEPHYR_TEST_CFLAGS`.

The failure was that the same flags also reached Zephyr's own CMake build through the process environment. In the NUCLEO-N657X0-Q benchmark action, the composed `--cflags="${inputs.cflags} ${inputs.archflags}"` becomes a single space when both inputs are empty. scripts/tests treats that as a non-empty value and exports `CFLAGS`. GNU make then appends mlkem-native's default warning policy to `CFLAGS` and, because `CFLAGS` originated in the environment, exports it to recipe children. CMake imports that environment `CFLAGS` value into the global Zephyr build, so Zephyr internals such as lib/heap/heap_constants.c are compiled under mlkem-native's pedantic warning policy and fail with `-Werror`.

This change does not weaken mlkem-native's flags, and does not stop forwarding them to the Zephyr app. Instead, it runs the Zephyr CMake configure and build steps with `CFLAGS`, `CXXFLAGS`, `CPPFLAGS` and `LDFLAGS` unset. This keeps the deliberate strict policy on the mlkem/test sources via `ZEPHYR_TEST_CFLAGS` while leaving Zephyr-controlled sources under Zephyr's own warning policy.

This did not fire in the Zephyr functional tests because that path passes an empty `--cflags=""` value, together with `--no-auto`, so scripts/tests does not export `CFLAGS`. The hardware benchmark path was different: its action always joins cflags and archflags with a space, producing the non-empty " " case that triggered the environment leak.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -28,69 +28,69 @@ jobs:
       fail-fast: true
       matrix:
        target:
-        - system: rpi4
-          name: Arm Cortex-A72 (Raspberry Pi 4) benchmarks
-          bench_pmu: PMU
-          archflags: -mcpu=cortex-a72 -DMLK_SYS_AARCH64_SLOW_BARREL_SHIFTER
-          cflags: "-flto -DMLK_FORCE_AARCH64"
-          ldflags: "-flto"
-          bench_extra_args: ""
-          nix_shell: bench
-          extra_makefile: ""
-          zephyr_target: ""
-        - system: rpi5
-          name: Arm Cortex-A76 (Raspberry Pi 5) benchmarks
-          bench_pmu: PERF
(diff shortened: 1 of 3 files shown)


mkannwischer:
Superseded by https://github.com/pq-code-package/mlkem-native/pull/1791. Thanks @bremoran for debugging!

PR #1791, opened 2026-07-08
zephyr: keep project CFLAGS off Zephyr internals

mkannwischer:
Fixing the benchmarks in main

- Hoisted out from #1790

### DIFF
--- a/.github/actions/bench/action.yml
+++ b/.github/actions/bench/action.yml
@@ -88,10 +88,6 @@ runs:
               --ldflags="${{ inputs.ldflags }}" \
               --opt=$([[ ${{ inputs.opt }} == "false" ]] && echo "no_opt" || echo "opt")  \
               -v  ${{ inputs.bench_extra_args }}
-    - name: Check namespace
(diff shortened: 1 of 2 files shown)


.github/actions/bench/action.yml
```diff
@@ -89,6 +92,7 @@ runs:
               --opt=$([[ ${{ inputs.opt }} == "false" ]] && echo "no_opt" || echo "opt")  \
               -v  ${{ inputs.bench_extra_args }}
     - name: Check namespace
+      if: ${{ inputs.check_namespace == 'true' }}
       shell: ${{ env.SHELL }}
       run: |
         ./scripts/check-namespace
```
hanno-becker: This should just be removed I think. It doesn't belong here.

hanno-becker (changes requested):
The CFLAGS fix is correct. Regarding the namespacing, I think we should just drop `check-namespace` from the bench action altogether -- it doesn't belong there.

hanno-becker (approved):

PR #1796, opened 2026-07-09
zephyr: fix NUCLEO benchmark timing and TCM layout

bremoran:
Use Zephyr's 64-bit cycle counter so long benchmark windows do not wrap, and make benchmark percentile sorting robust for large uint64_t values.

Link NUCLEO-N657X0-Q Zephyr images as XIP from ITCM with runtime data and stacks in DTCM. Force the STM32N6 VTOR setup to the ITCM vector table and scrub the expanded DTCM window in the early reset hook.

Keep project CFLAGS out of Zephyr's cached global CMake flags, and print the Zephyr build log when a suppressed Ninja build fails.

Tested:
- nix develop .#zephyr --command make run_func EXTRA_MAKEFILE=test/zephyr/platform.mk ZEPHYR_TARGET=mps3-an547 CYCLES=NO -j1
- nix develop .#zephyr --command make run_bench EXTRA_MAKEFILE=test/zephyr/platform.mk ZEPHYR_TARGET=nucleo-n657x0-q CYCLES=NO -j1

### DIFF
--- a/test/bench/bench_components_mlkem.c
+++ b/test/bench/bench_components_mlkem.c
@@ -32,7 +32,12 @@
 
 static int cmp_uint64_t(const void *a, const void *b)
 {
-  return (int)((*((const uint64_t *)a)) - (*((const uint64_t *)b)));
+  const uint64_t va = *((const uint64_t *)a);
+  const uint64_t vb = *((const uint64_t *)b);
+
+  /* Avoid subtracting uint64_t values: qsort only needs the sign, and the
+   * difference may exceed int range after long benchmark runs. */
+  return (va > vb) - (va < vb);
 }
 
 #define CHECK(x)                                              \
--- a/test/zephyr/app/nucleo_n657x0_q.overlay
+++ b/test/zephyr/app/nucleo_n657x0_q.overlay
@@ -4,6 +4,18 @@
  */
 
 / {
+    mlk_itcm: memory@10000000 {
+        compatible = "zephyr,memory-region", "arm,itcm";
+        reg = <0x10000000 0x00040000>;
+        zephyr,memory-region = "ITCM";
+    };
+
+    mlk_dtcm: memory@30000000 {
+        compatible = "zephyr,memory-region", "arm,dtcm";
+        reg = <0x30000000 0x00040000>;
+        zephyr,memory-region = "DTCM";
+    };
+
     mlk_axisram: memory@34080000 {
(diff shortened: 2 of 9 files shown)


test/bench/bench_components_mlkem.c
```diff
@@ -32,7 +32,10 @@
 
 static int cmp_uint64_t(const void *a, const void *b)
 {
-  return (int)((*((const uint64_t *)a)) - (*((const uint64_t *)b)));
+  const uint64_t va = *((const uint64_t *)a);
+  const uint64_t vb = *((const uint64_t *)b);
+
+  return (va > vb) - (va < vb);
```
hanno-becker, line 40: Why was this change made? Some comment would be useful.

test/zephyr/app/nucleo_n657x0_q.overlay
```diff
@@ -4,13 +4,26 @@
...
+        zephyr,memory-region = "ITCM";
+    };
+
+    mlk_dtcm: memory@30000000 {
+        compatible = "zephyr,memory-region", "arm,dtcm";
+        reg = <0x30000000 0x00040000>;
+        zephyr,memory-region = "DTCM";
+    };
+
     mlk_axisram: memory@34080000 {
```
hanno-becker, line 19: Is this still needed?

hanno-becker (commented):
Can you split the cycle counter changes and the memory layout changes in two separate commits?

test/zephyr/app/nucleo_n657x0_q.overlay
bremoran: Yes, AXISRAM is being used to host Zephyr bootargs.

test/bench/bench_components_mlkem.c
bremoran: The old approach left it open that, if the cycle count differed by more than 2^32, this would cause an integer overflow. There were two options: increase the return type to 64-bit, or return the sign. Only the sign is used. This approach preserves that.

test/zephyr/app/nucleo_n657x0_q.overlay
bremoran: This is documented in test/zephyr/nucleo_n657x0_q/README.md

mkannwischer (approved):
Thanks @bremoran. LGTM.
I have checked that this is a faithful port of the changes applied in https://github.com/pq-code-package/mldsa-native/pull/1269.

PR #1800, opened 2026-07-15
Correct name of poly_getnoise_eta1_4x_native harness file

rod-chapman:
This ensures that generated SMT filenames are unique for all C functions,
so that all SMT files can be collected into a single directory if necessary.


mkannwischer (approved):

mkannwischer:
@rod-chapman, I see you have been restarting jobs in this PR. Note that for the jobs that run on EC2, there is no point in restarting the "Run tests" job as it won't be able to run on any instance. You will instead have to start the corresponding "Start instance" job which will then automatically trigger the "Run Tests" job.
I've just done that for you.

Issue #1763, opened 2026-06-29
Wycheproof: Support new testvectors


mkannwischer:
New ML-KEM tests have landed in Wycheproof:

https://github.com/C2SP/wycheproof/commit/782238d476d85dbdbb51d49c7efab36304b84e71
https://github.com/C2SP/wycheproof/commit/7a87c8699e576c128693fb31795c0df520051d9c
https://github.com/C2SP/wycheproof/commit/ee7b4f7e611928cbe163dc6f5e54527bfd166f34

We should consider adding them.

PR #1801, opened 2026-07-16
CI: Track upstream Wycheproof ML-KEM vector updates

### LINKED ISSUE #1764: Wycheproof: Do not use outdated cached testvectors
Our Wycheproof tests cache the testvectors locally, but doesn't check if the cached testvectors correspond to the pinned Wycheproof commit. This isn't a problem in CI, but we should still improve it.
This was fixed in mldsa-native in https://github.com/pq-code-package/mldsa-native/pull/1235.

hanno-becker:
Add a daily workflow and helper that compare the pinned Wycheproof commit against upstream main, restricted to testvectors_v1/mlkem_* files, and open or refresh a tracking issue when they differ. Also pin wycheproof_client.py to a specific commit with a per-commit cache directory, since the diffing mechanism needs a pin to compare against.

Ports pq-code-package/mldsa-native#1286.

Resolves #1764

### DIFF
--- a/test/wycheproof/wycheproof_client.py
+++ b/test/wycheproof/wycheproof_client.py
@@ -45,6 +45,25 @@ PARAMETER_SET_TO_LEVEL = {
     "ML-KEM-1024": 1024,
 }
 
+# Fields each test-case handler knows how to interpret. If Wycheproof adds a
+# new field to a schema, an unrecognized key here means it may carry
+# information we are silently failing to check; fail loudly instead.
+KNOWN_FIELDS = {
+    "keygen_seed": {"tcId", "comment", "result", "seed", "ek", "dk"},
+    "encaps": {"tcId", "comment", "flags", "result", "ek", "m", "c", "K"},
+    "semi_expanded_decaps": {
+        "tcId",
+        "comment",
+        "flags",
+        "result",
+        "dk",
+        "c",
+        "ek",
+        "K",
+    },
+    "combined": {"tcId", "comment", "flags", "result", "seed", "ek", "c", "K"},
+}
+
 
 def err(msg, **kwargs):
     print(msg, file=sys.stderr, **kwargs)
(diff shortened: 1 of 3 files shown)


mkannwischer (changes requested):
There have been changes to the testvectors upstream recently (see #1763). 
Could you either track a commit that predates those, or look into what needs to change to adequately cover the changes. 
From my quick glance, I believe that we need to change the test on our side to properly  test https://github.com/C2SP/wycheproof/commit/782238d4.

test/wycheproof/wycheproof_client.py
```diff
@@ -180,6 +210,12 @@ def run_semi_expanded_decaps_test(data_file):
                 )
             elif tc["result"] == "valid":
                 assert "K" in out, f"missing K in output tcId={tc['tcId']}"
+                # K is only present in the vector from Wycheproof commit
+                # 782238d4 onwards; older pinned commits lack it.
+                if "K" in tc:
+                    assert out["K"].upper() == tc["K"].upper(), (
+                        f"K mismatch tcId={tc['tcId']}"
+                    )
```
mkannwischer: I don't see much value in trying to support older Wycheprooof commits.
I'd prefer if only support the latest version and simplify this code instead.

test/wycheproof/wycheproof_client.py
```diff
@@ -79,6 +98,14 @@ def download_wycheproof_files(data_dir):
...
 
 
+def check_known_fields(kind, tc):
+    unknown = set(tc.keys()) - KNOWN_FIELDS[kind]
+    assert not unknown, (
+        f"Unrecognized field(s) {sorted(unknown)} in {kind} tcId={tc['tcId']}; "
+        "Wycheproof schema may have grown a new field this client doesn't check"
+    )
+
+
```
mkannwischer, line 108: Great addition. This should be ported to mldsa-native. 
I've opened https://github.com/pq-code-package/mldsa-native/issues/1291 for that.

mkannwischer (changes requested):
Thanks for porting this @hanno-becker and for also covering the new Wycheproof vectors. 

This is also resolving #1763.

test/wycheproof/wycheproof_client.py
hanno-becker: Okay.

test/wycheproof/wycheproof_client.py
hanno-becker: Thank you!

mkannwischer (approved):
Thanks for the changes @hanno-becker. LGTM.

PR #1805, opened 2026-07-20
AArch64: Add runtime NEON capability gate to native backend

### LINKED ISSUE #1804: Add Neon capability
This allows resolving #1803

hanno-becker:
* Resolves #1804 

Previously, the native backend and ABI checker assumed that NEON was
present on every AArch64 system.

Add MLK_SYS_CAP_AARCH64_NEON and gate every NEON-dependent native and
FIPS202 entry point on it. This allows integrators to provide runtime
capability detection and fall back to C.

Mark NEON requirements in the assembly ABI metadata and teach generated
ABI checks to skip unsupported kernels.

Route ABI checks through C call-stub wrappers. On AArch64, pass the
runtime capability into the assembly stub so it only saves, seeds,
captures, and restores vector registers when NEON is available. Keep the
GPR self-test unconditional and run vector corrupters only when NEON is
supported.

Extend the custom ID-register test configuration to detect AdvSIMD
through ID_AA64PFR0_EL1.

### DIFF
--- a/dev/aarch64_clean/meta.h
+++ b/dev/aarch64_clean/meta.h
@@ -28,6 +28,10 @@
 MLK_MUST_CHECK_RETURN_VALUE
 static MLK_INLINE int mlk_ntt_native(int16_t data[MLKEM_N])
 {
+  if (!mlk_sys_check_capability(MLK_SYS_CAP_NEON))
+  {
+    return MLK_NATIVE_FUNC_FALLBACK;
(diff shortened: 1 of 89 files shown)


mkannwischer (approved):
Thanks @hanno-becker! This looks great to me.
We need to be careful not to forget renaming it to include the aarch64 prefix before the v2 release.

PR #1809, opened 2026-07-21
AArch64: Add compile-time guard for Neon

### LINKED ISSUE #1808: Build fails for `-march=armv8-a+nosimd`
We don't check for `__ARM_NEON` support. This needs to be dealt with in the same way as AVX2.

- [ ] Add `MLK_SYS_AARCH64_NEON`
- [ ] Replace `MLK_SYS_AARCH64` by `MLK_SYS_AARCH64_NEON` as appropriate (in most but not all backends; e.g. FIPS202 has a purely scalar backend as well)
- [ ] Test `armv8-a+nosimd` in CI

hanno-becker:
* Resolves #1808 

Previously, we assumed that every AArch64 system has Neon, and that the implementation has access to it.

With this commit, use of Neon is explicitly guarded by __ARM_NEON, similar to how all use of AVX2 is guarded by `__AVX2__`.


hanno-becker:
@mkannwischer There's a slight inconsistency here, partly pre-existing: We use `__ARM_FEATURE_SHA3` for SHA3 detection, but `MLK_SYS_AARCH64_NEON` for Neon. The latter is in line with `MLK_SYS_X86_64_AVX2`. We may want to introduce `MLK_SYS_AARCH64_SHA3` at some point, but I believe it can be a follow-up.

mkannwischer (approved):
Thanks @hanno-becker! Same non-blocking comments as in https://github.com/pq-code-package/mldsa-native/pull/1305 apply, but besides that this looks great!

PR #1811, opened 2026-07-22
ACVP: Support FIPS203-tr1 encapDecap (v1.1.0.43)

### LINKED ISSUE #1755: AVR: Port in-place decoding and argument placement for ACVP tests from mldsa-native
See https://github.com/pq-code-package/mldsa-native/pull/1158
In mldsa-native several adjustments were needed to make ACVP tests work in the limited memory on AVR including how inputs to ACVP tests are handed over. For ML-KEM those are not strictly needed, but to not diverge from mldsa-native, they can also be ported over.

### LINKED ISSUE #1807: ACVP: Support v1.1.0.43
From the release notes:

ML-KEM / encapDecap / FIPS203-tr1 and ML-DSA / sigGen / FIPS204-tr1 (Algo / Mode / Revision) - The new ML-KEM encapDecap and ML-DSA sigGen testing revisions add the possibility to use the seed, expanded private key format, or both in testing. This is accomplished by a new registration property: keyFormats.

Task:
- Extend the ACVP tests to additionally support the new tests types

mkannwischer:
The FIPS203-tr1 encapDecap revision adds keyFormat 'seed'/'expanded' groups: the decapsulation key may be supplied as a seed (d||z) to expand rather than as the expanded dk.

Add a decode_dk() entry point to the ACVP harness accepting either dk=HEX or seed=HEX (expanding the seed via crypto_kem_keypair_derand), and route the decapsulation function through it. keyFormat is inspected per group, so the client copes whether or not a given file carries it. Download and run the ML-KEM-encapDecap-FIPS203-tr1 vectors when the version ships them (v1.1.0.43+); keep the base FIPS203 dataset.

Two quirks in the v1.1.0.43 sample vectors are worth recording:

- decapsulationKeyCheck cases carry only a tcId with no dk, yet their expected results are a non-trivial mix of pass and fail, so they cannot be reproduced offline. The client drops decapsulationKeyCheck cases that lack a key, from prompt and expected alike; they run normally again once a key is present. This is a known upstream sample bug (usnistgov/ACVP-Server#459), where the maintainers confirm keyFormat there should be 'expanded' with the dk provided.

- The base ML-KEM-encapDecap-FIPS203 file also gained keyFormat at v1.1.0.43, whereas for ML-DSA only the -tr1 revision carries it. This looks like an unintended regeneration of the base file; the per-group handling above keeps the client correct whether or not it is fixed.

Bump the default ACVP version to v1.1.0.43 and roll the CI matrix forward to the three latest revisions.

 - Resolves #1807 
 - Resolves https://github.com/pq-code-package/mlkem-native/issues/1755

### DIFF
--- a/test/acvp/acvp_client.py
+++ b/test/acvp/acvp_client.py
@@ -240,6 +240,9 @@ def keyless_decap_key_check(tg, tc):
 
     The v1.1.0.43 sample vectors omit dk for these cases; the check cannot run
     without a key. They run normally again once a key is present.
+
+    TODO(#1842): drop this once we bump past v1.1.0.43. The upstream bug
+    (usnistgov/ACVP-Server#459) is fixed on master, but not in any tag yet.
     """
     if tg.get("function") == "decapsulationKeyCheck" and "dk" not in tc:
         return "decapsulationKeyCheck without a key"
@@ -269,6 +272,23 @@ def flush_warnings():
     err("")
     for msg in _warnings:
         err(msg)
+
+
+def parse_kv_output(stdout):
+    """Parse the binary's `key=value` output lines, ignoring anything else.
+
+    Baremetal harnesses route stderr into stdout over semihosting, so usage
+    and diagnostic messages can be interleaved with the results. Those never
+    have a bare identifier before the first '=', so they are skipped. The
+    {en,de}capsulationKeyCheck cases provoke such messages by design, since
+    they pass keys of incorrect length.
+    """
+    kv = {}
+    for line in stdout.splitlines():
+        key, sep, value = line.partition("=")
+        if sep and key.isidentifier():
+            kv[key] = value
+    return kv
 
 
 def get_acvp_binary(tg):
@@ -305,9 +325,7 @@ def run_encapDecap_test(tg, tc):
             err(result.stderr)
             exit(1)
         # Extract results
-        for line in result.stdout.splitlines():
-            (k, v) = line.split("=")
-            results[k] = v
+        results.update(parse_kv_output(result.stdout))
     elif tg["function"] == "decapsulation":
         acvp_bin = get_acvp_binary(tg)
         # keyFormat 'seed' provides d and z to expand into the key; 'expanded'
@@ -331,9 +349,7 @@ def run_encapDecap_test(tg, tc):
             err(result.stderr)
             exit(1)
         # Extract results
-        for line in result.stdout.splitlines():
-            (k, v) = line.split("=")
-            results[k] = v
+        results.update(parse_kv_output(result.stdout))
     elif tg["function"] == "encapsulationKeyCheck":
         acvp_bin = get_acvp_binary(tg)
         acvp_call = exec_prefix + [
@@ -350,8 +366,7 @@ def run_encapDecap_test(tg, tc):
             err(result.stderr)
             exit(1)
         # Extract results
-        for line in result.stdout.splitlines():
-            (k, v) = line.split("=")
+        for k, v in parse_kv_output(result.stdout).items():
             results[k] = v == "1"
 
     elif tg["function"] == "decapsulationKeyCheck":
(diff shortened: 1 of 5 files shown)


test/acvp/acvp_client.py
```diff
@@ -63,6 +80,38 @@ def download_acvp_files(version):
...
+    # ACVTS files wrap the payload as [{"acvVersion": ...}, {...}].
+    return data[1] if isinstance(data, list) else data
+
+
+def drop_keyless_decap_key_checks(data):
+    """Drop decapsulationKeyCheck cases that provide no key.
+
+    Some ML-KEM sample vectors omit dk for these cases; the check cannot run
+    without a key. Removed from prompt and expected so the comparison stays
+    consistent, and runs normally once a key is present.
```
hanno-becker: We should open an issue to remove this once upstream is fixed

test/acvp/acvp_mlkem.c
```diff
@@ -119,6 +119,38 @@ static int decode_hex(const char *prefix, unsigned char *out, size_t out_len,
...
+                                  unsigned char dk[CRYPTO_SECRETKEYBYTES])
+{
+  size_t seed_len = strlen("seed=");
+
+  /* Prefix check via memcmp; strncmp is unavailable on baremetal builds. */
+  if (strlen(arg) >= seed_len && memcmp(arg, "seed=", seed_len) == 0)
+  {
+    unsigned char ek[CRYPTO_PUBLICKEYBYTES];
+    unsigned char coins[2 * MLKEM_SYMBYTES];
+    if (decode_hex("seed", coins, sizeof(coins), arg) != 0)
```
hanno-becker: NB: We should align this with mldsa-native, where `decode_hex` operates in-place.

test/acvp/acvp_mlkem.c
```diff
@@ -119,6 +119,38 @@ static int decode_hex(const char *prefix, unsigned char *out, size_t out_len,
...
+    {
+      return 1;
+    }
+    if (crypto_kem_keypair_derand(ek, dk, coins) != 0)
```
hanno-becker: This relates to https://github.com/pq-code-package/mldsa-native/issues/1272. It would be nice to avoid the unused `ek` buffer here.

mkannwischer: Added https://github.com/pq-code-package/mlkem-native/issues/1841 to track and referenced it the code.

test/acvp/acvp_client.py
mkannwischer: Added https://github.com/pq-code-package/mlkem-native/issues/1842 to track and referenced it in the code. It's fixed in upstream master, but there is no new release for it yet.

test/acvp/acvp_mlkem.c
mkannwischer: Done.

test/acvp/acvp_client.py
```diff
@@ -457,6 +512,18 @@ def test(prompt, expected, output, version, supported_functions=None):
                 f"Unsupported functions: {fns}."
             )
 
+    # Seed key-format cases require keyGen to expand the key; drop them when the
+    # build does not provide it.
+    if supported_functions is not None and "keyGen" not in supported_functions:
```
hanno-becker, line 517: Pre-existing: It's a bit odd to ignore a missing ACVP bin and to interpret it as `all functions supported` _and_ represent that as `supported_functions == None`. I think we should fix that, but it's separate from this PR.

hanno-becker: https://github.com/pq-code-package/mlkem-native/issues/1861

test/acvp/acvp_mlkem.c
hanno-becker: Hm, no?

hanno-becker (dismissed):
There's still a minor misalignment with mldsa-native whereby `decode_hex` is not operating in-place as far as I see.

hanno-becker (approved):

PR #1813, opened 2026-07-23
Backends: Namespace assembly local labels with mlk_ prefix

hanno-becker:
Previously, local labels in the assembly backends were prefixed only by the name of the enclosing function, e.g. Lntt_layer123_start in ntt_aarch64_asm.S. Global entry symbols are namespaced per project via MLK_ASM_NAMESPACE(), but local labels are emitted as literal text and survive verbatim into the autogenerated mlkem/*.S files.

mldsa-native derives its assembly from mlkem-native and kept the same function and label names, so its local labels are byte-for-byte identical (Lntt_layer123_start, Lrej_uniform_loop48, ...). This is harmless as long as each backend is assembled into its own object file, but breaks when both backends are merged into a single link unit: AWS-LC's FIPS build concatenates all sources and runs them through the delocate tool, which flattens them into one symbol namespace and panics on the duplicate labels.

This commit adds a fixed mlk_ project prefix to every local label across all backends (AArch64 arithmetic and FIPS202, x86_64 AVX2, Armv8.1-M MVE, ppc64le), not just the ones that clash today, so mlkem-native emits Lmlk_ntt_layer123_start and a future mld_-prefixed mldsa-native drop can no longer collide, independent of the integrator or platform.

The change is made in the dev/ assembly sources and propagated to mlkem/*.S and the proofs/hol_light/* drops via scripts/simpasm. Local labels emit no object-code bytes, and branch targets encode PC-relative offsets rather than names, so the assembled .text is byte-identical; the byte-locked HOL Light and CBMC proofs are unaffected.

check_asm_loop_labels in scripts/autogen is updated to require the mlk_-prefixed function-derived prefix, so a newly added un-namespaced label fails CI.


mkannwischer (approved):
Thanks! LGTM. I checked that this is aligned with the corresponding mldsa-native PR.

PR #1821, opened 2026-08-02
Backends: Prefix assembly filenames with mlkem_

hanno-becker:
The backend .S files are self-contained and could be lifted into other projects on their own. However, a generic name like ntt_aarch64_asm.S says nothing about which scheme it implements. In fact, mlkem-native and mldsa-native happen to pick several identically.

This commit prefixes the arithmetic assembly under dev/, mlkem/ and proofs/hol_light/ so each file is self-identifying:

  dev/aarch64_opt/src/ntt_aarch64_asm.S
    -> dev/aarch64_opt/src/mlkem_ntt_aarch64_asm.S

FIPS202 assembly keeps its names: keccak_f1600_* already names the primitive. The two armv81m MVE helpers named neither scheme nor primitive, so they take the prefix of the symbols they export:

  state_xor_bytes_x4_mve.S
    -> keccak_f1600_x4_state_xor_bytes_mve.S

Filenames only: YAML `Name:` fields, exported symbols, `mlk_` local labels and the bytecode embedded in the proof scripts are untouched. Hand-updated the proofs/hol_light Makefiles, dump_bytecode.ml, hol_light.yml, README and the dev/ header back-references; the rest is autogenerated.


mkannwischer (approved):
LGTM. Thanks @hanno-becker!

PR #1823, opened 2026-08-02
test_stack: Use a valid secret key for the decaps measurement

### LINKED ISSUE #1822: Stack test measures a rejected decapsulation
`test_decaps_only()` in `test/src/test_stack.c` passes an all-zero secret key, so `crypto_kem_dec()` returns on the `H(ek)` check before decapsulating and `indcpa_dec()` plus the re-encryption never run. The reported decaps figure is therefore far too low — 7,680 bytes vs. 21,992 with a valid key on ML-KEM-768 — and since decaps is actually the largest consumer at every parameter set, `scripts/stack` also understates the overall peak by around 10%.

mkannwischer:
test_decaps_only() passed an all-zero secret key. crypto_kem_dec() compares H(ek) against the hash stored in dk and returns before decapsulating if they differ, so indcpa_dec() and the re-encryption never ran and went unmeasured.

Copy the decaps test vectors into the local buffers instead. The buffers stay local so the caller-side working set matches test_keygen_only() and test_encaps_only(), keeping the three numbers comparable.

ML-KEM-768, before -> after:

  KeyGen : 17,888 -> 17,888
  Encaps : 19,744 -> 19,744
  Decaps :  7,680 -> 21,992
  Peak   : 19,744 -> 21,992

Decaps is the largest consumer at every parameter set, so the peak was understated by around 10%.

- Resolves https://github.com/pq-code-package/mlkem-native/issues/1822

### DIFF
--- a/test/src/test_stack.c
+++ b/test/src/test_stack.c
@@ -8,6 +8,8 @@
 
 #include "mlkem_native.h"
 
+#include "../test_vectors/expected_test_vectors.h"
+
 static void print_info(void)
 {
 #if !defined(MLK_CONFIG_NO_KEYPAIR_API)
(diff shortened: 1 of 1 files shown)


hanno-becker (approved):
Good catch, thank you @mkannwischer!

PR #1831, opened 2026-08-02
CI: Test import into AWS-LC v5.4.0

hanno-becker:
* Builds on #1821 

v5.0.0 predates two upstream changes that current mlkem-native relies
on, so the integration test was pinned to a release that no longer
matches HEAD. This bumps it to v5.4.0 and reconciles the differences,
following https://github.com/aws/aws-lc/pull/3367:

  - AWS-LC shipped its own copy of the AArch64 backend meta.h to add a
    runtime NEON gate. mlkem-native provides that gate itself now, so
    the copy is a stale duplicate naming symbols (mlk_ntt_asm) that no
    longer exist. Consume the vendored meta.h instead.

  - MLK_SYS_CAP_NEON is an enumerator in mlkem/src/sys.h now, so AWS-LC
    defining it as a macro expands inside the enum declaration. Drop the
    macro and take mlk_sys_cap in the custom capability function.

  - The assembly source lists named each .S file, so the mlkem_ prefix
    left them pointing at files the importer no longer writes. Glob the
    backend directories instead, as the mldsa lists already do; future
    renames then need no edit here. This also replaces the filename
    rewrite table that previously lived in pre_import.patch.

pre_import.patch keeps the unifdef fix. Without it the reduced-API
guards survive the import and MLK_ARITH_BACKEND_AARCH64 is left in
place, which is false at build time once the include is rewritten to
s2n-bignum: intt, poly_tobytes and poly_tomont then assemble to nothing
while the C frontend still references them.

Verified against v5.4.0: importer, build, and crypto_test pass, with all
ten AArch64 assembly symbols defined and resolved. FIPS=1 is untested
locally, as static FIPS builds are Linux-only.

Once https://github.com/aws/aws-lc/pull/3367 lands, all three hunks become redundant and
post_import.patch can likely go away.

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -82,7 +82,7 @@ jobs:
     needs: [ base ]
     uses: ./.github/workflows/integration-awslc.yml
     with:
-      commit: v5.0.0
+      commit: v5.4.0
     secrets: inherit
   ct-test:
     name: Constant-time
--- a/integration/aws-lc/post_import.patch
+++ b/integration/aws-lc/post_import.patch
@@ -1,39 +1,26 @@
 # Copyright (c) The mlkem-native project authors
 # SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
 #
-# Map mlkem-native's AArch64 NEON capability to AWS-LC's runtime CPU
-# capability detection.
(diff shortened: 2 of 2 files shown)


mkannwischer (approved):
Thanks @hanno-becker. LGTM.

PR #1833, opened 2026-08-03
Fix stale filename references

mkannwischer:
The HOL-Light dependency lists for the five AArch64 keccak proofs named keccak_specs.ml, but the spec file is proofs/hol_light/common/keccak_spec.ml. Since `needs` drives a suffix match against the changed files, a change to keccak_spec.ml alone never triggered those proofs.

The monolithic assembly unit still pointed at mlkem_native_monobuild.c, which was renamed to mlkem/mlkem_native.c.


hanno-becker (approved):
Good catches!

PR #1840, opened 2026-08-03
Use runtime checks in Wycheproof client

### LINKED ISSUE #1827: Port: Use runtime checks in Wycheproof client
Port of pq-code-package/mldsa-native#1234.

mkannwischer:
The client checked test results with assert, which is stripped under python -O, so a corrupted or mismatching test vector was reported as passing. Replace them with fail()/require() helpers.

Also guard the output-dictionary lookups on the valid paths, so a missing key from the binary is reported as a test failure rather than a KeyError traceback.

- Resolves https://github.com/pq-code-package/mlkem-native/issues/1827
- Ports https://github.com/pq-code-package/mldsa-native/pull/1234

### DIFF
--- a/test/wycheproof/wycheproof_client.py
+++ b/test/wycheproof/wycheproof_client.py
@@ -69,6 +69,16 @@ def err(msg, **kwargs):
     print(msg, file=sys.stderr, **kwargs)
 
 
+def fail(msg):
+    err(msg)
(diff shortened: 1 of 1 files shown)


hanno-becker (approved):
LGTM. Thanks for adding further key checks.

PR #1850, opened 2026-08-05
CBMC: Check that outputs are zeroized or unmodified on error

### LINKED ISSUE #1849: CBMC: Verify that output buffers are unchanged or zeroized upon failure
This property is expressible in CBMC, though it's unclear how much it will affect proof times. We should try.

hanno-becker:
* Based on #1851 
* Resolves #1849 

Strengthens the CBMC specifications for the top-level API to include the API convention that output buffers be either zeroized or unmodified upon error, as documented in the API conventions.


mkannwischer (approved):
Thanks @hanno-becker! LGTM. I'm suprised this required so few changes.

mkannwischer (changes requested):
Actually the contract of zeroization isn't strong enough and this only works due to inlining of a memset right now, I believe. Can you strengthen that?

hanno-becker:
> Actually the contract of zeroization isn't strong enough and this only works due to inlining of a memset right now, I believe. Can you strengthen that?

Done, see last commit.

mkannwischer (approved):
Thanks @hanno-becker! LGTM.

PR #1852, opened 2026-08-05
Split ambiguous MLK_ERR_FAIL into specific error codes

### LINKED ISSUE #1846: Split ambiguous MLK_ERR_FAIL into separate error codes
Analogous to https://github.com/pq-code-package/mldsa-native/pull/1309. At least, PCT failure should be a separate error code.

hanno-becker:
* Resolves: https://github.com/pq-code-package/mlkem-native/issues/1846
 
Previously, MLK_ERR_FAIL was returned for several externally distinct failure conditions: a public key rejected by the FIPS 203 'modulus check', a secret key rejected by the FIPS 203 'hash check', and a failed pairwise consistency test (PCT). An integrator translating mlkem-native return codes into their own error space could not tell these apart without extra work.

This commit adds three dedicated error codes:

  MLK_ERR_INVALID_PK (-4)  check_pk / encaps rejected the public key
  MLK_ERR_INVALID_SK (-5)  check_sk / decaps rejected the secret key
  MLK_ERR_PCT_FAIL   (-6)  keygen PCT self-test failed

The existing codes (-1..-3) are unchanged. In particular, MLK_ERR_FAIL remains, but is no longer returned by any function.

Since the key pair tested by the PCT was just generated by this library, a key check rejecting it hints at a faulty implementation rather than at bad input; mlk_check_pct therefore maps MLK_ERR_INVALID_{PK,SK} to MLK_ERR_PCT_FAIL and passes platform failures through unmodified.

The commit also makes the CBMC ensures contracts specific instead of admitting MLK_ERR_FAIL everywhere, and drops it from the indcpa contracts, which only ever return 0 or MLK_ERR_OUT_OF_MEMORY.


mkannwischer (approved):
Thanks @hanno-becker! LGTM!

PR #1860, opened 2026-08-07
liboqs: Normalize return codes to OQS_STATUS

### LINKED ISSUE #1802: liboqs integration: normalize return codes to OQS_STATUS
Follow-up to liboqs [open-quantum-safe/liboqs#2469](https://github.com/open-quantum-safe/liboqs/pull/2469) / [#2437](https://github.com/open-quantum-safe/liboqs/issues/2437).

Our `ML-KEM-*_META.yml` points liboqs directly at the namespaced `mlk_kem_*` functions, which can return `MLK_ERR_OUT_OF_MEMORY (-2)` / `MLK_ERR_RNG_FAIL (-3)`. liboqs's `OQS_STATUS` is a closed enum (`OQS_SUCCESS = 0`, `OQS_ERROR = -1`) and the generated dispatcher casts our result straight through, so those out-of-contract values would reach consumers unchanged. It's latent today (the liboqs RNG shim can't fail and the default stack allocator can't return `NULL`), but `-2` becomes reachable with a custom allocator, and callers testing `if (ret == OQS_ERROR)` would miss it.

We should normalize at the `integration/liboqs/` boundary (map nonzero -> `OQS_ERROR`) via a small glue shim, analogous to mldsa-native's `sig_glue.c`.

Sibling issue: pq-code-package/mldsa-native#1296.

mkannwischer:
Collapse any nonzero result to OQS_ERROR in a new kem_glue.c, since OQS_STATUS is a closed enum and mlkem-native returns a wider error set. All five entry points liboqs links against are shimmed, and the META.yml files point at the shims.

~~Unlike in mldsa-native, this is latent today:
nothing but 0 and MLK_ERR_FAIL can reach liboqs,
since the default stack allocator rules out MLK_ERR_OUT_OF_MEMORY, and the liboqs randombytes shim wraps the void OQS_randombytes and always returns 0, ruling out MLK_ERR_RNG_FAIL.
However, as we may introduce additional error codes later on, it seems worthwhile to align this with mldsa-native now.~~
Key validation can still fail and that no longer returns -1 since https://github.com/pq-code-package/mlkem-native/pull/1852.

 - Resolves #1802

### DIFF
--- a/integration/liboqs/kem_glue.c
+++ b/integration/liboqs/kem_glue.c
@@ -1,0 +1,83 @@
+/*
+ * Copyright (c) The mlkem-native project authors
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+ */
+
+/*
+ * liboqs glue for mlkem-native.
+ *
+ * liboqs's OQS_STATUS is a closed enum of OQS_SUCCESS / OQS_ERROR, and the
+ * generated ML-KEM wrapper casts the result of the linked implementation
+ * straight through. mlkem-native returns a wider error set, so these thin
+ * shims collapse any nonzero result to OQS_ERROR. They are referenced from
+ * the per-level META.yml files.
+ *
+ * The shims are compiled once per build (parameter set and backend), so the
+ * namespaced symbol names match the names liboqs links against.
+ */
+
+#include <stdint.h>
+
+#include <oqs/common.h>
+
+#include "mlkem/mlkem_native.h"
+
+/*
+ * mlkem-native's public symbols are prefixed with MLK_CONFIG_NAMESPACE_PREFIX.
+ * mlkem_native.h builds these names via MLK_API_NAMESPACE(), but #undef's
+ * MLK_API_NAMESPACE_PREFIX at the end of the header, so MLK_API_NAMESPACE()
+ * cannot be reused here. Replicate the concatenation for both the shim symbols
+ * and their callees.
+ */
+#define MLK_OQS_CONCAT_(x, y) x##y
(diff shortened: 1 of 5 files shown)


integration/liboqs/kem_glue.c
```diff
@@ -0,0 +1,83 @@
...
+}
+
+int mlk_oqs_keypair_derand(uint8_t *pk, uint8_t *sk, const uint8_t *coins)
+{
+  return mlk_oqs_status(MLK_OQS_NS(keypair_derand)(pk, sk, coins));
+}
+
+int mlk_oqs_enc(uint8_t *ct, uint8_t *ss, const uint8_t *pk)
+{
+  return mlk_oqs_status(MLK_OQS_NS(enc)(ct, ss, pk));
```
hanno-becker, line 62: What's OQS' convention for outputs? Do they need zeroization upon failure?

hanno-becker (approved):
I did not test, but it LGTM. One should check though if OQS needs unconditional zeroization upon failure.

integration/liboqs/kem_glue.c
mkannwischer: I'm not aware of any guarantees required/provided by liboqs.

@bhess, @dstebila, do you require that output buffers are zeroized in case of failure? Our current contract is that on failure the output buffers are either unchanged or zerorized. If you require them to always be zero, we  could add an additional zeroization step to the glue code.

dstebila: No, we don't have any such requirement. Do you think we should?

PR #1862, opened 2026-08-07
Fix double level-suffixing of default namespace in multi-level builds

### LINKED ISSUE #1781: Default namespace subject to inconsistent level-suffixing in multi-level build
The default namespace is `PQCP_MLKEM_NATIVE_MLKEM{512,768,1024}_` but in a multi-level build, the prefix itself is suffixes by the level -- and what's more, the config is only imported once, for the 512-level. In effect, you'll get the namespace prefixes `PQCP_MLKEM_NATIVE_MLKEM512{512,768,1024}_XXX`, which makes no sense of course.

There's no functional issue here, but the naming is undesired and confusing.

hanno-becker:
The default namespace prefix embeds the parameter set (PQCP_MLKEM_NATIVE_MLKEM512/768/1024), which is correct for single-level builds but wrong in a multi-level build: the config is processed once (at the 512 level) and the namespacing machinery appends the parameter set again, yielding PQCP_MLKEM_NATIVE_MLKEM512{512,768, 1024}_. Use a level-independent default (PQCP_MLKEM_NATIVE_MLKEM) when MLK_CONFIG_MULTILEVEL_BUILD is set so the level is appended exactly once. Single-level builds are unchanged.

* Port of https://github.com/pq-code-package/mldsa-native/pull/1270 
* Resolves #1781

### DIFF
--- a/examples/basic_deterministic/mlkem_native/mlkem_native_config.h
+++ b/examples/basic_deterministic/mlkem_native/mlkem_native_config.h
@@ -684,7 +684,11 @@
  * e.g., PQCP_MLKEM_NATIVE_MLKEM512_
  */
 
-#if MLK_CONFIG_PARAMETER_SET == 512
+#if defined(MLK_CONFIG_MULTILEVEL_BUILD)
+/* In a multi-level build the parameter set is appended by the namespacing
+ * machinery, so the default prefix must not embed it. */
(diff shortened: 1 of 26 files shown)


mkannwischer (approved):
Thanks! LGTM and aligned with the changes in mldsa-native

PR #1866, opened 2026-08-09
Test: Run ABI checks as Zephyr applications

### LINKED ISSUE #1859: Port: Test: Run ABI checks as Zephyr applications
https://github.com/pq-code-package/mldsa-native/pull/1318

### LINKED ISSUE #1899: Port: Test: Run ABI checks as Zephyr applications
- Port of https://github.com/pq-code-package/mldsa-native/pull/1318

mkannwischer:
The ABI-check target bypassed the custom platform build path and linked an executable directly with $(LD). Under Zephyr that produced a host executable, which was then handed to Arm QEMU and faulted at startup.

Route the ABI-check target through the existing $(LINK)/CUSTOM_BUILD mechanism, so the ABI driver, self-tests, generated checks, and Armv8.1-M assembly sources build as a standalone Zephyr application. Select armv81m explicitly for mps3-an547 rather than inferring it from the generic arm-none-eabi compiler prefix, and reject ABI-check requests for other Zephyr targets explicitly. Preserve OPT and AUTO when scripts/tests invokes a run target, so the run step no longer re-enters the makefiles under a platform's own OPT default.

Track OPT, the selected FIPS202 backend, and the now-configurable test counts in the active build marker so changes to those inputs rebuild stale Zephyr binaries, name the native assembly amalgamation as an explicit dependency, and allow QEMU execution timeouts to be overridden.

- Ports pq-code-package/mldsa-native#1318.
- Fixes #1859  
- Fixes https://github.com/pq-code-package/mlkem-native/issues/1899

### DIFF
--- a/scripts/tests
+++ b/scripts/tests
@@ -538,6 +538,8 @@ class Tests:
         log = logger(test_type, scheme_str, self.args.cross_prefix, opt)
 
         args = ["make", test_type.make_run_target(scheme)]
+        if test_type.is_example() is False:
+            args += [f"OPT={int(opt)}", f"AUTO={int(self.args.auto)}"]
         if test_type.is_benchmark() is False and test_type.is_example() is False:
             args += self.make_j()
         if test_type.make_dir() != "":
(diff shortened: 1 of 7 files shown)


hanno-becker (commented):
Where/How is this exercised in CI?

PR #1877, opened 2026-08-24
Add example for the use of context parameters

hanno-becker:
This PR adds an example for the `MLK_CONFIG_CONTEXT_PARAMETER` option. As a concrete use case, illustrate how to pass a bump allocator context to a custom alloc/free implementation.

Because the example does include `mlkem_native.h` without `MLK_BUILD_INTERNAL`, this would also have caught https://github.com/pq-code-package/mlkem-native/issues/1874.


examples/context_parameter/example_context.c
```diff
@@ -0,0 +1,77 @@
...
+
+void example_context_free(example_context *context, void *ptr, size_t size)
+{
+  if (ptr == NULL)
+  {
+    return;
+  }
+
+  /* mlkem-native frees in FIFO order, so the to-be-freed region must be a
+   * tail of the used memory region. To free it, we merely move the pointer. */
```
mkannwischer: FIFO -> LIFO :-)

mkannwischer (changes requested):
Thanks @hanno-becker. One typo, rest looks good.

examples/context_parameter/example_context.c
hanno-becker: Argh!

hanno-becker: Fixed

mkannwischer (approved):

PR #1879, opened 2026-08-24
Use gcc and minimal libc headers for all CBMC proofs

rod-chapman:
1. CBMC Makefile.common now always uses gcc for pre-processing.

2. Specify -nostdinc for pre-processing, so host-dependent include files are NOT used, since this leads to proof instability across platforms.

3. Instead, specify and add our own minimal set the the standard LibC header files, but only what we need for mlkem-native and nothing else, all based solely on what is required by ISO C99 (7.17 - 7.21)

4. First two sets of header files are for

   4.1. The "LP64" data model, where "Long and Pointers are 64 bits", which is that used by common 64-bit platforms such as AArch64/Linux, x86_64/Linux and AArch64/macOS.

   4.2. The "ILP32" data model, for 32-bit systems such as 32-bit ARM, RV32, and most embedded 32-bit microcontrollers.

   A third set of headers for the LLP64 (64-bit Windows) will follow.

5. Environment variable CBMC_DM controls selection of the data-model in Makefile.common. It defaults to "LP64"

6. Update four proofs to use Z3 instead of bitwuzla, which turn out to be more efficient with CBMC 6.11 and these header files.

### DIFF
--- a/proofs/cbmc/Makefile.common
+++ b/proofs/cbmc/Makefile.common
@@ -143,11 +143,23 @@ default: report
 #
 MAKEFILE_PATHS = $(foreach makefile,$(MAKEFILE_LIST),$(abspath $(makefile)))
 PROOF_ROOT = $(dir $(filter %/Makefile.common,$(MAKEFILE_PATHS)))
-
 CBMC_ROOT = $(shell dirname $(PROOF_ROOT))
 PROOF_SOURCE = $(CBMC_ROOT)/sources
-PROOF_INCLUDE = $(CBMC_ROOT)/include
 PROOF_STUB = $(CBMC_ROOT)/stubs
+
+# Determine flags and include files for Data Model
+# which must be
+#     "LP64"  - 64-bit systems where "long and pointer are 64-bit"
+#
+# if CBMC_DM is not set, default to LP64
+CBMC_DM ?= LP64
+ifeq ($(strip $(CBMC_DM)),LP64)
+  GOTO_CC_DM_FLAG = --64
+  CBMC_DM_FLAG = --LP64
+  PROOF_INCLUDE = $(CBMC_ROOT)/cbmc/include/LP64
+else
+  $(error CBMC_DM must be LP64)
+endif
 
 # Flag to pass to goto-cc to make all static symbols globally visible. Leave it
 # unset or set it to --export-file-local-symbols to enable this behavior. Else,
@@ -331,7 +343,7 @@ NONDET_STATIC ?=
 # in order to get consistent header files on all platforms,
 # rather than using $(CC) (which might be preferred for compilation
 # on some platforms such as macOS/Darwin)
-COMPILE_FLAGS ?= -Wall -Werror --native-compiler gcc
+COMPILE_FLAGS ?= -Wall -Werror --native-compiler gcc -nostdinc -I$(PROOF_INCLUDE) $(GOTO_CC_DM_FLAG)
 LINK_FLAGS ?= -Wall -Werror
 EXPORT_FILE_LOCAL_SYMBOLS ?= --export-file-local-symbols
 
@@ -562,6 +574,7 @@ endif
 CBMCFLAGS += $(CBMC_FLAG_FLUSH)
 CBMCFLAGS += --object-bits $(CBMC_OBJECT_BITS)
 CBMCFLAGS += --slice-formula
+CBMCFLAGS += $(CBMC_DM_FLAG)
 
 ################################################################
 # Set C compiler defines
--- a/proofs/cbmc/Makefile.common
+++ b/proofs/cbmc/Makefile.common
@@ -152,13 +152,15 @@ PROOF_STUB = $(CBMC_ROOT)/stubs
 #     "LP64"  - 64-bit systems where "long and pointer are 64-bit"
 #
 # if CBMC_DM is not set, default to LP64
-CBMC_DM ?= LP64
+ifndef CBMC_DM
+  CBMC_DM = LP64
+endif
 ifeq ($(strip $(CBMC_DM)),LP64)
   GOTO_CC_DM_FLAG = --64
   CBMC_DM_FLAG = --LP64
   PROOF_INCLUDE = $(CBMC_ROOT)/cbmc/include/LP64
 else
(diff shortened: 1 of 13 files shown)


proofs/cbmc/poly_frommsg/Makefile
```diff
@@ -27,7 +27,7 @@ USE_DYNAMIC_FRAMES=1
 
 # Disable any setting of EXTERNAL_SAT_SOLVER, and choose SMT backend instead
 EXTERNAL_SAT_SOLVER=
-CBMCFLAGS=--bitwuzla
```
hanno-becker, line 30: Those solver changes seem unrelated or at least unexplained.

proofs/cbmc/Makefile.common
```diff
@@ -143,12 +143,29 @@ default: report
...
-PROOF_INCLUDE = $(CBMC_ROOT)/include
 PROOF_STUB = $(CBMC_ROOT)/stubs
 
+# Determine flags and include files for Data Model
+# which must be exactly one of
+#     "ILP32" - 32-bit systems where "int, long, pointer are 32-bit"
+# or  "LP64"  - 64-bit systems where "long and pointer are 64-bit"
+#
+# if CBMC_DM is not set, default to LP64
+CBMC_DM ?= LP64
```
hanno-becker: Can you plumb that into `tests cbmc ...` as well please?

proofs/cbmc/Makefile.common
```diff
@@ -143,12 +143,29 @@ default: report
...
+ifeq ($(strip $(CBMC_DM)),LP64)
+  GOTO_CC_DM_FLAG = --64
+  CBMC_DM_FLAG = --LP64
+  PROOF_INCLUDE = $(CBMC_ROOT)/cbmc/include/LP64
+else ifeq ($(strip $(CBMC_DM)),ILP32)
+  GOTO_CC_DM_FLAG = --32
```
hanno-becker: Does the 32-bit mode work?

hanno-becker (changes requested):
Thanks @rod-chapman, it's great to see progress in this direction.

Please can you confirm the status of the 32-bit model? I don't think we should include it without also testing it in CI. If we are not ready for that, yet, I suggest trimming down the PR to the addition of the infrastructure and the 64-bit header.

Please also integrate the choice of data model into `tests cbmc`.

Please also clean up the CI; the second commit message is insufficient, and there are solver changes in the first commit which are unrelated and/or unexplained.

rod-chapman:
Thanks for your thoughts.

Yes... we need the CI runners to change to add at least 1 32-bit proof run, but I wanted to discuss that separately, under a new PR. How many CI runs in 32-bit mode should we do? What will that cost in terms of runners? I thought that a policy discussion was warranted on those topics before going ahead.

With the proof changes - I noticed that several proofs became slower with bitwuzla. I tried switching them back to z3, and they got faster. I don't have an "explanation" for this per se... they're just faster with z3 now.

rod-chapman:
I see this PR as adding the plumbing to support multiple data models. I had hoped to add the "porcelain" layer (i.e. updating "tests cbmc" and CI actions) following requirements and policy decisions that I don't feel qualified to make alone.

proofs/cbmc/Makefile.common
rod-chapman: Yes... but running it locally on my Mac I observed some proofs became really slow. Some tweaking of proof Makefiles might be required (i.e. the setting of CBMC_OBJECT_BITS to suit the smaller value of SIZE_MAX), but I hope to address that under a new PR.

proofs/cbmc/poly_frommsg/Makefile
rod-chapman: Just maintaining reasonable performance for the _current_ CI runs seemed a reasonable thing to do under the auspices of this PR. As as I said below, adding _new_ CI runs to cover ILP32 requires some thinking and budget decisions that are best handled separately.

rod-chapman:
I am looking at the proof performance for ntt_butterfly_block() on my Grv4 machine now. I will update commit messages as soon as I have good data.

hanno-becker:
@rod-chapman Yes, it's fine to add the plumbing, but let's add 32-bit only when we can actually exercise at least a few functions with it. I think the management of the proof dimensions (function, solver, data model) could be piggy backed atop the draft https://github.com/pq-code-package/mldsa-native/pull/1122 for supporting multiple solvers (needs to be ported to mlkem-native; I can do that). Please use Claude, precedent, and your judgement, to make the necessary changes to `tests cbmc` -- I think a `--lp={lp64,...}` is all we need and should be simple to implement.

rod-chapman:
OK... I will update "tests" as part of the "plumbing" in this PR.

I think "multi-solver" support is still a long way off, and independent of this work on stabilization and multi-data-model support. As I said, my local benchmarks have cvc5 a long way behind z3 on CBMC-generated problems, but it would be a good way to drive improvement of cvc5 in the long term.

rod-chapman:
As I feared, proofs for some functions with ILP32 are very slow. For example - over 1 hour for indcpa_enc(). We may need assistance of CBMC team to diagnose this, so suggest we defer actual CI runs for ILP32 until a later PR.

hanno-becker:
> I think "multi-solver" support is still a long way off, and independent of this work on stabilization and multi-data-model support. As I said, my local benchmarks have cvc5 a long way behind z3 on CBMC-generated problems, but it would be a good way to drive improvement of cvc5 in the long term.

I don't think it's so far off -- one should not approach it as a monolithic change, but as a gradual one. And it relates to this PR insofar as it, too, adds a verification dimension to CBMC, and we need a way to manage them.

hanno-becker:
@mkannwischer Are you happy with this direction?

hanno-becker (approved):
Thanks @rod-chapman, this is a good first step towards 32-bit support.

hanno-becker:
(For the record: I took the liberty to implement a few fixes and cleanups myself in https://github.com/pq-code-package/mlkem-native/compare/7be7ed6563ef3b6223643a72466dc80178a97b1f..b1736c5d9c8c7f12758803e08c877f09c96fe2f5)

mkannwischer (approved):

PR #1881, opened 2026-08-28
AWS-LC: Bump integration test to v5.6.0 and drop patches

mkannwischer:


hanno-becker (approved):
That's the kind of PR I like :-)

PR #1883, opened 2026-08-28
CI: Bump EC2 AMIs to Ubuntu 26.04 LTS

mkannwischer:
The stock AMIs used by the EC2 runners were 24.04 images, the ones for the CI and benchmarking runners built as far back as August 2024. Bump them all to the current 26.04 LTS images in us-east-1 (release 20260819).

Also rename AMI_UBUNTU_LATEST_* to AMI_UBUNTU_* and the AMI selector strings from "ubuntu-latest (...)" to "ubuntu (...)": the AMIs are pinned, so "latest" was never accurate. The Ubuntu version, the architecture and the region are now recorded in a comment next to each AMI ID instead.

The custom AMI in ci.yml (docker images preinstalled) is unchanged.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -134,56 +134,56 @@ jobs:
         target:
           - name: Graviton2
             ec2_instance_type: t4g.small
-            ec2_ami: ubuntu-latest (aarch64)
+            ec2_ami: ubuntu (aarch64)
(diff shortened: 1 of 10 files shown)


hanno-becker (approved):
LGTM. Thank you for spotting and fixing the inaccuracy of the `x-latest` naming along the way.

PR #1886, opened 2026-09-01
sys.h: Add C-standard detection macros and CI snapshot check

hanno-becker:
The MLK_INLINE, MLK_ALWAYS_INLINE and MLK_RESTRICT guards tested `__STDC_VERSION__ >= 199901L` open-coded at each site, and there was no way to inspect the resolved sys.h configuration or to guard it against unintended change across the compiler and language-standard matrix.

Introduce cumulative MLK_SYS_C99, MLK_SYS_C11, MLK_SYS_C17 and MLK_SYS_C23 macros, each defined when the compiler reports that standard or later, plus MLK_SYS_CXX for C++. The three guards now test `defined(MLK_SYS_C99)` (the inline hints keep the `defined(inline)` fallback). Cumulative semantics keep "C99 or later" a single check that stays correct for future standards; "exactly Cxx" remains expressible as `defined(MLK_SYS_Cxx) && !defined(MLK_SYS_C<next>)`.

Add scripts/show-sys, which preprocesses a probe translation unit and prints the resolved MLK_* configuration (architecture, endianness, OS, C standard, inline/restrict/align spellings, SysV ABI), and scripts/check-sys, which diffs `make show-sys` against a checked-in snapshot. Expected snapshots for each platform and standard live under test/sys/expected and are verified from ci.yml, legacy-compilers.yml and the functest/multi-functest actions via a new sys_expected input.

MLK_SYS_C90 is intentionally omitted: under cumulative semantics every C build is at least C90, so pre-C99 is `!defined(MLK_SYS_C99)`. C++ builds define only MLK_SYS_CXX, so on GCC/Clang MLK_INLINE resolves to `__attribute__((unused))` as before.


mlkem/src/sys.h
```diff
@@ -124,6 +128,121 @@
...
+
+/*
+ * Detect the bit width of int.
+ *
+ * This is intentionally numeric rather than symbolic so it can be compared in
+ * preprocessor expressions.
+ */
+#if !defined(MLK_SYS_SIZEOF_INT)
+#if defined(__INT_WIDTH__)
+#define MLK_SYS_SIZEOF_INT __INT_WIDTH__
```
mkannwischer, line 166: When I read `SIZEOF`, I think `sizeof()` which is in bytes. Maybe we change this to `INT_WIDTH`?

mlkem/src/sys.h
```diff
@@ -124,6 +128,121 @@
...
+
+/*
+ * Detect the bit width of size_t.
+ *
+ * This is intentionally numeric rather than symbolic so it can be compared in
+ * preprocessor expressions.
+ */
+#if !defined(MLK_SYS_SIZEOF_SIZET)
+#if defined(__SIZE_WIDTH__)
+#define MLK_SYS_SIZEOF_SIZET __SIZE_WIDTH__
```
mkannwischer, line 198: same here.

mlkem/src/sys.h
```diff
@@ -124,6 +128,121 @@
...
+#if (__SIZEOF_SIZE_T__ * CHAR_BIT) == 16
+#define MLK_SYS_SIZEOF_SIZET 16
+#elif (__SIZEOF_SIZE_T__ * CHAR_BIT) == 32
+#define MLK_SYS_SIZEOF_SIZET 32
+#elif (__SIZEOF_SIZE_T__ * CHAR_BIT) == 64
+#define MLK_SYS_SIZEOF_SIZET 64
+#endif
+#elif defined(__SIZEOF_POINTER__) && !defined(__ASSEMBLER__)
+#if (__SIZEOF_POINTER__ * CHAR_BIT) == 16
+#define MLK_SYS_SIZEOF_SIZET 16
```
mkannwischer, line 209: Something is off here. this section is about sizet, not pointers.

mkannwischer (changes requested):

PR #1893, opened 2026-09-04
CI: Bump libOQS and Pavona pin

mkannwischer:
liboqs bump and other hygiene. See individual commits.

### DIFF
--- a/.github/workflows/integration-liboqs.yml
+++ b/.github/workflows/integration-liboqs.yml
@@ -40,7 +40,7 @@ jobs:
           packages: 'cmake python3-jinja2 python3-tabulate python3-git python3-pytest valgrind'
       - uses: ./.github/actions/setup-oqs
         with:
-          commit: 'f986aea60a9f3cb4055474aa212538bb0b14f1fe' # main (2026-05-27)
+          commit: 'c27c88b76473f67a8072ce5b66874d172287ff96' # main (2026-09-03)
           gh_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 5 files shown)


.github/workflows/integration-liboqs.yml
```diff
@@ -52,12 +52,8 @@ jobs:
           # TODO: Remove once patch is removed upstream
           # Remove patch
           sed -i "/name: mlkem-native/,/preserve_folder_structure/{/patches:/d}" scripts/copy_from_upstream/copy_from_upstream.yml
-          # TODO: Remove one it has been removed upstream
-          # Remove CT test suppressions
-          echo "" > tests/constant_time/kem/passes/ml_kem
           # Temporarily remove oldpqclean because of build failures in its SHA3 assembly
           yq e -i 'del(.kems[] | select(.name == "kyber"))' scripts/copy_from_upstream/copy_from_upstream.yml
-          yq e -i 'del(.sigs[] | select(.name == "dilithium"))' scripts/copy_from_upstream/copy_from_upstream.yml
```
hanno-becker, line 60: Why can dilithium be removed here, but not Kyber?

hanno-becker (approved):

.github/workflows/integration-liboqs.yml
mkannwischer: liboqs 0.15.0 dropped Dilithium. Kyber is still supported as of now.
I do not know if the problems we were having with that code are fixed.

PR #1894, opened 2026-09-04
Performance-measuring on macOS using clock_gettime_nsec_np() API

### LINKED ISSUE #1888: Benchmarks fail on AArch64/macOS 26.6.2
```
tests bench -r -c MAC
```
no longer works on recent macOS (e.g 26.6.2) builds. Apple have been restricting access to the kernel performance counter API, so some changes might be needed in test/hal/hal.c to restore functionality.

The benchmark_components.c program reports "kpc_set_config failed", which appears here:
https://github.com/pq-code-package/mlkem-native/blob/a1b94a1d903072d5ea1d8ecf434bd9d5518014c0/test/hal/hal.c#L315

Google AI summary says
"Apple Silicon Compatibility: On newer Apple Silicon architectures (such as the M4 chips running modern macOS versions like Sequoia/Tahoe releases), strict privilege levels and tighter security controls mean direct user-space or unprivileged calls to configure PMU via kpc_set_config or related kpc.config sysctl wrappers can fail and return -1 unless proper entitlements or debugging contexts are present."

and refers to 

https://developer.apple.com/documentation/kernel/1572437-kpc_set_config

rod-chapman:
Fixes #1888 

Adds performance-measuring for macOS from the clock_gettime_nsec_np() API.

This is intended to replace the KPC-based measurement (of clock cycles)
which no longer works on recent Apple Silicon (M4 and M5 CPU) or on
macOS 26.6.2 on all Apple Silicon machines.

In particular

1. The old "CYCLES=MAC" build option is renamed to "MAC_KPC" and remains
   for backwards compatibility for the time being.

2. New build options "CYCLES=MAC_NS" is added, with a new implementation
   of the HAL API, based on the clock_gettime_nsec_np() function on macOS.

3. The HAL API is extended to add a new get_cyclecounter_unit() API
   that returns a reference to a string giving the units of measurement -
   currently either "cycles" or "ns" so that the higher-level reporting
   scripts can report and process the units correctly.

4. Benchmarking test programs are updated to report units of measurement
   accordingly.

5. The scripts/tests script is updated to parse and record the measurement
   units, and to send the units on to the GitHub pages.

6. Update README files and BUILDING.md to cover MAC_KPC and MAC_NS options.

7. Update GitHub CI workflows and actions to use the MAC_NS option when
   running benchmarks on the Mac Mini (2020) test runner.

### DIFF
--- a/.github/actions/bench/action.yml
+++ b/.github/actions/bench/action.yml
@@ -9,7 +9,7 @@ inputs:
     description: Name for the benchmarking run
     required: true
   perf:
-    description: Method of obtaining PMU metrics
+    description: "HAL implementation used to measure runtime, passed to `tests bench -c`: NO | PMU | PERF | MAC_KPC | MAC_NS"
     required: true
     default: "PERF"
   cflags:
@@ -66,6 +66,7 @@ runs:
           cat >> $GITHUB_STEP_SUMMARY <<-EOF
             ## Setup
             Architecture: $ARCH
+            HAL implementation (CYCLES): ${{ inputs.perf }}
             - $(uname -a)
             - $(nix --version)
             - $(${{ matrix.target.cross_prefix }}gcc --version | grep -m1 "")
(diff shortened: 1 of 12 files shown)


hanno-becker:
I don't think this is still needed after #1895

PR #1895, opened 2026-09-04
hal: Fix macOS cycle counting on Apple M4 and macOS 26.6

### LINKED ISSUE #1888: Benchmarks fail on AArch64/macOS 26.6.2

mkannwischer:
The kpc-based cycle counter hardcodes 0x02 as selector of the CORE_ACTIVE_CYCLE event. Since M4, Apple Silicon uses 0x11 instead, and the macOS 26.6 kernel rejects configurations with selectors that are unknown for the running CPU, so kpc_set_config fails. Look up the configuration word in the kperfdata database of the running CPU instead of hardcoding it.

The macOS 26.6 kernel further only applies the configuration when counting is started, so set it before enabling counting.

Tested on Apple M1 (macOS 26.4), where cycle counts are unchanged. Also tested on Apple M4 (macOS 26.6).

 - Resolves #1888
 
 For the record, this has now been tested on:
 - Apple M1 (macOS 26.4)
 - Apple M1 (macOS 26.6)
 - Apple M4 (macOS 26.6)
 - Apple M5 (macOS 26.6)
 - x86_64 Mac (macOS 26.6)

### DIFF
--- a/test/hal/hal.c
+++ b/test/hal/hal.c
@@ -250,10 +250,23 @@ uint64_t get_cyclecounter(void)
   F(int, kperf_sample_get, int *)              \
   F(int, kpc_get_thread_counters, int, unsigned, void *)
 
+typedef struct kpep_db kpep_db;
+typedef struct kpep_config kpep_config;
+typedef struct kpep_event kpep_event;
+
+#define KPERFDATA_LIST                                                  \
+  F(int, kpep_db_create, const char *, kpep_db **)                      \
+  F(int, kpep_db_event, kpep_db *, const char *, kpep_event **)         \
+  F(int, kpep_config_create, kpep_db *, kpep_config **)                 \
+  F(int, kpep_config_add_event, kpep_config *, kpep_event **, uint32_t, \
+    uint32_t *)                                                         \
+  F(int, kpep_config_kpc, kpep_config *, uint64_t *, size_t)
+
 #define F(ret, name, ...)              \
   typedef ret name##proc(__VA_ARGS__); \
   static name##proc *name;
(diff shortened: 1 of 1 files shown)


rod-chapman (commented):
Please detect and reject an illegal setting of CYCLES at build time, or at least issue a warning that it has been set to "NO".

hanno-becker (approved):
I confirm that this work on Apple M5.

mkannwischer:
> Please detect and reject an illegal setting of CYCLES at build time, or at least issue a warning that it has been set to "NO".

I think this is an orthogonal improvement. I don't have time right now to implement it. Feel free to open an issue or PR for it.

PR #1896, opened 2026-09-05
CI: Use the main Ubuntu archive for Pavona dependencies

mkannwischer:
The regional EC2 mirror us-east-1.ec2.archive.ubuntu.com intermittently
stalls, leaving apt retrying for tens of minutes. Point the archive
sources at archive.ubuntu.com instead; security.ubuntu.com is unchanged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -632,6 +632,21 @@ jobs:
     container:
       ${{ matrix.container.id }}
     steps:
+      # On 2026-09-05, Bullseye's live mirrors advertised packages whose
+      # download URLs returned 404s, breaking CI dependency installation.
+      # Use a fixed snapshot to keep Bullseye compatibility tests working.
+      # Configure APT before checkout, which needs to install Git.
+      - name: Configure Bullseye package snapshot
+        if: ${{ matrix.container.id == 'debian:bullseye' }}
+        shell: bash
+        run: |
+          snapshot=20260824T000000Z
+          printf '%s\n' \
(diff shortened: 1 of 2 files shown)


hanno-becker (changes requested):
This does not seem to address the CI issue.

mkannwischer:
> This does not seem to address the CI issue.

Switched to  https://archive.ubuntu.com/ now which seems to be a Cloudflare CDN and looks more reliable than Ubuntu's EC2 mirror. That  seems to solve it.

mkannwischer:
debian:bullseye tests have an unrelated failure when installing git via apt:
```
E: Failed to fetch http://deb.debian.org/debian-security/pool/updates/main/p/perl/libperl5.32_5.32.1-4%2bdeb11u5_amd64.deb  404  Not Found [IP: 146.75.38.132 80]
```

I fixed it by switching to an older Debian snapshot.

.github/workflows/ci.yml
```diff
@@ -632,6 +632,21 @@ jobs:
     container:
       ${{ matrix.container.id }}
     steps:
+      # On 2026-09-05, Bullseye's live mirrors advertised packages whose
```
hanno-becker, line 635: Perhaps create an issue to track this

hanno-becker (approved):

.github/workflows/ci.yml
mkannwischer: Hmm, Bullseye is EOL anyway since 2026-08-31. I'm not convinced we need to track it.

PR #1905, opened 2026-09-10
CI: gate SLOTHY and add label to trigger it 

L-series:
- **CI: Only run SLOTHY when its inputs change**
- **CI: Allow requesting SLOTHY with the slothy-tests label**

Ported from: https://github.com/pq-code-package/mldsa-native/pull/1384


mkannwischer (approved):
Thanks @L-series! I've checked that this is a faithful port of https://github.com/pq-code-package/mldsa-native/pull/1384.

PR #1906, opened 2026-09-14
nix: Take hol_light and its OCaml toolchain from nixpkgs-unstable

mkannwischer:
Testing the latest HOL-Light + OCaml version that's going to be shipped in nixpkgs 26.11.


mkannwischer:
This works fine now and will arrive with the nixpkgs 26.11 update in November/December.

PR #1907, opened 2026-09-18
CI: Add xcode-27 runner image

mkannwischer:
GitHub's new macOS images are keyed to the major Xcode version rather than the OS. The first of them, xcode-27, is in public preview, runs macOS 27, and is available on arm64 only.


hanno-becker (approved):
LGTM

